PR #5, opened 2014-03-15
Add Twitter, CloudFlare, Coinbase. Add support for Authy

zach-taylor:
The three sites that I added still need images.

Authy is an alternative to Google Authenticator and has a lot more features so I thought it would be a  good idea to add it as well.


jdavis:
For Twitter, you can generate a 2FA code with the app?

zach-taylor:
Yeah, it seems that way

screenshot 2014-03-15 18 41 44

jdavis:
Ahhh, I was just looking around for it but missed that. Awesome.

jdavis:
Awesome. Before you merge into `gh-pages`, I saw a little issue with one of the tables, but I'll fix that up. Other than that, A++++ would merge again.

zach-taylor:
Sounds good!

PR #6, opened 2014-03-16
Adding Charles Schwab

mpdavis:
Adds Charles Schwab to the list of finance offerings.

Charles Schwab offers two factor via a hardware fob that they will send you for free.  You can read about it at the bottom of the website [here.](http://www.schwab.com/public/schwab/nn/legal_compliance/schwabsafe/your_questions_answered)

I wasn't sure if I should add another column for hardware two factor auth or not.  I am happy to do that if you want.


jdavis:
<3 Thanks Michael!

Yeah, we might have to add a column for hardware solutions. We'll have to wait to see how common it is.

PR #7, opened 2014-03-16
Digitalocean

mpdavis:


jdavis:
They don't have SMS? I would check but I'm away from my desktop.

mpdavis:
They use SMS as a way to unlock the account should you lose your TFA device, but as far as I can tell they don't offer it as a TFA option.

jdavis:
Ahh gotcha. I read something about SMS and was just wondering.

Also if you could change the link to this: https://www.digitalocean.com/company/blog/introducing-two-factor-authentication/

I think the Docs link should point to a place where people can read about it. Then it can be merged!

Also, I added you as a Collaborator. Feel free to merge in your own or commit directly to master. I'm using master as the development branch and gh-pages as the production.

PR #10, opened 2014-03-16
Adding LinkedIn and Evernote

jeffgodwyll:


jdavis:
This looks excellent, Jeff!

One question that I have from looking at it, do both the apps provide a possible way to provide a key for 2FA?

If so, then it should work similarly to [Facebook's Code Generator](https://www.facebook.com/help/www/270942386330392).

jeffgodwyll:
Yeah both apps require keys. But just so I'm sure I get you correctly, the apps included should have 2FA as well right?

Evernote's app requires an authenticator app (such as Google Authenticator App) installed on user's device. You then have to scan a QR code it'll generate, etc. There are backup options too.

LinkedIn's is pretty much straightforward. You enter the key received via sms to enter your account.

jdavis:
Ahh, I should be more clear in the README or something like that especially for the **Custom** section. But basically, for the custom section, I have that reserved for apps that provide their "own" 2FA service.

I used Facebook as an example because there are a few ways to get a code to login to Facebook. The first is through text, of course. The second is to open up the Facebook application, then use the built in Code Generator to generate a 2FA key. Basically the **Custom** section is for an app that doesn't use SMS, Google Auth, or Authy.

Another example is that Steam uses 2FA but only through email. So the second factor comes from sending out an email. If you look at Steam on the website, it has an X everywhere except Custom and there it has an Email icon.

Does that make sense?

jeffgodwyll:
Yeah it's clear now thanks.

I will like to put a variation of this("Basically the Custom section is for an app that doesn't use SMS, Google Auth, or Authy.") somewhere in the readme. Are you okay with that?

jdavis:
Yeah, I think clarifying it in the README would be great. Feel free to fork
it again and it'll be merged!

jeffgodwyll:
Great

PR #29, opened 2014-03-17
Adding Blizzard

smarkwell:


zach-taylor:
@smarkwell your iTunes link redirects to https://itunes.apple.com/app/battle.net-mobile-authenticator/id306862897 so might as well use that.

Will also need to add an icon. @jdavis and I have just been grabbing the site's favicon.ico and using http://www.convertico.com/ to convert it to png. 16x16 should be sufficient

smarkwell:
Can I'll redo the pull then.  Need to actually dl a github client to add the png file.

smarkwell:
Ok, should be good now based on your feedback.

PR #31, opened 2014-03-17
OptionsHouse

ilyakatz:


computmaxer:
@ilyakatz This PR has merge conflicts, please pull in the latest master to resolve.

zach-taylor:
Will also want to remove .DS_store from the gitignore. That should should go in your .gitignore_global file rather than each individual repository.

jdavis:
I left it in there for now, I agree in that it should go in the global. But some people might not ignore it... I also added the iOS for OptionsHouse. Thanks for the PR though, @ilyakatz

PR #37, opened 2014-03-17
Add VeriSign VIP Access.

ab:
Many prominent sites, including PayPal and name.com, support two factor
authentication using VeriSign's VIP Access system.

https://idprotect.vip.symantec.com/learnmore.v

I didn't test this at all, so let me know if it looks horrendous / if I can do anything to improve it.

### DIFF
--- a/_data/main.yml
+++ b/_data/main.yml
@@ -181,6 +181,7 @@ sections:
           url: https://www.paypal.com/
           tfa: Yes
           sms: Yes
+          verisign: Yes
           doc: https://www.paypal.com/us/cgi-bin?cmd=xpt/Marketing_CommandDriven/securitycenter/PayPalSecurityKey-outside&bn_r=o
 
         - name: Stripe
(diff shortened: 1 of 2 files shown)


sudeepkodavati:
Yes I think we have to have VeriSign. I just added name.com to the data.yml. Placed veriSign links for Android and Iphone in Custom element.

jdavis:
I'm taking a look at it now. How many sites do you think use this?

sudeepkodavati:
From the website I can see major sites like name.com, ebay, box.net, paypal, Merrill Lynch, adobe systems and others. On total the website shows 160 companies which are using Symantec two factor authentication (also called VeriSign). From the list we can see more than half of them are based out side of States. 

http://m.vip.symantec.com/wheretouse.v

ab:
Yeah I don't have concrete numbers, but my impression is that it's quite widely used — probably as much as plain TOTP is.

jdavis:
Okay! I think I'm ready to give VIP the go ahead to be added to the site. I've had a lot of requests for it and based off what you say, I think it looks great.

Unfortunately we've changed the data files around so it can no longer be merged. If you could fix that, that's great. If you can't, no worries; one of us will. Just let us know.

jdavis:
Finally it has been merged in. Thanks for your PR, @ab.

PR #42, opened 2014-03-17
Added Gandi (domain registrar)

matthewgall:
Added data for Gandi (https://gandi.net) a domain registrar and an image (credit: https://gandi.net/favicon.ico)


jdavis:
Nice work, pal!

PR #46, opened 2014-03-17
Added Cryptsy

Vanhouttethomas:
Cryptocurrency instead of just Bitcoin?

### DIFF
--- a/_data/main.yml
+++ b/_data/main.yml
@@ -343,7 +343,7 @@ sections:
 #
 
   - id: bitcoin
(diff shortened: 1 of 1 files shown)


bsdlp:
:+1: on cryptocurrency, as you'll likely see other exchanges that handle multiple cryptocurrencies instead of just bitcoin.

jdavis:
I added in Cryptsy's logo, but other than that, the suggestion was great.

PR #48, opened 2014-03-17
Add DreamHost entry

csuhta:
They support email confirmation and Google Auth


jdavis:
Sorry for not merging this, we ended up getting a different PR, #76  with a better icon. That guy is ginormous. Thanks a lot though!

PR #51, opened 2014-03-17
Add League of Legend and Guild Wars 2

csuhta:
GW2 supports Google Auth and email
League of Legends supports nothing :(


jdavis:
Nice work, champ!

PR #53, opened 2014-03-17
Added WordPress.com

Terrance:
WordPress.com supports it: http://en.support.wordpress.com/security/two-step-authentication/


jdavis:
<3's all around

PR #54, opened 2014-03-17
Added name.com

sudeepkodavati:
*Added Name.com 
*It uses VeriSign for tfa.


jdavis:
Awesome. Would you be able to grab the Name.com favicon to add? http://www.convertico.com/ works pretty well!

sudeepkodavati:
*Added the png for name.com.

jdavis:
I had to double check what the heck it was.. That is a strange 2FA solution.

jdavis:
Nice work though! Cheers!

PR #58, opened 2014-03-17
Addition of Communications area and Skype w/ Icon

STLCajun:
Recently had my account compromised and was told they offered no additional protection other than recommending strong passwords.


computmaxer:
:+1:

PR #60, opened 2014-03-17
Outlook should support Authy

Terrance:
Using it right now.


jdavis:
More love your way :heart:

PR #64, opened 2014-03-17
Added Humble Bundle

Terrance:
Humble Bundle supports it: http://support.humblebundle.com/customer/portal/articles/1412106


mathstuf:
FWIW, when I asked them about generic TOTP support, they said to keep an eye on their blog.

Issue #35, opened 2014-03-17
A way to represent cross-functional systems such as Google, Microsoft, Yahoo, etc


smarkwell:
The current category system doesn't really cover the breadth of larger cross functional authentication systems such as Google and Microsoft.  For example, should G+, Blogger and Gmail be separate systems or group in some other way?  As they all rely on Google's 2FA solution when enabled.

Same is true of Microsoft, but how that exactly works for each property is a little inconsistent.

victorhaggqvist:
That could maybe be achieved by a "Corporations" section or something of that nature. Then adding a possibility to bind services to a corporation.

jdavis:
I totally understand. I've been saying this to a lot of issues, but I need to think on it and let it sit in my head to see how to best represent it. I do think something needs to be done though.

jdavis:
We're on our way to being able to represent this in #208. I'm closing this in favor of that one. Thanks for the suggestion.

PR #65, opened 2014-03-17
Add Google Cloud Platform to developers section

icco:


mpdavis:
Is there any reason to have this listed separately than Gmail as it is the same account?  Perhaps combine them to just a Google account?  Otherwise the list would be littered with entries for Gmail, Google Cloud, G+, and every other service Google offers.

computmaxer:
I agree with @mpdavis on this one.

zach-taylor:
Yeah I think its being discussed on #35 how to deal with that kind of stuff.

jdavis:
I'm going to add it for right now as soon as I add a icon.

As I mentioned in #35, I'm going to think on it and see what I can come up with.

PR #68, opened 2014-03-17
tfa yes on spideroak

victorhaggqvist:
forgot set tfa to yes on spideroak, my bad :P


computmaxer:
Thanks for finding that!

PR #71, opened 2014-03-17
Digital Ocean call themselves DigitalOcean (without the space).

matthewgall:
Corrected this in _data/main.yml


jdavis:
Awesome. Thanks for the correction.

PR #73, opened 2014-03-17
Updated HootSuite and Guild Wars 2 icon to avoid awkward icon rendering ...

matthewgall:
...on website. I thought this would probably repair it. Pull request number three of the night!


jdavis:
Nice additions. The GW icon is way better.

PR #74, opened 2014-03-17
add chase sms auth

jfocht:
Chase does require two factor authentication for adding new devices. I wasn't able to find documentation of the feature on Chase's online security guide, but here is a blog post about it: http://fppad.com/2011/11/29/boost-your-online-security-with-two-factor-authentication/. I can confirm that I need two factor authentication whenever accessing chase from a new device.

### DIFF
--- a/_data/main.yml
+++ b/_data/main.yml
@@ -192,7 +192,8 @@ sections:
           url: https://www.chase.com/
           twitter: chase
           img: chase.png
-          tfa: No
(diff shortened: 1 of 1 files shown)


jdavis:
Awesome.

For some reason there are a few sites that don't have great documentation for it either and we've used blog posts. Still it's better than nothing.

jdavis:
Thanks Jordan =D

PR #75, opened 2014-03-17
Added More Blizzard custom options

RichJeanes:
Added Windows Phone market and Blackberry AppWorld links for the Blizzard Battle.net Mobile Authenticator app as found at https://us.battle.net/account/support/mobile-auth-download.html


mpdavis:
Looks good.  +1

PR #81, opened 2014-03-18
"Custom" column should not appear negative

### LINKED ISSUE #25: "Custom" field
The custom field currently displays a big red X if the service does not provide a custom 2fa solution. However, it shouldn't be considered a negative, as google auth, sms and authy support is Good Enough™.

I propose that the custom field be blank if N/A or otherwise more neutral.

computmaxer:
*Adding neutral class.
*Custom column is now "neutral" with a minus icon instead of red with an 'x'

**Example:**
neutral

fixes #25

### DIFF
--- a/css/base.css
+++ b/css/base.css
@@ -98,6 +98,19 @@ td img.icon {
     background-color: #f2dede !important;
 }
 
+.ui.table tr.neutral td,
+.ui.table td.neutral {
+    background-color: #ffffff !important;
+}
+
+.ui.celled.table tr.neutral:hover td,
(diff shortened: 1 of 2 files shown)


computmaxer:
@jdavis Here is the code for the custom column change.

jdavis:
I love the final result. I went with the [disabled](http://semantic-ui.com/collections/table.html) class built into Semantic which was basically the same result. The only difference is that the text is a bit lighter. None the less though, it wouldn't have looked that good without you <3

PR #93, opened 2014-03-18
Add wordpress custom 2FA plugin

ryanralph:


mpdavis:
I don't think that the wordpress.org plugin for 2FA belongs attached to wordpress.com, which is a completely different auth system of their own. 

The plugin is a way for wordpress sites to be able to offer 2FA and is never used to login to wordpress.com

ryanralph:
You're completely right though this does apply to all self hosted wordpress sites. Does this warrant it being listed?

mpdavis:
I'm not sure that it does.

All of us know that there are resources available to implement 2FA.  That is what makes it so sad that so few sites have implemented it.

I could see adding some section of the website pointing out ways for developers to implement 2FA that includes things such as the wordpress plugin, but simply adding it to Wordpress.com's row doesn't make sense.

ryanralph:
No worries

PR #94, opened 2014-03-18
Added Health section and websites

JimmyW7:


jdavis:
I really like the idea of the Health section, I just can't think of anything that would fit well. Possibly 23andme.com?

Would WebMD and Drugs be appropriate for having 2FA?

Also isn't HealthVault Microsoft? We are looking at how to handle companies with multiple services per account in #35.

JimmyW7:
I was looking for services that store health records. Those are the first 3 that came to mind, but there might be others. I would like to see sites like these supporting 2FA.

23andme.com is a good idea also.

jdavis:
I know of another one! Possibly [Healthcare.gov](https://www.healthcare.gov/)? I don't know how much info they have, but I feel it could be pretty important to have 2FA.

Do you think you could add 23andme and Healthcare.gov? If so, I'd merge this =]

zach-taylor:
We could use "doctor", "hospital", "medkit", or "ambulance" for the icon rather than "health" since it isn't an icon in Semantic.

jdavis:
Yeah, that'd be cool and I think either of those icons would work. I like the doctor and think it represents things well.

The more I think about it, the more I think we really should have a health section. There are other websites like http://www.myfitnesspal.com/, fitbit.com and the like that I feel people would like 2FA for because there's some info that they would want to protect.

jdavis:
Nevermind. The doctor icon is harder and harder to understand the more I look at it. Possibly the medkit or ambulance since there's a visible health cross on it.

JimmyW7:
Ok, have changed the icon to medkit and added a few other sites, with twitter accounts.

jdavis:
Awesome, Jimmy. We just need to add some icons and I think it'd be perfect to merge into master. I need to take a break and get some homework done, but if it isn't done by then, I'll do it myself as I want to see this get addded soon.

computmaxer:
@JimmyW7 This has a merge conflict due to the changes merged in #142. You'll want to move this new section's websites list to a `health.yml` file instead of `main.yml`

JimmyW7:
Reworked this as a lot had changed since this fork. See #158

Issue #84, opened 2014-03-18
Consider increasing the standard icon resolution from 16x16


computmaxer:
This would help save the eyes of high-DPI screen users like the Retina Macbook pro from seeing pixelated icons.

jdavis:
Oh poo, they look pixelated on retina displays?

jdavis:
I don't have a Retina display (pity me), but does 32x32 look fine? If so, we could go with that and add it to the README.

computmaxer:
So I just realized part of the problem is that a lot of the icons are 16x16 but are being displayed at 24x24.

From what I've read about retina support, you generally want to double the resolution.  So perhaps we want 48x48.

RichJeanes:
It looks like it's been settled in #103 at 32x32 and added to the README. Closing this for the sake of tidiness.

PR #103, opened 2014-03-18
Updated site icons

RichJeanes:
Updated any images larger than 24x24 to that size since that is what is displayed on the site anyway. Any images not in png were converted to match as well.

If you find this superfluous, feel free to close the PR without merging. Just a bit of me indulging in my OCD. Even made sure to keep the changes branched from my fork master.


computmaxer:
I agree that we should increase the standard icon size (see #84) but I think we should pick a size and stick with it instead of having a bunch of random sized icons.  Something like 32x32 should be fine.

jdavis:
I'm going to leave this open until we decide on what image size to go to. If we go to 32x32, this might make this commit obsolete but when it is decided, I'll revisit this and if you redid it for 32x32, that'd be awesome.

mpdavis:
I would agree on 32x32

jdavis:
32x32 it is! @RichJeanes If you could update them, that'd be great. However, if you can't, no worries. I'll merge it in for right now for temporary improvement.

PR #106, opened 2014-03-18
Adding Register4Less

smarkwell:
Adding domain register, Register4Less


jdavis:
I removed some whitespace, but other than that, it has been merged =]

PR #107, opened 2014-03-18
Added Network Solutions

joehallissey:
I should say that Network Solutions does have a quasi-TFA, but they charge for it:
https://www.networksolutions.com/blog/2014/01/weblock-security-program-information/


jdavis:
Awesome. I cleaned up the whitespace and added their Twitter handle: https://twitter.com/netsolcares

Great work.

PR #108, opened 2014-03-18
security

pixelpaulaus:
added security section and ssl certificate site


jdavis:
Can you think of other sites that might go under Security as well? I'd only want to add a new section if more than one can be in it. I'm going to switch SaaS to Other, so if you can't find another, it can go under there.

pixelpaulaus:
Yeah I can get a few together

pixelpaulaus:
hmm, looks like a lot of the security website don't actually have it. So maybe put this ssltrust.com.au under other for now. Would you like me to do the changes with a new pull request? or will you make the changes?

jdavis:
If you could do it, that'd be awesome. Much love :heart:

zach-taylor:
Some ideas: Geotrust, Verisign, GlobalSign, Comodo, DigiCert

jdavis:
So basically make Security be anything that enhances security on the web,
like SSL certs do. I approve.

pixelpaulaus:
i had a look at geotrust, symantec and comodo but none of them actually have two factor auth on their own user accounts. which is weird as some of them offer it as a product/solution. i'll have another look now.

victorhaggqvist:
https://www.startssl.com/, though they use cert authentication which is not tfa but a bit more secure than just plan user/pass login. Bu I guess that still goes by no tfa, or?

pixelpaulaus:
i have looked at soooo many security websites, but non so far offer it. even tho some have some access to your website files so some really should get it. Maybe if we get the security section up, some people will see it and then if they know of some security tool websites, they can add/suggest some?

zach-taylor:
@pixelpaulaus the goal is to display both sites with two-factor and ones without. Certificate sites like Geotrust and Comodo should definitely be added.

RichJeanes:
I feel like a "Security" section would be a good addition with sites like LastPass and Dashlane to add (Dashlane isn't mentioned in an issue yet but it's on my own list).

pixelpaulaus:
i am updating the list now

pixelpaulaus:
not to sure about LastPass and Dashlane as i have not used them but are they just desktops apps?

anyways i think we are good for now. i will add more later. and let me know if you want me to change anything.

RichJeanes:
While the contribution is appreciated, it looks like you're still working off the pre- #142 files. The source was reorganized into separate yaml files for each section. Try a pull and merge from jdavis/master.
In regards to LastPass and Dashlane, yeah they're desktop/web based password managers. LastPass is pending in #161 and Dashlane is on my to-do list so you don't have to worry about those.

PR #109, opened 2014-03-18
Update main.yml

cmcky:
Added Authy support for Digital Ocean


jdavis:
This and #122 are basically the same. I'm waiting to get confirmation that Authy works with the actual DO account, not just securing a box with SSH.

RichJeanes:
Resolved in #266

PR #110, opened 2014-03-18
adding OneDrive

alexiip:
Adding OneDrive to storage section.


alexiip:
Sorry for the tabs, should be fixed now.

jdavis:
This looks great, Alex. I took out the Custom section because we are using like this: https://github.com/jdavis/twofactorauth#custom

Thanks for fixing the tabs as well!

PR #116, opened 2014-03-18
Add Justcoin Exchange to the Bitcoin section

abrkn:


jdavis:
Great! The only thing that is missing is a link to documentation confirming it. I saw it mentioned in the [FAQ](https://justcoin.com/en/faq) but is there a better URL?

abrkn:
How about this? https://twitter.com/jstcoin/status/445900548743897088

jdavis:
That'll work for now. But for your user's sake, you should document it really, really well =]

I can add the link if you can't. Just let me know.

abrkn:
Link added. Our users are very tech savvy, but I guess we can always be more helpful.

links

jdavis:
Yeah, I totally understand, most users of bitcoin are. It was just a suggestion =]

Thanks for your contributions with this. It looks great.

PR #118, opened 2014-03-18
Add Skype TFA support

tathamoddie:
Skype usernames are being increasingly migrated to Microsoft Account based sign in instead. When using this newer authentication method, all of the Microsoft Account protections are used. (TOTP, SMS, recovery codes, single use passwords, app passwords, etc.)

### DIFF
--- a/_data/main.yml
+++ b/_data/main.yml
@@ -666,11 +666,15 @@ sections:
     icon: communication
 
     websites:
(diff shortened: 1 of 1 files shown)


jdavis:
Cool, thanks for the info. I changed it to "w/ Microsoft Account" to shorten it a bit.

Issue #30, opened 2014-03-17
"Google Auth" and "Authy" aren't descriptive or useful.


onethumb:
Both Google Authenticator and Authy implement  the same algorithm TOTP.  See:  http://en.wikipedia.org/wiki/Time-based_One-time_Password_Algorithm

The only difference is GA tends to use 6-digit tokens and Authy uses 8-digit tokens.  Same algorithm, same implementation.

More importantly, there are lot of other TOTP apps & services that work great (in some cases better) and are deployed at large organizations, like Duo:  https://www.duosecurity.com/

Many of the people using those other great 2FA services won't know what "Google Auth" or "Authy" mean, when in reality, those apps & services work just fine with all the things on this list.

Maybe a more appropriate heading would be "TOTP" and the values could be 6 digits, 8 digits, or 6-8 digits, depending on the service?

onethumb:
See Hacker News comments to the safe effect:  https://news.ycombinator.com/item?id=7414945

onethumb:
Given how important it is to keep it simple, on further reflection I'd just change it to "TOTP" and checkmarks, rather than getting into digit length.  It's not really relevant.

ilyakatz:
TOTP could be confusing to non-technical user, so If you do change to TOTP or something like that, I would suggest to link it to a description that provides links to google authenticator, authy, duo and others

ryankearney:
@ilyakatz Do you really think non-technical users are going to http://twofactorauth.org/ to check if their favorite website supports two-factor authentication?

ilyakatz:
@ryankearney 2-step authentication is becoming more and more common. As far as I know, banks in US are required it set it up now. So over time2-step will become a more common place phenomenon (unless something better comes along), so I think it is very reasonable to work on this project with assumption that it will be used by non-tech people

jdavis:
Hmm, I totally think the two should be merged into a single column if possible. I'm going to think on this to see if there is a good way to visually indicate that Authy/GA/others are all compatible.

Let me know if any of you guys have more ideas; I'd love to hear them.

mathstuf:
The problem with Authy-only sites is that the user never gets the key to use non-Authy applications (unless I missed something or I can't see that it's possible without having an Authy account), so compatibility is a one-way street.

[...]

PR #119, opened 2014-03-18
Rename Google Auth to something more accurate

tathamoddie:
Google Auth is a single, specific implementation, on a fixed number of platforms. Really, it's just an implementation of RFC6238 which other apps also support, like Microsoft's Authenticator: http://www.windowsphone.com/en-us/store/app/authenticator/e7994dbc-2336-4950-91ba-ca22d653759b


jdavis:
This is great and something we are thinking about. It was mentioned in #30 and we're addressing it there. Right now I think at the sacrifice of having two columns (even if they do the same thing), "Google Auth" or "Authy" makes more sense to people than TOTP right now, unfortunately.

PR #120, opened 2014-03-18
eBay & PayPal support Verisign Identity Protection

masterful:
PayPal (and, by extension, eBay), both support 2FA using the [Verisign Identity Protection](https://idprotect.verisign.com/mainmenu.v) app/token.

Admittedly, they could do a lot better in terms of documenting how to add it to your account. But the existing links in the repository point out that it is possible, I'm merely adding in the links present on the name.com entry (since they're the same app).

### DIFF
--- a/_data/main.yml
+++ b/_data/main.yml
@@ -123,6 +123,11 @@ sections:
           tfa: Yes
           sms: Yes
           doc: https://www.paypal.com/cgi-bin/webscr?cmd=xpt/Marketing_CommandDriven/securitycenter/PayPalSecurityKey-outside
+          custom:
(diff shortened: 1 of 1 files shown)


jdavis:
We might add Verisign Id Prot to the table as #37 notes. If so, we'll come back to this.

mpdavis:
Fixed in #37

masterful:
It's not actually fixed :(

But I'll submit a new pull request :)

Edit: It is fixed in the data, but the page has not been updated?

mpdavis:
Correct, it is fixed in master, but not out on the page yet.

PR #122, opened 2014-03-18
add authy to digital ocean

devinschulz:


jdavis:
Can I get a link to this or some sort of screenshot of configuration just to be sure?

mxxcon:
@iDevSchulz perhaps update your commit to use this url since it's more useful than just an announcement post.. https://www.digitalocean.com/community/articles/how-to-install-authy-and-configure-two-factor-authentication-for-ssh

jdavis:
I came across that but that's for securing your actual box SSH access with
Authy, not your actual accont on DO.

jdavis:
If I could get a confirmation that Authy works for the actual account, I'll merge this. In the meantime, I'm setting it to pending.

CameronBanga:
I personally use Authy with Digital Ocean, although I can't find any official blog post, etc. I'll look a bit later, but even if unofficial, DO does work with Authy currently.

computmaxer:
Added Authy support for DigitalOcean in #266

PR #123, opened 2014-03-18
Add Airbrake.io

tijmenb:
The people at @airbrake need some encouragement :)


phumpal:
@tijmenb we're a small nimble team but we now support 2FA.

PTAL https://github.com/2factorauth/twofactorauth/pull/3204

PR #125, opened 2014-03-18
Added directnic.com to config and added icon to img/domains

ralfonso-directnic:


ralfonso-directnic:
Yeah sorry about that, I have no idea where that came from, I was just as surprised.

PR #126, opened 2014-03-18
Update Outlook

LLBDdesign:
Added .com designation for Outlook to help differentiate between web service and Outlook client.


jdavis:
+1 :+1:

PR #134, opened 2014-03-18
Alphabetizing each section

mpdavis:
I had to go through and reorder each section to be alphabetical.  It was bugging me.


jdavis:
I tried to order them by what might interest people. And really, by that, I mean I put Social at the top and everything else was haphazard. But yeah, I think alphabetized makes more sense. I'll pull things in and do a manual merge to make sure things look good.

mpdavis:
This doesn't change the order of the sections, only the order of the items in each section.

victorhaggqvist:
It might be a better idea to do this programmatically instead, if it can be easily done, to reduce the maintenance work and help keep it up

jdavis:
Oh duh. I should have looked at the actual commits... Do you just mean that it bugged you while looking through the actual data? And not the actual website?

I added in sorting to the template. It sorted them by A-Z,a-z. I'd prefer case insensitive but it looks like sort can't do insensitive as I was using it here: https://github.com/jdavis/twofactorauth/blob/master/index.html#L70

mpdavis:
Ah, I have just been looking at main.yml too much.  I didn't realize it sorted them already.  Closing.

jdavis:
Ahh, well still, sorting in the yaml isn't bad either :P It makes it look a bit nicer. We can still merge it in.

PR #137, opened 2014-03-18
google drive and wallet

victorhaggqvist:


jdavis:
Darn, it looks like there's an indentation issue as the build status notes.

victorhaggqvist:
now it should be good, even lited it my self

RichJeanes:
I feel like this falls under the "Gmail" entry seeing as both Drive and Wallet use the same Google Account.

jdavis:
Yeah, I'm merging it for right now, but the idea is still churning as I discussed in #35

Issue #140, opened 2014-03-18
Fidelity


hypatia:
Apparently if you call up fidelity they will set up a Symantec VIP token for you: http://thefinancebuff.com/security-token-fidelity-schwab-etrade.html

Not sure you want to handle that since it's not publicly documented anywhere official, so I filed an issue rather than just adding it myself :)

hypatia:
Here's more documentation of how to do this: http://www.bogleheads.org/forum/viewtopic.php?f=2&t=108267

I've seen reports in some places that they will only do this if you have more than some $ value invested with them, which is pretty wack if true.

jdavis:
That's painful and sad. We could just add Fidelity to the site and maybe they'd get the message that it needs to be improved.

hypatia:
the twitters tell me they do support it kinda sorta: see "transaction security" on this page: https://www.fidelity.com/security/how-fidelity-keeps-you-safe
so it's not 2fa right on login, but close enough for me to feel better :)

warmlogic:
I actually just turned on Fidelity's two-factor authentication yesterday (by calling them), and I have to say that it works well. It really is 2FA on login, as the first link you posted says. You can use the VeriSign VIP Access phone app (iOS and Android), which is the same thing as the "key" icon link for e*trade on http://twofactorauth.org. PayPal and eBay use the VIP Access app as well. No idea about account balances...

hypatia:
@warmlogic so it sounds like if you want the login auth you have to call them, but the transaction auth anyone can get.

jdavis:
We addressed this in #148, thanks for pointing it out, @hypatia!

reillyeon:
The feature is also not available for workplace accounts but you have to call them to find that out.

PR #148, opened 2014-03-19
Add Fidelity Investments

gfbarros:


computmaxer:
This is related to #140.
@jdavis what do you think, should we merge this?

zach-taylor:
Looks good to me

computmaxer:
Reason I was questioning it is because it sounds like they have some kind of solution if you call them to turn it on and this PR shows that they do not support it at all.

gfbarros:
Unless they already offer it to everyone, shouldn't we push for that?

Pardon my grammar, sent from a mobile device.

jdavis:
Yeah, I think this is fine to merge. It's ridiculous to hide a security feature behind a phone call.

PR #153, opened 2014-03-19
Add FastMail to Email category

robn:
The second "custom" link is supposed to refer to our paper-based OTP option. I'm not really thrilled about the choice of icon, but I didn't see anything better. Suggestion: allow a `title` option to add an attribute to the link.


computmaxer:
Thanks! :+1:

ngocphamm:
I'm not sure if it's relevant but I just wonder whether the current "method" provided by Fastmail is actually 2-factor authentication or not. Fastmail, in fact, always allow the Master password and they just call the (assuming) 2FA "Alternative login".

> Note you can't disable your master password. Alternative logins are additional ways to login to your account on machines you believe may be less secure, or where you want to allow only restricted access.

robn:
Yes, it's still 2FA. Our recommendation is that you set a very complex master password and store it securely (eg in a safe), and never use it except for situations where you lose your regular login (eg stolen phone). Conceptually it's the same idea as Google's backup codes, for example.

ngocphamm:
I think I'm still confused with this. Basically this is what I understand about Google (and other) 2FA:

You have a master password, probably a _secure-enough_ one. For accessing from an _unauthorized computer_, you also need the token from Google Authenticator. This way, even if your master password is revealed to someone, they **won't** be able to login to your account unless they also have the token. Google's backup codes is only for the second part, the token. If the attacker only have one of the two pieces (master password + token), they won't be able to do anything.

In the case of Fastmail, once one has your master password, you are done! All the other methods for 2FA are for _alternative_ login, while the less secured master password still can be used for top-level privilege access.

Please correct me if I'm wrong. I'm pretty sure that I don't understand it enough. :confused:

PR #155, opened 2014-03-19
Adding Ally Bank, Copy, and Feedbin

zchrykng:
Adds Ally Bank, Copy, and Feedbin. All lacking two factor authentication.

There is a long standing thread on Ally's support site requesting it that has been basically ignored for a long time now.

---

Update: Removed Feedbin and news section.


computmaxer:
Thanks for the Ally Bank and Copy additions!

Should we really add a News section? Do those sites hold any confidential information?

zchrykng:
True. Was not sure of the scope we were shooting for. It contains the last four digits of your credit card number, but I think that is it.

PR #156, opened 2014-03-19
Add PNC Bank to finance (they lack tfa support).

linville:
This adds PNC Bank to the finance category (unfortunately they lack tfa support).


computmaxer:
Looks good, thanks :+1:

PR #157, opened 2014-03-19
Removing margin from the icons in the tables so that they are centered.

mpdavis:
There is a .25em right margin on icons that appears to be used to separate them from text that is placed next to them.  I removed this from the icons in the tables so that they are centered in the column.


mpdavis:
Before:



After:

jdavis:
I like your style.

PR #162, opened 2014-03-19
add hpcloud.com and launchpad.net and add VIP to PayPal and eBay

fallenpegasus:
add hpcloud (has two 2fa), launchpad.net (has google auth 2fa, but not well known), and PayPal and eBay (which use the "Verisign Validation & ID Protection (VIP)" token)


RichJeanes:
If you could change the png to just a 32x32 of the icon that would be fantastic!

RichJeanes:
Also, adding Verisign to top level status with Google and Authy is a work in progress right now  (#37), so it might just be easier to hold off until that is added.

RichJeanes:
Looks like you HP CLoud made it in in #284 and Ebay and PayPal have been updated. If you still want to add launchpad, go for another PR.

PR #173, opened 2014-03-19
Update domains.yml

stannie:
Add DNSmadeeady


computmaxer:
This was addressed in #191

Issue #172, opened 2014-03-19
Teamviewer.com supports 2fa


fatalicus:
Teamviewer.com supports 2fa with Google Auth, Authy and others: https://www.teamviewer.com/en/help/399-Welche-Apps-kann-ich-verwenden-um-einen-Sicherheitscode-f-uuml-r-die-Zwei-Faktor-Authentifizierung-zu-erstellen.aspx (no idea why they have that on a german link...)

i'm new to this whole Github thing, so if anyone could add it, that would be nice.

RichJeanes:
Being addressed in #177.

PR #177, opened 2014-03-19
new section remote control and add teamviewer

victorhaggqvist:
closing #172


jdavis:
Hey Victor, by remote control do you mean like VNC viewer type apps?

victorhaggqvist:
Yep, like LogMeIn, TeamViewer, VNC and RDP stuff. Maybe it could be "Remote Control & VPN" also.

RichJeanes:
@jdavis So do we want to add a "Remote Access" section or lump these into Developer for now?

jdavis:
I think Remote Access has enough usage that it could be another category.

jdavis:
I think the 'desktop' icon would be a good choice for it as well: http://semantic-ui.com/elements/icon.html

jdavis:
It looks like all of these are added. Sorry for not getting to your PR first.

PR #178, opened 2014-03-19
Added Dynadot domain registrar.

seisatsu:
I added the Dynadot domain registrar, which supports SMS and app-based two-factor authentication. Please review.


jdavis:
Yup, add an Apple icon with the URL pointing to the AppStore, then an Android icon pointing to the Play Store. @sudeepkodavati nice job.

sudeepkodavati:
Now we have Verisign enabled, we can delete the custom element and add verisign: Yes to pull request.

mpdavis:
I will merge and make the update to Verisign.

PR #184, opened 2014-03-19
Add USAA

bgaid:
USAA has tfa through Verisign VIP and Pin on mobile apps.


jdavis:
This looks great, @briangaid. My only question is regarding the Custom. Does the USAA app have a custom way to do 2FA? Also check out the changes that #37 added; VeriSign VIP is now available as a column.

bgaid:
Nice, I see the VeriSign VIP now. According to the security page on the website the Apps use a compination of the VIP system and a separate PIN code by the user. Is that considered custom?

jdavis:
Yup, I think so!

PR #188, opened 2014-03-19
Resorted developer entries and added PythonEverywhere service.

mxxcon:


zach-taylor:
Hi @mxxcon, it looks like it should PythonAnywhere rather than PythonEverywhere, is that right?

mxxcon:
@zach-taylor wow, i'm a dope. :D

mxxcon:
Why did it fail this commit on health.yml, a file that I didn't touch...?

mpdavis:
I believe there was a broken master for a short amount of time causing a few builds to fail.  PR has merge conflicts now, so after you fix them, another build should run and pass.

PR #192, opened 2014-03-19
Add Mailchimp

smholloway:
Mailchimp is an email marketing service provider; basically, they make it easy to email groups of people. They're quite progressive in terms of 2fa, which is cool.

The email-as-a-service didn't seem to fit nicely into the existing categories, so I offer Mailchimp as an Other.


RichJeanes:
If you could make that a 32x32 logo as per the README Guidelines, that'd be fantastic. Also, I'd suggest adding YubiKey (key icon) under custom as well.
@jdavis What do you think about this? They offer so many options, but I think linking to each app for each product might be a bit much. On the other hand, it doesn't look like the Toopher site has a link to download their app on it. And while Duo does have links on their site, there's four of them. Thoughts?

rublon-mw:
MailChimp is actually using their own solution, AlterEgo, which also serves as a "pass through" for other services like Google Authenticator and YubiKey: https://blog.mailchimp.com/alterego-now-works-with-google-authenticator-and-yubi-key/

I would just put a green check mark in the "Google Auth" column and link to http://kb.mailchimp.com/article/what-is-alterego in the "Custom" column.

smholloway:
I've updated the image dimensions. The README suggests that contributors use [Google's favicon service](https://www.google.com/s2/favicons?domain=http://example.com), but that failed me. I see several pull requests getting dinged for the same thing, so I think it's time to update the README. Either remove the favicon suggestion (ideally replacing it with a reliable source of 32x32 images) or explain what to do after downloading the image. The process is not so straightforward; for example, with this icon I first tried to expand the previous image, however, it was too low resolution, so I had to download MailChimp's assets and scale the image appropriately.

The larger issue is that an open source project is dictating how a brand's logo should be used. When a brand doesn't publish a 32x32 logo, the project contributors will have to decide what looks best, which could annoy the brand (see the [MailChimp design guidelines](http://mailchimp.com/about/brand-assets/). It's not fair for you guys to have to review design guidelines before accepting a pull request. I'd suggest a policy of deferring to images contributed by people who verifiably work for the image's company. Accept decent looking images on the initial request, but leave it up to brands to get things just right.

smholloway:
Looking at [the site](http://twofactorauth.org/) now, I am concerned about the proliferation of columns. I'd love to add Toopher, and I imagine Duo, Yubico, and various other 2fa providers would also like a fair shake. Where do you draw the line? There are tens of 2fa providers, but tens of columns would break the beautiful design of the site. 

I think the correct way forward is to have columns for more abstract ideas, like SMS, [OTP](http://en.wikipedia.org/wiki/One-time_password), and Custom. Custom would include homegrown solutions, Authy, Verisign, Toopher, Duo, Yubico, Clef, Authentify, StrongAuth, Rublon, DigiPASS, etc. Many of these providers can also handle OTPs, but I think it's wrong to check the box everywhere Google Authenticator is checked simply because the provider also implements [RFC 6238](http://tools.ietf.org/html/rfc6238). (Google Authenticator only handles OTPs as defined by [RFC 6238](http://tools.ietf.org/html/rfc6238), so calling Google Authenticator out specifically feels too specific to me.)

I will submit a pull request that creates a page dedicated to providers.

troyjfarrell:
@smholloway As you can see in issue #30, we're discussing a similar problem.  I'd like to see the Authy column removed, since it's a proprietary solution.

smholloway:
Closing in favor of PR #381 -- built around the new and improved format.

Issue #198, opened 2014-03-19
Box shows up 3 times


jdavis:
I think this might have to do with the lowercase stuff that @mpdavis did.

@mwww pointed it out.

mpdavis:
Yeah, this is due to how I am matching names to websites.  Liquid's string manipulation leaves something to be desired.  I'll see if I can figure something out.

jdavis:
Hahha, yup. Liquid is great, it just isn't great for the kind of voodoo that we are doing without using a real plugin.

PR #200, opened 2014-03-19
Quick fix for removing duplicate rows

mpdavis:
This is an easy quick fix for the duplicate rows, but it requires completely removing indentation from a portion of the template.

The reason for this is when creating the list of downcased website names when the template code is indented adds some weird whitespace that is not removed with the calls to strip or strip_newlines.

I was using a contains comparator to get around this, but that caused Box to match with Box, Dropbox and FrostBox.

Alternatively, we could ask Box to change their name to something more unique.


mpdavis:
Fixes #198 for now

jdavis:
I'll send a Tweet out to Box :wink:. In the meanwhile, let's fix this for the short term.

Issue #170, opened 2014-03-19
Paper codes based two-factor?


miohtama:
Would there be sense to add an option paper codes based for two-factor (HOTP: An HMAC-Based One-Time Password Algorithm)? E.g. you print / receive by mail a set of one-time codes which you need to enter for two-factor actions. Maybe an icon to custom column? 

All banks in Scandinavia use this (probably a requirement by regulation).

LocalBitcoins does this: https://localbitcoins.com/forums/#!/general-discussion#paper-codes-based-two-facto

This works everywhere (no smartphone needed) and is resilent against mobile malware attacks where your two-factor SMS is hijacked: http://blog.kaspersky.com/an-android-that-robbed-your-bank-account/

However I don't know how popular paper codes are around the globe.

Related Python project which does both Google Authenticator and paper codes based two-factor: https://github.com/miohtama/django-twofactor

RichJeanes:
I know paper codes are used by Google, but they're meant as backup codes for situations when you temporarily lose access to your authenticator, like when your phone dies, and other sites (so h as Dropbox) offer a single "recovery code" in case you lose your authenticator. I'd never actually heard of it as a primary source of OTPs.

jdavis:
Yeah, I agree with @RichJeanes. Paper codes aren't the primary mechanism for 2FA. They are for if  the 2FA device/phone stops working and access is still needed. Since this has to do with 2FA recovery options, I don't think we hae a place for it on 2FA.org.

Thanks for bringing it up though but I'm closing this for now.

kangasbros:
Only maybe 1/10th of the human population has access to smartphones, and SMS 2FA is expensive. Paper 2FA codes could be very potential innovation for large part of the world. I really don't like that you dismiss it like that.

jdavis:
Dismissing it? I'm not dismissing the idea, just the fact that it isn't being used now except for backup.

mpdavis:
@kangasbros Do you have any sort of evidence that paper 2FA codes are in use anywhere in the world?

thanksmister:
In Argentina we use a version of this for banking, they provide the codes on a plastic credit card.  It's used for online transactions, though not exactly the same I think it does demonstrate a low-tech solution.

kangasbros:
mpdavis, every bank in Finland uses paper based 2FA auth - ok, its a small country, but still.

Also our startup LocalBitcoins.com uses them, we have about 200k users (growing currently at a rate of about 600 new users daily).

[...]

PR #208, opened 2014-03-19
Add seperate page for 2fa providers, allowing for a cleaner homepage

smholloway:
This pull request adds a page dedicated to cataloging the numerous 2fa providers. This was mentioned in [PR 192](https://github.com/jdavis/twofactorauth/pull/192) as an alternative to having an ever-expanding list of providers with a column each. I think this is more scaleable and less political (because you don't have to decide who gets their own column).

From here, I would suggest that the main site generalize Google Authenticator to "OTP" or "RFC6238" then put all options (Google Authenticator, Authy, Verisign, Toopher, Duo, Authentify, Clef, etc) into the custom column.


tehroot:
I think it should stick to the most common 2fa providers, i.e., the ones that people are most likely to use, e.g., authy, google authenticator.

rublon-mw:
I like your suggestion concerning the simplification of the columns. TOTP (Time-Based One-Time Password), which falls under OTP (One-Time Password), is a standard that is being supported by several mobile apps (http://tools.ietf.org/html/rfc6238). SMS also is OTP.

I would suggest the following format:

Icon | Name | Docs | SMS (check mark) | TOTP (check mark) | Custom (icon)

This would allow us to showcase the two most popular standards (SMS and TOTP) and would also give us the possibility to include self-made solutions.

smholloway:
@TehRoot Do you have some data about the number of people using the various 2fa solutions? If you show every 2fa provider, you will have tens (soon hundreds) of columns, which is intractable. If you choose the top n--for example, three--providers, providers will rotate off the list, which requires a lot of maintenance--do you want to update every yaml file and scrub the now-fourth-place-formerly-third-place entry? If you simply decide based on your feelings, you shatter all hopes of objectivity and wreck the spirit of this site. Deciding who deserves a column is a draining political discussion that does not benefit the community. See for example, my reply, followed by your reply...

@mwww I like your proposed format. I think that scales better and captures all the information at an appropriate level of abstraction.

tehroot:
I like @mwww solution as well, but I feel (take this with a grain of salt, or whatever) that the people that would most benefit from a better understanding, should be directly linked to products that are easy to setup and use. e.g., Google Authenticator or Authy, or to setup SMS OTP.

There isn't a reason to get all weird and nonsensical over it, when the solution presented is elegant enough and conveys the best amount of information for 95% of users.

rublon-mw:
We already have the "Docs" column that links to documentation on how to set up 2FA on a particular site.

Another proposal:

Icon | Name | Docs | SMS (check mark) | TOTP (check mark) | Other (check mark) | Apps (icons)

SMS/TOTP/Other would indicate what kind of standard is being used. We would place the icons of supported mobile authenticator apps inside the Apps column. This would allow us to place links to mobile apps that support TOTP (e.g. Google Authenticator) directly inside a row of a site with TOTP support.

Example based on App.net:

Icon | App.net | Link | X | V | X | [Google Authenticator]

We would create a separate page on twofactorauth.org for each app that would have download links for each mobile platform. This would also allow us to create comparisons between apps in the future.

rublon-mw:
After getting familiar with #170 my proposal is:

Icon | Name | How-to | SMS (check mark) | TOTP (check mark) | Other (check mark) | Tool (icons)

Docs -> How-to
Apps -> Tool

This would allow us to include icons in the "Tool" column that would link to descriptions of:
a) mobile apps,
b) hardware devices (e.g. RSA token, YubiKey, mobile phone),
c) other tools (e.g. papers or cards with printed OTPs).

Edit:
In the "Tool" column, we would post links to the highest layer that is needed to make the 2FA work. A solution that supports SMS does not need a mobile app to work, so we can post a mobile phone icon with a link to a page that would describe how the SMS OTP 2FA solution works using a mobile phone. A site using TOTP would require the Google Authenticator app, so we would just post an icon of this app instead of a mobile phone (while you have to have a mobile phone, you also have to have the app, which is the highest layer).

smholloway:
I could get on board with that proposal @mwww. 

A couple questions:
- Does SMS get checked if you're using a third-party provider who handles SMS?
- There are a number of providers that can generate TOTPs (for example, the Toopher apps implement RFC6238, so it can be used in place of Google Authenticator); do you think each provider that implements the TOTP spec should be listed anytime TOTP is checked?

rublon-mw:
> Does SMS get checked if you're using a third-party provider who handles SMS?

Yes, since SMS is supported in such a case. It does not matter who physically sends the SMS.

> There are a number of providers that can generate TOTPs (for example, the Toopher apps implement RFC6238, so it can be used in place of Google Authenticator); do you think each provider that implements the TOTP spec should be listed anytime TOTP is checked?

There are indeed many TOTP app providers. I believe that TwoFactorAuth.org should not favor any provider in order to stay objective. We have to have rules that dictate when a provider should be added to the column. E.g. Google Authenticator is very popular, so it should be present whenever TOTP is checked and it supports this app. I believe that one provider is enough, so in such a case Google Authenticator should be the only icon that shows up, if it is supported. On the dedicated Google Authenticator page we should show alternatives (e.g. Authy and others). If TOTP is implemented using the Authy method, then only the Authy icon should show up. The same system should be used for any other method (e.g. Toopher; I didn't know that they also support RFC6238). This will result in a clear page without too many icons (= options). Let's keep it simple.

(This has been edited.)

smholloway:
Damn, I've got questions and you've got answers! I love it, thanks. That sounds really good to me. 

A site like [LastPass who supports a ton of 2fa providers](https://helpdesk.lastpass.com/security-options/multifactor-authentication-options/) would have an icon for each provider, right?

rublon-mw:
> A site like LastPass who supports a ton of 2fa providers would have an icon for each provider, right?

Exactly.

smholloway:
:thumbsup:

jdavis:
I told @mwww this but I'll have limited time until this weekend for these _big_ changes (small PRs I can do easily) but hope to get a lot done then. Hold tight and I'll revisit this issue soon.

jdavis:
Since the community that we have seems to love 2FA as much as I do, I want the next "version" of 2FA.org to be a joint effort. Each and everyone of you has some really great ideas (@mwww with the providers, @stefanwallin with the todolist, @lbrunsmann, @mxxcon with the international sections). Then of course our awesome maintainers: @RichJeanes, @mpdavis, @computmaxer, @zach-taylor 

I'm creating a new [Awesome 1.0 Status issue #320](https://github.com/jdavis/twofactorauth/issues/320). It has a todo list of the things to do as well as a link to where the code lives and the people working on it.

If you'd like to help, head on over there and just comment saying what you'd like to work on and where the code will live. I think this is going to be fun =]

mxxcon:
Maybe we can categorize it even more generically?
I can think of 5 distinct methods: `sms`, `email`, `phone call`, `hardware implementation`, `software implementation`. I see no real benefit in listing specific applications since you would have to go to the listed site's relevant page anyway and will see what needs to do to enable/active 2FA...

smholloway:
@mxxcon I like that. Do you have an example of someone doing 2fa via email? I'm not sure I understand how that would work.

I'm working on a PR to convert to more generic headers now.

RobFreiburger:
Most of the gaming section do email-based 2fa. You login, it sends you an email with a code, you put the code in where prompted.

smholloway:
Interesting. Sounds @mxxcon's proposal is a great start then.

In my new PR I'm reworking the main site to use these columns:

```
| site | docs | sms | email | phone call | hardware token | software implementation |
```

jdavis:
:+1:

smholloway:
See PR #333 and PR #334 for my proposed changes.

smholloway:
Merged with current master and refactored to match the style of the homepage. Here's what this pull request will add to `twofactorauth.org/providers/`

Imgur

If you're okay with the new page, I think this is ready to merge. Then providers can be added and adjusted.

jdavis:
Yeah, that looks awesome, Seth. I added an Authy icon and just merged it.

smholloway:
Thanks, you rock! (Apologies for forgetting the Authy icon.)

mxxcon:
@jdavis and @smholloway I'm a bit confused for the purpose of listing 2fa solutions?
Are we trying to accomplish something by listing them or just as an information resource?

jdavis:
I think everything 2FA is important. Being as we used to have a list of providers in the table, this is a better alternative.

As has been said before, educating people on 2FA is what we are aiming for. So I think having a section of different providers for 2FA fits perfectly within that goal.

Issue #204, opened 2014-03-19
SpiderOak does NOT support 2 Factor Authentication


sumocomputers:
SpiderOak does NOT support 2 Factor Authentication.

They rolled it out I think in 2011, and that is one of the reasons why I went with them.  They very soon revoked support for 2FA, and as of my communication with them on March 2014, they still have not reactivated it.  Shame really.

RobFreiburger:
They seem to only support 2FA in the US according to the blog comments.

sumocomputers:
I am in the US.  Here is a response I got from SpiderOak just a few days ago (note the part about 2FA being offline "...at the moment...".  It has actually been offline since 2011:

"This is Adam from SpiderOak support. Thanks for contacting us. 2-factor
authentication is still evolving. The developers are working to make this
accessible to all users (not just those in the US and Canada as at present) and
one of the solutions they are looking at is Google Authenticator. We haven't
been given a timeline when this will be implemented though. The 2-factor system
is offline at the moment, which is why you didn't receive a token. I would
suggest disabling 2-factor authentication for now so you don't get locked out
of your account. If you need help doing that please let me know.

We're very sorry for any inconvenience this has caused you.

Thanks,
Adam"

computmaxer:
This was fixed in #211

PR #211, opened 2014-03-19
Changes SpiderOak to no 2FA.

RobFreiburger:
Resolves issue #204.


mpdavis:
Travis marks this build as failed, but reading the log it looks like it was caused by another change that has already been fixed.  I'll merge it and watch it closely.

sumocomputers:
Not to be picky, but shouldn't Spideroak read "Tell them to support 2FA", rather than a bunch of red x marks?

PR #212, opened 2014-03-19
add TD Ameritrade. also update and reuse the TD Bank icon.

snarfed:
from what I can tell, TD Ameritrade supports the linked hardware OTP generator, but TD Bank doesn't. happy to be corrected if I'm wrong!


snarfed:
btw, out of curiosity, is there a way to generate the site's html? i was hoping to do that so i could see the end result of this change, but couldn't find how. did i miss it?

mpdavis:
The site runs using [jekyll](http://jekyllrb.com/), which you can install locally and use to serve the site locally in order to view your changes, assuming you have a Linux, Unix, or OSX based OS.

mpdavis:
Looks good.

PR #213, opened 2014-03-19
Paypal supports Verisign VIP Access

gboudreau:
Paypal supports VeriSign Identity Protection (VIP); it's not clear from the public documentation page, but here's what appears when you try to active 2FA: http://cl.ly/image/3g030A082F3r
I also simplified the link to the documentation; the previous one linked to a US-specific version, while this one should be generic.

### DIFF
--- a/_data/payments.yml
+++ b/_data/payments.yml
@@ -10,7 +10,8 @@ websites:
       url: https://www.paypal.com/
       tfa: Yes
       sms: Yes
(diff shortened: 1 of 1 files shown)


mpdavis:
@gboudreau Is there a reason that you changed the doc link?  The old one points to relevant info, while the new link requires me to be logged in to view.

gboudreau:
True. I was sure I tested in an anonymous tab to make sure! Looks like I didn't.
I was trying to remove the /us/ from the URL, since that could send non-US users to a page that could contain US-specific information.
I simply removed the /us/ from the URL now, and it works like before.
Thanks for the pointer.

mpdavis:
Awesome, thanks.

PR #218, opened 2014-03-19
Update other.yml

Ritchie007:
Added LastPass service.

Don't know how to had other methods of 2FA that lastpass supports.


jdavis:
#230 won out with this, worry Ritchie!

PR #220, opened 2014-03-19
Added Twilio

benjholla:


jdavis:
Bah, I didn't look close enough. Could you fix the doc URL and submit another request? The link you provided requires being logged in. Support docs are preferred.

benjholla:
Could I suggest adding a differentiation (another column perhaps) for links to where to enable it and docs describing it?

PR #221, opened 2014-03-19
Fixed Twilio Support Docs Link

benjholla:


benjholla:
Actually Icon size is wrong...I can't rtfm.  closing

PR #224, opened 2014-03-19
Adds Walmart to retail category.

RobFreiburger:


mpdavis:
That looks better, I will merge it in after a passing Travis build.

PR #232, opened 2014-03-19
Reduces wording on thanks Twitter badge.

RobFreiburger:
Badge felt long compared to the “Tell them to support 2FA” badge, so perhaps this is a good compromise.


jdavis:
... Beautiful!

PR #230, opened 2014-03-19
Add Capital One and LastPass

ghost:

PR #234, opened 2014-03-19
Add LastPass

Terrance:
NB: actually supports a lot more options, see https://helpdesk.lastpass.com/security-options/multifactor-authentication-options/.


jdavis:
#230 beat ya, sorry Ollie! Good work though!

PR #236, opened 2014-03-19
Various 32x32 icon updates

Terrance:
Updated some of the icons from 16x16 to 32x32.


jdavis:
You're a champ, Ollie. Great work.

PR #244, opened 2014-03-20
Updated tweet message to read better

jtheriault:
A minor change to the tweet message to read better, but which shouldn't blow past the 140 character limit.


computmaxer:
Thanks! :+1:

PR #246, opened 2014-03-20
Added Cloudshards and CrazyDomains

ramo-j:
Added Cloud shards to developers and crazydomains to domains.

It might be wise to create a separate section for hosting? Cloudshards would go there, as would things like AWS, rackspace, azure, and probably others.


computmaxer:
I like the idea of moving some of them into a hosting section. @jdavis what do you think?

jdavis:
@computmaxer I've been thinking about a Hosting section for awhile now. The hard part is that the dichotomy isn't exactly perfect. A lot of hosting sites offer domains and a lot of domain sites have hosting.

RichJeanes:
Bump.
Any thoughts on this? I will say, the Developer and Domain sections seem to be getting a little out of control, not to mention Finance.  I can't even get the whole Finance list in one screen anymore.

jdavis:
Yeah, we should break things up a bit.

PR #252, opened 2014-03-20
Todolist Application

StefanWallin:
Hi!

I just added a feature to the website. It's something I think will be useful. It's a todo manager that saves your progress in local storage.

The idea is this: A visitor comes along, realizes that 2FA is the shit and need to enroll. So first the visitor goes through them and creates a list of 2FA=yes-sites that he/she uses. Then goes through the docs and enrolls on each site, checking them off as we go.

Take it or leave it. It was a great evening hacking anyway.

### DIFF
--- a/css/base.css
+++ b/css/base.css
@@ -64,7 +64,8 @@ td.twitter {
     text-align: center;
 }
 
-td img.icon {
+td img.icon,
+.task img.icon {
     float: left;
     border-radius: 500em!important;
     padding: 4px 4px!important;
@@ -75,15 +76,17 @@ td img.icon {
     background-color: white;
     height: 16px;
 }
-
+.task img.icon {
+    margin-top: -3px !important;
+}
+.ui.celled.table tr.positive:hover td,
+.ui.celled.table tr:hover td.positive,
+.ui.table tr.positive:hover td,
 .ui.table tr.positive td,
 .ui.table td.positive {
     background-color: #dff0d8 !important;
 }
 
-.ui.celled.table tr.positive:hover td,
-.ui.celled.table tr:hover td.positive,
-.ui.table tr.positive:hover td,
 .ui.table td:hover.positive,
 .ui.table th:hover.positive {
     background-color: #dff0d8 !important;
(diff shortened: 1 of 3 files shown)


jdavis:
Wow, that's a neat idea, Stefan. I actually really like the idea.

Sorry for not looking at it sooner, I think the other collaborators and myself saw it and moved onto the shorter ones that didn't require the mental load that this one does (mental load in a good way, as in we have to think about this feature add)

I think it'd be need a bit tidying up and a bit more planning, but I actually wouldn't mind seeing this on the site. I have had various people tell me that they have turned on 2FA for sites over the last few days.

It might possibly be better on another page? Like next to Menu (it probably should be Categories), put like "..." (I can't think of what to call it now), and then instead of using a sidebar, use the whole page.

The great thing is that you are using localstorage so obviously it doesn't need anything else for how simple/basic the feature is.

What does everyone else think? @mpdavis @computmaxer @mwww @RichJeanes 

Also, to help out the other collaborators, it might help to screenshot it.

jdavis:
I'm going to add this to the Awesome 1.0 Milestone. Basically I want to keep the site simple but add an aggregation page for all the tweets and this. After that, I don't think it'd need anything but maintenance/updating.

StefanWallin:
Great that you like it, just a small evening hack. So I took some screenshots and noted them below:
1. Page loaded - New green button, clicking it.
   skarmavbild 2014-03-21 kl 06 58 33
2. Todolist loaded(if already created, it would load from localstorage
   skarmavbild 2014-03-21 kl 06 58 37
3. All lines change to a darker color on hover and gives a pointer cursor
   skarmavbild 2014-03-21 kl 06 58 42
4. Click adds it to the sidebar.
   skarmavbild 2014-03-21 kl 06 58 48
5. Can add more, note that they're hidden in the main column.
   skarmavbild 2014-03-21 kl 06 59 01
6. Added all under email to todolist, so entire section is now hidden.
   skarmavbild 2014-03-21 kl 06 59 06
7. I've completed 2FA setup on Facebook and Twitter(separate flow from implementation obviously) so they're lined.
   skarmavbild 2014-03-21 kl 07 00 03
8. And now I removed some that was added by mistake by clicking the X-button. The Email section appears again.
   skarmavbild 2014-03-21 kl 06 59 20
- Any action concluded will be placed on a undo queue and clear the redo queue.
- Any undid action(CTRL+Z || ⌘+Z) will be placed on a redo queue.
- Any redone action(CTRL+SHIFT+Z || ⌘+SHIFT+Z) will be placed on the undo queue.

PR #254, opened 2014-03-20
Ally Bank supports 2FA via SMS

reschly:
They call them 'Security Codes' in the linked FAQ.


jdavis:
#249 took care of this.

PR #260, opened 2014-03-20
Add NearlyFreeSpeech.net

joffotron:


jdavis:
NearlyFreeSpeech is already there in the Developer section because of the hosting they offer.

jdavis:
Thanks for the PR though! :heart:

PR #261, opened 2014-03-20
Added BMO Harris and enom

rk25862:
Added BMO Harris and enom


RichJeanes:
Looks like there's a merge conflict. Think you can sort that out? Also, 861bc58 seems to refer to a non-existent png.

PR #270, opened 2014-03-20
Adds First Direct to Finance

meatballs:


rublon-mw:
It looks like First Direct is or will be offering two-factor authentication:
http://www1.firstdirect.com/1/2/securekey
http://www1.firstdirect.com/1/2/securekey/options

meatballs:
Hmmm... they've been saying that for months. I reckon 'Not implemented' is closer to the truth than 'In Progress'

RichJeanes:
Adding the status link is appreciated. This site is about being open about security practices and info.

RichJeanes:
Interestingly, it looks like even when they roll out 2FA, it will be **optional** for login, but certain features will be disabled without it. Not sure how I feel about this or how it meshes with the goals of this site.
@jdavis?

jdavis:
It looks like features are only added with the usage of 2FA. I'm assuming that this is because they consider doing it without 2FA a security risk.

It seems reasonable as the features that are added all should require extra verification.

Unless I'm reading it wrongly, I think that seems okay?

meatballs:
I think it's as a result of the bad publicity the parent company, HSBC, got when they implemented TFA:

http://www.telegraph.co.uk/finance/personalfinance/consumertips/banking/8725302/Facebook-campaign-by-angry-HSBC-customers-over-new-online-security-key.html

Personally, I prefer HSBC's approach, but my fellow countrymen don't necessarily agree!

ilyakatz:
@meatballs if there are limitations on 2FA usage (eg, it only works for certain features). Could you please document this with a warning such as dfe1d07f72a7a06eed9d00de7f6a1c21ac897ca4, after that I think we'll be able to merge it in

RichJeanes:
@meatballs Do you care to pursue this? If yes, please pull your fork up to date and add the "exception" tag as noted. If not, this PR will be closed for housekeeping.

jamcat22:
This issue is superseded by #1544, and will be removed from the backlog.

PR #271, opened 2014-03-20
Adds Zen Internet to Domains

meatballs:


rublon-mw:
@meatballs, thank you for your effort. Could you please add a link to official documentation on how to activate Zen Internet two-factor authentication and recommit? I couldn't find any.

meatballs:
They don't have TFA. That's why I've added them with tfa: No

rublon-mw:
True, sorry for that :-)

PR #286, opened 2014-03-20
Adding Trillian IM, does not have TFA

smarkwell:


jdavis:
Hey @smarkwell. Could you upgrade this to a 32x32 icon? If so, that'd be perfect.

computmaxer:
Does Trillian have their own service? AFAIK it is just a chat client to connect to other services like Google Chat/XMPP, IRC, etc. I'm not sure there is anything Trillian could do about those other services not supporting two-factor.

smarkwell:
Are we upscaling in any particular way?  16x16 was the source of their favicon.

mpdavis:
I would also argue that Trillian IM doesn't belong in the list.  They have no authority over 2FA for the people that use their software.

jdavis:
@mpdavis and @computmaxer, keepin' my trigger happy pull requesting normalized.

I agree with them after looking into it. It doesn't look like it has its own service.

smarkwell:
Oddly enough Trillian has both, they have both their own network and provide proxy interconnect to other chat services (AIM/ICQ/XMPP/etc).  They run their own services and store your credentials in their servers for chat services.  I'm ok not including them either way.

jdavis:
K, let's not include them for now. If we find a compelling reason to, we can reconsider it.

PR #287, opened 2014-03-20
Added Bitcasa

dy-dx:


RichJeanes:
Looks good. Think you could find a support article to add for the doc link?

dy-dx:
@RichJeanes I thought a doc link is only for if they support 2FA?

RichJeanes:
Haha Good point there. Sorry about that.
Good work :thumbsup:

PR #288, opened 2014-03-20
Add Yodlee MoneyCenter financial aggregation site

marvingreenberg:
Add Yodlee Labs - Beta/incubator version of MoneyCenter


ilyakatz:
I would say this isn't exactly true, Yodlee does asks for answers to verification questions when you login from a new device

marvingreenberg:
So does vanguard and some of the other financial web sites. But it isn't 2FA since it isn't requiring me to HAVE something (like my phone or a key fob) in addition to knowing something.

ilyakatz:
@marvingreenberg  cool, thanks

RichJeanes:
So these sites have two independent account systems even though one was spawned from the other?

marvingreenberg:
The newer yodleemoneycenter.com needs a new account set up - I think yodlee is trying to redesign from ground up, but kept the older renamed site (thankfully) for existing users.

ilyakatz:
@marvingreenberg this PR is not mergeable, would you mind updating it and we'll merge it in?

jdavis:
Hmm, is there a way to just have one of these instead of both sites? If yodleemoneycenter.com is the newer one, could we just have that one?

jdavis:
Thanks for that update, @marvingreenberg. I just noticed it now.

PR #292, opened 2014-03-20
Added CIBC to Finance

strattondev:
Don't edit on the master branch >.>
Also 3rd times the charm?


jdavis:
Oh, don't worry about editing on the master branch. It's your fork so it isn't a big deal.

PR #297, opened 2014-03-20
add Boeing Employee Credit Union

fallenpegasus:
Boeing Employee Credit Union is the largest credit union in the State of Washington in the US.

They do not yet have 2FA security for their online banking.


RichJeanes:
Looks like that link doesn't work. Seems like it needs to be www.becu.org, not .com.
Also, they don't have their SSL certificates signed for becu.org, only www.becu.org. Not relevant to this, but I find it funny. haha

RichJeanes:
GitHub is also complaining about a merge conflict. So if you can fix that and update the URL, this would be a welcome addition.

jdavis:
Thanks, Mark!

PR #298, opened 2014-03-20
fix healthvault

ilyakatz:
MS HealthVault info was incorrect, it uses MS account so has 2fa


RichJeanes:
Verified HealthVault uses MS Live login. Fair enough. Thanks for the correction! :thumbsup:

PR #300, opened 2014-03-20
add CAcert.org 

fallenpegasus:
CAcert.org is a community-driven Certificate Authority that issues certificates to the public at large for free.

They do not have 2FA login into their website at this time.


RichJeanes:
Hmm, that's an interesting site. Hadn't heard of them before. Though it seems they haven't actually gained a foothold yet. Chrome warned me that their root certificate was not trusted meaning that they haven't been accepted as a trusted cert signing authority yet, but still interesting.
But everything on the commit looks good! :thumbsup:

PR #302, opened 2014-03-21
Removed fixed width from container

mslepko:
More mobile friendly look

### DIFF
--- a/css/base.css
+++ b/css/base.css
@@ -17,7 +17,8 @@ body {
 }
 
 .container {
-    width: 840px;
+    max-width: 840px;
+    width: 100%;
     margin: 0 auto;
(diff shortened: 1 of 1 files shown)


jdavis:
For ease of merging and historical purposes, could you post a screenshot of before and after? I think mobile is important and would love to see what it looks like.

mslepko:
It's not perfect but I think it look a lot better.

Current site
before

After my change
after

computmaxer:
Nice. :+1: Does Semantic UI have any kind of responsive support like Bootstrap?

jdavis:
Yes, it does a bit. I actually ripped it out because of how janky it was. It rendered the tables completely unreadable.

PR #304, opened 2014-03-21
Added Barebones SSO Server/Client

cubiclesoft:
The Barebones SSO Server/Client is a self-hosted, standalone solution written in PHP that has various forms of two-factor authentication built-in, but the 2FA features are optional.  The Generic Login provider has Google Authenticator and e-mail 2FA support.

http://barebonescms.com/documentation/sso/


jdavis:
Hmm, it's self hosted? I am a bit hesitant to throw it onto the list because self-hosted implies a lot of technical background knowledge.

Since that is more than likely the case, I think it would be most likely that the person setting up the self-hosted solution would be comfortable with 2FA and know if/when they want to use it.

cubiclesoft:
What is the purpose of the website?  You've got links to "tell this business to support 2FA" but there's nothing on the site to say to those developers who show up as a result of a Tweet, "Hey, you want to implement 2FA, that's great!  Here's a simple set of resources to get you started..."  Part of those resources might be software solutions that replace their entire sign in system with something more up-to-date that has the support built-in (e.g. the Barebones SSO server/client).  And the other part of the resources should be educational materials to get up to speed on how to implement 2FA themselves.  I was trying to work within the system you've already built before rocking the boat too much.  :)

Of the 2FA options you have in the list, Google Authenticator is the only free option.  I'm not sure why it isn't listed before SMS.  Implementing Google Authenticator has absolutely nothing to do with Google - it is just the name of an app that implements the HOTP/TOTP RFCs.  Microsoft has a similarly named app for Windows phones that functions identically to Google Authenticator.  However, developers not familiar with 2FA are going to go through a LOT of research before discovering this fact because it isn't obvious.  The name of the app implies being related to Google accounts (and their app does request the "Full network access" and various "account" privileges - not sure why they need that).  When developers think 2FA, they tend to think SMS messaging or physical tokens (e.g. SecurID).  Google Authenticator is not the norm among users despite being relatively easy to implement and uses software available for free on smartphones and tablets.

SMS sounds cool and does target a few more users, but the cheapest SMS solution I know of - short of rolling a service with a GSM modem - is Twilio and that still costs money per SMS.  SMS is fine for organizations that can eat the cost.

By comparison, Authy gets pricey pretty fast and I'm not even sure what the pricing structure for Verisign VIP is, which is a tad frightening.  Authy + SSH is downright scary since they can hold an entire business hostage until it agrees to pay for the next tier.

smholloway:
See PR #208. I think SSOs fit better on a dedicated providers page.

cubiclesoft:
Okay.  Thanks for keeping me updated on this.

jdavis:
We've added a Providers page to 2fa.org. So if you think it'd work there, feel free to make a pull request so we can review it.

I'm closing this for now. :smile:

Issue #241, opened 2014-03-20
Creation of International Sections


lbrunsmann:
This website is getting a lot of attention in Australia. I have also noticed that a company from Brazil is wanting to be added. In order to curb confusion, some sort of internationalization needs to be created.

mxxcon:
What's wrong with adding sites to the current list? Why would there be confusion? Is there ever a situation where some sites offers tfa in one country but not another?

lbrunsmann:
@mxxcon issue #166 address this. Citibank in Australia supports 2FA. Possibly a way to differentiate an Australian company to an American one. Different country laws have different security requirements.

RichJeanes:
@jdavis This seems like something that could use some consideration. Maybe putting flags on the ones that are country specific (eg. US flag on PayPal and Apple as per #272)?

rublon-mw:
In such cases I would just add an "alert" icon next to a check mark (under SMS and other columns; see #208). Clicking on this icon would show us the restrictions that the site implemented on each method. The user could get informed that only specific countries are supported or that 2FA is only available for customers of certain paid service plans.

In most implementations, 2FA will be available internationally and across all service plans. This is why we should not introduce drastic changes to the template.

Alert icon illustration (example):
http://premium.wpmudev.org/blog/wp-content/uploads/2012/08/wordpress-malware-alert.jpg

jdavis:
I like the idea of an international section and like what @mwww proposes with the icon. Instead of putting it on the main page (things are getting crowded), why not make a separate international.html page that has all the exceptions and country specific solutions that offer 2FA. We can leave the homepage to be global ones.

So in the case of like #272, put a little [warning icon](http://semantic-ui.com/elements/icon.html) next to companies that dont' work in specific countries and/or exceptions.

How does that sound to everyone?

RichJeanes:
Sounds good to me.

rublon-mw:
Showing only the global ones on the homepage sounds good. What do we do with sites that only offer 2FA for paid accounts?

mxxcon:
@mwww grr. Why are you complicating things? :)
I say we shame them as no 2FA. It's a despicable company if it does not value their users' safety and privacy, even on free accounts.

rublon-mw:
I'm just trying to find an optimal solution :-) The real world is complicated. Our mission should be to catalog the 2FA landscape and to present it in the most user-friendly way.

[...]

PR #305, opened 2014-03-21
Adds Halifax Bank to Finance

meatballs:
Halifax Bank has no 2FA


meatballs:
Why the international label? #241 seems to apply to services where the provision of TFA differs from one country to another.

To my mind, that is a different situation to a service only being available in some countries but not others. For this site, as long as the TFA provision is consistent across all countries where the service is provided, surely #241 does not apply (and that seems to be the thinking with the American Banks on the site so far).

PR #312, opened 2014-03-22
Sorted payments.yml and added Square.

mxxcon:
Square is that smartphone credit card swipe thingie.


jdavis:
For some reason the icon doesn't load for me...

mxxcon:
@jdavis seems like usual github glitches :/
That image is here https://github.com/mxxcon/twofactorauth/blob/Square/img/payments/squareup.png

jdavis:
Ahh, okay. Strange.

jdavis:
It loads now too, btw. Thanks for this @mxxcon. You're pretty awesome and very close to becoming an official Collaborator =]

PR #314, opened 2014-03-22
Namespro

nicwaller:
NamesPro is a great Canadian registrar, but they'd be even better with 2FA.


jdavis:
Is it only for .ca domains?

nicwaller:
Nope, they do all common TLD's. But for .ca they are the best in my opinion.--
Nic Waller

PR #319, opened 2014-03-23
add citibusiness payroll.

rajiv:


ilyakatz:
@rajiv when i click on one of the two links to Sign in, i'm taking to citibank.com so is this covered by the citibank entry?

rajiv:
it seems that the actual sign in link is not on the citibusiness payroll homepage. citibusiness payroll has a limited audience, and i'm no longer using it, so i'm closing this pr.

PR #323, opened 2014-03-23
Add education section, with Edline and Blackboard to start.

davidlohle:
Edline and Blackboard are portals where students, parents, and teachers have access to a student's grades and progress during the school year. Two factor authentication is desperately needed in this area, as access to an account means an access to a student's very private information, ranging from their grades, schedules, and conversations between teachers.

If an attacker got hold of a teacher's account, ruining student's data is only clicks away as well, including grades and attendance.


jdavis:
:+1: for this. Sorry for the delay in addressing this. We have switched the layout a bit. The README is up to date with what to expect, if you could update this to that, I'll pull this in.

RichJeanes:
Hmm... I don't know how I feel about putting Blackboard on this list. I know Bb uses individual deployments to customize the product on a per-school basis, including authentication system. We use Blackboard at my school (Northern Arizona University) and it's integrated with our own login system (Central Authentication Service, CAS) so it is up to NAU to implement 2FA (though to their credit, they are actually working on it). Does Bb provide login functionality for schools that don't have their own system? Even then, it seems like it becomes a case-by-case issue.

jdavis:
Yeah, you're right, Rich. Here at ISU, it's the exact same way.

jdavis:
I'm closing this in the meantime because of the comment that @RichJeanes brought up. Each University has to have their own instance of BB. So it is a bit tricky to say "BB has 2FA" when some schools might be running an older version.

PR #324, opened 2014-03-23
Added Banco do Brasil to finance.yml

tiagobugarin:
Banco do Brasil is one of the most popular banks in the country (Brazil) and until this day it doesn't support 2FA. :/


jdavis:
We're working on organizing/adding an international section; hold tight and we'll add this there when the time is right!

Carlgo11:
@tiagobugarin if you still think this is relevant we'd be more than happy to add 'Banco do Brasil' to the list.
But before we can do that I'd need you to update

Carlgo11:
Closing since the merge branch is deleted

PR #326, opened 2014-03-23
Dreamhost do support Authy

tiagobugarin:
I am using Authy to deal with Dreamhost 2FA w/o a problem. Made the change to reflect that.


jdavis:
=D Thanks!

Issue #320, opened 2014-03-23
Awesome 1.0 Status


jdavis:
Overall status for Awesome 1.0 and who is working on it:
- [ ] [Search Bot](https://github.com/jdavis/twofactorauth/tree/bot) (@jdavis, ...) #269 
- [ ] [Stream Bot](https://github.com/jdavis/twofactorauth/tree/bot) (@jdavis, ...) #268 
- [ ] International Sections (@lbrunsmann, ...) #241 
- [x] Cleaner 2FA Providers (@smholloway, ...) #208

smholloway:
PR #208 should work towards the `Cleaner 2FA Providers` todo. Did you have something else in mind?

If there's not one already, I'll work on a pull request to rework the main page with the columns that @mwww and I discussed (moving all 2fa providers into the custom column and keeping the columns more abstract, like SMS and TOTP).

jdavis:
Awesome, I added you both.

jdavis:
Does anyone think this is too overkill for getting the message across? xD

![screen shot 2014-03-23 at 2 00 59 pm](https://f.cloud.github.com/assets/157801/2494259/a3bdba52-b2bd-11e3-95da-3e28928ad797.png)

StefanWallin:
I love that tweetview! It might be a bit overkill, but I like it!

jeffgodwyll:
@jdavis What's the todo list about?

rublon-mw:
@jdavis If you could display profile photos instead of names, I think that such a page could look better.

jdavis:
@jeffgodwyll Back in #252 @StefanWallin had the idea of allowing people to make a list of the services they are using and track it while they add 2FA. I thought it was a really neat idea.

Lately though I've been thinking about if that's something we want to add though. While I like the idea, we might not want to add it to keep the site simple.

Also, @mwww, I totally agree that images would be great. I was showing profile images but the browser trying to load hundreds at once was pretty intense (naive implementation). If we do add profile images, we'll have to do some JavaScript stuff to lazily load them. It's definitely something I want to look into though.

jdavis:
I'm going to take the todo list off of this version. We can add it later if people agree on it.

Still though, @StefanWallin, I like the idea, don't get me wrong =]

PR #333, opened 2014-03-24
Rework site to show the two-factor usage method

smholloway:
This pull request changes the columns from 

```
Site | Docs | SMS | Google Auth | Authy | VeriSign VIP | Custom
```

to

```
Site | Docs | SMS | Phone Call | Email | Hardware Token | Software Implementation
```

Imgur

See [PR #320](https://github.com/jdavis/twofactorauth/issues/320) and [PR #208](https://github.com/jdavis/twofactorauth/issues/208) for more context about these changes.

I updated every `.yaml` file; anywhere there was a `Yes` for `authy` or `goog` or `verisign` there should now simply be `software: Yes`. For `custom` entries I added the appropriate `hardware` or `software` entry. I visited every `doc` link and tried to capture the data, but I welcome the community to fix anything I missed. I also removed stray white space and capitalized `Yes` and `No` where appropriate.

I think it turned out well.

---

Now that I've done all this work, I have a new idea that might be cleaner:

```
Site | Docs | Implementation
```

We then place the appropriate icons for each of the five categories. Perhaps

```
SMS: mobile icon
Phone Call: phone icon
Email: mail icon
Hardware Token: lock icon
Software Implementation: shield icon
```

Here's what that might look like:

Imgur

See PR #334 for the code.

### DIFF
--- a/_data/email.yml
+++ b/_data/email.yml
@@ -3,7 +3,7 @@ websites:
       url: https://www.fastmail.fm/
       img: fastmail.png
       tfa: Yes
-      sms: No
+      sms: Yes
       software: Yes
       doc: https://www.fastmail.fm/help/features_alternative_logins.html
 
--- a/_data/email.yml
+++ b/_data/email.yml
@@ -3,40 +3,31 @@ websites:
       url: https://www.fastmail.fm/
       img: fastmail.png
       tfa: Yes
-      goog: Yes
-      sms: Yes
+      sms: No
+      software: Yes
       doc: https://www.fastmail.fm/help/features_alternative_logins.html
-      custom:
-          - icon: key
-            url: https://www.fastmail.fm/help/login_yubikey.html
-          - icon: text file outline
-            url: https://www.fastmail.fm/help/features_alternative_logins.html#idp2602704
 
     - name: Gmail
       url: https://gmail.com
       img: gmail.png
       tfa: Yes
-      goog: Yes
-      authy: Yes
       sms: Yes
+      software: Yes
       doc: https://support.google.com/accounts/answer/180744?hl=en
 
     - name: Outlook.com
       url: https://outlook.com
       img: outlook.png
       tfa: Yes
-      goog: Yes
-      authy: Yes
       sms: Yes
+      software: Yes
       doc: http://windows.microsoft.com/en-us/windows/two-step-verification-faq
 
     - name: Pobox
       url: https://www.pobox.com/
       img: pobox.png
       tfa: Yes
-      goog: Yes
-      authy: Yes
-      sms: No
+      software: Yes
       doc: https://pobox.com/helpspot/index.php?pg=kb.chapter&id=65
 
     - name: Yahoo Mail
@@ -50,7 +41,6 @@ websites:
       url: https://www.zoho.com/mail/
       img: zoho.png
       tfa: Yes
-      goog: Yes
-      authy: No
       sms: Yes
+      software: Yes
       doc: https://adminconsole.wiki.zoho.com/mail/Two-Factor-Authentication.html
--- a/_config.yml
+++ b/_config.yml
@@ -6,8 +6,8 @@ markdown: liquid
 title: Two Factor Auth List
 author: Josh Davis
 description:
-    List of sites with Two Factor Auth support which includes SMS, Google
-    Authenticator, and other methods.
+    List of sites with Two Factor Auth support which includes SMS, email,
+    phone calls, hardware, and software.
 url: http://twofactorauth.org
 repo: https://github.com/jdavis/twofactorauth
 img: img/logo.png
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -17,7 +17,7 @@ websites:
     img: copy.png
     tfa: No
     twitter: copyapp
-    
+
   - name: CrashPlan
     url: http://www.crashplan.com
     img: crashplan.png
@@ -29,8 +29,7 @@ websites:
     url: https://www.dropbox.com
     tfa: Yes
     sms: Yes
-    goog: Yes
-    authy: Yes
+    software: Yes
     doc: https://www.dropbox.com/help/363/en
 
   - name: Evernote
@@ -38,29 +37,20 @@ websites:
     url: https://evernote.com
     tfa: Yes
     sms: Yes
-    goog: Yes
-    authy: Yes
+    software: Yes
     doc: http://blog.evernote.com/blog/2013/10/04/two-step-verification-available-to-all-users/
 
   - name: Frostbox
     url: http://frostbox.com
     img: frostbox.png
     tfa: Yes
-    sms: No
-    authy: No
-    goog: No
-    custom:
-      - icon: shield
-        url: "https://rublon.com/"
-      - icon: text file outline
-        url: "https://app.frostbox.com/main.php?action=2fa"
+    software: Yes
 
   - name: OneDrive
     img: onedrive.png
     url: https://www.onedrive.com
     tfa: Yes
-    goog: Yes
-    authy: Yes
+    software: Yes
     sms: Yes
     doc: http://windows.microsoft.com/en-us/windows/two-step-verification-faq
 
@@ -69,8 +59,6 @@ websites:
     url: https://spideroak.com
     tfa: Yes
     sms: Yes
-    goog: No
-    authy: No
     doc: https://spideroak.com/blog/20110620235134-2-factor-authentication-to-your-spideroak-account
 
   - name: SugarSync
@@ -83,7 +71,6 @@ websites:
     url: https://drive.google.com
     img: drive.png
     tfa: Yes
-    goog: Yes
-    authy: Yes
     sms: Yes
+    software: Yes
     doc: https://support.google.com/accounts/answer/180744?hl=en
--- a/_data/bitcoin.yml
+++ b/_data/bitcoin.yml
@@ -4,7 +4,7 @@ websites:
       url: https://bips.me/
       tfa: Yes
       sms: No
-      email: Yes
+      software: Yes
       doc: https://bips.me/security
 
     - name: Bitstamp
@@ -21,6 +21,7 @@ websites:
       tfa: Yes
       sms: Yes
       software: Yes
+      hardware: Yes
       doc: https://blockchain.info/wallet/support-pages
 
     - name: Coinbase
--- a/_data/bitcoin.yml
+++ b/_data/bitcoin.yml
@@ -4,20 +4,15 @@ websites:
       url: https://bips.me/
       tfa: Yes
       sms: No
-      goog: Yes
-      authy: No
+      email: Yes
       doc: https://bips.me/security
-      custom:
-          - icon: key
-            url: https://bips.me/security
 
     - name: Bitstamp
       img: bitstamp.png
       url: https://bitstamp.net
       tfa: Yes
       sms: No
-      goog: Yes
-      authy: Yes
+      software: Yes
       doc: https://www.bitstamp.net/s/documents/bitstamp_2_factor_authentication_guide.pdf
 
     - name: Blockchain
@@ -25,19 +20,15 @@ websites:
       url: https://blockchain.info/
       tfa: Yes
       sms: Yes
-      goog: Yes
+      software: Yes
       doc: https://blockchain.info/wallet/support-pages
-      custom:
-          - icon: key
-            url: https://blockchain.info/wallet/yubikey
 
     - name: Coinbase
       url: https://coinbase.com/
       img: coinbase.png
       tfa: Yes
       sms: Yes
-      goog: Yes
-      authy: Yes
+      software: Yes
       doc: http://blog.coinbase.com/post/25677574019/coinbase-now-offers-two-factor-authentication
 
     - name: Cryptsy
@@ -45,8 +36,7 @@ websites:
       img: cryptsy.png
       tfa: Yes
       sms: No
-      goog: Yes
-      authy: Yes
+      software: Yes
       doc: https://www.cryptsy.com/pages/security
 
     - name: LocalBitcoins
@@ -54,8 +44,7 @@ websites:
       img: localbitcoins.png
       tfa: Yes
       sms: No
-      goog: Yes
-      authy: No
+      software: Yes
       doc: http://localbitcoins.blogspot.fi/2013/05/using-two-factor-authentication-on.html
 
     - name: Justcoin Exchange
@@ -63,8 +52,7 @@ websites:
       img: justcoinexchange.png
       tfa: Yes
       sms: No
-      goog: Yes
-      authy: No
+      software: Yes
       doc: https://twitter.com/jstcoin/status/445900548743897088
 
     - name: WeMineLTC.com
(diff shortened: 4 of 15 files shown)


jdavis:
Cool. I really like the first one because it still provides a tonnn of information yet it is easy to read. The problem with the icons is that it doesn't give a good way to compare like the first one does.

jdavis:
Also, I don't know if this is overkill, but what if there was a way to see which hardware and which software solutions are offered either through a hover over or similar feature?

mxxcon:
@jdavis But is "comparing" a relevant problem that we are solving here? Do people come to the site to _compare_ implementations of different sites or do they come to _find out_ if a service they use has 2FA? Do you have any user feedback or insight on that?

smholloway:
If we want to show the solutions offered, I think the custom column was a good guide. After integrating PR #208 we could show the provider icon and link to the provider url. The code might look something like this:

```
{% for item in website.software %}
  {% if item.url %}
    <a href="{{ item.url }}">
      <img src="/img/providers/{{ item.name }}.png" class="icon" alt="{{ item.name }}">
    </a>
  {% else %}
    <i class="lock small icon"></i>
  {% endif %}
{% endfor %}
```

My concern with the first design is that it could foster competition. When looking at the list, I have to wonder if the best site is the one with the most checkmarks. If other people think that way, then the site could incentivize companies to check every box, and, while it might be cool to have more sites offering 2fa via SMS, email, voice calls, hardware tokens, and custom software, I'm not sure the point of the site is to encourage companies to support more modes of 2fa. 

Actually, I think that logic would apply to any design that could show multiple icons per row. The only way to avoid the feature-list-checkmark mania is to simply say, "Yes, this site supports 2fa" or "Tell them to support 2fa". Is that better? Is the additional information useful? I'd have to see how people are actually using the site to know if I think this is worthwhile. My hunch is that people will scan the list and find a site they use, then click into the docs to figure out how to set 2fa up on that site.

Edit: @mxxcon said it more succinctly and more quickly :)

jdavis:
I understand all the concern and know we'll settle on something that works really well.

First, before I respond.. @mxxcon, when you say compare, are you saying comparing different software solutions (like Google Auth and Authy)? Or comparing two sites (say Namecheap and Network Solutions)?

mxxcon:
@jdavis comparing two sites..isn't that what you meant?

@smholloway Well, I guess that's why the line is green once any of those are supported. I don't think this site is important enough (yet?) for some service to change its strategy to make sure they check off all boxes. :)
I imagine each site will decide on its own which 2FA are relevant and feasible to them.

Having said that, personally I'm not too found of the icons you used. They are a bit too obtuse/obscure. I guess @jdavis encourages the use of fontawesome, but if the font does not have descriptive enough imagery, I hope it's not an absolute roadblock to use a standalone file. I imagine we can find some acceptable free icon library to use.

jdavis:
@mxxcon Yeah, I was referring to sites. I just wanted to make sure that we understood each other correctly.

So regarding your question, it's hard for me to speak for other people, but the reason I created this site is because I wanted a place where I could go to see who had 2FA and then compare alternatives when making a decision. Like my blog says, I was transferring domains and wanted a registrar that had 2FA.

I'm moving across the country in a few months and am going to use the list to decide which financial institution to use and it all depends on what kind of 2FA and service they provide.

For example, I'd take SMS/TOTP 2FA over hardware or email any day. And the checkmark/X design really makes it easy to see that. The icon layout (the 2nd that @smholloway showed) isn't as easy.

Regarding @smholloway point about fostering competition, I don't necessarily agree. I don't think (and hope for that matter) that a site will be daft enough to think that.

If the consumer thinks that more checks is better, then that is a valid concern. Maybe we could get rid of the Red/Green combo all together and just do the --- like we do in custom and make it white.

What do you guys think of that?

mxxcon:
I do like red/green combo. That makes bad sites stand out and gives a warning to the visitor...Public shaming. :)

smholloway:
@jdavis Thanks for sharing your use case. For banks and domain registrars, it definitely makes sense. I'm not sure I'd switch from LoL to WoW just because of 2fa, but I would like to see 2fa everywhere and this site is a step in that direction (BTW, thanks!). At worst, a lagging company can see that their competitors have already added 2fa, thus motivating them to follow suit.

@mxxcon I'd argue that "bad sites" in this case are the ones with no 2fa in place. That being said, I like the red/green combo too.

Rather than speculate, I tried a few things out. Here are a few thousand words:

### Red and green

Imgur

### White Exes

Imgur

### White dashes

Imgur

Any of those strike your fancy? I'm currently drawn to the first with wholly red or wholly green rows and checkmarks where appropriate.

mxxcon:
I like 1st and last. What about full green rows, but just empty boxes?

jdavis:
Yup, I really like the first one as well. It is easier on the eyes without the white/red breaking things up and it gets across the point about it not being bad if a method of 2FA isn't supported.

smholloway:
@mxxcon Here's green rows with empty boxes:

Imgur

And here's one with checkboxes:

Imgur

Better or worse?

jdavis:
You'd make a great optometrist, @smholloway ;)

Also, I reallly really really like the last one with blank checkboxes. It is what I should have done in the first place, really.

smholloway:
Cool. I committed the changed to make entire rows either red or green and to represent capabilities with checkboxes. I also made the `Docs` icon `large` so it would match the size of the neighbor columns. Here's a sample:

Imgur

jdavis:
Yeah, I think that looks awesome.

jamcat22:
I agree. We should definitely change it so that the methods are sms, phone call, email, software implementation, and hardware token because most people are confused with the Authy vs. Google Authenticatior because anything that works with Google works with Authy, but not all things that work with Authy work with Google. (Also you would then have to get into Duo.)

I could help convert everything.
^EDIT: I guess this pull request already has that. (Impressive)

smholloway:
@jdavis Awesome! Anything else you'd like to see in this PR? I think this would play nicely with PR #208.

jdavis:
If we all think that the last image that you posted is what looks/functions the best, I think we'd be ready to merge this into the final release candidate for the new version.

I still would like to address the international issues from #241 and possibly the feature I've been working on when gathering/visualizing tweets before we "release" it.

jamcat22:
@smholloway You did a great job at changing all of that! I found a couple of mistakes though.

E*Trade - has software (Verisign VIP)
Fastmail - has SMS (maybe - the docs made it questionable) and hardware (Yubikey)
Dreamhost - has hardware (Yubikey)
Blockchain - has hardware (Yubikey)
BIPS - has software (Google Auth/Authy/Duo/every other app that allows Google Auth)

Let me know if I made any mistakes in my thinking.

smholloway:
Thanks, @jamcat22! I've updated those sites.

jdavis:
@smholloway Should I create a new rc branch for merging your changes into?

jamcat22:
Could we do anything to add a column for paid only providers? #354

smholloway:
@jdavis I'm ready for the rc branch if you are :smile: Let's :shipit:

jdavis:
Woo, cool.

I just created a branch called `dev` that we can use to do the merging and various checking to make sure all the data is up to date, then we can just merge it into master.

jdavis:
Also, I have two large homework things coming up this week and are due Friday and Saturday. So I'm going to be a bit strapped for time.

I added you as a Collaborator, @smholloway. Feel free to merge these into the dev branch and push them. Then myself and the others can take a look.

rublon-mw:
Haven't been here for a while. Progress looks very good.

My suggestions for column name optimization:
Docs -> How-to
Phone Call -> Call
Hardware token -> Key fob
Software implementation -> Mobile app

_data/email.yml
```diff
@@ -3,40 +3,31 @@ websites:
       url: https://www.fastmail.fm/
       img: fastmail.png
       tfa: Yes
-      goog: Yes
       sms: Yes
+      software: Yes
       doc: https://www.fastmail.fm/help/features_alternative_logins.html
-      custom:
-          - icon: key
-            url: https://www.fastmail.fm/help/login_yubikey.html
```
jamcat22, line 11: Fastmail has hardware (Yubikey).

mxxcon:
@mwww "fob" is just a...bleh name.
see https://en.wikipedia.org/wiki/Security_token vs https://en.wikipedia.org/wiki/Key_fob
plus stuff like yubikey does not generally fit "fob" concepts..

There's another method that I can think of sites use: text grid...Perhaps it should not be a separate column, but maybe we should leave some flexibility for "other" field?

jamcat22:
If I have other websites to add to the list, should I comment the changes on the file on this request? Or should I create a new pull request (and if so should I use the new format or the old one)?

juju4:
As said in #352, It would be nice to have a field for comment or extra-protection
either like a wikipedia foot note, or a comments column, cant always summarize by checkbox
As previously said, most French banks don't use 2FA for login, but they use it (since a long time) for adding transfer account, which limit impact of a compromise (outside of other protections like virtual keyboard, trusteer software, ...)

You could also precise if it's an option or default/mandatory as it seems part of americans are not familiar w 2FA and not pushing for it
www.infosecurity-magazine.com/view/33199/despite-security-fears-most-americans-dont-use-twofactor-authentication/

smholloway:
@jamcat22 Create a new pull request. However, if you can, wait until we get this live.

smholloway:
@juju4 I'm not sure how we could add a comment field without cluttering the interface. I'm afraid that a comment section would be spammy, long, and controversial; I could imagine the comment leading to more questions than answers.

I don't think twofactorauth.org is exclusively for login 2fa. I appreciate that companies are using a second factor for actions other than logins (in fact, my employer, Toopher, prides itself on letting implementers require user interaction for any action--not just logins) and personally think they should be added.

At work, many people are forced to use 2fa; however, I haven't heard of any consumer site requiring 2fa. Do you have an example of a consumer site where 2fa is mandatory?

I think your link helps show why twofactorauth.org is so important. People won't use 2fa until they know about it and this site has already been viewed by thousands of people--a step in the right direction.

juju4:
Yeah I understand. In best case, the doc link should provide info but probably not all the case
Still, in this case, you should make somewhere a clear definition of what is considered 2FA and which conditions (using for login, for some part, optional/mandatory, free/include or extra cost, ...)

I would say for consumer, outside of gmail, I can only think about internet payment w Visa 3D Secure/Verified by visa or Mastercard SecureCode with SMS (old one was using password like)
http://en.wikipedia.org/wiki/3-D_Secure
http://www.visaeurope.com/en/cardholders/verified_by_visa.aspx
http://www.mastercard.us/securecode.html

Yeah, knowledge is better but allowing to push it w twitter and the like is definitely the way to go!

gitcnd:
One big problem with this approach: Stifling progress: Most 2FA tech is TOTP, invented around 1984 - but there are new and far more secure offerings in today's marketplace, including things that block phishing and malware (TOTP was designed only to block offline keylogger password theft) and other things that make stuff easy for users (eg: PUSH or other automation helpers).

The problem - is that someone is going to decide what should, or should not, be in the table - and the 2FA industry is very wealthy, so I'm pretty sure there's going to be vendors out there blocking their competitors who have better products from adding this into the table...

gitcnd:
Also - I think "software" should be re-named to "mobile app" - since there's desktop 2FA plugin software out there which is a totally different thing, but everyone knows what "mobile app" means.

mxxcon:
wow @gitcnd what kind of tinfoil hat are you wearing? :/
The primary reason why there will be column names change is specifically to make it vendor agnostic! Right now there is a preference given to Google Authenticator, Authy and VeriSign VIP. All other solutions are relegated to "Other" category.
Soon we will have just broad categories and you will get more specific details once you go to each service's/site's relevant page.

There is nothing on this page "stifling progress". Nothing here prevents you from making your own 2nd factor authentication solution and convincing sites/services to implement yours.
This site just lists which sites/services support 2FA, regardless of what it is.

This git repo is public, therefore every single commit, merge and pull request are public. So if you claim that some "very wealthy" vendor will be blocking competitors, you can easily see and prove it from repo's history.
`Software` is generic enough to also include "desktop software". I'm not saying I'm aware of every single desktop solution, but the ones I'm aware of are just reimplementation of "mobile app" as you put it. They are not popular enough to justify a separate column.
Again, you have to click on the relevant site's `docs` link anyway to enable 2FA on your account, so you will see details of the specific implementation regardless.

smholloway:
@gitcnd I had similar concerns about impartiality; my solution is PR #208 and PR #333. With a page dedicated to providers, people can quickly get a glimpse of the 2fa landscape--without that information cluttering the main site (the main site simply shows whether 2fa is available on various sites). As @mxxcon said, the site is open source so every PR is public: anyone can freely add sites.

jdavis:
Nice comments, @smholloway and @mxxcon. :heart:

jdavis:
Hmm, I'm merging everything now. Would Verisign VIP go under hardware + software? It looks like they have tokens and mobile versions: https://idprotect.verisign.com/mainmenu.v

smholloway:
Yep. Old school RSA fobs were only hardware, then a number of people came up with "soft tokens" (software). Now, I think most third party 2fa providers have some kind of hardware token option in addition to software. I read the docs and tried to check `hardware`, `software` or both.

jdavis:
Okay, I think I finished the merge. If a few people could take a look, that'd be awesome. @RichJeanes @mxxcon

jdavis:
@mwww  you should take a look as well

jdavis:
I'd like to make the merge tonight if a few more people look at it =]

cc: @RichJeanes @mxxcon @mwww

jdavis:
Btw, the branch dev is where it all lives: https://github.com/jdavis/twofactorauth/tree/dev

konklone:
I'm clearly arriving quite late! But FWIW, I think the empty checkboxes do create an expectation that "they should be checked", even if the background is green. I also think it makes the table look more visually busy and cluttered, since you've already got borders around each table cell. So I suggest removing the background checkbox. That's all!

jdavis:
Awesome. Thanks for the feedback, @konklone.

mxxcon:
@jdavis I'd like to help but I don't exactly understand what do you want us to look at?

jdavis:
If there are any issues or if I missed anything =]

mxxcon:
@jdavis Is there a need for an explicit `tfa:` `No|Yes` field?
If there are no relevant fields present, then assume `No`.
If any field is present, assume `Yes`.
I looked through a bunch of files, and to me they look correct, but I'm still not quiet sure what to look for..

RichJeanes:
The html looks fine to me, but cleaning up the yaml files and I just found Network Solutions (Domains) says tfa:no and verisign:yes with no doc link. Anyone know the right answer? A quick search of their support site doesn't show anything.

RichJeanes:
I'll open up a new issue for Network Solutions instead of zombie-ing this thread.

jdavis:
@RichJeanes Cool, good idea.

smholloway:
Fantastic work, @jdavis. The site looks great!

jdavis:
It was all you, @smholloway. Haha.

computmaxer:
I'm just catching up on this stuff. Looks great guys!

PR #334, opened 2014-03-24
Generalize 2fa methods using icons

smholloway:
Here's the code represent implementations with icons only. This is an alternative to PR #333.

Imgur


RichJeanes:
I think this can be closed with the merging of #333, yes?

smholloway:
Yessir. Good call!

PR #335, opened 2014-03-24
added MakePlans (online appointment booking)

espen:


espen:
Two-facktor authentication using Time-based One-time Password (TOTP) algorithm specified in RFC 6238.

espen:
Done. Sorry I didn't notice the order in the file and it wasn't mentioned in the readme.

jdavis:
What is MakePlans? I've never heard of it or used it before?

espen:
@jdavis it is an online appointment booking SaaS. Can be used by doctors, hairdressers, spas or anything which require an appointment to be booked. Each client has it own booking site which can be customized. More info at http://www.makeplans.net/

jdavis:
I think it's a great site but I don't think it falls under #394 which is trying to figure out the criteria to be on 2FA.org.

I don't think any of the data is that critical if compromised.

espen:
MakePlans stores sensitive and personal health data and payment information. The application is certified by the Norwegian Department of Health (Norsk Helsenett) to store this type of data.

PR #337, opened 2014-03-24
Tumblr now supports 2FA

SeanHood:
Tumblr announced today that they now support two factor auth via SMS and Google Authenticator. (http://staff.tumblr.com/post/80583764030)


jdavis:
That was fast. I'm waiting for the ok from Travis and then I'll merge.

PR #339, opened 2014-03-24
Add Patelco Credit Union

jaimeohm:


RichJeanes:
Thanks for the submission! Is the bank's mobile app used to enable 2FA in any way? According to the docs they use SMS, email, and automated phone calls.
So I'd make 2 suggestions:
1) If the mobile apps don't do anything to provide 2FA, they can be removed
2) Add email to the custom column (check out other entries for the icon, I think it's just "mail")
Bonus 3) Why not add the "phone" icon under custom while you're at it. I know Chase supports that as well, it just hasn't been added.

jdavis:
In addition to the issues Rich mentioned above, we changed the format so this needs to be revisited.

jaimeohm:
I followed your suggestions, @RichJeanes and @jdavis:
1) Removed the mobile apps (the iOS app has never prompted me for more than my password)
2) Added email and phone as "YES"

It builds happily. Let me know if you need anything else.

jdavis:
Great.

PR #341, opened 2014-03-24
Added ADP.

mxxcon:
Looks like they don't support 2FA. :/


jdavis:
:+1: for ADP, however, it looks like your YAML has some issues: https://travis-ci.org/jdavis/twofactorauth/builds/21464937

It says you used tabs. HOW COULD YOU USE TABS?! :wink:

PR #345, opened 2014-03-25
tumblr works with authy as well

smisk:
Tumblr's two factor authentication works well with authy as well.  It was missing from the original pull request.


jdavis:
Hey @smisk, we've changed the site a bit but none the less, I've fixed Tumblr. Thanks for heads up.

PR #346, opened 2014-03-25
Moved Status.io icon to developer category.

joet3ch:
Status.io was moved to the Developer category, but the icon was left in the Other category. This fixes that.


mxxcon:
That was my fault. sorry. :)

joet3ch:
@mxxcon No prob :)

PR #353, opened 2014-03-25
Adding SingleHop to list of Developer section

daviddunlap:
This change adds SingleHop to the developer section. Included is the developer.yml change along with the inclusion of SingleHop logo created to spec. Thank you very much for the opportunity! Let me know if anything needs to be changed. Thank you very much!


ilyakatz:
Seems this pull request slipped under the radar. @daviddunlap would you mind updating it with the new convention - looks like it would be in cloud.yml. let me know if you need help. thanks

jdavis:
It looks like #546 is taking care of this. Thanks for the ping on it, @ilyakatz.

PR #357, opened 2014-03-26
Added Joyent to Developer

SeanHood:
They also support Duo mobile, however not sure whether to add that as a custom one? Google, Authy and Duo all seem to work in very similar ways.


smholloway:
The providers supported has been the source of much debate. There are a lot of 2fa providers like Authy, Duo, Toopher, Rublon, and many more. This site should present the options plainly and fairly. With PR #333 I've changed the columns to be more generic (SMS, phone call, email hardware, and software) and PR #208 adds a separate page to list providers. These combined should make the site easier to update and more scalable.

jdavis:
Like before, @SeanHood, if you updated it, this can be merged as well. I'd like to see Joyent in the Developer section.

SeanHood:
Image isn't playing well, will fix this later.

SeanHood:
Original image should work now.

PR #364, opened 2014-03-26
Added GoCardless to Payments

SeanHood:
They don't support 2FA :(


jdavis:
Hey @SeanHood, we're back at PRs. We changed up the site a bit, so if you could adhere to that, we would be able to merge this.

SeanHood:
Okay, will get this updated as with #357

jdavis:
Hmm, it looks like the image isn't working correctly. I even downloaded it but was still unable to view it.

SeanHood:
Re done the images for this and #357

jdavis:
Great. It works now. Sorry for being so picky, @SeanHood.

Issue #367, opened 2014-03-27
No More Merges Until #333 Is Resolved


jdavis:
To cut down on the number of merge issues that we'll have when we switch over to the new format, I'm stopping merge requests until this weekend when I'll have time to work with @smholloway to switch over to the new categories.

Have a look at the issue, #333, to see what you think.

smholloway:
:thumbsup: Good call, good sir.

jdavis:
Cool, thanks for the patience @smholloway. I wish I could do this before the weekend; my mind is telling me yes but my school is telling me no.

I probably better listen to school :P

smholloway:
No worries; real life comes first. You've already done something great and [open source shouldn't make you feel guilty](https://www.youtube.com/watch?v=UIDb6VBO9os).

jdavis:
As much as I dislike Jacob's demeanor/style most of the time, I really liked the take away and idea. Thanks for sharing that, Seth.

jdavis:
#333 has been closed so I'm closing this as well.

Back to PRs/Issues!

jdavis:
Also, @smholloway, @mpdavis, @zach-taylor, @computmaxer, @RichJeanes, @mwww, feel free to start merging again.

PR #370, opened 2014-03-27
Added Check Inc to Finance

kd7bay:


jdavis:
We'll revisit this after we finish #333 as it says in #367

PR #377, opened 2014-03-30
Added Yammer and OpenShift

ahelpingchip:
Both currently have no two-factor authentication support.

[Yammer](https://yammer.com) is an enterprise social network by Microsoft.
[OpenShift](https://openshift.com) is cloud hosting / PaaS by Red Hat.


jdavis:
Cool, we're almost finished with #333 and then we'll get back to PRs/issues.

PR #381, opened 2014-04-01
Add Mailchimp

smholloway:
Mailchimp is an email marketing service provider; basically, they make it easy to email groups of people. They're quite progressive in terms of 2fa, which is cool.


smholloway:
I just added the Mailchimp to the `other.yaml` rather than try to update PR #192.

PR #387, opened 2014-04-01
International players: France, Canada

juju4:
adding banks and others actors on French and Canadian markets, also e-commerce platform (some have optional/pay tfa which I considered as No => security should be default/included)
I removed the iphone/android link as custom field as been removed.

Please validate yaml file. I will check after for image files. are icon file (favicon) good enough?

Thanks


juju4:
Any feedback on those modifications? things to review?
Normally, I merge w upstream and there shouldn't be any conflicts. Not sure if the label is coming from an automated check or something else. The verify.rb seems also to be ok.

jdavis:
There are some changes that I agree with interlaced with ones that I disagree with. Would you be able to separate this huge pull request into smaller ones that only address a single thing?

Like you have the e-commerce platforms, which I don't think belong on the site. Then you have a few minor fix ups of wording (like with Ubisoft), then all the additions and other things.

juju4:
Ok but how to do it simply? seems git rebase but looks can't split one commit + few files, for example extract banks and their images.

jdavis:
One way would be to cherry-pick commits.

So make a new branch for each individual component, and then for example, you could just `cherry-pick` the commits for those specific changes.

juju4:
Ok. After a long rebase and cherry-picks, I split in 4 pull requests: finance-fr-ca, retail-fr, social-fr, backup-synology
I skipped eplatforms as you are not interested.

mxxcon:
Oh my.. I'm not an expert on git, but this does not seem correct...All those commits should not be in your branch..Even Github itself says that this pull request can not be cleanly merged.
I think you are better off preserving your changes elsewhere, deleting this pull request and that branch and starting over with a current up-to-date `master`...

juju4:
No, this one can be closed (will do it). rebase was a bit painful but necessary to extract what was appropriate for cherry-picks. just use the others pull request.

PR #391, opened 2014-04-01
Added Nest to Other

kd7bay:


jdavis:
How critical is it that Nest has 2FA? What kind of damage could be done if an account were compromised? (I'm not arguing, I'm just curious and want to be sure we only add sites that have sensitive info)

kd7bay:
Perhaps Nest is not so critical and more of me just getting carried away with adding services...  The site doesn't store personal info (name, address, phone, payment, etc) that I'm aware of.

A compromised account would allow an attacker access to heating and cooling controls, which could be an issue if, say, heating controls were turned off during the cold season in some places.  There's potential for damage to the heating and/or cooling systems.  There's potential for increased utility bills.

A compromised Nest account may lead more to a feeling of having one's personal space violated than actual loss.

That's about all I've got other than being able to see temperature, smoke alarm status and zip code...  I won't be all that upset if it doesn't qualify for the list, though perhaps it will in the future as they expand their line, along with other home automation companies that provide remote controls like Control4, Nexia, Vera, etc.

jdavis:
No, I think it's great that we iron out the specifics of what should go on 2FA.org. I'm going to make a new Issue for it now too because it has been coming up in a few other scenarios.

jamcat22:
Should this pull request be closed?

jdavis:
I think for right now we'll close this. If Nest adds more to their line up of things that can be controlled, like if they develop a home security feature, then we can come back to this.

PR #392, opened 2014-04-02
added LendingClub

ilyakatz:


seanmcelroy:
It looks like the URL is incorrect.  Should it not be https://www.lendingclub.com/ ?

ilyakatz:
@seanmcelroy absolutely correct, updated

PR #393, opened 2014-04-02
Added MotifInvensting

ilyakatz:


mxxcon:
I don't see their /security link mention anything about phone calls.. Do they have better link explaining what they offer?

ilyakatz:
I had a surprisingly difficult time finding anything about this on Motif website. I don't remember setting this up separately - I think it may have been part of the sign up process.

But they do provide it, here is a screenshot of what happens when I login

http://take.ms/lIaUA

jdavis:
Thanks for the screenshot. Also it looks like they barely mention it:

> Two-factor authentication (also known as 2-step verification) relies on something you know (like a password) and something you have (like a **cell phone**). When you are logging in using an unrecognized browser, we'll send a code to a device you pre-register. You'll enter that online code to complete authentication process.

They don't exactly say that they have it but I'll merge this for now.

Issue #384, opened 2014-04-01
Password Managers


RichJeanes:
I'm going to add Dashlane (supports 2FA) but now that we have a Security section, I feel like password managers fall into that category (right now, LastPass is in Other).
Thoughts?

smholloway:
I wondered where LastPass should fit in. I like the idea of moving it to Security.

jdavis:
I think Security is a good place for it, good call.

zach-taylor:
I like the idea of moving LastPass to Security. Okta would probably need to move there since it is very similar, just focused on enterprise. I wonder if an Identity Management section would make sense if we could come up with a handful?

jdavis:
Now we have 3.

mxxcon:
I think there are sufficient number of players(more than 3), it is a distinct and important enough category that it deserves a separate section...or at the very least upgraded from "Other" :)

jdavis:
Yup, I agree with @mxxcon.

clsy:
Create section called "Identity/Password Management"? There are at least a dozen legitimate players in that space.

jdavis:
I'm going to close this as we added the Identity section in #384. Feel free to add Dashlane or anything else, @RichJeanes.

RichJeanes:
Will do. Sorry I haven't been contributing. Been busy between Startup Weekend now and LA Hacks next weekend.

jdavis:
Real life should come before open source :smile:

PR #401, opened 2014-04-04
Make new section: Identity Management (per #384)

clsy:
Added CloudEntr, moved LastPass and Okta to new identity management
section. Per #384.


jdavis:
This looks great.

The only thing I noticed is the icon you selected: `identity`. I couldn't find it in the Semantic list of icons: http://semantic-ui.com/elements/icon.html

We could possibly use the icon called: `user`. We are using `users` for Social but I think the singular will work in this case.

mxxcon:
@jdavis for _Social_ you are using `Community` icon, not `Users`
How about `Hide` icon or `Id`

jdavis:
We're using `users` for `social.yml`: https://github.com/jdavis/twofactorauth/blob/master/_data/main.yml#L4

jdavis:
Also, the icons you are looking at @mxxcon, are part of the basic icon set.

But yes, I like that ID icon from the basic set.

To use them, the value should be:

``` yml
icon: basic id
```

mxxcon:
Ah, yes. sorry. I was looking at "Basic" set's `Community`, which looks very similar to "Standard" set's users. :)

jdavis:
Haha, no worries =]

But good eye with spotting the `basic id` icon. I really like it.

@clsygemalto If you could update it to `basic id`, we'll merge this. If you don't get to it before one of the collaborators that is. Either way :+1: for your changes.

clsy:
Done. Ready for merge. Thanks.

PR #408, opened 2014-04-06
Meraki logo should in "other", not "developer".

mxxcon:


jdavis:
I've noticed this for awhile, thanks for taking care of it :+1:

PR #417, opened 2014-04-07
Add WiKID Systems

nowen:
added WiKID. Not sure about whether to add 2nd listing for the open-source community version, please feel free to edit as you see fit.


smholloway:
I'd say that if you're using the same image and both have the same capabilities, they should be collapsed into one entry. In this case, there's also a link to the community edition from the main WiKiD site, which clearly ties the two together.

Almost every provider has different versions and plans--listing them all won't scale. This also leads to a draining political discussion about why one provider gets multiple entries while another doesn't.

nowen:
sounds good - and agreed about draining discussions.
ᐧ

On Mon, Apr 7, 2014 at 12:47 PM, Seth Holloway notifications@github.comwrote:

> I'd say that if you're using the same image and both have the same
> capabilities, they should be collapsed into one entry. In this case,
> there's also a link to the community edition from the main WiKiD site,
> which clearly ties the two together.
> 
> Almost every provider has different versions and plans--listing them all
> won't scale. This also leads to a draining political discussion about why
> one provider gets multiple entries while another doesn't.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/jdavis/twofactorauth/pull/417#issuecomment-39754336
> .

## 

Nick Owen
WiKID Systems, Inc.
http://www.wikidsystems.com
Commercial/Open Source Two-Factor Authentication
http://twitter.com/wikidsystems | #wikid on freenode,net

mxxcon:
@nowen your commit is missing logo image file..(32x32 else it will look strange)

nowen:
wikid

ok - sorry - right spot and size, hopefully

mxxcon:
@nowen close, but embedding it in comment here is not the same thing :)
go your own fork of this repo, switch to "patch-1" branch and add your image there. It will automagically appear here.

nowen:
I think I got it that time. obviously not very active one github. thanks for the patience.

mxxcon:
Good try, but still not there. :)
If you take a look at the file directly in your repo https://github.com/nowen/twofactorauth/blob/patch-1/img/providers/wikid.png you'll see that it is actually a plain text string rather than a real image file..

jdavis:
Would combining the two solutions into one work? I'm definitely in favor of that.

nowen:
Certainly.

nowen:
I just removed the sourceforge reference.

jdavis:
Looks good!

PR #421, opened 2014-04-08
Update backup.yml

zmarties:
Google Drive supports phone call

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -58,6 +58,7 @@ websites:
     img: drive.png
     tfa: Yes
     sms: Yes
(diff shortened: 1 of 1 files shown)


mxxcon:
@zmarties should probably update other google services as well..
And I think this link is better for `docs`: http://www.google.com/intl/en-US/landing/2step/features.html

jdavis:
Looks like someone beat you to it. And yes, the other Google services should be updated as well.

jdavis:
I did all that in this commit: eed9e9712819cd3d656909d76c5800a048738e55

PR #423, opened 2014-04-10
Add Mandrill, Mixpanel and add an image check 

tijmenb:
This PR moves Basecamp from Other to Communication and adds Mandrill (email service) and Mixpanel (stats). Also adds a test to the verifier so that it complains when you forget to add an image or make a typo.


jdavis:
All the sites you added look great, and thanks for reclassifying Basecamp. :+1:

PR #433, opened 2014-04-14
Add Xero and Mailgun

edwardabraham:
Added the cloud accounting service xero (www.xero.com) to the finance section. They don't offer two-factor authentication, even though there has been a feature request for this for over a year. 

Added mailgun.com to the email section. Not sure if this is the correct section. Mailgun are all about email, but they are more aimed at developers.


jdavis:
Great.

I'll merge this for now but let us know if you think it belongs more in the Developer section.

Also, FYI, I want to break up the Developer section a bit so stay tuned for that. I'll happen soon; school is just keeping me a bit occupied.

Issue #301, opened 2014-03-20
Fidelity - Kind of Two Factor Authentication


wooster11:
Fidelity's response to the tweets has been the following:

We have strong protections in place: http://go.fidelity.com/safe .  Customers can also request two-factor authentication.

Not exactly we're going to fix this but not exactly wrong either.

Although they don't support Two Factor Authentication in the purest sense - as a requirement for logging in - they do offer something similar called Transaction Security which requires a code sent via text to be entered before allowing a transaction to take place.

I don't know if there's a place on your site to indicate this or not, but it might be something worth mentioning.

The main problem with their implementation is that it still allows whoever is logged in to still see balances and holdings - and allows for access to lots of other stuff.

jdavis:
Yeah, I'm aware of this. I got a GuerillaMail yesterday from someone saying the exact same thing.

However, this has come up in the last few days and I won't budge as I didn't in #190.

Issue #190, opened 2014-03-19
Network Solutions *does* offer 2-factor authentication support


billcurnow:
Network Solutions offers 2-factor authentication support via Symantec's VIP Access app.  However, you must contact customer service to get support for 2FA added to your account.  Once you've jumped through the hoops it works quite well.

jdavis:
Awesome. However, to be "green" on 2FA.org, I think it's important that there be no "hoops" to jump through. Security shouldn't be hidden by loops.

billcurnow:
I agree that their implementation is terrible, and that they do not advertise their support of 2FA anywhere on their website, but the fact remains they do support it.  Saying otherwise is just disingenuous.  

If you were to flag their record appropriately there'd be 3 red Xs followed by a green "Call Support" under the custom option.  This should be more than sufficient to demonstrate that they're doing their own thing, which should please the NetSol haters out there.

RichJeanes:
I feel like that would be us going out of the way to support their poor policies. I agree with @jdavis on this one. I think we're better off having a link to Tweet at them until they're more transparent on their practices. Since they already have the system in place, all we'd need to see is some public facing documentation of it. Just a support page. That's it. I don't think that's too much to ask of them.

jdavis:
We've had to make this stand a few different times and we'll stay in the same position: unless a customer can turn on 2FA without contacting support, they won't be in the green on 2FA.org. Most users won't contact support.

Being disingenuous goes both ways here: if the company cares about their users, they'll do all they can to provide adequate security without the "hoops". Having a "hoop" is disingenuous to the progress that customer's deserve and expect.

rublon-mw:
Generally I believe that adding Network Solutions the way that @billcurnow described it would benefit the visitors of TwoFactorAuth.org. Maybe this way we could make some Network Solutions clients aware that they can protect their account with two-factor authentication, even if it's so hard to set up. In the end it's just a custom 2FA solution. On the other hand, there seems to be no official information about this on their website. In that case there is no easy way to check the status of 2FA support at Network Solutions. Like @RichJeanes said, there should be public documentation available for any site that we add to the list.

In the beginning of this year Network Solution introduced some kind of WebLock program:
https://www.networksolutions.com/blog/2014/01/weblock-security-program-information/
http://domainnamewire.com/2014/01/21/network-solutions-auto-enrolls-customer-into-1850-weblock-service/

fleish:
[...]

PR #436, opened 2014-04-15
Fidelity now supports two factor authentication

aleksa:
Fidelity doesn’t advertise security tokens on its website, but if you call customer service, they will tell you how to do it.


jdavis:
Fidelity has come up before as well as other sites that don't advertise 2FA: #190, #301, #148, #140.

Until it is clear on their site and accessible without a phone call, they won't be "green".

PR #441, opened 2014-04-16
Add Synology

juju4:
Synology NAS have a capacity to use 2FA w Google Authentificator.
Definitely worth noting as so many big companies don't seem to be able to do it.


mxxcon:
Image in 910fc3f is not 32x32...

juju4:
Thanks

PR #442, opened 2014-04-16
Update Authy

jamcat22:
If I'm not mistaken, Authy can text you, call you (for some sites like Coinbase), use software, and give you hardware tokens (for some sites).

### DIFF
--- a/_data/providers.yml
+++ b/_data/providers.yml
@@ -3,11 +3,21 @@
 #
 
 providers:
-
+  - name: Company Name
+    url: https://example.com
+    img: company.png
+    sms: Yes
+    email: Yes
+    phone: Yes
+    software: Yes
+    hardware: Yes
   - name: Authy
     url: https://authy.com
     img: authy.png
+    sms: Yes
+    phone: Yes
     software: Yes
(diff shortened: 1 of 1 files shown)


smholloway:
Looking at their website I don't see those capabilities... can you point me towards documentation that says they can do it all?

This brings up an interesting question: how do we determine a provider's capabilities? I suspect any provider would do anything given a large enough client, but that seems like a poor metric. My gut is to emphasize their strengths (most excel at software, with hardware tokens as a gimped fallback), but that's very subjective. I fear that PRs like this one will simply inspire an arms race where each provider will claim to do it all. We've still given providers a safe place to strut their stuff, but the information is less meaningful if they're all the same. Perhaps we should remove the capability columns and just make the interface an icon, name, and link.

mxxcon:
@smholloway this image shows phone call, sms and hardware token https://www.authy.com/assets/pages/index/devices-daa43fe72730e5ae873112b64cce864f.png

smholloway:
Bah! Who would look at the front page? :)

Looks like they do it all. Cool :shipit:

Issue #394, opened 2014-04-02
Determine Criteria for What Should have 2FA


jdavis:
We've come across this a few different times now (#391 #366) and I've decided we should create our criteria of what websites should go on 2FA.org.

I came up with a really quick set of criteria in a [comment](https://github.com/jdavis/twofactorauth/pull/366#issuecomment-39338476) we should iterate on:
1. Personal Info/Image*
2. Data (sensitive/important)
3. Money
4. Control

The idea that I had behind control includes things like remote desktop, hosting servers, and basically anything that would allow someone to control something that could cause damage to one of the first three criteria. It's almost like Asimov's Robotics Rules, haha.

\* - Personal Image was added at @dansmcgowan's request.

What thoughts does everyone have regarding this? Suggestions? Comments?

zach-taylor:
These may need to be fleshed out a bit so it's a little more clear on what should go on the site. Data is pretty vague. A to-do app technically has "data" but is it important? I guess sensitive and/or important would need some clarification.

mxxcon:
@jdavis what kind of sites would you _not_ want to be listed here? do you have specific examples?

And in the global scope, don't we want 2fa everywhere?

danmcgowan:
@mxxcon What would global 2FA really buy us though? We have to keep in mind that security controls should be commensurate with the assets / data they protect. 

For example, do I really want to ask the developers of my favorite recipe aggregation service to spend time and effort on developing (and then maintaining and _potentially_ paying on-going costs such those associated with an SMS gateway/service or an extra VM to support 2FA requests)? What's the value of the assets that the service is protecting and whats the damage potential? Pretty darn low. I'd much rather they add a new feature or keep the service free instead. 

That being said, I would recommend the addition of "Public Image" to the list of criteria @jdavis has proposed for consideration. This could include things such as major blog services that might not necessarily have a lot of personal information (or "personal data" as defined by EU privacy legislation) beyond an email address, but could negatively impact one's public image if the account was compromised and defaced or used for unintended purposes.

jdavis:
@mxxcon A site like Imgur, or a RSS site or a news site account. Something like that. Basically an account where compromise doesn't lead to terrible consequences.

@dansmcgowan Yeah, that's another good example with the recipes website.

Also, @dansmcgowan, if we amended Personal Info to be "Personal Info/Image", would that work?

I'm ready to start enforcing this and add it to the README so I'm just moving forward with it.

[...]

PR #443, opened 2014-04-16
Added Cloudinary.com an image hosting company to 'Other'

dmin7b5:


jdavis:
Does an image site warrant 2FA as discussed in #394 ? I'm not doubting it, just trying to keep a critical eye on things that get added. =]

dmin7b5:
Just finished reading the thread.  Since many people use image sites to host their blogs images, myself included, I certainly think it is warranted.  I think the main concern would be the deletion of images from an account since most use unique urls that can't be reused.

ilyakatz:
@jdavis I agree with @dmin7b5 , cloudinary is used by businesses not only to process images but to store them. if someone gets access to an account, they can delete all image assets which can be mission critical, Imagine pinterest without images :)

PR #447, opened 2014-04-18
Update E*Trade

jamcat22:
E*Trade supports software (with Verisign). I also updated the docs to show both methods.


jdavis:
:+1:

PR #448, opened 2014-04-18
20 Sites Added

jamcat22:
I added 20 sites to the list (sadly I couldn't get logos to work on my computer :frowning:. Sorry).

I also corrected some capitalization, and deleted a logo for 1Password which isn't needed anymore because it was removed from the list.

I was going a little fast so please tell me if I missed something!


mxxcon:
In the future please submit separate PR (at least per-file).. :(
What do you mean you couldn't get logos to "work" on your computer? They are simply image files...

jamcat22:
I have restrictions on my computer that don't let me download them. I'll just brute force myself into the admin account and check group policy, applocker, local security policy, and some other administrative applications to get around this. 

I'll just make them separate now.

PR #451, opened 2014-04-19
Add vanillaforums.com as "in progress"

kaecyra:


mxxcon:
"on our roadmap" is not the same as "in progress".

kaecyra:
It has been specced, all key technology components have been chosen, and a prototype is already under construction. If more progress than that is required, I'm happy to wait.

mxxcon:
Yes, _I think_ it's better to wait until they have something further along. I wouldn't mind seeing this entry added, but not with 'in progress' status.

jdavis:
I've finally finished writing up the Criteria for what belongs on 2FA.org here: https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria

Vanilla Forms look ridiculously cool. The only thing i question is how important 2FA is for the site and whether it belongs on 2FA.org.

I'm not trying to shoot it down; I just have a bar to hold =]

kaecyra:
No problem, I understand your desire to curate the list. I'll shed some light and you make the call.

Our plan for 2FA on Vanilla Forums is, at least initially, to prompt forum administrators for a second factor when they try to enter the admin dashboard. This should help to make "hacking" the forum more difficult, since the vast majority are the result of the continued use (by admin users) of compromised passwords. These passwords are usually compromised when other sites that the user frequents have their data stolen and dumped, or when the forum is migrated to our service after being compromised on an older platform.

We have had requests from several customers for this particular feature, and it seems to make sense. With wordpress.com listed, I figured we belong as well. Literally the same idea, but different technology.

kaecyra:
To be clear, I'm talking about building 2FA into the end product that our customers and their end users use, not just the vanillaforums.com website.

jdavis:
Sorry for letting this sit for so long, @kaecyra. But I'm going to merge this as you mentioned the Wordpress similarity.

PR #453, opened 2014-04-20
Update Heroku - status

rajiv:
add link to heroku discussions with note about beta 2fa support.

### DIFF
--- a/_data/developer.yml
+++ b/_data/developer.yml
@@ -50,6 +50,7 @@ websites:
       twitter: heroku
       img: heroku.png
       tfa: No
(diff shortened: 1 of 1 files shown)


jdavis:
Would this possibly be a better link? https://beta.heroku.com/2fa

rajiv:
i thought about that and specifically linked to the post on the discussion forums for a few reasons. the post on the forum is from a heroku employee, and at some point prior to production 2fa support, the beta might close and the beta link may stop working.

jdavis:
Good points; I agree. Thanks for this!

PR #455, opened 2014-04-21
Update heroku status; they have started a private beta

sdamashek:
Heroku started a private beta for two factor authentication. I updated the status with the link to sign up.


sdamashek:
Already committed.

PR #458, opened 2014-04-21
Added United States Social Security Administration

jamcat22:
See #448


jamcat22:
Please tell me if you can find a better logo. That was the only one I could find.

PR #463, opened 2014-04-22
Updating Duo Security entry

mstanislav:
Just a quick edit to list our formal name, "Duo Security" rather than the existing shorthand, "Duo". Also, URL pedantry.


jdavis:
:heart:

jdavis:
Thanks =]

PR #464, opened 2014-04-25
Add Provider LaunchKey

degan:
LaunchKey is a multi-factor provider that has Authenticator OTP support additionally protected by geo-fencing, device and knowledge factors. The LaunchKey platform itself can also be implemented as a 2nd factor.


degan:
I see pull requests after mine being accepted. Here is more information on LaunchKey's two factor OTP application: https://launchkey.com/app/guide/otp

jdavis:
Sorry about that.

Thanks for adding the information about LaunchKey =]

PR #469, opened 2014-04-28
League of Legends and Rift Update

rice2007:
First time doing a pull request, so I hope I did this correctly.


mxxcon:
@rice2007 is it only for Rift or all Trionworld games? If so, perhaps it's better to make a more generic Trionworld entry, rather than game-specific one?

rice2007:
@mxxcon I just went through their other titles, and it looks like ArcheAge has one as well. TrionWorlds has all of its account security settings under different subdomains that are structured the same: https://GAMETITLE.trionworlds.com/account/security/security-flow

If 2FA is available for the title, there's a link to it on that specific page.

jdavis:
@rice2007 So how should we reflect this on 2FA.org? It'd be nice if we could collapse all of TrionWorld's games into one like we do for Blizzard.

Could you do some more research to see if there is a way to do this? And if all the current games offered by Trion support 2FA?

rice2007:
@mxxcon After days of waiting for a response from their support, I received a generic, "We have a lot of tickets. We made a lot of changes to our website and hope that your issues was solved in those changes. Sorry," non-answer. I opted for further assistance. Maybe I'll get a better answer in the days to come.

I also noticed that both Rift and ArcheAge use the same authenticator for both titles, so if you have it for one title, then it is automatically activated for the other. In spite of this, there still is not a singular page to manage the authenticator; i.e you have to visit that game's specific subdomain to add/remove it. In the mean time, I see four ways to handle this: Implement the titles Rift and ArcheAge as separate elements, linked to only title (I think Rift is the more popular title) to manage 2FA, add an expanding element for TrionWorlds to list both titles, or wait for a response from their support. Linking to only one title isn't <i>that</i> bad, because you can add a game to your account with only a few mouse clicks and no downloading. This is your project, so I'm totally fine with implementing it however you see fit.

Also, TW has four titles under their umbrella. Rift and ArcheAge both support 2FA while Defiance and Trove do not.

Edit: Now that I'm looking at the website, we could just link to the main page with all of the titles listed: https://store.trionworlds.com/account/profile/manage-game!input.action?dd=1. When a user selects a title, there is a security tab specific to that title in the bar at the top. Would you be alright with this method?

jdavis:
I think that'd be good enough, @rice2007. The fact that it is available (for some of the games) is what is most important.

rice2007:
@jdavis By the way, everything <em>should</em> be good in this build. If anything else need to happen, I'll update.

jdavis:
K cool. I'm going to pull it down locally and take a look. Then I'll merge.

jdavis:
Cool, looks good.

PR #470, opened 2014-04-29
Mandrill has 2FA now

tijmenb:
Don't know when they implemented it, but there it is!

:v:


tijmenb:
Woops, I didn't see #468

PR #476, opened 2014-05-04
update heroku link to official (but still beta) documentation.

rajiv:


jdavis:
Way to keep up to date on this =]

Issue #395, opened 2014-04-02
New Categories!


jdavis:
We are starting to blur the line on a few places. So I think it'd be great to introduce a few new categories to keep the clutter down.

Here are a few of my ideas:
- Banking: To remove some of the banks from finance. Finance would remain one that deal with personal finance.
- Investing: Investment sites
- Hosting/VPS: this would split up Domains/Developer a lot more
- Cloud Services possibly
- Project Management (Trello, Wunderlist, etc)

Any other ideas? Those are the two largest categories along with Domains which I'm not sure what to do with.

seanmcelroy:
The banking-finance distinction would be hard for some of the very large financial institutions that offer both services under one login.  I'm all for the Hosting/VPS separation.

RichJeanes:
Though it's like you said before, a lot of domain registrars also offer hosting (eg Namecheap).

jdavis:
Yup, in that case, we could go by what they are most known for and/or started with. Like Namecheap is domains, and then Dreamhost would be hosting.

Maybe..? Hmm.

mxxcon:
Maybe then somehow allow cross-linking of categories?
It will become harder for contributors to find the "source" file, but for the users a site will be listed in all relevant categories.

jdavis:
What do you mean by cross-linking categories, @mxxcon?

Do you mean like having a section for "Hosting", a section for "Domains", and then "Hosting + Domains"? Or do you mean having a site listed multiple times?

I have some time now that I want to play a bit with the categories so any feedback would be great.

mxxcon:
Let's take Godaddy as an example. They offer a mile long list of services. It is obviously not a good idea to duplicate godaddy in every single category where they fit. It will be pita to update. Perhaps implement a system of tags that will automatically put each entry into appropriate category?

jdavis:
In addition to the commit above, I broke them up into other sections:
- Added a Project/Task Management section: 0aad2ef699cffdb466150e7ab6f7538594d22b6e
- Broke up Finance into Personal Finance, Investing, Banking (it actually worked nicely): 45b7bcefb68197c42d32eef9b29fa94ad37b5234
- Added a Hosting/VPS section: 2df32cdf32ce5ba1a5935e98cba8a53276316647

[...]

PR #479, opened 2014-05-06
Add new categories as in #395

jdavis:
If one or two people agree with these changes, we can merge this into the live version.

Thanks! :cat:


jdavis:
Here, @mxxcon.

jdavis:
... Okay.

mxxcon:
@jdavis I've been meaning to comment, kept putting it off and now... too late.(hopefully not?)
`Security` category name is kinda ambiguous.. Do you see something else go there beside SSL cert vendors?
I think `Cloud Computing` and `Hosting/VPS` should be combined into the same category. They are essentially the same thing and people that know nuanced difference it doesn't matter. Or if you still think they should be separate, perhaps rename cloud computing into `PaaS/IaaS`?

jdavis:
Yup yup, you can still make comments. No worries. I can't think of anything else that would go there but yeah, the name is a bit ambiguous.

Okay, I'll think on the `Cloud Computing` and `Hosting/VPS` sections and see what I think.

PR #483, opened 2014-05-08
Added identity managemet and provider

wakwanza:
Added new identity management service and provider


mxxcon:
@wakwanza It seems like both of them should be under `Providers`. Both sites seem to offer various two-factor authentication clients.
`Identity Management` section is more for password managers...
Also please make sure that logos are located in correct directories.

wakwanza:
Sure i have updated the logos directories and will push that change.I will update the Providers  although from saaspass functionality seems akin to a passsword/login wallet for various services more like lastpass.Correct me if im wrong of the difference between ıd management vs provider.

mxxcon:
Identity Management are _consumers_ of two-factor authentication. You would login _into_ `Identity Management` using solutions from `Providers` section.
`Providers` are various two-factor authentication clients...Google Authenticator, other OTP clients, physical RSA tokens, etc.
Saaspass seems to support TOTP and bluetooth proximity authentication. It does not store your actual usernames and passwords like LastPass does.

wakwanza:
Great.Pushing changes according to saaspass as provider.

wakwanza:
Any other changes to get the commit ready to merge.Im not sure about the size of the images is there a specific px ratio.

jdavis:
Thanks for this @wakwanza, the gh-pages branch will be updated soon. Thanks for your patience as well.

PR #488, opened 2014-05-13
Added Squarespace

mxxcon:
However, I'm not sure if this and wordpress.**com** fit into `Social` category..They are more of website/blog hosting platforms..


jdavis:
Yeah, I think Hosting is a better section for them. If you want to update that and then do a `git push --force`, it will update this pull request. It's what we've been doing with Neovim pull requests.

mxxcon:
SourceTree client makes working with git repos a breeze. :) If you haven't used it yet, you haven't lived. :palm_tree:

jdavis:
Hah, I know people that like it. What do you love about it? I'm quite fond of the command line, plus my [gitconfig](https://github.com/jdavis/dotfiles/blob/master/.gitconfig) is like a finely tuned engine.

mxxcon:
I'm not a pro with git. Probably if you can remember my 1st few commits/pullrequests were botched and I had to start over from scratch. As you can see from my stats this project is actually the most I ever commited to. So using a visual client simplifies a lot of things. Yes knowing raw git commands is useful in any situation, but from what I can see with SourceTree you can do beginner to higher-intermediate difficulty operations...And perhaps even more advanced stuff that I don't know/never needed to. Also it looks like it has great support for gitflow if you are into that.
If I can avoid vim, I will...It is the most illogical and convoluted tool one can use. Neovim might be a nice project, but to me it's like trying to convert a steam locomotive to run on fuel cells. :) Time to scrap it and start with something better.

PR #489, opened 2014-05-13
Added SyferLock GridGuard provider

syferlock:
Added SyferLock GridGuard provider to twofactorauth.org


mxxcon:
@syferlock you call that 32x32? :)

syferlock:
Sorry. I just noticed I upload loaded the raw file. Reject this pull and I'll send the version with the proper png file.

mxxcon:
@syferlock no need for such drastic measures. just commit an updated image file to the same branch and it'll be appended here.

PR #495, opened 2014-05-14
CapitalOne360: Add notice about ability to view sensitive personal information

rcj4747:
Capital One 360 accounts are secured by a 6-10 digit, self-selected PIN.  Accounts can not be fully closed / locked.  Access to the account leaks sensitive personal information (SPI) through the process of opening a new sub-account including SSN.  This information can not be removed or locked down.  These issues have been added to notes for this banking service.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -45,6 +45,7 @@ websites:
       twitter: CapitalOne
       img: capitalone.png
       tfa: No
+      doc: http://twofactorauth.org/notes/capitalone360/
 
     - name: Capital One 360
       url: https://home.capitalone360.com/
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -45,13 +45,13 @@ websites:
       twitter: CapitalOne
       img: capitalone.png
       tfa: No
-      doc: http://twofactorauth.org/notes/capitalone360/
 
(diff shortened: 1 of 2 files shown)


mxxcon:
That is pretty grim. :/

rcj4747:
The notes are now associated with the correct bank (CapOne360 not CapOne) with latest commit in this merge request.

rcj4747:
jdavis: The source of this information was a call I placed to Capital One 360 support.  I tried to keep the tone neutral but this issue is close to me so if I'm won't be offended by any editorial work.  I just thought it was important to surface this issue to consider the security implications of opening an account with Capital One 360 without 2FA.

ilyakatz:
@rcj4747 unfortunately I think this notice applies to many organizations. From my experience, closing accounts at HSBC and Citi has the same effect - online account with all the info still remains. 

Although I understand your frustration, I'm not sure that the goal of this project is to cover all security flaws of companies but just the 2 factor authentications. @jdavis @mxxcon @smholloway @mpdavis Would like your thoughts on this

Carlgo11:
I agree with @ilyakatz

rcj4747:
Sounds fine, I'll pull the merge request.  Thanks for the review.

PR #522, opened 2014-05-27
Namecheap: Added Yes to Phone support

joeykrim:
Phone support indicated on the same documentation link already provided: https://www.namecheap.com/support/knowledgebase/article.aspx/9253/45/how-to-two-factor-authentication


jdavis:
Great catch. Thanks for this!

PR #527, opened 2014-05-29
Add EngineYard

ilyakatz:


mxxcon:
I don't like "currently in beta" part. Without any type of dating system no way to know how current this _currently_ is. Perhaps a better wording would be something like `as of late May 2014....`

ilyakatz:
@mxxcon well isn't everything "currently in " status? For example, any site that has an "In progress" marker will become outdated when the company completes their work. So we would need to update it at that time (hopefully, the community will pick it up sooner rather than later)

mxxcon:
True, however with the current 'in progress' there's a link to a tweet which has some sort of date reference.
Also "available by request" to where?

ilyakatz:
Ok, makes sense. Will update in a different pull request, as well as wording of by request via a support ticket

mxxcon:
:+1:

jdavis:
Great point, @mxxcon. :+1:

PR #530, opened 2014-06-01
Adding internet.bs to domains

tresni:
I can't find any specific page or tweet that indicates their support for Two Factor Auth.  But it is very present in the account setup page once you sign up.

Example


ilyakatz:
@tresni i tried creating an account and it didn't ask me to setup 2FA. And I couldn't find how to enable it after a few minutes of browsing my new account. Would you be able to describe the procedure in a note similar to this https://github.com/jdavis/twofactorauth/blob/master/notes/fidelity.html ?

ilyakatz:
@tresni ah ok, i do see it in my profile. Still suggesting to create a note since they do not have any docs about it. What do you think?

jdavis:
Yeah, I think a note would be the best way to document it if they don't have customer facing documentation.

tresni:
I can do that.  I'll have to sign up for a new account as I don't know if the steps are exactly the same for setting it up initially as it is if you already had it before.

jdavis:
Cool. Thanks for doing that, @tresni.

Feel free to update this issue when the note is finished or if you need advice/assistance.

tresni:
Okay, I think I've done this right.  Had a hard time testing the gh-pages, but tried to follow the example for Fidelity as close as possible

ilyakatz:
@tresni you can preview your changes locally using the [jekyll](https://github.com/jekyll/jekyll) gem, which is what github uses for the pages under the hood
- `gem install jekyll`
- cd into the root for your twofactorauth
- `jekyll serve`
- Your page will be accessible via http://localhost:4000/notes/internetbs/

tresni:
Thanks @ilyakatz .  Looks good to me!

jdavis:
Beautiful. Nice work, @tresni. And thanks for helping him out, @ilyakatz.

PR #533, opened 2014-06-04
Add MePIN

hasa:


mxxcon:
Are you sure about "email" support?

hasa:
@mxxcon Yes

smholloway:
https://www.mepin.com/technical-overview/ says nothing about email--can you point me to some documentation, please?

hasa:
It was mentioned as legacy method in previous version of developer doc ( in developer.mepin.com ), but not present in latest version. MePIN does not recommend E-Mail 2FA although it's supported, so I can also leave it out if you feel having a checkmark there is a problem.

jdavis:
Yeah, @hasa, if you left off email, I think that'd be for the best. As @smholloway mentioned, since it says nothing about it and is only available in the previous version, I don't think it should be there.

Once you update that, it can be merged.

hasa:
Thanks for input, I fixed the feature list to follow documentation.

jdavis:
Great! Thanks.

PR #535, opened 2014-06-05
Added Enjin

jamcat22:
Also corrected a capitalization error. See #448


mxxcon:
@jdavis another example, should this be under `gaming` or `hosting`? :wink:
@jamcat22 why did you delete trion worlds?

jamcat22:
@mxxcon Because it is already on the website. I deleted it from my pull request, not from the site.

jdavis:
Oh, that was strange. I was very confused.

I pulled it locally and it looks like you re-added it once it was there. Therefore you removed one of the occurrences which left it still there.

Strange!

jdavis:
I manually merged it with these commits:
- 096955e4937c6d02e02b71b70299899f0b9e10a2
- c06ba5e4845ae00b20d1e1346f01d1e2b216c709
- e9523789f34b8591463969c6171b40f92a757b79

I try not to modify the history but I did a cherry-pick of each of the commits because I was confused about the deletion of Trion until I had realized what happened.

PR #545, opened 2014-06-05
Update Ally Bank - email

ilyakatz:
Ally bank also allows email 2-step-authentication, here is a screenshot

Alt text


jdavis:
Awesome, I should have noticed this; I just signed up for an Ally account.

Issue #285, opened 2014-03-20
Add SAML/OpenID connect support for self-hosted 2FA


nowen:
SAML/OpenID connect would allow you or your organization to control the authentication server.  Google Apps allows this, for example.  

This provides some protection and far greater control.

mxxcon:
There's now `Identity Management` section, so I imagine your WiKID would go there. Feel free to contribute accurate information. :)

RichJeanes:
I feel like this has come up before but I can't seem to find the issue thread... @jdavis I seem to recall that you shot down the idea of listing services that allow you to host your own 2FA...
Please correct me if I'm wrong.

mxxcon:
@RichJeanes service that support such integration is different from services that provide such integration.. I think SAML services that support 2FA are a good match for `Identity Management` section.
Don't you think so?

jdavis:
Yeah, I don't agree with having self-hosted solutions that support 2FA on the site unless I heard a compelling argument in favor of it.

Hmm, I'm a bit ignorant of SAML and might have to read up more on it to decide.

mxxcon:
Bumping this up again..With commits like #547 we should probably make a decision one way or the other when it comes to services that support SAML intergration.

smholloway:
My initial reaction is that SAML integrations and SSO tie-ins are outside the scope of this site. Theoretically, any site could support third party authentication systems, but configuring a single sign-on then enabling 2fa feels very different than simply enabling 2fa. Perhaps the criteria is something like this: is 2fa part of the standard user login and can it be enabled in one step?

ben-dearmond:
"Security is important, @RightScale. We'd like it if you supported two factor auth. http://twofactorauth.org  #SupportTwoFactorAuth"

It _is_ possible to implement a 2FA solution with SAML SSO on RightScale, as username/password authentication can be disabled, so the veracity of the twitter share "depends on what your definition of is is".

mxxcon:
SAML is just a mechanism for a site to ask some designated destination if this user is allowed in or not. How that destination verifies that user is irrelevant to the site. In some cases it can be 2fa, in other it's a colonic map :wink:
It's up to the SSO provider to support 2fa.
@smholloway So I think a more accurate statement would be "any site that support SAML/SSO could support 2fa".

jdavis:
I agree with @smholloway regarding this.

[...]

PR #547, opened 2014-06-05
Update RightScale

ben-dearmond:
It is possible to use 2FA with RightScale, once SAML SSO is configured. Related ticket: https://github.com/jdavis/twofactorauth/issues/450

### DIFF
--- a/_data/cloud.yml
+++ b/_data/cloud.yml
@@ -65,7 +65,10 @@ websites:
     - name: RightScale
       url: http://www.rightscale.com/
       img: rightscale.png
-      tfa: No
+      tfa: Yes
+      exceptions:
+          text: " 2fa through SAML SSO"
+      doc: "http://support.rightscale.com/12-Guides/Dashboard_Users_Guide/Settings/User/Actions/Getting_Started_with_SAML#Enable_SAML-based_Single_Sign-On"
       twitter: RightScale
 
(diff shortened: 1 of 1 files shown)


jdavis:
Hmm, I don't understand?

Isn't this for people _using_ RightScale to host their service? Rather than for users logging onto RightScale through this link: https://my.rightscale.com/session/new

mxxcon:
RightScale is a cloud-management platform.
It obstructs cloud-specific implementations allowing you to (relatively) transparently have the same infrastructure setup regardless of what kind of cloud provider you use.

RightScale themselves would not use SSO to manage your cloud accounts. They would use cloud-specific management APIs.

ben-dearmond:
RightScale SAML support overrides the username/password authentication at https://my.rightscale.com/session/new

Example screenshot when enabled (note that username/password auth can be disabled):

screen shot 2014-06-11 at 9 51 03 am

As mxxcon wrote, communication between RightScale and cloud providers uses cloud-specific APIs.

jdavis:
As discussed in #285, I feel that in its current form, SAML & SSO don't belong on 2FA.org. I'm closing this.

ben-dearmond:
Thanks for considering. If you ever revisit this, maybe implement a "Bring your own" flag similar to "In Progress".

PR #550, opened 2014-06-06
Added Norton

jamcat22:
See #448


mxxcon:
Should this be named just "Norton"?
We have "Symantec/Verisign VIP" in the `providers` section..

jdavis:
Has it been updated? All I see is "Norton" as the name.

jamcat22:
@mxxcon and @jdavis  While Norton is a brand of Symantec, a Norton Account is separate from Symantec's VIP and Symantec's SymAccount. Norton is the consumer division of Symantec.

This means you can have two accounts:
• A Norton account (https://account.norton.com) which manages your consumer Norton branded products such as Norton 360, Norton Family, and Norton Antitheft.
• A Symantec SymAccount (https://symaccount.symantec.com) which manages your business Symantec branded products such as MySymantec, MySupport, Symantec Licensing Portal, PartnerNet, Customer Care and Symantec Connect.

Neither of them are providers of 2FA.

The two accounts are completely seperate. You cannot login to Norton Family with a Symantec SymAccount and you cannot login to Symantec Licensing Portal with a Norton account.

So to avoid confusion, I suggest we name it either "Norton" or "Norton Account". I do not think we should put Symantec anywhere in the name.

PR #552, opened 2014-06-06
Adding FreeOTP

denilsonsa:
FreeOTP is an open-source OTP/TFA app for iOS/Android provided by Red Hat/Fedora.

If it is okay to add FreeOTP to the list, I can make the extra commit of adding the icon too.


mxxcon:
I see no reason why not to add it

denilsonsa:
Added the icon.

PR #554, opened 2014-06-06
Minor CSS change, no visible change.

denilsonsa:
`border-radius: 50%` is enough to achieve a circle.

Also, `!important` is not needed.


jdavis:
Thanks!

PR #557, opened 2014-06-07
Added Altra Federal Credit Union to banking

Mitlik:
Altra _does_ have a 2 level authentication, but both levels are from within the accessing user agent. Their second tier consists of asking security questions from which the answer must be pulled from a drop down with a shifting answer order. The user selects 3 questions from a group of 6 or 7. 2 of those questions are asked at log in. All questions have a predetermined set of possible answers.

I feel they do not live up to the 2FA criteria as it doesn't use a second channel to establish credentials.

Edit: Added minor detail to the 2nd level auth step and fixed an omitted word.


smholloway:
While security questions on top of a password could slow attackers, I'd agree that Altra FCU is not providing 2fa. They're only using a single factor: knowledge. (The possible factors are something you know (knowledge), something you have (possession), and something you are (inherence).)

ilyakatz:
We had similar discussions before (eg. #288) and agreed that knowledge based protection is appropriate for this site. @Mitlik would you mind updating the file and we'll merge it in? Thanks for your pull request!

mxxcon:
@ilyakatz Reading 288 does not seem to show that it was agreed. I have to agree with @smholloway here. Security questions are still "something you know", not "something you have"...

ilyakatz:
@mxxcon i'm not sure I understand what you're saying? Are you suggesting that knowledge based question are valid for this project?

Mitlik:
@mxxcon, I took the discussion in 288 as supporting the idea that additional knowledge was _not_ TFA. After all the change was accepted with "tfa: no".

dsoegijono:
I think we're agreeing that security questions is "something you know" hence it's not TFA? Can we merge this?

Mitlik:
Thanks for the accept!

Issue #256, opened 2014-03-20
IDriveSync


Ritchie007:
ANother for the Backup section that doesn't support 2FA

jdavis:
Want to make the pull request, @Ritchie007?

Ritchie007:
See my problem to create a propre pull request explained in issue #281 !!

mxxcon:
@Ritchie007 
1) Pixelmator is supposedly really good. It's a commercial software, but for just $30 you'll get virtually the same features as Adobe Photoshop. Also take a look at http://alternativeto.net/software/adobe-photoshop/?platform=mac to see what other tools you can use. Your needs are relatively simple: crop an image, resize it and preferably preserve transparency layer.
2) You can create text files directly from github's website. Binary files such as images have to be uploaded using a desktop git client.
3) See this answer how to resync your fork with this repo http://stackoverflow.com/a/23853061/949859
4) I personally use SourceTree git client. You can take a look at it, but it might be a bit of an overkill for your needs. GitHub has https://mac.github.com/ but I don't know how good it is. Also take a look at http://alternativeto.net/software/sourcetree/?platform=mac to see what else is out there.

dsoegijono:
Is [IDriveSync](https://www.idrivesync.com/) a part of [IDrive](https://www.idrive.com/) or a different service entirely? Thought it would be better to add IDrive instead of IDriveSync if IDriveSync is a part of IDrive.

@Ritchie007 To add to @mxxcon's suggestions, if you only need to resize an image, just google a free online tool (I use http://www.picresize.com/). Or if you need something more powerful, [GIMP](http://www.gimp.org/) is a great free alternative to Photoshop.

Ritchie007:
Hi,

iDriveSync is a sync service supplied by iDrive web site.
iDrive is a Backup service supplied by iDrive web site.

I landed a job, so I won't have time to help anymore.

Cheers and Long Live to the 2FA site.

Ritchie

---

Richard Migneron
📧 : richard@migneron.com
iMessage: richard@migneron.com
Skype: rmigneron
📱 : +33 6 52 22 84 76
about.me/richardmigneron

dsoegijono:
> iDriveSync is a sync service supplied by iDrive web site.
> iDrive is a Backup service supplied by iDrive web site.

Should both be added to the website or just one?

jdavis:
It looks like iDrive is like Backblaze/Carbonite while iDriveSync is like Dropbox.

[...]

PR #558, opened 2014-06-08
Add IDrive to Backup

dsoegijono:
As per #256


jdavis:
Great, thanks for that!

PR #559, opened 2014-06-09
Update Envato

digihash:
I looked at the website and envato doesn't support tfa and certainly not with software.

### DIFF
--- a/_data/other.yml
+++ b/_data/other.yml
@@ -23,8 +23,8 @@ websites:
     - name: Envato
       url: http://envato.com
       img: envato.png
-      tfa: Yes
-      software: Yes
+      tfa: No
+      software: No
 
     - name: IFTTT
       url: https://ifttt.com


ilyakatz:
@Ibby looks like you were the author of the original Envato entry in this [commit](https://github.com/digihash/twofactorauth/commit/4a42e1337f0e8e825aaec1e4eea432a03b8e1887). Were you able to setup 2FA setup to your account? Do you know if there are docs that explains it? ir is it undocumented?

mxxcon:
@digihash what about https://account.envato.com/account/two_factor mentioned in the original commit?

digihash:
@mxxcon that page doesn't exist anymore and if you search for two factor in google in the envato site, only scripts that are made by other people are found to implement two factor in your own website.

mxxcon:
btw if `tfa:no` then you can remove `software:` line completely and instead should add their twitter `@envato`

jdavis:
I just sent @Ibby an email as he had already been in contact with me when he was first wanting to contribute. Hopefully he'll see this soon.

jdavis:
Here's what @ibby said:

> I think that was my bad, I have got staff account and totally forgot that 2FA is not available for all the members yet

So it turns out he has a staff account and so that's why he can see 2FA.

I'm going to merge this until it is available for all members.

PR #562, opened 2014-06-10
Remove DigitalOcean

ilyakatz:
there is already an entry under hosting


ilyakatz:
yay, the build is green again!

PR #569, opened 2014-06-12
Add Customer.io

ilyakatz:


jdavis:
I get a HTTPS certificate error when trying to go to their page...

ilyakatz:
good catch, the homepage is not https, management area is. updated

jdavis:
Ahh, gotcha.

PR #454, opened 2014-04-20
Added Aha!, aTech Media, Cloud 66, and MongoHQ

jamcat22:
See #448

PR #570, opened 2014-06-12
Added aTech Media

jamcat22:
See #454 and #448


ilyakatz:
@jamcat22 maybe this is a better link for the doc

https://atechmedia.com/blog/general/launches/two-factor-authentication

jamcat22:
@ilyakatz Couldn't agree more :smile:!

PR #572, opened 2014-06-12
Added Envato Twitter

jamcat22:
The red "X"s scared me on the page so I added their Twitter to replace them with a nice button.


ilyakatz:
Thanks. Twitter is also useful so that people can easily ask the company to add 2fa, but yes, those red x's are scary :)

jdavis:
Hahah, scary is good in this case =]

PR #581, opened 2014-06-15
Update Keeper doc link

dsoegijono:
Should be "doc" instead of "docs"

The old link only took me to their FAQ page, not to a specific Q/A. The correct FAQ link is https://keepersecurity.com/support/faq#faq36802 but in case it gets moved around (maybe why the old link no longer works), I used their 2FA page.

### DIFF
--- a/_data/identity.yml
+++ b/_data/identity.yml
@@ -27,7 +27,7 @@ websites:
       software: Yes
       phone: Yes
       sms: Yes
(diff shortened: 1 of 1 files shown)


jdavis:
By the way, sorry for not merging this sooner. I was out of town for the weekend and didn't have access to much internet. =]

PR #584, opened 2014-06-16
Add Bitly status link

jdavis:
I'm guessing this and the [security breach](http://blog.bitly.com/post/85169217199/urgent-security-update-regarding-your-bitly-account) go hand in hand.


mxxcon:
Everybody is complacent until they get burned. :disappointed:

PR #591, opened 2014-06-19
Adding 101domain to list of supporting Domain Registrars

tresni:
I use them for one domain so figured I should find out.  The software clients for VIP Access work (even though everything says "Security Card.")


jdavis:
Cool, this looks good to me, :+1:. I'll give it a bit of time to let others see it.

mxxcon:
It looks like they are using Symantec VIP system, so maybe match its features?

jdavis:
What do you mean, @mxxcon? VIP has hardware and software and both are added for 101domain.

mxxcon:
my fault. for some reason i remembered they also do sms and phone. sorry. :)

jdavis:
Ah, I gotcha. I actually had to double check VIP as well. I was a little unsure on the software side based on 101's docs.

PR #592, opened 2014-06-19
Adding Fabulous.com to list of TFA supporting Domain Registrars

tresni:
Found these guys in a lost of registrars supporting Two Factor Auth, can't really confirm myself as my domain portfolio doesn't meet their requirements, but all the documentation points to supporting a hardware 2FA device (Security Key.)


jdavis:
This looks good to me as well, :+1:

mxxcon:
Sounds like they use something similar to Yubikey https://secure.fabulous.com/informationcenter/index.htm?formdata[qid]=1194
Also maybe this link is better for doc? https://secure.fabulous.com/informationcenter/index.htm?formdata[qid]=1325

jdavis:
I think your previous link is better: http://fabulous.com/informationcenter/index.htm?formdata[qid]=1194

The second link assumes the person already has it while this link actually says how to get one and has the link you suggest for documentation at the bottom.

jdavis:
Cool, looks good to me :+1:. If someone else sees this (like @mxxcon and @ilyakatz), we can merge.

PR #596, opened 2014-06-21
Add BLACKLIST.md to show removed sites

jdavis:
I think this will help provide documentation on why we remove/don't list a site.


mxxcon:
"blacklist" sounds like they are being punished...Unless you want to add such site there as well?

jdavis:
Hmm, I suppose I can see that. Any ideas on what we can call it then?

mxxcon:
excluded?
not applicable?
unlisted? (delisted also sounds negative)

jdavis:
Yeah, I like excluded/exclusion. How's that?

mxxcon:
:+1:

PR #598, opened 2014-06-23
Corrected Windows Azure (now called Microsoft Azure)

jamcat22:
Windows Azure is now called Microsoft Azure. Their two factor authentication supports phone calls too. Also, the docs link was for a Microsoft Account, not Microsoft Azure.


mxxcon:
Is Azure account different from microsoft/hotmail/outlook account? Always?

For example Gmail is always a Google account, but Google account doesn't always have to be Gmail.

jamcat22:
Yes. They are completely separate. One is a corporate account and one is a consumer account.

They have completely separate logon systems and completely separate passwords.

jdavis:
Cool, thanks for that @jamcat22. I want to get a confirmation by someone else before I continue. Possibly some Microsoft insiders like @alexiip or @theMagicalKarp :smile:

theMagicalKarp:
I checked and everything looks correct. +1 ( ﾟヮﾟ)
It's a little confusing since the management and accounts portal both say **Windows Azure** but it's fo sho **Microsoft Azure** now.

jdavis:
You're awesome, @theMagicalKarp! Thanks dude.

Issue #242, opened 2014-03-20
Clarification on Wells Fargo Status


todd:
cc @TehRoot as it appears that you were responsible for the commit that makes note of Advanced Access.

How exactly are we defining 2FA here? Is it a strict adherence to the idea that you need a secondary authorization scheme for accessing a resource or the idea that only some actions on a protected resource require the additional security? I.e., are we saying you need a token to access an account or that you only need a token to perform certain actions on the account? Advanced Access is the latter (and, I might add, as a Wells Fargo customer, I've never been prompted to use it, regardless of having it set up).

tehroot:
I've had AA prompt me with large money transfers or withdrawls through the web interface. I put it under custom since it's a custom form of 2fa. This is supposed to let people know what kind of authentication, if any, institutions have, it's also why I included the direct link to wells fargo's documentation.

I don't think there was a rigorous look into the defining characteristics of 2fa besides whether it fit in custom, or in a normal OTP app like Authy or Authenticator.

todd:
I think it's still worth having a discussion about since it's not what I would consider 2FA (others may disagree). I would agree that having some additional form of security in the way that Advanced Access provides is good, but I don't think it's necessarily what 2FA has colloquially become known as.

@jdavis Thoughts?

todd:
Bump.

jdavis:
Thanks for the bump, @todd. We're working on a big change in #333 and then hopefully we can come back to fixing up all the issues/PRs.

Custom is no longer there, instead we are having a section for Hardware, Software, SMS, Phone Call, and Email.

Regarding AA, I've never used it so could use some more info (their site isn't that helpful: https://www.wellsfargo.com/help/faqs/advanced-access/)

How is it being used?

todd:
@jdavis Totally understood.

AA is only used when Wells Fargo determines you're performing a high-risk operation (like transferring a large amount of money). It works pretty much the same way as traditional 2FA - you're sent a code and asked to enter it on the site before you can complete your request.

It differs from 2FA in that it's only used in certain situations and you can still gain access to an account without having used it. As I mentioned above, I've never been prompted to use AA, despite having it set up. This is a huge issue because you can still access and perform all sorts of sensitive actions/information by just logging into the account (i.e., account numbers, statements, bill pay). In instances like that, AA does not come into play.

[...]

PR #599, opened 2014-06-23
Add Clef as a 2FA provider

landakram:
[Clef](https://getclef.com) is a two-factor replacement for passwords. We use possession of the phone and knowledge of a PIN as the two factors. Authentication is done by verifying digital signatures generated by a private key stored in the smartphone app.


jessepollak:
:+1:

rublon-mw:
Like you said, Clef is a replacement for passwords — it is a single sign-on solution, not two-factor authentication. It replaces one factor with a different factor. It doesn't matter what you do with the user to authenticate him with Clef. What matters is how many factors the web service needs to check before it authenticates a user. In Clef's case, the web service depends only on one factor (= Clef). Adding Clef to TwoFactorAuth.org would mean that we should also have to add every social login solution that supports 2FA on their side (Login with Facebook, Twitter, LinkedIn, Google+ etc.). Once you're signed into Clef, you can access other Clef-supported sites with a single click (hence single sign-on). Great solution, but solves a different problem.

landakram:
By both the definition given by twofactorauth.org [here](https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions) and standard definitions around the web, Clef is definitively a two-factor authentication system. It uses (1) something you know and (2) something you possess to verify you.

In terms of the social login point, the difference between Clef and those social logins is that Clef _always_ requires two-factor — it's not just an option that is turned off by default.

jdavis:
I follow @jessepollak on Twitter so I see Clef all the time =] I never thought about it as a 2FA solution.

I understand what @mwww is pointing out in that it looks like the only factor is possession of the mobile device.

@landakram Could you clarify what part of Clef is the first factor, something you know?

landakram:
Yep, there's a PIN you use to unlock the app itself.

jdavis:
Okay, is the pin required, _always_?

Also, I'd like a few other people to weigh in on this, like @mxxcon, @dsoegijono, @ilyakatz, and possibly even @smholloway. Let me know what you guys think; I'm not too familiar with this area of auth.

ilyakatz:
Hm, it does look like Clef fits the bill for the two properties. One question I have as a user, so when I log in, it says I'm logged in for an hour, does it mean I'm logged in for that hour to any site that uses Clef?

landakram:
@ilyakatz that's right. Sorry if that isn't more clear! 

@jdavis the PIN is always required, though we're looking to replace it with TouchID for phones that support it :)

jdavis:
Okay cool. Well, there definitely are two factors for Clef then: the device, and your brain/finger.

Any objections to adding it then? I'll give it 12 hours or so and we can merge this if no one objects :smile:

landakram:
just checking in — is this good to go?

jdavis:
There was some concern raised privately by an individual.

The issue was that Clef is more of a replacement of passwords than a 2FA solution. While you guys do offer the PIN feature, it is only 4 numbers which doesn't give the same security as a password especially if a phone were to be compromised.

If passwords were still required + using Clef, I'd totally agree in that it should be added to the site. I thought that the argument raised was a valid one.

Edit: Also, the section for providers is more for a solution that could be added on top of passwords as opposed to replace them.

landakram:
We market Clef as a replacement for passwords because users hate passwords, but Clef is two factor by the book and in practice. We use possession as the primary factor and knowledge as secondary, which lets us use a shorter PIN as the knowledge component since it cannot be brute forced without already possessing the phone.

If we really care about two factor, than making it a) easier to implement and b) easier for users should be our top priorities, and Clef is easier to implement than most other options and much easier for users.

ilyakatz:
@jdavis i think we're deviating from the ["constitution"](https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions). It doesn't anything about our evaluation of makes a 2FA implementation less or more secure. How do we know that companies that use SMS for 2FA don't send the same number to everyone? Or how can we guarantee that  companies' custom 2FA solutions are secure and not bogus? Clef fits the requirements so should be added. Or, we should update the requirements.

ilyakatz:
FIY, i'm not affiliated with with Clef is anyway, I guess I just take Supreme Court cases close to heart :)

ibenrodriguez:
Is the verification of the pin performed by an app in the phone or through
some other hosted service?

1) on the phone  - not 2fa
2) somewhere else - yes 2fa

If both the factors are in the phone then its not 2fa.

Take an ATM debit card for example.

If the pin was stored in the magstripe of the card that would make it
potentially possible to retrieve the pin.

Plus, with an ATM I can change the pin at the bank and no change to the
card is needed as it's not stored in the card.

jamcat22:
Well, if we already have Launchkey, we should add Clef.

Here is why:
1. Launchkey is like Clef and is marketed as a replacement for passwords.
2. Like Clef, they have a PIN feature, but theirs is optional and Clef's is required.
3. They are very similar in their features.

@ibenrodriguez Both. When setting up the app for the first time, you are required to make a PIN. When you setup the app on a new device, you are required to enter your name, email, and that same PIN to restore your account. If the PINs do not match, then your account isn't restored. While you are offline, you can still use Clef after entering your PIN, which suggests that the PIN is verified on a registered phone and their servers.

jdavis:
Okay, sorry for the delay. I'm going to take one more look at this but I hope to have it resolved in the next few days.

josephwegner:
I'd like to chime in here, in favor of adding Clef to the list. As far as I'm concerned, Clef definitely fits the bill for being defined as 2FA.

It seems that the two main issues blocking Clef in this discussion are:
1. Is a pin secure enough to be considered 2FA?
2. Is it really 2FA if both factors are on the user's phone?

@ilyakatz makes the good point for the first issue. 2FA, as a definition, is not concerned with the security of the implementation. 2FA is only concerned in the login flow following a pattern that requires two different forms of authentication. It doesn't really matter what those two factors are - a password, a pin, a thumbprint, or screaming a secret phrase at my computer. 2FA just means that I have two different factors.

The second question about both pieces of authentication happening on the phone is a bit trickier. The requirement here is that each of the two factors should be unique enough that if one is compromised, the other factor is still secure. Having the keys to get through one factor should not decrease the security of the second factor. I believe this is the case for Clef. The first factor is possession of my phone, and the second is the knowledge of my pin. If someone were to steal my phone, they still do not know my PIN, and it becomes no easier for them to gain access to it. If someone knows my PIN, it similarly does not become easier for them to steal my phone. Each of the factors here are unique and entirely unrelated to eachother from an authentication standpoint.

Just my thoughts. +1

_disclaimer: I worked pretty closely with Clef to develop [Waltz](https://getwaltz.com)_

lolux:
Since the purpose of this site is to raise general awareness on authentication security by informing the public about what websites support 2FA and what companies provide 2FA solutions, then there should be no problem including companies that are doing the latter. 

So :+1: on adding Clef to the list of 2FA providers without overthinking it or applying extra vetting beyond adherence to the Wikipedia definition cited on the site and the discussion referenced in #242; for Clef meets the requirements of both.

jdavis:
The Clef army is in full force ;)

I'm waiting for a reply through email and then I'll finish this up. I promise; I'm just trying to be thorough :smile:

jessepollak:
Always happy to see our supporters :)

No rush

mxxcon:
I'm not talking about technical aspects of clef here, however I get a feeling that such concerted efforts by companies to be listed on the site is turning more into marketing opportunity for companies rather than an informational resource. I'm uneasy about that. :/ (feel free to delete this if offtopic).

jdavis:
I agree with you, @mxxcon in that the marketing stuff is getting a bit worrisome. Especially regarding that other Issue that I won't name that was very blatant and that you commented on.

Regarding Clef though, I'm okay with accepting this for the time being. Especially since @jamcat22 brought up the fact that we already have LaunchKey in the list.

All disagreement should either be emailed to me or brought up in a new issue.

PR #600, opened 2014-06-24
Add Clavid IdP

thr0wable:
Added Clavid as 2FA provider


ilyakatz:
looking good, @thr0wable, is there a link to instructions/documentation for how to enable 2FA on Clavid?

thr0wable:
glad that it's looking good @ilyakatz. There is documentation available for a few authentication methods at: http://www.clavid.com/index.php?option=com_content&view=article&id=85&Itemid=128&lang=en but you usually don't need that, as wizards guide you through registering your token or OTP application.

If you signup with a token at: https://www.clavid.com/registration/ 2FA is immediately active.

Is that what you're looking for?

mxxcon:
I don't think Clavid should be added to providers.
It was decided that we would not list SSO/SAML services.
And for 2fa Clavid uses https://tiqr.org/, http://www.syferlock.com/, (which should be added to providers) YubiKey and google authenticator-compatible totp.
Clavid themselves do not provide 2fa.

thr0wable:
Whatever you think is best but from what I saw by quickly glimpsing through the other providers a few of them provide SSO as well. But maybe they have a custom app for their service which qualifies them as 2FA provider for you if I understood this correctly.

mxxcon:
Well, it seem like Clavid is primarily SSO. They support 2fa for authenticating into Clavid, but they themselves don't provide 2fa like Yubikey or Duo or similar.
Clavid would probably be more suitable for `Identity Management` section rather than providers.

thr0wable:
I think that makes sense now. You want me to move it there?

mxxcon:
(I think yes, it should be moved there)

thr0wable:
Ok, I moved Clavid to the IdM section. Do I have to create that pull request again now or is that still working?

dsoegijono:
@thr0wable This one works. Looks good.

Issue #601, opened 2014-06-25
Add mailbox.org


generalmanager:
Please add this German e-mail service which is very security focused (supports DANE, auto-encryption of incoming mails with PGP, sending only over encrypted connections and so on):

```
    - name: mailbox.org
      url: https://www.mailbox.org/
      twitter: mailbox_org
      img: site.png
      tfa: Yes
      hardware: Yes
      doc: https://mailbox.org/en/one-time-passwords-for-secure-login-to-our-webmailer/
```

jdavis:
Do you have a 32x32 image that we could use? Feel free to add it to a comment so someone can make the changes without having to hunt for an image.

generalmanager:
The official logo is available here: https://mailbox.org/en/press/
I grabbed the RGB logo and modified it to resemble the favicon:
![logo_mailbox](https://cloud.githubusercontent.com/assets/6797943/3386620/43fa337c-fc76-11e3-80b9-8e881803889f.png)

I hope that helps ;)

jamcat22:
Created a pull request for this. #603.

generalmanager:
Greatly appreciated, thanks!

PR #603, opened 2014-06-25
Added mailbox.org

jamcat22:
Also reordered the list in alphabetical order. See #601. (mailbox.org is lowercase because that is how the company stylies it.)


jdavis:
I was wondering why it was lowercase but went to the site and realized that's how they brand themselves =]

This looks good to me. Once another person signs off, it can be merged.

PR #604, opened 2014-06-25
add stackexchange

fm4d:


jdavis:
Social is a decent place for it. It also fits in Developer pretty well. My vote would be for developer as it is  more about the code than connecting with others.

mxxcon:
Stackoverflow or serverfault are developer centric.
But they also have a ton of other sites https://stackexchange.com/sites
Not the end of the world if it goes into either. :smiley:

fm4d:
Well, a lot of these sites are related with developing, but there are also Martial Arts, Sports, History etc. so I think that Social fits better than Developer.

jdavis:
I forgot about all of their non-developer sites _sigh_

This looks good to me then!

PR #606, opened 2014-06-26
Added AOL

jamcat22:
AOL had 2FA before everyone else, but they removed it a few years ago. :frowning: http://securitywatch.pcmag.com/security-software/284240-aol-dropping-2-factor-authentication-support


jamcat22:
Ok added the picture.

jdavis:
Looks good to me, :+1: I still can't believe people still use AOL though.

If you disagree, please email me at: [chunkylover53@aol.com](http://en.wikipedia.org/wiki/The_Dad_Who_Knew_Too_Little#Homer.27s_email)

PR #612, opened 2014-06-29
Added exception to Square Enix

jamcat22:
Also corrected a capitalization error.


jdavis:
Cool, thanks for this @jamcat22. If anyone wants to confirm it, this restriction is listed at the bottom of the page of the doc site: http://www.square-enix.com/na/account/otp/

PR #613, opened 2014-06-30
Update Bitly - tfa

jehiah:
Bitly now happily supports 2FA. 

http://blog.bitly.com/post/90381427894/introducing-2-step-verification-for-all-users


jdavis:
That's great news, @jehiah. Great work over at Bitly.

ilyakatz:
@jehiah wow, that was really fast. Thank you!

PR #615, opened 2014-07-01
Cleaned up identity.yml

jamcat22:
Deleted extra whitespace, capitalized "Yes" and "No" and removed unneeded "No"s and "twitter:"s.


mxxcon:
I would keep "unneeded" entries. It doesn't hurt to be more explicit. Perhaps twitter handle is also useful to keep in case we'll do something with it.

PR #621, opened 2014-07-04
Optimized and replaced logos

mxxcon:
There were still images of wrong dimensions and not in png format.
Also re-compressed everything once again to save a few more kb.


jdavis:
:+1: Looks good to me.

Issue #623, opened 2014-07-06
Apple supports 2FA in more countries than just the US


Matt3o12:
Apple supports 2FA in more than just the US. Here is a full list of all countries and the supported providers: ["Apple ID: About SMS and two-step verification"](http://support.apple.com/kb/HT5593)

jdavis:
Great, thanks for that info, @Matt3o12. Take a look at #624 that @mxxcon opened up and let us know if that suffices.

PR #624, opened 2014-07-06
List of supported carriers by Apple SMS.

mxxcon:
Addresses #623

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -12,7 +12,7 @@ websites:
       sms: Yes
       software: Yes
       exceptions:
-          text: "SMS only available in the US."
+          text: "See http://support.apple.com/kb/HT5593 for a list of supported carriers."
       doc: http://support.apple.com/kb/ht5570
 
     - name: Comixology


jdavis:
Should we format this link as clickable? Putting `<a>` tags around the link should work. It's just up to us whether or not we want that in the YAML...

mxxcon:
I don't think clickable link will work with the current implementation.
The moment you move mouse away from ! popup disappears....people can't even copy/paste plain text from it. :/

jdavis:
Ah yes, I forgot I picked "on hover" instead of on click. I tried out both.

The popups can be configured to either be on click or on hover: http://semantic-ui.com/modules/popup.html#/settings

mxxcon:
Is this still relevant or was it resolved in another PR?

jdavis:
As far as I can tell it is still relevant. I didn't merge it because I didn't know if you wanted to address the link issue. One possible solution is to allow configuration of the popover to be on click instead of on hover.

I could do this if you didn't want to; I just won't get to it until this weekend as I just started a new job on Monday. Just let me know, @mxxcon. :smile:

mxxcon:
Can we make it so that on hover text box expands, but also be able to attach a url to it for click?
Or would that be too obscure for users to realize?

jdavis:
What about just making it "onclick" instead of "onhover"? Then the icon would be clicked on and wouldn't go away until a click outside of the box?

Take a look at the first example with the button "Click Me": http://semantic-ui.com/modules/popup.html#/examples

mxxcon:
Is it going to be for all or only the ones that have links?

jdavis:
We could have it be configurable. Do you think it'll be too strange to have different behaviors for identical icons though?

mxxcon:
Yes, perhaps. I'm just worried about discoverability of on-click action. It's hard for me to judge what a new user would experience.

dsoegijono:
Can it be both? On-click and on-hover opens the popup, so it's the same as others but also clickable.

Issue #619, opened 2014-07-01
Should we add iCloud?


dsoegijono:
http://www.theverge.com/2014/7/1/5860436/apple-brings-two-step-verification-to-icloud-website

We already have Apple under Retail, and an iCloud account is an Apple ID. However, there are people (me included) who have separate iTunes account and iCloud account. Or maybe, instead of creating a separate entry for iCloud, we can just note it somewhere? Because without this article I wouldn't have realized that I only turned on TFA for one of my Apple IDs.

mxxcon:
I think it's fine to add icloud as a separate entry.
Google and Microsoft are listed multiple times for various products. Until we figure out how to link/connect multiple entries together, i think it's fine to add them.

jdavis:
:+1:

jdavis:
Closed by #620.

PR #626, opened 2014-07-08
Clarify iCloud exceptions/restrictions

reshleman:
I've added some clarification to how Apple applies 2FA to iCloud data. The list currently implies that personal information stored in iCloud is protected by 2FA, but that's not the case. Per http://support.apple.com/kb/ht5570, 2FA only applies when you take one of these actions with an Apple ID:
- Sign in to My Apple ID to manage your account
- Make an iTunes, App Store, or iBooks Store purchase from a new device
- Get Apple ID related support from Apple

In particular, signing in to iCloud.com on an untrusted device, signing into iCloud via Settings on an untrusted iOS device (to synchronize data), and attempting to restore an iCloud backup onto an untrusted device **will not** trigger 2FA verification. (Additional explanation and implications [here](http://arstechnica.com/security/2013/05/icloud-users-take-note-apple-two-step-protection-wont-protect-your-data/).)

I've added some wording to the "Exceptions/Restrictions" section for iCloud to clarify this. I'm open to rephrasing for clarity, if you think it needs it.

(I've also slightly modified the wording for Apple's SMS support, per #623 and #624. This matches the current wording for Twitter.)

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -75,7 +75,7 @@ websites:
     sms: Yes
     software: Yes
     exceptions:
-        text: "SMS only available on select providers. 2FA protects Apple ID account changes and iTunes/App Store purchases only. 2FA does not protect iCloud.com logins, synchronized iCloud data, or iCloud backup data."
+        text: "SMS only available on select providers. 2FA for iCloud data is available for select accounts only."
     doc: http://support.apple.com/kb/ht5570
 
   - name: IDrive
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -75,7 +75,7 @@ websites:
     sms: Yes
     software: Yes
     exceptions:
-        text: "SMS only available in the US."
+        text: "SMS only available on select providers. 2FA protects Apple ID account changes and iTunes/App Store purchases only. 2FA does not protect iCloud.com logins, synchronized iCloud data, or iCloud backup data."
     doc: http://support.apple.com/kb/ht5570
 
   - name: IDrive
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -12,7 +12,7 @@ websites:
       sms: Yes
       software: Yes
       exceptions:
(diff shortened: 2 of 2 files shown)


dsoegijono:
[This](http://www.theverge.com/2014/7/1/5860436/apple-brings-two-step-verification-to-icloud-website) is the article I linked to in #619. It's more recent than the article you linked. Apple is rolling out 2FA for iCloud apps:

> users who already have two-step verification enabled on their Apple IDs may see their iCloud web apps — Contacts, Mail, Calendar, Notes, etc. — grayed out and "locked" until they enter in the code sent to their devices.

Not sure if it covers restoring from iCloud backup.

Edit: they only mention accessing iCloud apps from iCloud.com, not iCloud apps downloaded on iDevices.

jamcat22:
Maybe we can put "iCloud protection currently in beta.". It shows up for me on one of my accounts but not the other, even though 2FA is enabled on both accounts.

dsoegijono:
@jamcat22 Agree. They are still being rolled out, so not available on all accounts yet. @reshleman Can you change the exception to show that?

reshleman:
Ah, good call. I thought I had remembered seeing that, but couldn't put my finger on where. MacRumors has [this article](http://www.macrumors.com/2014/06/30/apple-two-step-verification-icloud/), too.

There's also this caveat:

> At this time, it is unknown if Apple is simply testing the feature with some users or working on a wider rollout for all iCloud.com users with two-factor verification enabled.

Let me know if the wording in 340d8bd works. (I'm hesitant to use the word "beta" because that implies opt-in, whereas there's no way to request this feature on a specific account, AFAICT.)

dsoegijono:
@reshleman looks good to me! :)

jdavis:
If you're done with it, @reshleman, this looks good to me and can be merged.

reshleman:
:thumbsup: Yep, thanks!

PR #627, opened 2014-07-08
Twitter no longer requires SMS

reshleman:
Twitter supports SW-based 2FA via their Android/iOS App that does not require SMS. ([Details](https://blog.twitter.com/2013/improvements-to-login-verification-photos-and-more) - it's actually quite slick.)

Updating the Exceptions field for Twitter to reflect this.

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -80,7 +80,7 @@ websites:
       tfa: Yes
       sms: Yes
       exceptions:
-          text: "SMS only available on select providers, SMS required for 2FA."
+          text: "SMS only available on select providers."
       software: Yes
       doc: https://support.twitter.com/articles/20170388
 


mxxcon:
Interesting approach to bake 2fa directly into their app.

jdavis:
The reason we added that SMS required message is that in order to get to that stage, you need to enter a SMS number to activate it.

So when trying to activate 2FA from the Twitter app it will say "You must have a phone associated with your Twitter account to enable login verification."

Then if you enter a phone number that is already in use by another account on Twitter, you get: "The phone number you gave us, ‎+1 867-5309, is currently used by another Twitter account. Only one account can be used with a mobile phone at a time."

Thus the restriction still holds.

reshleman:
Interesting. The post I linked above, from the original release of their app-based authentication, explicitly says that you don't need a phone number:

> By using push messaging and in-application approvals, you no longer need to provide your phone number to use login verification. If you manage multiple Twitter accounts, but only have one phone number, you can now opt all of them into login verification.

And:

> Now, all you need is an Internet connection and one of our supported apps to enroll in login verification.

Looks like the actual implementation is different, so I'll close this.

jdavis:
That is weird. I actually just looked at the timestamp of the article you sent me and posted the information because I knew I tried to use 2FA with the same phone number on two different accounts, so I posted that.

But you are completely right, unless the person writing that article didn't know the details or they've changed the implementation since, it is wrong.

PR #642, opened 2014-07-26
Fixes broken YAML array

konklone:
This was mingling reddit and Stack Exchange data, and hiding Reddit from the list.


jdavis:
Nice find! That's actually a really interesting side effect of that syntax specification in YAML.

PR #653, opened 2014-07-31
added Netflix and iTunes

Carlgo11:


mxxcon:
Do we need itunes if we already have apple?

Carlgo11:
uhm yeah I think so

jdavis:
iTunes and an Apple ID are the same. I don't think iTunes should be on there.

Carlgo11:
@jdavis Okay. Anything else?

mxxcon:
@Carlgo11 is Netflix the logo correct? For me github shows it as broken image.

dsoegijono:
@mxxcon Netflix logo looks right to me.

PR #655, opened 2014-08-05
Update providers.yml

HiDeveloperDev:
Thanks for quick reply :) - done suggested change.


dsoegijono:
It should be sorted alphabetically (although I see a few of them aren't in their right place). I might go in and fix them after you're done.

You should also fix the indentation of your entry.

You can also remove `tfa: Yes` because it's not needed for providers.

dsoegijono:
Moved to #656

PR #656, opened 2014-08-06
Add SecSign

HiDeveloperDev:

PR #657, opened 2014-08-06
Add image, for #656

dsoegijono:


mxxcon:
@dsoegijono what you could've done instead of this separate PR was to send a PR to `HiDeveloperDev:patch-5`. Once he accepted it, it would've showed up in PR #656.

dsoegijono:
Got it. Thanks 😀

PR #659, opened 2014-08-06
Adding Wealthfront

gregschwartz:
`ruby verify.rb` shows no errors, and testing it looks right.


mxxcon:
should it be under `banking`, `finance` or `investing`? :confused:

gregschwartz:
Hmm. Actually you're right, it should be under investing. (Looks like
you're using "finance" more for aggregation than individual services. And I
don't think it's banking; you cannot, for example, write checks from
there.)

gregschwartz:
Updated.

jdavis:
Yup, the investing section is a better fit. This looks good to me even though the icon is 34x34 instead of 32x32 :smile:

gregschwartz:
Whoops! I must have made a typo when I resized it. 

Thanks!

Issue #641, opened 2014-07-26
OpenDNS - enterprise only & bitly URL


techmagus:
For the OpenDNS entry, it's only available for enterprise accounts. If you're a free account user, it is not possible to access the 2FA setting.

For Bitly, the URL can be updated to: https://bitly.com

Carlgo11:
@dsoegijono Can you close this issue since the pr #666 already have been merged?

PR #666, opened 2014-08-08
Add exception to OpenDNS

dsoegijono:
For issue #641. I'm taking @jcsesecuneta's word for it as I don't have an account myself to check.


jdavis:
This was confirmed at the time it was added, we just didn't have exceptions then: https://github.com/jdavis/twofactorauth/pull/510#issuecomment-43946236

Issue #670, opened 2014-08-09
HTTPS!


jdavis:
Even though our site doesn't transmit any sensitive information, I think it would still be great if we by default used https.

I setup https with Cloudflare and you can view it here: https://twofactorauth.org

Everyone should check it out and let me know if you see any issues. GitHub doesn't allow https so this is only through Cloudflare's services. I contacted GitHub Staff about this, maybe it will change and we can completely support SSL.

Here's a link to a SSL test: https://www.ssllabs.com/ssltest/analyze.html?d=twofactorauth.org

Cloudflare announced it is going to have free SSL soon and with Google's announcement of "encouraging" people to use SSL, I couldn't agree more. SSL should be the standard; regardless of the content.

So yeah, let me know if there are any concerns/issues!

mxxcon:
In Firefox debug console: `Blocked loading mixed active content "http://fonts.googleapis.com/css?family=Open+Sans:400,700"`
I guess need to go through all files/templates and update them to https or make protocol neutral.
I don't know if CloudFlare supports it, but if they do, enable HSTS header.
Another benefit of CloudFlare is that they support SPDY protocol, so our page with a mile long list of requests should load faster now

rublon-mw:
Great move. SSL should indeed be a standard all over the web.

jdavis:
@mxxcon I'll look into that HSTS header. And yeah, SPDY is a great feature as well that will help with loads. Although, I still would like to cut down on the huge list of requests like we talked about :smile:

Do you guys think it would be okay to _only_ use HTTPS now? I just need to setup a rule in CloudFlare and it would be enabled.

mxxcon:
> Do you guys think it would be okay to only use HTTPS now? I just need to setup a rule in CloudFlare and it would be enabled.

That's what HSTS is for.

jdavis:
It looks like there is only support for page rules, not direct HSTS support.

jamcat22:
@jdavis Since you are hosting the site on GitHub Pages and can't use .htaccess or modify the webserver, I do not think HSTS is possible. Best way to force would be CloudFlare rules. +1 for forcing HTTPS.

jdavis:
I've setup the page rule for now. If GitHub Pages ever fully supports HTTPS, then we can look at this more.

The new URL: https://twofactorauth.org

PR #671, opened 2014-08-09
Fixed external font reference to be protocol independent(for SSL).

mxxcon:
Address issue discovered in #670


jdavis:
Excellent. Looks good.

PR #673, opened 2014-08-09
Added conversation with Mojang support

Carlgo11:
I recently contacted Mojang support about 2FA.
Thought it would be good to show their reply.


dsoegijono:
I don't think this is necessary. They don't have TFA and don't have plans to implement it. I think the `tfa: No` is enough.

Carlgo11:
ok

jdavis:
Great thought though, @Carlgo11. :smile:

PR #674, opened 2014-08-09
Add Hulu, Last.fm, and Grooveshark to Entertainment

dsoegijono:
Last.fm and Grooveshark users can upload their music. Not sure about Hulu but Netflix was added and they're similar.


Carlgo11:
Not really sure about last.fm and Grooveshark since they're free to use.
unauthorized access wouldn't be the end of the world on those sites.

dsoegijono:
@Carlgo11 As I mentioned, last.fm and Grooveshark also allow users to upload music. I only use last.fm for scrobbles but Grooveshark is used by a few indie artists to share their music. I'm pretty sure most of the sites listed are free to use.

Carlgo11:
You have a good point there. I'd merge it if I where you :)

PR #679, opened 2014-08-12
Added ProtonMail

Carlgo11:


mxxcon:
I don't know if it's correct to include that pastebin..
They plan to implement it, that's great. But at the moment they don't have it. When they'll add it, we'll update the site.

Carlgo11:
@mxxcon ok

Carlgo11:
But actually, we have other sites that don't already have 2fa.

dsoegijono:
I think @mxxcon means (and I agree if that's true) that the entry is fine, but the `status:` part is unnecessary. We'll just change it to `tfa: Yes` when they've implemented it.

mxxcon:
yes, what @dsoegijono said. :smile:

dsoegijono:
Ping @Carlgo11. Can you remove the pastebin? I think it'll be good to merge then.

Carlgo11:
@dsoegijono Ok. had to make a new pr for that though. Long story...
#693

PR #677, opened 2014-08-11
Added the Entersekt provider

michaelmatthee:

PR #680, opened 2014-08-12
Added Transakt

Carlgo11:
As referred in #677


Carlgo11:
So are we merging this?

jdavis:
Yes.

PR #682, opened 2014-08-12
Update providers.yml

Transformagic:


mxxcon:
1st thing, @Transformagic please rename that image and its reference to something more descriptive than ia.png. `intensityanalytics.png` is a good choice.

I wonder if this solution can be considered as MFA/TFA.
You still have username/password, however the additional factor is that this system monitors HOW you type them in. To some degree it qualifies "something you have/you are"...However, I'm still not sure.

_data/providers.yml
```diff
@@ -53,6 +53,12 @@ providers:
     url: http://www.google.com/landing/2step/
     img: google.png
     software: Yes
+    
+  - name: Intensity Analytics
+    url: http://intensityanalytics.com/cvmetrics
+    img: ia.png
```
dsoegijono, line 59: You need to fix this to `intensityanalytics.png`

Carlgo11:
I agree with @mxxcon. Closing the PR

Issue #273, opened 2014-03-20
gmx.com, gmx.net, mail.com in email


lpn25:
Gmx.com, gmx.net and mail.com in Email, all owned by United Internet. None of them supports 2FA. Gmx.net and .com are rather large email providers in Europe.

jdavis:
Being that they are from Europe, that explains why I've never heard of them. Let's hold off on adding these until we get the International section done.

lpn25:
Makes sense, although there are already some international services listed (fastmail in email, gandi in domains).

jdavis:
Cool, we'll have to address those once we figure things out in #241

dsoegijono:
Added with #684

PR #684, opened 2014-08-13
Add GMX and Mail.com

dsoegijono:
Add GMX, Mail.com, MailChimp


dsoegijono:
I think this covers #273

jdavis:
MailChimp is in the Communication section as is Mailgun and Sendloop. Since they aren't "email" clients, we added them there.

dsoegijono:
@jdavis Oops! I've removed it. Are the other two good?

jdavis:
It looks good to me. I've never used either service though so unless they use "undercover" 2FA that doesn't have documentation (which is unlikely), then I think the values of `No` are good.

jamcat22:
I have email accounts at gmx and mail.com (and almost every other service to check for 2FA) and they do not have any form of 2FA.

jdavis:
Cool, thanks for the info, @jamcat22. This looks good.

PR #687, opened 2014-08-14
Added AeroFS

zarvox:


mxxcon:
Almost... Logo should be 32x32. :smile:

zarvox:
Good catch!  Amended with the appropriate 32x32 icon. :)

mxxcon:
:+1: lgtm

PR #702, opened 2014-08-20
Removed Loopia status

Carlgo11:
Loopia support says that they're thinking of adding 2FA but it haven't left the planning stage.

### DIFF
--- a/_data/domains.yml
+++ b/_data/domains.yml
@@ -154,7 +154,6 @@ websites:
       twitter: loopia
       img: loopia.png
       tfa: No
-      status: https://twitter.com/loopia/statuses/446223091161911296
 
     - name: Metaname
       url: http://metaname.net/


Carlgo11:
For those of you that can read Swedish, here's the link to my conversation with Loopia.se: https://twitter.com/carlgo11/status/502142278191030272

For those of you who can't, I'll try to summarize it below.

Me: "@loopia Hi what's the status on your TFA? Any ETA?"
Loopia: "@carlgo11 Hi, at the moment we don't have an ETA but it's still something we're working on."
Me: "@loopia so are you actively working on adding TFA?"
Loopia: "No it's still just on the planning stage and we're researching possible solutions." 

Since we've decided not to put a status on sites that haven't actually started working on TFA I think it would be good to do the same here.

jdavis:
Good work.

PR #703, opened 2014-08-20
Added MarkMonitor to domains.yml

mxxcon:
Just got an email from our account manager saying that as of September 30, 2014 all accounts will be required to use 2FA. They use Symantec VIP.


mxxcon:
@jdavis TravisCI puked...

jdavis:
Wow. That it did puke. Let's see if it runs again.

jdavis:
That was strange. It looks good now.

Issue #704, opened 2014-08-21
Bypassing Humble Bundle's TFA


dsoegijono:
I found out about a week ago and I've emailed their customer support about it, but have yet to receive a response (got an email a couple days ago saying that they're busy).

Basically, their TFA is set up through Authy, which doesn't work where I am. It somehow lets me set it up even though I can't use Authy. So I was stuck. I clicked on "forgot password" and reseting my password logged me in and I was able to turn off TFA through my account's settings.

Just letting you guys know.

Carlgo11:
How about an exception note?

dsoegijono:
Can someone confirm that this can be done? Don't want to add an incorrect warning.

Carlgo11:
It might be best if we don't add it at all actually. You said you contacted their support right? Let's not advertise a vulnerability

mxxcon:
@Carlgo11 security through obscurity is not security. Make it public shaming otherwise they will take forever to fix it.

Carlgo11:
Can we at least wait a few days to give them a chance to fix it before everyone tries to exploit the bug?
I know some security firms wait 5 days after contacting the owners of the product.

dsoegijono:
I agree with @mxxcon.

@Carlgo11 I already gave them a week between emailing and posting this issue here. Got their automated response of "Thank you... We will be following up with you shortly" on Aug 14.

dsoegijono:
Update: Got an email from them saying that they're pushing it to engineering to see what's up. They're finally looking into it!

jamcat22:
Fixed. See #708.

PR #708, opened 2014-08-22
Add Humble Bundle exception

dsoegijono:
For #704. It would be great if someone could confirm that TFA can be bypassed using the forget password flow.


jamcat22:
Just tried resetting my password and I can confirm that password resets do not bypass 2FA anymore.
image

Issue #710, opened 2014-08-22
Unfuddle now supports 2fa


thedoofa:
As the title suggests - Unfuddle is now onboard :D https://unfuddle.com/blog/2014/07/140721-two-factor-authentication

dsoegijono:
#711 merged

Issue #519, opened 2014-05-26
A "Recently Added" section?


dsoegijono:
A section on the website that shows sites that recently add 2FA. Do you think this would be beneficial? I'm thinking something like a page/section that lists sites that recently implemented 2FA, since the last time the website was updated or in the last week/month. Might be beneficial if people want to come back to the website to see if there are sites they use that now implement 2FA.

mxxcon:
I think it's not that simple to implement because the site a fully pre-rendered static HTML. The source of data are yaml files which do not record timestamp of when entries were added.
Unless I'm missing something, I think your best bet right now is to simply look at repo's changelog at https://github.com/jdavis/twofactorauth/pulse

dsoegijono:
@mxxcon Yeah, I couldn't figure out a way to implement this. That's a pretty good workaround, but I don't think the website's users should have to look at github. Then again, I'm not sure people other than me go to the site and look for this info.

I've never used the GitHub API before, but maybe there's a way to do this by pulling closed issues with a certain label? I don't see how this would be accurate though.

EDIT: Or maybe checking merges to master and figure out which ones update 2FA implementation info for sites?

jdavis:
As far as Jekyll variables go, we do have these available to use: https://help.github.com/articles/repository-metadata-on-github-pages but it doesn't look like there's anything useful there.

If we wanted to use the API, we could just do it client side, which would be neat. There's a JS library for that: https://github.com/k33g/gh3 Although it doesn't look like it can filter issues which isn't a problem, we'd just have to write our own.

The idea of using a specific label for it is interesting and probably the best way to handle it. We could have a little section somewhere that just lists the last sites added in the last week based on a issue tag called like "update" or something.

If we did this, I'd consider it a solution to #99 as I like the idea of the JS client-side as opposed to RSS.

dsoegijono:
Doing it client side sounds great.

Using a specific label might also be useful if a site adds/removes a 2FA method. Not sure if this happens. If so, maybe instead of "Recently Added" it should be called "Recently Updated."

RichJeanes:
Rezzing this thread.
Perhaps a hook to trigger a script when gh-pages is committed to to generate a table of Recently Added/Modified?

smiley:
This could just be done client-side, tracking the last visit date and asking the GitHub API for changes, then rendering a small table at the top. (Or highlighting the new entries, at the very least)

Carlgo11:
[...]

PR #711, opened 2014-08-22
Update Unfuddle

Carlgo11:
As requested in #710

### DIFF
--- a/_data/developer.yml
+++ b/_data/developer.yml
@@ -96,7 +96,11 @@ websites:
       url: https://unfuddle.com/
       twitter: unfuddle
       img: unfuddle.png
-      tfa: No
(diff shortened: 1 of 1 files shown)


dsoegijono:
@jdavis Requesting a new label (something like "update") because this is a good example for an updated site, for #519 (Recently Added section).

jdavis:
@dsoegijono Hmm, Collaborators can't add new labels? I've just added one btw.

dsoegijono:
@jdavis I don't think so. I can add/remove labels, but I don't see how to create a new label. Thanks!

jdavis:
@dsoegijono Interesting. No problem though!

PR #718, opened 2014-08-25
Compress images

dsoegijono:


mxxcon:
How much did you save?

dsoegijono:
@mxxcon The significant one was token2.png, which was ~28KB now down to ~1-2KB. The others were between 3-6KB compressed to about 1-2KB.

PR #722, opened 2014-08-26
Added email support to token2

token2:
Email based OTP deliver is offered, but used for debugging only and not recommended . But since you have that field in the table, should be mentioned for token2

### DIFF
--- a/_data/providers.yml
+++ b/_data/providers.yml
@@ -147,6 +147,7 @@ providers:
   - name: Token2
     url: https://token2.com
     img: token2.png
+    email: Yes
     software: Yes
     hardware: Yes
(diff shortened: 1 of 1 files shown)


dsoegijono:
If it's for debugging only, I don't think it should be added.

token2:
@dsoegijono there is no technical restriction on how email based otp delivery can be used. Using it for debug only is a recommendation and not a restriction.

mxxcon:
I agree with @dsoegijono here. :-1:

Carlgo11:
And I agree with @dsoegijono and @mxxcon as well.
Closing the pr since everyone is against it

PR #729, opened 2014-08-27
Added Entersekt provider

altusvantonder:
Please use the PNG image (entersekt.png) commited by michaelmatthee in #677

PR #731, opened 2014-08-28
Added Entersekt

altusvantonder:
I discussed this at length with @Carlgo11


mxxcon:
Ok...so what exactly did you discuss?

jdavis:
@altusvantonder Are you referring to the email you sent to @Carlgo11?

altusvantonder:
@jdavis that is correct; pls merge.

mxxcon:
@altusvantonder This is not how opensource and peer-reviewed development works. You can't just email somebody behind the closed doors and expect everybody else to fall in line.
As I said in the previous PR, "marketing positioning" is irrelevant to this site. Being a useful resource to users is the primary concern.
If there's a justifiable reason beyond just marketing, post it here. Otherwise :-1:

Carlgo11:
I'd just like to point out that I'm completely nutral in this question.

altusvantonder:
@mxxcon I am afraid you are missing the point. Exactly why I did not open the discussion here - there were too many people trying to form opinions when they in fact have no affiliation with Entersekt and have no understanding of the products being discussed.

This is not about marketing at all, or any other discussion we could open. This is about the current info on the site that is simply wrong. That is the one and only point. 

I engaged via email to users who seemed to understand and explained that the current info is neither correct nor helpful. We appreciate the efforts to make the web safer and inform the general public about security and especially two factor solutions. This is definitely needed and well received. Hence our need to correct the incorrect info. Transakt is merely a phone app - is does nothing on it's own. So it cannot be a provider or help with two factor in any way. Entersekt is the provider of the two factor system which entails a lot more than the phone app alone.

So, yet again: the only point to discuss is rectifying the incorrect branding (let's not call it marketing then) so that the website provides correct information about the applicable two factor solutions.

I trust you will find this helpful and we can put this to rest. Thanks.

dsoegijono:
I think you're being unnecessarily combative. All @mxxcon and I wanted to know is the reason why we should use Entersekt instead of Transakt. And you refused to give is a reason while forcing a merge. If you just said what you said above, we would consider it and resolve this.

altusvantonder:
I appreciate the comment. Although I explained all of this in #729 already, yet my slight frustration. Apologies if it came over too harsh - it was not the intention.

So I gather that this makes sense now?

mxxcon:
Transakt is merely a phone app. But so is Google Authenticator or FreeOTP. They also do nothing on their own. Yet there's no question for them to be there.
If I, as a user, were to come to https://www.entersekt.com/ I'd leave it in about 40 seconds in frustration because I couldn't find useful/relevant information there. All I see are marketing blurbs.
If I were to come to https://gettransakt.com I see 4 clear buttons to each platform and I'm done.

@altusvantonder your explanation in 729 is

> It goes against our marketing (branding) positioning. 

And at least for me that's not a good enough reason to confuse and frustrate users.

altusvantonder:
@mxxcon I can see your reasoning. However, this is not how we define our brand and products. So from Entersekt's perspective this is not correct and it is not how any of our clients and customers view it. The provider should be Entersekt, and Transakt is merely the phone app.

And again this is not about marketing. If you are using our brand in any way on a site then we would appreciate it being used correctly. The point is surely to convey the correct information?

Google does a myriad of other things apart from two factor. Entersekt specializes in only two factor. So it is indeed a bit different.

Carlgo11:
Well, Yubico is listed on the site and not YubiKey.

altusvantonder:
Yet again I see your reasoning; and we can come up with countless more examples supposedly for and against.

However, the only matter of importance here is that our brand is used incorrectly on the website. As the brand and trademark owner, Entersekt has the sole discretion on what constitutes correct brand usage. Entersekt's branding standpoints have been made very clear throughout this discussion and it won't change. Respectfully meant, any other opinion on the matter from external parties is simply irrelevant when determining correct usage.

I acknowledge that you may not agree (and may never agree), but your agreement is not required - only compliance is required when you do in fact use a brand and trademarks not owned by you on any website. I trust in your comprehension of the matter.

mxxcon:
@altusvantonder If you are going to incorrectly flaunt trademark and copyright laws and try to bully and threaten this site then _I feel_ it's best to completely remove all references to Entersekt/Transakt and both of us be on our way.

altusvantonder:
@mxxcon I fear everyone has been too emotional about this from the start. Nobody is flaunting anything, nor threatening anybody. This has only been a discussion and I would like it to remain as such.

The core matter which you fail to comprehend is that you are using the Entersekt brand incorrectly. I completely fail to understand why you deliberately insist to go against all whishes expressed and insist on driving your own opinion. It is simply about right and wrong, not about anything else you seem to read into the discussion. Please refrain from diverting the discussion to all kinds of sinister and unnecessary sub-matters.

From the interaction you seem like an intelligent person and as I stated numerous times I see your arguments and they have definite merit. However, in the usage of Entersekt's name and brands you simply do not dictate the positioning. So it is in no way disrespectful to your points of view, it is simply how it works.

Carlgo11:
This discussion is getting out of hand. I'm locking this thread

Issue #733, opened 2014-08-30
Add Roboform


jdavis:
I received an email mentioning Roboform:

> I would like to let you know that Roboform,  Version 7.9.8, the password manager, has implemented 2FA for logon to its own online site for "Everywhere" accounts.  Only email is supported but allows for multiple devices.  Documentation is at:  http://www.roboform.com

PaulMoore2014:
You should also know that SiberSystems (creators of Roboform) are actually changing their marketing / website to remove the phrase 2FA and use the correct term, 2SV...

Just a heads up ;)

PR #734, opened 2014-08-30
Add Roboform to Identity

dsoegijono:
For #733. I actually can't find a good page for `doc`. They mention OTP support on their release notes (like [this](http://www.roboform.com/download/news-windows) one), but not more information on it or how to enable it.

Also, rearranged OneLogin's entry by key so it's similar to others.


Carlgo11:
:+1:

PR #737, opened 2014-08-30
Updates to Symantec VIP offering

kunala:
Added compatibility changes as seen on Symantec website

### DIFF
--- a/_data/providers.yml
+++ b/_data/providers.yml
@@ -128,9 +128,12 @@ providers:
 
   - name: Symantec/Verisign VIP
     url: https://idprotect.verisign.com/
+    email: Yes
     img: symantec.png
     hardware: yes
     software: yes
+    sms: Yes
+    phone: Yes
 
   - name: TeleSign
     url: https://www.telesign.com


Carlgo11:
Can you give us a link where they mention this?

kunala:
Here is the link: http://www.symantec.com/content/en/us/enterprise/fact_sheets/b-vip_intelligent_authentication_DS_21213685.en-us.pdf

mxxcon:
Hmm..that document is from 2011, I wonder if it is still current or if they removed those methods?
https://idprotect.vip.symantec.com/orderstart.v lists only mobile app, desktop app and physical tokens, no mention of sms, phone call or email
Then I also found https://knowledge.verisign.com/support/identity-protection-support/index?page=content&id=SO19454
But again, it's from 2012..not sure if still current. I guess we need to contact Symantec to find out for sure.

kunala:
Yea there's some unclarity with the specs. Actually I work at Symantec, and
the product does support the out of band. :)

jdavis:
Okay, if @kunala works at Symantec and can testify that the data is accurate, I'd say this looks good to me? :+1: 

@kunala You should contact the team to get some up-to-date documentation for your users =]

kunala:
Yea, will do. Trying to revamp a lot of the usability on the products. Really open to suggestions for this to make it a lot better and easier to integrate.

PR #738, opened 2014-08-31
Added Domain.com to Domains

tjbenator:


dsoegijono:
The site supports https, so we should use `url: https://domain.com/`.

mxxcon:
:+1:

Issue #636, opened 2014-07-18
eBay has a software implementation


arikb:
Same as Paypal, eBay can use the Verisign / Symantec VIP app on phones.

mxxcon:
Indeed, looks like they use the same system.
@arikb would you like to make pull request for this?

arikb:
Done and done.

PR #744, opened 2014-09-02
Update Ebay in retail.yml

seancohen:
Removed software option from eBay.  eBay relies on PayPal, but does not support their software implementation, only their hardware tokens.

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -26,7 +26,6 @@ websites:
       img: ebay.png
       tfa: Yes
       hardware: Yes
-      software: Yes
       doc: https://www.paypal.com/cgi-bin/webscr?cmd=xpt/Marketing_CommandDriven/securitycenter/PayPalSecurityKey-outside
 
     - name: Etsy


jdavis:
It was changed in #636 and #637. Do you have a source for this? Possibly @arikb can chime in here as well.

seancohen:
Yep, I've been beating my head against it.

Paypal (owned by ebay) supports 2FA by both RSA hardware token, and SMS.

eBay can be set up to use the paypal hardware token only, but doesn't
support SMS or auth app, only the hardware token.

Sources:

 1) Chat transcript with ebay support (attached)

 2) Email reply from Paypal support (below, actual response came from ebay
rep)

hope this is enough! :)

Cheers,

Sean.

[EMAIL FROM PAYPAL SUPPORT:]
Hi Justin,

Thank you for contact eBay about on how to secure your account.

Unlike PayPal there is no security option to help you secure your account.
If you want you can activate your security token on your PayPal account to
help you secure your eBay account.

To activate a PayPal Security Key on eBay, you must activate it on your
PayPal account, add your eBay ID to the PayPal account, and then follow
these steps:

Log in to your PayPal account.
Select "Profile" at the top of the page.
Select the "PayPal Security Key" link in the "Account Information" column.
Select PayPal Security Key to activate on eBay and click "Activate on
eBay."
The serial number is automatically entered. If not, enter the number from
the back of the key.
Press the button on the PayPal Security Key.
Enter the code that appears.
Wait 30 seconds for the display to clear and press the button.
Enter the code that appears.
Click "Activate."

Thank you for choosing eBay.

Warm regards,

Marilyn O.

eBay Customer Support

[THREAD ID:1-FPH7B37]

On 3 September 2014 14:05, Josh Davis notifications@github.com wrote:
It was changed in #636 and #637. Do you have a source for this? Possibly
@arikb can chime in here as well.

techmagus:
Ebay and PayPal use Verisign's VIP Access software 2FA. However, PayPal only supports 2FA for US accounts but for eBay, we can use Verisign's VIP Access software.

I no longer have my various conversations with PayPal, it was 2 to 3 years ago. But for accounts here in ASEAN, they do not support 2FA "yet", to this day.

For ebay, it is possible, but one must activate it via the main ebay website. Local/regional ebay sites doesn't have a copy of the files (html, scripts) for activating (and resetting) 2FA. But their login supports it fine.

I haven't checked lately if they fixed the missing files since I removed my 2FA software. (VIP Access 2FA doesn't have a good process / implementation if you no longer have access. In any case it's beyond the scope of the project.)

seancohen:
I'm in Australia and I can confirm that PayPal fully supports both hardware and software 2FA for local accounts, but eBay only supports hardware tokens.

It's software support from eBay that I was looking for.  The desktop site on eBay to connect it to a token linked to your paypal account only supports hardware tokens.

Here's the only eBay page I can find, and it insists on a hardware token ID:

https://scgi.ebay.com.au/ws/eBayISAPI.dll?ActivateSecurityToken

Cheers,

Sean.

techmagus:
Ahh, that's interesting, so they're really fragmented when it comes to security. For ASEAN accounts, especially Philippines accounts, our situation is:

PayPal: No 2FA support, in any format
Ebay: hardware and software; personally I've tried software, they're using VIP Access, but must set it up via ebay.com not for example, eBay.ph

Here are my bookmarks:
Recovery: https://signin.ebay.com/ws/eBayISAPI.dll?LostOrForgottenToken

Activiation: https://scgi.ebay.com/ws/eBayISAPI.dll?ActivateSecurityToken
                  https://signin.ebay.com/ws/eBayISAPI.dll?ActivateSecurityToken

Manage keys: https://signin.ebay.com/ws/ebayISAPI.dll?ManageMySecurityTokens

Last I tried the above links was a month ago, still working for us here, and I can add software token (VIP Access = different token per software installation, so you have to add each installation to your ebay account.)

Anyway, this would mean not everyone will get 2FA support because each region are in a way, autonomous from each other.

arikb:
So, I suggest that all of these details are added as comments to the eBay
section. If no one volunteers I'll go over these emails, summarise and
create a pull request. If someone else wants to do a coherent summary, it
would help.

-- Arik

mxxcon:
@arikb Yes, that would be greatly appreciated.
Perhaps word it in such a way that we would not have to re-edit this entry with each newly discovered excluded country.

jdavis:
:+1: to what @mxxcon said.

seancohen:
If the intent is to drive awareness and action on the part of the reader,
then we want to make sure it's exposed well in the UI.

What options do we have in the UI with regards to displaying "partial"
support?

We want to make sure the twitter button is still in place, so people can
still poke the company for the regions that aren't supported.

Cheers,

Sean.

seancohen:
Ah, I see your suggestion works for ebay Australia as well.

Use this page to register:
https://scgi.ebay.com/ws/eBayISAPI.dll?ActivateSecurityToken
and use the Credential ID from the app as your Security Key ID.

No mention of the verisign app on the web site, but it definitely works.

Good extra data point.

arikb:
Pull request #750, hope you like it

jdavis:
@arikb @seancohen Just so I understand: the Verisign app does work, right? Which means this pull request is void? But #750 is still correct, right?

seancohen:
Correct, though I would also add a note that the Verisign app may work even
in locales where it is not promoted or advertised (like for me in .au)

arikb:
My experience is with the US PayPal and eBay - both have no mention of the
software app, but if you use the serial number of the app - they don't
complain.

Unlike OATH, which store the server secret inside the app and verify the
OTP with the server - the Verisign / Symantec solution brokers the
authentication requests. This is how my app can work with PayPal, eBay and
E*TRADE without learning a secret code from each. That is also why wherever
they ask for a serial number off of the back of a token - giving the soft
token serial works as well, because they send the details to Verisign /
Symantec anyway without knowing it actually came from the soft token.

For this reason, software 2FA based on this solution should work wherever a
hardware 2FA is advertised. E*TRADE is another example that I personally
use.

-- Arik

jdavis:
I believe this can be closed because of #750.

PR #753, opened 2014-09-05
Add BlueHost

cbiesinger:


mxxcon:
I think bluehost is better suited for "Hosting" category rather than "domains", wouldn't you agree?

cbiesinger:
_Sigh_ I confused myself with the categories.  Fixed.

mxxcon:
lgtm :+1:

PR #764, opened 2014-09-10
Added Azure MFA

berga0d:
azure
I've attached "azure.png" to this change request description

### DIFF
--- a/_data/providers.yml
+++ b/_data/providers.yml
@@ -15,6 +15,13 @@ providers:
     img: authenticatorplus.png
     software: Yes
 
+  - name: Azure Multi-Factor Authentication
+    url: http://azure.microsoft.com/en-us/services/multi-factor-authentication/
+    img: azure.png
+    phone: Yes
+    sms: Yes
+    software: Yes
+
   - name: Authentify Inc
     url: http://www.Authentify.com
     img: authentify.png


Carlgo11:
Could you add the logo yourself?

Carlgo11:
Also, I don't think you can call that a provider. They only protect their own service

mxxcon:
Azure is already listed under https://twofactorauth.org/#cloud

jdavis:
No, @berga0d is right. Azure is providing a service specifically for 2FA: http://azure.microsoft.com/en-us/services/multi-factor-authentication/

It isn't just 2FA for their cloud, but rather for people that build on top of their cloud.

Carlgo11:
Okay. We still need that icon though.
The png has to be exactly 32 x 32px.

Push another commit with the icon to patch-1 and the error will be fixed.

berga0d:
Is there a way to upload it through Githib website? It seems I can only create code files.

Inviata dal mio Windows Phone

---

Da: Carlgo11mailto:notifications@github.com
Inviato: ‎11/‎09/‎2014 11:12
A: jdavis/twofactorauthmailto:twofactorauth@noreply.github.com
Cc: berga0dmailto:berga_d@hotmail.com
Oggetto: Re: [twofactorauth] Added Azure MFA (#764)

Okay. We still need that icon though.
The png has to be exactly 32 x 32px.

Push another commit with the icon to patch-1 and the error will be fixed.

---

mxxcon:
@berga0d no, you have to use a desktop git client to do that. Their website does not allow uploading of images.

(ignore label, my mistake)

jdavis:
I'm sure GitHub has had that request before, but we really should contact them and let them know that being able to upload files through the web interface would be super nice.

PR #768, opened 2014-09-14
Finish adding Azure to providers list

jdavis:
Add the image to finalize Azure as a provider for cloud services. Closes #764


jdavis:
The image is a bit ugly, but it will suffice for now.

PR #770, opened 2014-09-15
added PhraseApp to list of services

docstun:
Added PhraseApp to list of services that support MFA, added to category "other"


mxxcon:
@docstun while that's great that you support 2FA, do you think PhraseApp fits into https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria ?

smholloway:
Does PhraseApp meet the [site criteria](https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria)?

> The following is a rough criteria and explanations for what sites should be on 2FA.org. If one of the following Criteria is met, it belongs on 2FA.org:
> 1. Personal Info/Image: Any site that deals with personal info or a person's image. An example of a site with Personal Info would be their Amazon account and a site regarding Personal Image would be one like Twitter.
> 2. Data: This means data that is either important or sensitive. It also is any data relating to Criteria 1.
> 3. Money: Any site that deals with money.
> 4. Control: This is a more general Criteria that includes sites that give access to things that might infringe upon Criteria 1, 2, and 3. An example of this is a site that allows remote access.

It's a useful service, but I don't think it matches 1, 2, 3, or 4.

docstun:
I guess if (project) management apps like Basecamp and Slack or Developer Tools like Mandrill, Mailgun or Github qualify, why should our translation management system (widely integrated into CI workflows) not qualify for this overview?

Carlgo11:
I think it does fit in the site criteria. Don't see why it wouldn't?

From what I can see someone could mess around and destroy/delete the app content and the user's personal data.

jamcat22:
Agree with @docstun and @Carlgo11 on this one.

Originally, I thought the criteria only applied to listing websites that don't offer 2FA...

PR #774, opened 2014-09-17
Added Zetta.net

matthew-gerstman:
Zetta.net is a cloud backup for enterprise solution that supports a software implementation of Two Factor Authentication


mxxcon:
@mattgerstman could you please run that PNG file through some optimizer? 8kb is pretty large for that logo and we'd like to keep homepage as light as possible.

matthew-gerstman:
Sure. Sorry about that. The new one is just under 2K. I hope thats better.

PR #776, opened 2014-09-18
Add WWPass

w3ua:
I could not figure out how to add logo... Please advice.


mxxcon:
@w3ua You can't upload images directly from github website. You have to use a desktop git client to add files.

Additionally, looking at your linked address, disappointingly https://www.wwpass.com/ link directs back to regular http webiste. :disappointed: If this is a bug, it should be fixed.

Furthermore, looking at what kind of products wwwpass offers, I think perhaps a better fit would be https://twofactorauth.org/#identity category, rather than `providers`.

w3ua:
Hope this will work

w3ua:
Thanks. Uploaded. Also will check with our IT if they can switch the 
whole public site into the https...

Talking about the proper classification... Honestly, I always thought 
that when we put on "authentication" hat (WWPass is much more than that 
-- basically, we offer unbreakable storage, which can be used for any 
purpose.... including credential storage) we act as "third party 
identity provider". So if we are more "identity" then "provider" (or 
vice versa) -- you decide as the list owner ;) . At the moment I 
uploaded our logo to "providers" directory, but if you want to move 
us... Go ahead. However, we position ourselves competing with RSA, 
Vasco, NagraID, Duo, SecurAuth, and not with Lastpass or 1Pass (which I 
believe soon be extinct, along with passwords themselves).

Indeed, when we do authentication, we can be "multi-factor" -- with 1, 
2, 3 or more factors, depending on the customer's needs. We differ from 
other "traditional" ones, because we can be not only "additional" factor 
_after_ the username/password (serial authentication), but we can work 
"in parallel" with traditional authentication (when user decided if he 
wants to use his old credentials OR WWPass), and even completely replace 
usernames and passwords (and eliminate associated vulnerabilities for 
good). Then the customer (company which uses our authentication) 
decides, if it is happy with one factor (our token/phone app), requires 
two factors (token plus access code), or even adds biometrics (as many 
factors as it wishes -- behavioristic, fingerprints, retina scan, DNA, 
whatever)

mxxcon:
To some degree there can be some overlap between tfa providers and identity management categories.
Let's see what other folks think as well.

> we position ourselves competing with RSA, Vasco, NagraID, Duo, SecurAuth, and not with Lastpass or 1Pass (which I believe soon be extinct, along with passwords themselves).

To play devil's advocate, there's something to be said about having a single point of failure in your auth system as well :wink:

w3ua:
Sorry for typos -- Sent from my mobile phone with tiny touchscreen keyboard.
Oh sure. If someone puts a gun to your head and "politely" asks you to provide the access code, then logs on to your bank account and transfers all your billions to his account in Nigeria... Wwpass can not help. We only promise that without your passkey your data is unobtainable by design. So far, we are aware of only one alternative with similar claim -- common access card with proper secure element. But cost and inconvenience of that technology is horrific, and at the same time it can only hold a very small amount of secure data, while capacity of our system is virtually unlimited...

The other "weakness" (or strength) - if you lost all your service keys, you lost control over your account. In case availability is more important than security, you need to install a back door. But since you don't use it daily, it can be very inconvenient (and secure).

At the end of the day world is not perfect, and one size never fits all.

Carlgo11:
So @w3ua does this Passkey work with any of the sites listed on <a href="https://twofactorauth.org/">2fa.org</a> or is it just to access a business system?

w3ua:
The Passkey CAN work with ANY site, if the site owner decides to add "Login with WWPass". The addition can take a few minutes for knowledgeable programmer adding it to well supported platform (Wordpress, Drupal, Apache), to a couple of weeks for complex proprietary systems. For example, adding it to a sandbox of one of the members of your "Investing" page (sorry, being under NDA can't disclose the name) took 4 days (some java coding to accommodate some "non-traditional" architecture).
Beauty of WWPass is that you need just ONE token (hardware or phone app) to access ALL WWPass-enabled systems, with guaranteed zero crosstalk (one application would never know anything about another one). If you lose one, you disable it -- and it has no access to anything. Then you create a new one -- and you got back access to everything you had before.

Carlgo11:
hm, well the idea seems awesome but I personally wouldn't trust a security company that doesn't even have https on their site.

Anyway, I don't think your product could fit in the providers criteria since it operates outside of the normal tfa services that the listed <a href="https://twofactorauth.org">sites</a> provides.

From what I understand Passkey's TFA is just to login to the passkey service, correct?
If so then I personally think it would fit better in the <a href="https://twofactorauth.org/#identity">Identity</a> category.

w3ua:
On 9/22/2014 4:41 PM, Carlgo11 wrote:

> hm, well the idea seems awesome but I personally wouldn't trust a
> security company that doesn't even have https on their site.

Kinda short sighted statement, since the web site you see with HTTP is 
purely public, with zero sensitive content. If you go to other our sites 
like https://developers.wwpass.com/ -- they are usually SSL-protected. 
And good luck accessing our intranet https://corp.wwpass.com or other 
resources on http://www.wwpass.com/applications/

Security is not about securing everything -- it's about securing what 
need to be secure ;) Just because there is a war in Syria and Ukraine, 
are you wearing bulletproof vest going out in Seattle?

However, you may be not be the only one falling victim of the idea that 
everything should be encrypted; so thanks for the suggestion, I'll ask 
my web designers to move it to the encrypted domain. For us it makes no 
difference, but for some it may.

> Anyway, I don't think your product could fit in the providers criteria
> since it operates outside of the normal tfa services that the listed
> sites https://twofactorauth.org provides.

But of course! Most of the others are simple OTP add-ons atop of the 
traditional username/password. We offer support of multiple scenarios, 
where username/passwords can be completely eliminated, or kept as back 
door; with 1,2,3 and more various factors...

By the way, what is "providers criteria"? I couldn't find one. I based 
my suggestion purely on similarity of what we are doing with what other 
companies on the list are doing. IQ-test exercise. From what I know 
about services of companies in "identity" list and in "provider" list, 
we are 70 percent aligned with "providers" and 30 percent with "identity".

>  From what I understand Passkey's TFA is just to login to the passkey
> service, correct?

Incorrect. Passkey is used to authenticate the key holder for each 
service provider which uses WWPass service. It is always 3-party game -- 
key holder, service provider (anyone from 2fa.org list plus 1,000,000 
others), and WWPass as third party.

When a key holder presents WWPass at the service provider's web site (or 
door lock, or payment terminal, or any other device where user with 
Passkey need to be authenticated), that web site (or other device) 
establishes connection with both Passkey (which is a cryptocomputer) and 
WWPass servers in the cloud. When both user AND service provider got 
authenticated with WWPass "user front end" and "service provider front 
end" servers respectively, (through 100 percent encrypted channels, with 
private keys never leaving Passkey and Service Provider's premises), 
location of multiple chunks of data, related to the particular pair (key 
holder + service provider) is determined, and the pieces of that data 
get delivered to the service provider's computer. The final data is 
rebuilt on service provider's computer. WWPass is never aware of the 
content of this data. If this data is used for authentication purposes 
(and it is one, but not the only use of WWPass technology -- we can also 
keep encryption keys for data, certificates and other secret stuff), 
then the service provider processes it as user credentials. How they do 
that -- it's their business. We just provide them with SDK, so they can 
integrate WWPass-based authentication with their systems. And then, 
every time a user logs on with WWpass, they will be receiving 
credentials from WWPass and process them accordingly to log the user in 
(or not).

One example - CAS. If you run CAS, you can add authentication with 
WWPass in less than a day -- and then your university will manage all 
logins with WWPass. In this case CAS will be identity manager, while 
WWPass -- authentication engine, if you wish. WWPass has no idea what 
are the identities of the users.

http://www.wwpass.com/wp-content/uploads/WWPass_WhitePaper-HowItWorks.pdf has 
more details

Anyway, it's your list, and you can put us anywhere you want. Or nowhere 
at all if you still don't get it. I tried to explain who we really are 
-- but if I failed ... sorry about that. I understand that our 
technology is more complex than Duo's, Lanchpad's and SecurAuth's, and 
it is my fault if I can not explain it in simple terms. I used to be a 
university professor, and know that if students do not understand the 
lecturer, it's 100 percent lecturer's fault. Students are customers, and 
customer is always right. And at the end of the day, English is not my 
native tongue.

But in any case we are NOT a password manager like 1Password, Lastpass 
and such.... although for support of sites which are not yet 
WWPass-enabled we have our own password manager (which will be trashed 
as soon as people finally realize that authentication without WWPass is 
bad bad bad). Just an interim patch for the imperfect world. It is a 
free gizmo for anyone with Android phone (or real key set, of course). 
IMHO it is not worth mentioning anywhere, but a couple of my programmers 
think otherwise, so I let them build it, so we have it. Like other free 
stuff, including PSS for files, certificate storage for VPN access, 
e-mail encryption and Microsoft domain credentials, TrueCrypt and other 
apps...

Why don't you just try our stuff? It's free to boot. Get an app, 
register as service provider, and add our authentication to some of your 
web sites. Seeing is believing.

w3ua:
Bottom line --- thanks to your comments, we finally moved the whole site to https... two hours worth of work which was postponed for last 2 years ;). it didn't add to security, but at least makes some web crawlers happier. I heard that Google ranks https sites higher in its searches.

Any other suggestions, please... We want to look good in everybody's eyes, even if we do not share the same view on who is a provider.

Carlgo11:
And thank you @w3ua for your patience.
You've been very cooperative. That's something we surely don't see every day.

Btw, I also just changed to https ;) #httpseverywhere!

w3ua:
Being cooperative is a part of job description. As long as it is not against our basic principles and beliefs. In privacy we trust.

PR #782, opened 2014-09-25
Change Mojang's Status to "In Progress" using link

ghost:


mxxcon:
Contradicts a more recent message https://twitter.com/MojangSupport/status/506714959753924609 stating they are only "looking into it", not actively working on it... :confused: 
Those ^MW and ^OZ need to get their story straight.

Carlgo11:
Yup. Until they actually start working on it we won't add a status link.

Hopefully Microsoft implements it now when they own the games

PR #784, opened 2014-09-27
Update Mixpanel - tfa

jbwyme:
https://pbs.twimg.com/media/ByggSiqCIAA9ELf.png

### DIFF
--- a/_data/other.yml
+++ b/_data/other.yml
@@ -67,6 +67,7 @@ websites:
       img: mixpanel.png
       tfa: Yes
       sms: Yes
+      software: Yes
       twitter: mixpanel
 
     - name: Office 365
--- a/_data/other.yml
+++ b/_data/other.yml
@@ -65,7 +65,8 @@ websites:
     - name: Mixpanel
       url: https://mixpanel.com/
       img: mixpanel.png
-      tfa: No
+      tfa: Yes
+      sms: Yes
       twitter: mixpanel
 
     - name: Office 365


mxxcon:
"generated on mobile device" sounds like a mobile app rather than `sms`. @jbwyme can you confirm?

jbwyme:
We use Authy (https://www.authy.com/) but we force the SMS (
https://docs.authy.com/#section-Forcing_SMS) when sending the auth tokens.

mxxcon:
So a user would need both Authy app and support sms or only sms?

jbwyme:
Only sms. You can optionally use the Authy app if you have it installed but it's not required and you'll still receive an sms.

Carlgo11:
@jbwyme so you can use the Authy app instead of reciving texts?

mxxcon:
@Carlgo11 no, SMS is a required part.

jbwyme:
@Carlgo11 @mxxcon With the normal Authy workflow it will only send SMS when the user doesn't have the Authy app installed (and you'd use the app instead). However, you can "force" the SMS which means it'll get sent even if the user does have the Authy app installed. We chose the "force" option so that we have a consistent experience which is particularly helpful for customer support.

dsoegijono:
Shouldn't `software` be marked as `yes` seeing as it works with the Authy app? Even if the SMS is still sent.

mxxcon:
@dsoegijono but then again, it seems to also work w/o..
Afaik we've never had a situation where one of the methods was optional and another mandatory.. :confused:

dsoegijono:
@mxxcon Hmm I think Software should be marked Yes, but add an exception that SMS is mandatory? I think pretty much all of the options listed on the site are optional (like Authy here). This issue is similar as Twitter, and we have listed Software even though SMS is mandatory there too.

jbwyme:
I can mark Software as yes... If you have the Authy app you can ignore the
SMS and use the code the app gives you so I suppose that makes SMS optional
(with the exception that you still received one).

On Sat, Sep 27, 2014 at 11:59 PM, Daphne Soegijono <notifications@github.com

Carlgo11:
I agree with @dsoegijono. Add software

Carlgo11:
In the meantime I'll remove the looks good tag so no one will accidentally merge. Like me for example ;)

jbwyme:
Alright updated so that `software: true`

Carlgo11:
Is there any documentation on how to use Authy on your site? Most sites have some kind of document explaining for their customers how to enable TFA.

jbwyme:
Not at the moment, it just lives in the project settings. If active, it'll
prompt everyone on the project to setup their TFA when they login if they
haven't already.

Carlgo11:
Well I guess it's good to go then.
Remember to add the doc when you make one

Walkman100:
Can't the image ( https://pbs.twimg.com/media/ByggSiqCIAA9ELf.png ) be linked to as the doc?

PR #794, opened 2014-09-30
Add RainLoop and ownCloud

trockenasche:


Carlgo11:
1. OwnCloud doesn't support TFA. There's a plugin that supports it.
   Also, I wouldn't consider OwnCloud to be on the site since it's a program that people have to setup and host themselves. It's not a site therefor owncloud.org doesn't support TFA in any way.
2. No documentation on TFA on RainLoop?
3. All links are http. Use https (if available.)

PR #795, opened 2014-09-30
Add RainLoop and ownCloud

trockenasche:


Carlgo11:
To quote #794: 
1. OwnCloud doesn't support TFA. There's a plugin that supports it.
Also, I wouldn't consider OwnCloud to be on the site since it's a program that people have to setup and host themselves. It's not a site therefore owncloud.org doesn't support TFA in any way.
2. No documentation on TFA on RainLoop?
3. All links are http. Use https (if available.)

trockenasche:
1. I linked the plugin for the doc.
   I be live self hosted program should be right. Rain Loop is also a self hosted program. I want to submit also SSH for TFA. In my opinion is self hosted program or local program also security relevant, thought? Maybe for a new category?
2. No documentation for TFA on RainLoop.
3. The ownCloud plugin link over https bring a crappy page and RainLoop is just over http availble. Only https://ownlcoud.org is useful.

smholloway:
I don't think these sites fit the [site criteria](https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria).

We do not include sites that support 2FA via a third-party plugin. We do not include sites that do not document that they have 2FA -- if we cannot tell that a site offers 2FA, how could anyone else? 

We are looking to promote sites that offer simple, easily accessed 2FA to all users, and I don't think ownCloud or RainLoop do that. I vote to close.

ZacharyDuBois:
(Sorry for popping this PR back up. I would say that there should be a section for software such as ownCloud and RainLoop as with the amount of libraries for MFA, there is no reason they shouldn't have it integrated in the application core.

jamcat22:
Hi there @ZacharyDuBois!
Discussion on this topic should be moved to #1592.

ChristophWurst:
> there is no reason they shouldn't have it integrated in the application core.

http://blog.wuc.me/2016/05/30/adding-two-factor-auth-to-owncloud.html :grinning: :rocket:

ZacharyDuBois:
@ChristophWurst, doesn't matter. They have declared self hosted apps should not be listed via #1592.

PR #799, opened 2014-10-02
Looker doc link update

jsiegz:
We made our docs public, so this pull request updates the link to the two factor page.


Carlgo11:
Looks good.
But your site should really consider changing to https

PR #800, opened 2014-10-02
Added SmartSign

nick-smith-microcosm:


Carlgo11:
What sites use SmartSign?

Carlgo11:
@nick-smith-microcosm any comments? If you don't reply within a couple of days then I'll close this PR because of inactivity.

nick-smith-microcosm:
Apologies for the delay. CopyMinder (http://www.copyminder.com) uses SmartSign 2FA login.

Carlgo11:
@nick-smith-microcosm yes but that's the same company that makes the tfa product right?
Are there any sites that's listed on our site that uses your TFA product?

RichJeanes:
I'm tempted to close this due to inactivity, but then again

Even if there is only one site that supports the SmartSign 2FA right now, I don't see why we should discriminate against them for that. It would just be another line on the providers page and falls under the Software Implementation and email categories.

As far as the CopyMinder site, I don't see a need to add that as a site

mxxcon:
@RichJeanes I agree.

PR #801, opened 2014-10-02
Added America First Credit Union info and logo

seanlane:
Added America First Credit Union info and logo. Their website states that they support "two-factor authentication", by which they mean they will show an image that was previously selected by the account holder along with a message from the account holder, as a means to show that the login portal is legitimate. They also suggest that security questions are another way they implement two-factor authentication, but they are just another variation of a password, something the user knows, rather than requiring access to an authentication app, phone, email account, etc. Their security documentation can be found here: https://www.americafirst.com/about/security/online-mobile-atm/online-security.cfm


jdavis:
Oh dear. That's slightly amusing actually.

PR #802, opened 2014-10-03
Change Amazon's twitter handle

Carlgo11:
I don't think Amazon listens to the Amazon handle.

Let's spam their support instead.

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -1,7 +1,7 @@
 websites:
     - name: Amazon
       url: https://amazon.com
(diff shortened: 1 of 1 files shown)


jdavis:
Let's give it a bit longer. In 2 months we can change it.

Carlgo11:
@jdavis But they haven't replied to a single one of the tweets people have sent them. We use the support accounts on other services listed on the site so why not on amazon?
Plus, isn't the support account the right account to turn to if one want to contact Amazon?

PR #806, opened 2014-10-07
Correct CONTRIBUTING.md

jamcat22:
I don't think "oppropriate" is a word. Correct me if I'm wrong. Changed to "appropriate" :)


mxxcon:
crap. i didn't notice it committed into this master rather than my own :disappointed:

jamcat22:
@mxxcon It's all good! :smile:

PR #809, opened 2014-10-13
Add Fastly CDN

simonwistow:
Full disclosure, I work for Fastly


Carlgo11:
Everything looks good on the first commit. Wow that gotta be a world record! :)

PR #811, opened 2014-10-14
"Copy" twitter account correction

kokotron:
The @copyapp account is not the correct one. The correct one is simply @copy.


copy:
Right

PR #816, opened 2014-10-14
Image verification

ilyakatz:
Add a check to the verification script to make sure that all images are 32x32


jdavis:
A great little addition that I think will help a lot. Great idea @ilyakatz.

Carlgo11:
yay :+1:

ilyakatz:
cool, glad you like it. i added because i messed it up every other time :)

mxxcon:
Is it possible to somehow make those error messages different color or standout a bit more along all the other output?

Carlgo11:
But what about providers? They are often more nooby than the people adding websites.

PR #817, opened 2014-10-15
Addition to #816

Carlgo11:
To start with, I haven't used ruby that much before so I'm still a bit of a noob. So sorry about that.
Never the less it looks like it's doing what it's supposed to do.
1. Added extension check for images. Will fail if not .png (the default extension we seem to be using)
2. Pasted the #816 code and my new one to the providers section.
3. A bit of formatting. Hope no one is against that. (Did it automatically #BlameCarlgo11)
4. updated the images that didn't qualify for the new checks.


jamcat22:
This is awesome!

Carlgo11:
Oh thank you @jamcat22! :D

Carlgo11:
@ilyakatz you seem to be the ruby pro here. Do you approve?

ilyakatz:
@Carlgo11 nice additions! I don't see anything major, maybe would suggest to decrease scope of  output variable from global `$output` to instance `@output`, but it really doesn't matter for this simple script :+1:

Carlgo11:
@ilyakatz so `looks good`?

ilyakatz:
@Carlgo11 yups, added

Carlgo11:
Thank you ^_^

jdavis:
Nice work, @Carlgo11. All these new tests from you and @ilyakatz really simplify our work =]

PR #822, opened 2014-10-16
Update USAA - hardware

cem-:
The Symantec VIP tokens can be either hardware or in software

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -255,6 +255,7 @@ websites:
       img: usaa.png
       tfa: Yes
       sms: Yes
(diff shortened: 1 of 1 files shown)


jamcat22:
:+1:

mxxcon:
It can be, but afaik companies can also pick which method they support. @cem- do you know for sure that they support hardware tokens?

cem-:
@mxxcon yes, I'm 100% sure that they support all the different Symantec VIP tokens as [listed here](https://idprotect.vip.symantec.com/orderstart.v)

Issue #644, opened 2014-07-28
Chase Docs


ChrisMcClendon:
The "Docs" section for Chase follows a mobile link. If you try to convert this to a standard browser link it gives a 404.

mxxcon:
Yes, that's a problem with Chase's confusing website. :confused:
Do you know of a better link we can use?

ChrisMcClendon:
Actually, it appears as though it isn't really two factor. I guess technically it is, but it ONLY works when going through their mobile site and you have no control over the process otherwise.

I'm not quite sure what to do about this one.

jdavis:
One option is to create a Note for this detailing it: https://github.com/jdavis/twofactorauth/tree/master/notes

mxxcon:
@ChrisMcClendon you are mistaken. See #91 and #263

ChrisMcClendon:
Thank you for the clarification. Would a note slightly describing this 'issue' still be appropriate as per @jdavis 's comment?

jdavis:
Some of the people that Chase that have confirmed it would be better people to ping for this. This includes @jfocht and @RichJeanes.

If the current documentation link doesn't suffice for Chase's 2FA, then a custom Note would be better.

jdavis:
I just got an email from Charlie Sobel saying this:

> You have Chase listed on your site as supporting 2FA, but you may want to update that. According to the docs you linked to, they only support 2FA for their mobile site and thru their mobile app, not via the primary website. That's a pretty big oversight on their part.

Like Charlie mentions, the [doc](https://mobilebanking.chase.com/Public/Docs/Faq?nodeId=1&itemId=2) only says that it has 2FA on mobile. Is there anyone with Chase again that can look into this?

jamcat22:
@jdavis Chase supports 2FA on all platforms. All their apps, websites, and mobile websites require 2FA to login from a new computer.

The only doc available is the mobile web doc, but 2FA is required for all users (you cannot opt out and you are "opted in" right when you sign up) on all devices.=

jamcat22:
@jdavis See #91 and #263

mxxcon:
@jdavis This is what 2fa on Chase desktop looks like
![clipboard01](https://cloud.githubusercontent.com/assets/847140/4695784/0bebbb2c-57de-11e4-8cf0-dfa26dc3ba96.png)
[...]

PR #825, opened 2014-10-20
Add new Chase doc

jamcat22:
That old one was confusing. See #644.


jdavis:
Very nice. Sometime we should put some of the [boilerplate](https://github.com/jamcat22/twofactorauth/blob/ab32d7606f6c308da8cd7cbf843bcda57ca955ba/notes/chase.html#L5-L33) into an `_include/` file (there's nothing wrong with this PR, I just happened to notice it).

If someone with Chase can give it a look over and confirm, I think it looks good.

jamcat22:
Does nobody use Chase? :smile:

Issue #832, opened 2014-10-21
Google just added hardware security key support


beriberikix:
http://googleonlinesecurity.blogspot.com/2014/10/strengthening-2-step-verification-with.html

jamcat22:
Thanks very much @beriberikix! Do you want to make a pull request, or would you like me to?

:+1:

mxxcon:
btw, this brings up a point for potential new column. It seems like that FIDO standard is getting some traction. Perhaps it makes sense to tag services that use Fido-compatible 2fa?
(I personally have no opinion yet since I didn't read the specifics.) But maybe this is something we should consider.

beriberikix:
Feel free @jamcat22 :)

jamcat22:
Just created a pr (#834). Thanks @beriberikix!

PR #834, opened 2014-10-23
Add hardware to Google

jamcat22:
As per #832. Tell me if I missed one.


Carlgo11:
Merge all these commits into one and upload that PR.

PR #835, opened 2014-10-23
Add Salesforce

ilyakatz:
I am not entirely sure that the category is correct since SF is so many things


mxxcon:
@ilyakatz Maybe use this icon? It has less jpeg compression artifacts
https://www.sfapex.com/images/sm-appexchange-icon.png

Also in the linked video it says "text message or email".

jdavis:
Of the specific categories, I think Communication is the best. Unless we wanted to make it more general and put it into Other since Salesforce encompasses a lot...

jamcat22:
Also, shouldn't Salesforce go before Sendgrid?

ilyakatz:
@jamcat22 you're completely right. updated

ilyakatz:
@jamcat22 you inspired me to do this #837

ilyakatz:
@mxxcon updated

PR #837, opened 2014-10-24
Sorting validation

ilyakatz:
This PR is an attempt to keep sanity in two ways
- sort all websites in each section
- add validation to the source code to make sure future pull requests keep things nice and organized

Let me know if you think either or both are an overkill

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -19,3 +19,9 @@ exclude:
     - README.md
     - Rakefile
     - verify.rb
+
+jekyll_sort:
+    - src: websites
+      by: name
+      direction: up
+      dest: websites_sorted
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -13,16 +13,16 @@ websites:
       img: altrafcu.png
       tfa: No
 
-    - name: American Express
-      url: https://www.americanexpress.com/
-      twitter: AmericanExpress
-      img: americanexpress.png
-      tfa: No
-
     - name: America First Credit Union
       url: https://www.americafirst.com/
       twitter: AFCU
       img: americafirstcu.png
+      tfa: No
+
+    - name: American Express
+      url: https://www.americanexpress.com/
+      twitter: AmericanExpress
+      img: americanexpress.png
       tfa: No
 
     - name: Bank of America
--- a/_data/bitcoin.yml
+++ b/_data/bitcoin.yml
@@ -36,13 +36,6 @@ websites:
       software: Yes
       doc: https://www.bitstamp.net/s/documents/bitstamp_2_factor_authentication_guide.pdf
 
-    - name: BTC-E
-      url: https://btc-e.com
-      img: btc-e.png
-      tfa: Yes
-      software: Yes
-      doc: https://support.btc-e.com/index.php?/Knowledgebase/Article/View/97/25/how-do-i-set-2-factor-authentication-to-my-btc-e-account
-
     - name: Blockchain
       url: https://blockchain.info/
       img: blockchain.png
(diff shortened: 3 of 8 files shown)


jamcat22:
Thank you for making this! Helps keep me from exploding when things aren't in order :smile:.

Carlgo11:
@ilyakatz don't forget to do the same with providers. Also, looks awesome 😃

mxxcon:
I think fail Travis test for this is a bit of an overkill. When the page gets generated it's sorted alphabetically so that's good for users. But _requiring_ it to be the same in the source code makes contributing less user friendly. It's a "nice to have" but I wouldn't fail a check for this.

ilyakatz:
i'm ok removing the build failures. one advantage is that since this project has a lot of contributors with various degrees of abilities/desires/patience to keep things organized it could help point people in the right direction. 

any strong opinions?

mxxcon:
I agree, but instead of failure can Travis give warnings? A commit can still go in but if user wants they can correct that issue.
Unless I'm mistaken, malformed Yaml block will fail the same way as unsorted entry.. I think that's bit extreme.

Carlgo11:
@mxxcon give warnings? Well that's kind of what it does already doesn't it? And you CAN merge a PR even if travis failed.

jamcat22:
Maybe remove the failures and just have it put the unsorted items under

<------------ WARNING ------------>

instead of 

Carlgo11:
But does anyone really look at the Travis output if it passes?

jamcat22:
I do ;)

Carlgo11:
But we want to keep all the sites alphabetical don't we? And we can still merge even if travis fails. It would just be a question of how important it is to keep that specific PR alphabetical.
I think we should merge

mxxcon:
I have no idea if travis or github supports that, but maybe instead of green checkmark we could have a yellow one? or maybe green checkmark with `*` next to it?
Or just leave it as a manual process...

Carlgo11:
Pretty sure that's not possible. At least not in the current way github services are built

ilyakatz:
although, i implemented it, for me personally, the order in the yml is not that important. i think of it, when i put records into database, the ordering may or may not be what i expect. what is important is what the user sees. just my two cents.

but we need some mediation here, since the house seems divided we need the executive branch, @jdavis, break the tie please. sorry, i'm influenced all the election talk in US yesterday :)

jdavis:
I'm okay with sorting all current ones which you did in this commit: fedc5a9aee389e1f9827c74b7a8f77c4ce478c5f

Regarding sorting sites using a Jekyll plugin, that isn't possible using GitHub Pages unless we started pre-building our pages into the `gh-pages` branch. Since a Jekyll plugin is arbitrary Ruby code, GitHub doesn't support running plugins beyond what they officially support: https://help.github.com/articles/using-jekyll-plugins-with-github-pages/

Lastly, I think the sorting in `verify.rb` is good, but instead of running it every single time and requiring it for CI to pass, what if we instead made it a command line argument (say `--sort`) that looks to see if any sites are unsorted and then prints it out. That way a contributor can just run it locally, fix things, and then issue a PR instead of forcing other contributors to keep things sorted. I think contributing to 2fa.org should be easy, so as long as we validate yaml and the image, we shouldn't go beyond that as any additional stuff is to our (the collaborator's) benefit and not the user (who we should be focusing on).

ilyakatz:
ah, I didn't realize the github pages had these limits. well, in that case, how about i close this request and we work on a different way to sort the entries in the yml files with @Carlgo11 in our secret project?

Carlgo11:
lol "secret" :smile: 
I think all the people in this conversation have been informed about the script already :wink:

ilyakatz:
great, secret's out, it's not fun anymore :)
anyways, closing this request. thanks for your input, everyone

jdavis:
@ilyakatz Feel free to open a new PR with this commit: fedc5a9aee389e1f9827c74b7a8f77c4ce478c5f. You can just do:

``` bash
$ git checkout -b pr <whatever head you want here>
$ git cherry-pick fedc5a9aee389e1f9827c74b7a8f77c4ce478c5f
```

Then push it and issue a pull request.

ilyakatz:
@jdavis that's ok, @Carlgo11 and my script will take care of this stuff, so no need to do a little patch now

jamcat22:
But what if people don't use the script... :D

ilyakatz:
next time someone uses it, it will reorder the file

PR #843, opened 2014-10-24
Adding CryptoPhoto as 2FA provider !

gitmanju:
Hope this works now. Keeping fingers crossed !


Carlgo11:
@gitmanju in what ways does CryptoPhoto offer hardware and phone tfa?

gitmanju:
It provides hardware printed card with the same one that gets shown in the phone. This can be printed on  credit/debit cards in case of banking and in others a plastic card to be carried in the wallet !

Carlgo11:
I wouldn't really call that hardware. What do you think @mxxcon? 

And phone? Does CryptoPhoto call the user?

gitmanju:
Cryptophoto uses phone and shows images on the phone ! Phone has the crypto photo app installed !! The same can be replicated in a printed card !

Carlgo11:
Then you don't support `phone`. That's software. And a photo on a plastic card is not hardware.
Yubikey is hardware, the small machines you get from your bank that gives you a timebased code is hardware.
Please remove hardware and phone

gitmanju:
@carlgo11 what does phone support mean then ? Can you explain ? I understand hardware as in hardware tokens.

Carlgo11:
The `phone` tag is for phone calls. (See <a href="http://twofactorauth.org/">the website</a>).
Hardware is only for hardware tokens which I don't classify your cards as.

gitmanju:
@Carlgo11  let me speak with our CTO and get back ! I want to be sure of this ! Thanks.

mxxcon:
From the video on your website it looks like it's a smartphone app. 
`Phone:` tag is if 2FA handled by a voice phone call where you either talk or press keys. It's also different from `sms:`
If your whole shtick is about showing images, I don't see how phone calls come into play here. :smiley: 
Also @gitmanju go easy on exclamation points, no need to shout. :wink:

gitmanju:
@mxxcon @Carlgo11 Thanks for the info. First of all I am not clear how exclamations got added to my sentences, I just use full stop most of the time. Second I wasn't aware the exclamations meant shouting. Sorry wasn't trying to shout at all :-), it gives me headaches.  
So I guess as per these definitions, we don't support phone/hardware.  Will update the file and resubmit. Just thinking aloud, should the comparison chart be based on attacks it mitigates rather than various forms it supports ? Is there a literature on why these parameters were chosen ? Thanks !

mxxcon:
@gitmanju what do you mean by "chart based on attacks it mitigates"?

We settled on these categories because they cover all the possible 2FA methods we could think of, yet they are general enough that we wouldn't have to create new columns as new solutions/companies/projects appear on the market.

We strive for this site to be first and foremost a useful resource to general public. For people to quickly check if services they use support 2FA, what kind and what steps they need to take to enable it, and if not, publicly shame and request services to support 2FA.

gitmanju:
@mxxcon Understood. Thanks. As you are aware, the strength/usefulness of 2FA is determined by number of attacks (MITM, MITB, dos) it mitigates along with number of channels/techniques it supports. So I was wondering if the 2FA providers chart should depict that information rather than just the modes (like hardware, sms) which may or may not be indicators of level of security/ease of use. But I do understand if you think this is too complex for the purposes the site serves.

mxxcon:
Other than LastPass, I can't think of any other site/service that offers users multiple choice of 2FA providers.
What difference does it make if I know that RFC 6238 doesn't protect against denial of service attacks? It's either take it or leave..Use 2FA that they offer or don't use it at all.

gitmanju:
@mxxcon this will be useful from the 2FA customers (businesses) or CISOs to be precise. It would be useful to see a menu of all providers and how many will address their needs of various security issues. 

I understand the users(end users who login) may not care too much about it if it doesn't affect them directly.

smholloway:
> the strength/usefulness of 2FA is determined by number of attacks (MITM, MITB, dos) it mitigates along with number of channels/techniques it supports.

That is an overly simplified opinion of security and it ignores several factors, including the likelihood of an attack and the likelihood that people are properly using the product. 2FA that doesn't get used is not useful, so I think the ease of use is most important. Every provider will have their own definition of the strength that plays to their strengths.

Thinking about the logistics of adding this additional information, how would we objectively determine which attacks should be listed? [Wikipedia has a very coarse list](http://en.wikipedia.org/wiki/Attack_%28computing%29#Types_of_attacks), [OWASP has an opinion about the most probable attacks](https://www.owasp.org/index.php/Top10#OWASP_Top_10_for_2013), and [Schneier presents another perspective based on the attacker's goal](https://www.schneier.com/paper-attacktrees-ddj-ft.html). So, just creating a fair list of attacks would be hard, then we have to determine if each provider can truly mitigate that attack. How would we objectively determine if a provider mitigates the attack? Anyone could say that they protect against any arbitrary attack. Policing this would be draining.

This site is not meant to make a CISO's job easier. If you'd like such a site you can fork this repository and create `TwoFactorAuthProviders.org`, which would likely be thought of as `CryptoPhotosBiasedListOfTwoFactorAuthProviders.org`. This site is useful because we strive to be useful and unbiased.

gitmanju:
@smholloway completely unwarranted last two lines. You could have stopped at "This site is not meant to make CISCO's job easier". Just trying to have a discussion with an intent to understand.  Thanks.

jamcat22:
Before this gets out of hand :smile:, is this pull request ready to merge?

gitmanju:
@jamcat22 :-), yes !

gitmanju:
@mxxcon thank. I am assuming it will take few hours for the changes to show up. Just checking, thanks.

mxxcon:
It's a manual process for Josh to publish the committed data on the actual website.
Depending on his free time it could be now, it could be tomorrow, it could be in a couple of days. We are all volunteers here and do it whenever we have time/desire.

PR #847, opened 2014-10-26
Add LevelUp

ilyakatz:


_data/finance.yml
```diff
@@ -35,6 +35,12 @@ websites:
       img: kiva.png
       tfa: No
 
+    - name: LevelUp
+      url: https://www.thelevelup.com
+      twitter: levelupcom
```
Carlgo11: That's not the right twitter handle. That account is for some Spanish gaming news site (http://www.levelup.com/).

ilyakatz:
yes, you're correct. I must be seeing things, even the color scheme is drastically different :) fixed

Carlgo11:
Eh we all make mistakes. Btw, I sent you a mail (I think). Can you confirm you got it?

ilyakatz:
@Carlgo11 yups, I see it. Replying.

PR #857, opened 2014-11-01
Added TARGOBANK

inverse:
Title says it all.


_data/banking.yml
```diff
@@ -235,6 +235,12 @@ websites:
       software: Yes
       doc: http://hjalp.swedbank.se/logga-in/index.htm
 
+    - name: TARGOBANK
+      url: https://www.targobank.de/de/index.html
```
Carlgo11: How about changing this to `https://www.targobank.de/` in case they decide to switch from html.

inverse: Agreed :+1:

PR #859, opened 2014-11-02
Edit addsite.rb script logic

fpigerre:
Edit addsite.rb script logic to ensure that links typed include an http prefix, remove unused brackets, place results inside an array.


Carlgo11:
Awesome, thanks :)

PR #860, opened 2014-11-02
Add Gemfile.lock to .gitignore

fpigerre:
Add Gemfile.lock to .gitignore to prevent committing Gem parameters specific to certain platforms.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ _site/
 *.DS_Store
 .idea/
 */Thumbs.db
+Gemfile.lock


ilyakatz:
Actually, i think it should be added to git since the idea is to use the same versions of gems as used in production to avoid conflicts

jdavis:
Yup, I agree with @ilyakatz. Here's a write up about it: http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

jdavis:
Thanks for the idea though =]

PR #861, opened 2014-11-04
Add Aerohive Networks

jdavis:
I received an email requesting that they be added to the list.


jamcat22:
:+1: http://getsatisfaction.com/aerohive/topics/does-aerohive-support-two-factor-authentication

PR #867, opened 2014-11-13
Added email provider Posteo.

rose85dol:


Carlgo11:
`"Alternatively, you could use special hardware, such as a Yubikey."` Looks like they support hardware as well

rose85dol:
It's TOTP, so yes, you can use TOTP (RFC6238) compatible hardware.

Would love to have the columns "FIDO U2F ready" and "TOTP (RFC6238)"!

Carlgo11:
This has been brought up already and we've decided not to add this (at least for now).

mxxcon:
@Carlgo11 I'd be hesitant to add `hardware` parameter.
While it says it support Yubikey, it sounds more like Yubikey implemented some mechanism to program TOTP into your yubikey device...at least that's my understanding. As such I'm not sure if we'd have to mark every other site with `hardware: yes`.

Carlgo11:
@mxxcon Companies add the craziest "security" ideas nowadays don't they?
What happened to the, "don't try to make your own login system"?!

And they describe their security in vague and incomprehensible ways.
I've even got the response from big companies "your password is safe but we can't tell you how". Right...

RichJeanes:
Not sure whether this is a concern for the purposes of this site, but they allow for multiple devices to be set up by showing the secret key again. I've never seen any other site that allows this. I guess if you set up multiple devices at once, there's no way they could stop you, but I've never seen or heard of a site re-revealing your secret key to you...

Carlgo11:
@RichJeanes developers do the craziest things :wink:

Carlgo11:
As far as 2FA the site seems to have it tho so I'd say it's ready to merge.

RichJeanes:
I would have to agree.

Carlgo11:
We sure are productive today!

RichJeanes:
For sure. I've been away for a while but recently got a job with some downtime so I'm trying to get back into programming and contributing to projects.

beerisgood:
Using a Nitrokey Pro works too!

So i would say the checkbox for hardware tokens can be set on the list.

Issue #763, opened 2014-09-09
OVH Don't support 2FA


ghost:
Despite the link in the site saying OVH have 2FA - This is wrong.

I got on to them about this, and here's their email.

Put a button/link there asking them to implement this!

![...](https://s3.amazonaws.com/f.cl.ly/items/3Q0W0j0L042n2v3x2J3P/a.png)

mxxcon:
@SoHiggo can you ask them what's the meaning of https://www.ovh.ie/aboutus/security.xml#double-authentification then?

ghost:
I think that's only for certain types of services they offer, like VPS. The standard hosting cpanel does not offer 2FA :(

mxxcon:
@SoHiggo Would you be able to get a bit more details on that so that we can update the site with accurate exception?

ghost:
I sent this http://s.higg.im/image/0S41052X2641

Just because it says it on the site doesn't mean it's accurate. It could be they once had it, but disabled it because I think they're migrating all customers to a new control panel...

Also - if enough people bombard them about this issue, they might feverishly push forward with it:

https://www.ovh.ie/support/

/ @jdavis

RichJeanes:
This thread would seem to be superseded by #870

PR #870, opened 2014-11-14
update OVH exceptions

pierreprinetti:
OVH only provides secured accounts to its French customers.

This was confirmed by their Italian support team, on a private ticket I opened 2014-10-15.

I was tempted to simply switch "tfa" to "No", given they sell their IaaS worldwide.

### DIFF
--- a/_data/hosting.yml
+++ b/_data/hosting.yml
@@ -93,7 +93,7 @@ websites:
       tfa: Yes
       sms: Yes
       exceptions:
-        text: "2FA is only available for the customers who registered on the french version of the website."
+        text: "2FA is only available for the customers who registered on the French version of the website."
       doc: http://www.ovh.com/us/about-us/security.xml#double-authentification
 
     - name: Rackspace
--- a/_data/hosting.yml
+++ b/_data/hosting.yml
@@ -92,6 +92,8 @@ websites:
       img: ovh.png
       tfa: Yes
       sms: Yes
+      exceptions:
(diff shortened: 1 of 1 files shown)


pierreprinetti:
et voilà

mxxcon:
It seems like OVH is inconsistent about what they support in different countries. See #763 and #583

pierreprinetti:
> It seems like OVH is inconsistent about what they support in different countries.

@mxxcon that's precisely the reason for this pull request. I had a conversation about this with their support. They told me TFA was only available for French customers (i.e. people who registered on the French portal), and that they don't have any plan to extend TFA to other countries.

The Italian law prevents me from posting here a screenshot of the conversation I had with the support (anyway that would have been in Italian).

The point is: OVH implemented TFA only for a fraction of their customers (advertising it as a feature in the process, but that's another story). So, should we mark it as implemented, but with a note (as in my pull request), or directly mark the feature as non implemented?

RichJeanes:
I would say that this change looks good.

@mxxcon As far as your comment from #763, your link to https://www.ovh.ie/aboutus/security.xml#double-authentification seems to lead to a non-existent 
section... Makes me think that maybe they're trying to clear up their 2FA support.

mxxcon:
Yes, there used to be a 2fa section between "Data confidentiality" and "Find out more on video". see https://web.archive.org/web/20140905101412/http://www.ovh.com/us/about-us/security.xml

PR #872, opened 2014-11-16
Standardize date formatting

cmahns:
Changed the date format for an exception in gaming.yml and in cloud.yml. Not everyone reading the site is an American, and using American or European date standards can lead to confusion between geographic areas (09/04/2014 for example -- is this saying September 4th or Ninth of April?). It's better to use ISO 8601 to alleviate this :)


mxxcon:
I agree except the 2 things you changed are pretty unambiguous. :smiley: 
You can't confuse `Jun 5, 2014` with anything else. Similarly, we don't have 29th month. :wink: 
For places where sorting might apply, yes, consistent date stamp is important. I don't know if it matters as much for plain text fields like exceptions. As a matter of fact, I think `Jun(e) 5, 2014` is even more user-friendly, than `2014/07/05`.

cmahns:
The only two things I could change were those two examples, most of the text has (luckily) lacked dates so far!

I agree that Jun 5, 2014 would be more user-friendly than using 8601, but having the month before the date is still something that is done in a certain geographic location. My friends in Europe would argue that this example should be 05 Jun 2014. My rationale for using the ISO standard is that it's the international standard of course ;)

Should this should be up for discussion as to how this should be handled? The initial reason for this commit was to have all dates use the same convention and to provide a standard for this site. While I would like everything to be using the ISO, as long as it's consistent I don't care.

Carlgo11:
@mxxcon you crazy muricans still resist the logic system I see. DD/MM/YYYY makes much more since than MM/DD/YYYY don't you think? :wink:

dsoegijono:
I also prefer June 5, 2014 for easier reading. IMO it's better to use something tied to certain geographic location while still understood by others, than using a standard that isn't really used by anyone. I think that the ISO standard would still be confusing to most people. Is 2014/09/04 YYYY/MM/DD or YYYY/DD/MM? Most people wouldn't know what an ISO standard is.

cmahns:
Updated the two files to match with the feedback from above, and changed the issue name to be more in line with what was discussed.

PR #873, opened 2014-11-17
Ally has disabled 2FA?

jdavis:
I was informed through an email that Ally has removed 2FA.

I verified by using a completely different browser, turning on a VPN tunnel and logging in. Unless they are using another way to fingerprint "new computers", I can't get Ally to trigger 2FA.

Ally is also working on a new redesign so I wonder if this is part of that.

I might email them as well since I use them and picked Ally because they were online only and had 2FA... :/

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -1,11 +1,9 @@
 websites:
     - name: Ally Bank
       url: http://www.ally.com/
+      twitter: allybank
       img: allybank.png
-      tfa: Yes
-      sms: Yes
-      email: Yes
-      doc: http://www.ally.com/messages/sms/index.html
+      tfa: No
 
     - name: Altra Federal Credit Union
       url: https://www.altra.org/


mxxcon:
So is it clear to merge or wait until you contact their support?

jdavis:
Let's wait until I hear back.

ilyakatz:
So annoying. Yes, I think they screwed it up. I called them about this, obviously the representative had no idea. Someone even called me, and when I logged in, I was presented to with a way to reset how "codes" are delivered. So apparently they just switched my account and didn't know what I was talking about. I think it's safe to say they no longer support 2FA :(

jdavis:
I almost think it might be worth it to add an exception icon next to them just noting that they have recently disabled it.

Either way, I'm going to wait to see what my contact with them says before I continue.

jdavis:
Maybe some of you can decipher what this means. I asked why 2FA had been disabled for my account:

> I apologize for the late response to your message. When you set up the security for your account it should have asked you if you would like a security code sent to your email or via text message. It would have then asked if the device you are using is a private or public device. If you select that it is not private, it should be sending a security code each time you log in. You would have to enter this code along with your log in information to have access to your online account. You can have this reset to ensure that you receive a security code when logging in. Please give us a call and we will be able to reset this security feature so when you log in, you will be able to select how you want to receive a code and make sure you select that it is not a private device.
> 
> The new security code feature helps enhance our security program to safeguard your account information, to keep pace with the most advanced security technology, and our customers? banking experience. We want to offer the best protection for our customers as the online landscape has changed. You are still secure on the Ally Bank website.

dsoegijono:
@jdavis Did you have it set up for email or text message previously? The way I understand it is that they recently changed how it works (now it's email vs text message), therefore if you set it up before the change they don't have that info saved for your account. So when you log in, not knowing if they should send the code via email or text message, they think that you didn't have 2FA set up, or something like that? So I think you should try setting it up again for your account.

ilyakatz:
It seems to work in some cases, I just tried from my phone

Alt text

Carlgo11:
Maybe it's regional based?

jdavis:
It worked for me the other day too. I'm not sure; I'm closing this.

PR #877, opened 2014-11-20
Add Coinapult

g-p-g:


g-p-g:
If it helps, here's a tweet about it: https://twitter.com/Coinapult/status/535281341890592768. 2FA is used during login and also after login for specific operations like withdrawals.

PR #881, opened 2014-11-24
Add SecurePass provider

gpaterno:
Added SecurePass as a 2fa provider


mxxcon:
Or should it be under https://twofactorauth.org/#identity ?

gpaterno:
No, in identity are listed more password management, this is a more comprehensive solution and can be used as a 2FA in the cloud.

That would be http://www.bankofpasswords.ch/ that is from the same company.

mxxcon:
Well, as you can see from the category name, it's called "Identity Management", it's more than just password managers.

gpaterno:
Do you think it can be in both of them? If yes, I will update both.

jdavis:
This to me looks like it belongs more in the provider section than the identity section. All I see on the site is that it supports 2FA/OTP and provides a SSO solution while making it easy to manage.

That's just what I gathered from looking through the site so I totally could be wrong. If that is the case though, I'd say this `looks good`.

mxxcon:
¯_(ツ)_/¯

jdavis:
@mxxcon You usually have more knowledge about these things than I do, haha. I just took a cursory look at the site so if you disagree, I'd love to hear it =]

PR #885, opened 2014-11-27
Add hardware 2FA note to PayPal

### LINKED ISSUE #880: Paypal no longer offers 2FA hardware
The Paypal hardware two-factor auth token is no longer for sale.  It continues to work for those who still have one, but they are not distributing new or replacement untis. They aren't clear about this on the web site, but I emailed customer service and got a straight answer.

They say they're lookign at other options for the future and will put you on a notification list (perhaps that's in reference to the FIDO U2F stuff), but there needs to be a note added to their entry on this site explaining that the hardware option has been discontinued. At the very least; it probably should be marked as a "no".

fpigerre:
Add a note to PayPal's section in _data\payments.yml, explaining that
hardware based two-factor authentication is no longer available. Fixes
#880.

### DIFF
--- a/_data/payments.yml
+++ b/_data/payments.yml
@@ -50,7 +50,7 @@ websites:
       software: Yes
       hardware: Yes
       exceptions:
-          text: "2FA only available in the US and Canada. SMS only available in the US."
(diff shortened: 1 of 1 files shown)


Carlgo11:
or just remove `hardware: Yes`?

PR #887, opened 2014-11-27
Add TeddyID

warlight:
adding TeddyID info


Carlgo11:
You need to add a logo. See the guidelines: https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#guidelines

Carlgo11:
They call it <a href="https://www.teddyid.com/sites.php#1.5factor">1.5 factor authentication</a>. I would guess it doesn't qualify as 2 factor authentication.

warlight:
1.5 factor authentication is as option. TeddyID has option of half and full integration. 
https://www.teddyid.com/developers.php

warlight:
Also you can look for index page. TeddyID uses two factors. For example - browser&phone. I have trouble to add icon at web.
http://pikucha.ru/idOFi

Carlgo11:
You can't add images via github.com. Clone the branch in a git client locally and add the image there

warlight:
And now, I pushed png. Everything is ok?

Carlgo11:
You need to push to `warlight:patch-1`

Carlgo11:
now you've removed the other lines

jamcat22:
Updated title to Add TeddyID. Carry on. :D

warlight:
We have some problems with our goverment that blocked github for 2 days...)

Carlgo11:
lol. China? :smile:

warlight:
Russia...

Carlgo11:
In soviet Russia GitHub fork youuuu!
Please use VPN or Proxy when in countries like Russia, China, Cuba etc...
Don't let them control the web!

warlight:
Back to USSR..

Carlgo11:
I'll stop. Please don't invade us with those "submarines" :wink: 


Nah but the PR looks good. I'd perfer a doc link tho. Is there one you can add?

warlight:
May be this?
http://teddyid.local/downloads/TeddyAPI-en.docx

Carlgo11:
uhm "teddyid.local"

warlight:
https://www.teddyid.com/downloads/TeddyAPI-en.docx
sorry )

Carlgo11:
oh right. it's a provider. never mind then.

warlight:
It seems to me that TeddyID also is Identity Management as LastPass and OneLogin.

RichJeanes:
From what I can see on the site, this is not 2FA at all, but rather a password replacement system/password manager side-product. They claim it is secure because it requires your phone to log in to a site on your computer, but it only requires one authentication factor, your phone OR your password if your phone is dead/lost/etc.
It would seem that the only time they require more than one is if you have lost access to your account and are trying to get it back. (http://www.teddyid.com/ under "Lost Access?" -> Learn More)

warlight:
RichJeanes, you are wrong.
This is one option of use TeddyID. Read please the whole options:
https://www.teddyid.com/developers.php
https://www.teddyid.com/downloads/TeddyAPI-en.docx

Carlgo11:
@warlight does any sites actually use this? Keep in mind that our site is meant for consumers and not developers.

warlight:
Some sites use it. For example https://www.billinger.ru/auth/

Carlgo11:
https://www.billinger.ru/ is an empty page with a beer bottle on it.
Is there any sites that's listed on our site that use teddyid?

warlight:
Really?)
https://www.billinger.ru/auth/ but not https://www.billinger.ru/
http://www.masterwebs.ru/

Carlgo11:
"Is there any sites that's listed on our site that use teddyid?"

warlight:
Any site, that uses openid (for example github)

Carlgo11:
I mean native support. For example Authy supports GA otp and so it supports almost all TFA sites. and so then we'd mark every single site on the list as that it supports phone calls and SMS. We've chosen only to do that to the sites that require Authy and not just GA that could be used with Authy.

warlight:
What does "GA" mean?

Carlgo11:
Google Authenticator

warlight:
But where is it written that TFA provider has to be in list? Where can I read the demands to be in list of TFA providers?

Carlgo11:
@warlight It's not a 'rule' in that matter but showing a list of websites supporting TFA and then telling the people to use one of the TFA providers of these sites kind of defeats the purpose if the TFA providers aren't supported by any site listed there.

mxxcon:
@Carlgo11, I'm not sure how I feel about approach. With that mindset we'd be favoring only larger and established providers and not new upcoming ones... :confused:

jamcat22:
Agree with @mxxcon on this one.

Carlgo11:
okay 2vs1 on this one. You win :wink:

Walkman100:
image
https://www.billinger.ru/auth/
It appears you have to go to another page then click the back button for it to show.

RichJeanes:
I still don't think this is 2FA. I'm sticking with my previous point that this doesn't actually require TWO factors. @Walkman100's screenshot kinda proves my point where it says "**Or** log in with password" (emphasis mine). Requiring only one or the other is only one factor.

I have looked over the documentation you've provided and none of it refers to 2FA. OAuth, SAML and OpenID are not 2FA schemes, they are cross-domain authentication mechanisms. This list of supported protocols is for sites to communicate with the TeddyID servers for user authentication.

Please see [HERE](http://en.wikipedia.org/wiki/Two_factor_authentication#Components) for what are considered "factors" in 2FA. Quote (emphasis mine) - "The use of two-factor authentication to prove one's identity is based on the approach that **both the required factors must be used** and must be correct"

Since TeddyID only requires your password OR your phone (**not both**), it is only one factor. 
No where on their site do they claim to be a 2FA provider. Even on the table comparing their password manager functionality ("1.5 factor" as they call it) to their full feature set, they label it as "Integration with Teddy API or plugins", not Two Factor. While outsourcing user authentication (a la free-to-use OAuth, SAML, OpenID, or TeddyID's own proprietary API) is generally considered more secure for small sites to do instead of trying to write their own authentication mechanisms, it is still not 2FA.

warlight:
RichJeanes, These protocols are options. The same is using site authorisation without TeddyID.
2FA documentation is https://www.teddyid.com/downloads/TeddyAPI-en.docx
billinger uses two ways auth and one of them - 2FA TeddyID. May be you will try it?

RichJeanes:
I've looked over this documentation and it only describes how to use Teddy's API instead of one of the other cross-domain authentication protocols. If you can point to exactly where in there it mentions 2FA, I would appreciate that because I can't find it.
The Teddy scheme as far as I understand it works similar to Authy (pushing a notification to your phone), but it doesn't require an account password first, removing an authentication factor (that of knowledge) and replacing it with another (that of possession), instead of supplementing it (which is called 2FA).
I have looked at the Billinger site and it uses Teddy exactly as I understand it. You can have the login attempt pushed to your phone via the Teddy app OR login with your username and password. Again, each login type only requires ONE factor to log in.

warlight:
Yes, TeddyID doesn't demand your password. But if it is so - it means that you have already login, and your browser has cookies. TeddyID uses as factor your browser (cookies), your device (you need to confirm action by tapping), your email.
If you want to login by TeddyID on another browser - you will go through the steps of authorization from begining (QR-code, confirming email).
We think that using login and password - very uncomfortably for users. Try TeddyID on the private mode on your browser. Or try it on another browser.

Carlgo11:
What you're basically saying is that you think the first factor in two factor authentication is bad and that teddyid uses 1FA. Am I right?

warlight:
1 - browser (authorisation on teddyid.com, confirming email)
2 - device.
Why 1FA?

RichJeanes:
Confirming an email address is not an authentication factor. A cookie in a browser is not an authentication factor (even if it was, the cookie identifies the device it is on, a factor of possession, duplicating that factor from having you phone, still not two factor).
Although Wikipedia lists a username as a knowledge factor, usernames (and email addresses that are treated as such) do not provide authentication security because they are not secret and known only by their owner, therefore do not count as a factor for the purposes of this site.

warlight:
Password also may be known to a third party.
So, you think, that password is mandatory factor for 2FA?

mxxcon:
Passwords would be known only after a security compromise, where as usernames not necessarily.

jamcat22:
Thought we should have a label for providers too.

Carry on.

warlight:
Now TeddyID app has opportunity to set pincode. If you think, that authorizated site on browser is not sufficient factor.

Carlgo11:
I don't see how an app pin is a second factor of authentication. If it where a pin for the actual sites then I would consider it that but it's basically one factor with another layer on that factor.

Like a door for example, say it has a padlock on it that requires a key. It doesn't have a biometrics scanner but the key for the padlock is stored in a box with a biometris scanner.
Sure it makes the key less likely to be used unauthorized but if you break the padlock on the door it doesn't help that the key used this extra layer.

tonyteddyid:
@Carlgo11 the padlock example doesn't apply here because if PIN is enabled in TeddyID, you can't bypass it. The PIN is not stored anywhere, it is sourced into the secrets used to authenticate the device to the server. So you have to both know the PIN (knowledge factor) and have the mobile device (possession factor) - strictly in line with the academic definition of 2FA.

Taking broader and more practical view, we are in the business of protecting user data, and all factors have their value. Our primary use case is: computer (technically, browser cookie) + mobile device (technically, secrets stored in the app) + optional PIN code. Even without the PIN code, we have two possession factors - albeit not academic 2FA, the two factors do their job of protecting user data as long as the two possession factors are **independent**.

Re "log in with password" option, some users don't have a password and can't use this option. Even if you use it, we'll check your browser (a possession factor) and if it is not paired to your account, we'll ask you to confirm this and that after you enter the correct password. Think of it as account recovery option that is integrated into sign in flow.

warlight:
@Carlgo11, what do you think now?

Carlgo11:
@tonyteddyid I'm not talking about storage but about which party the PIN is required for. If the website using TeddyID only uses that then I don't see that it uses two factor authentication. On the other hand if the website would require a PIN then it is two factor authentication (even tho it's a really weak one).

What I'm trying to say is that sites are vulnerable to attacks. Sure, maybe your product is secure from everything (highly unlikely) but the way the sites integrate your service means that it's still just one authentication.  Getting one go ahead by the TeddyID API still means that it's got <b>one</b> go ahead.

Please note that I'm not saying TeddyID isn't a TFA provider (I'll let the others decide that). I'm just saying the PIN isn't a second factor.

Also, please note that almost all other TFA apps have an option to enable PIN protection.
Here for example is my TFA provider of choice: Authy

tonyteddyid:
@Carlgo11 I think you misunderstood. The website that relies on our authentication doesn't receive the PIN. In fact they don't know anything about the PIN. The PIN is used to unlock the app, so without the PIN, the app is unusable. You have to both know the PIN and hold the phone with the app installed - 2 factors.

Carlgo11:
No I understand that.

tonyteddyid:
@Carlgo11 then I don't understand why you say PIN isn't a second factor..

Carlgo11:
PIN is a second factor but not when used on the app. If the service using TeddyID asked for a PIN it would be a very weak but still a second factor. Now the site just asks for TeddyID to verify and if it gets the verification from TeddyID you're in. That verification is sent once is it not? I see that as one factor.

On a theoretical level, if a normal site with tfa asked for a password and you somehow bypassed the password checking you would need to enter the tfa. If you bypass the TeddyID verification you're in. There's no second factor stopping the attacker. It's just a TFA client asking the user for the PIN.

I think that as long a PIN requirement is on the client side and not at all enforced at the server side it's just one factor.

mxxcon:
In other words, this situation would be something long as the lines of a site using regular username/password, and then considering master password to LastPass or 1Password, which store that username/password to be that "2nd factor".
You are authenticating in different places.

tonyteddyid:
@Carlgo11 thanks I see what you mean now.
When a site delegates authentication to TeddyID, there are at least 2 customer facing factors, both verified by Teddy, but the site receives only one "OK" - from TeddyID. You say this one link can be attacked and broken. Remember two things:
1. The site is free to add as many additional verifications as it wants, TeddyID can be just one of them. Think of TeddyID as provider of the second factor, like GA or Authy. But it is not really required, see next point.
2. The point of 2FA is to protect against human error. Humans use weak passwords, are easy victims to phishing attacks, install malware on their computers, etc. In those situations, a second factor definitely helps. That's why I think when talking about MFA we should count **customer facing factors**. The TeddyID-to-site link is machine-to-machine communication and if implemented properly it is immune to the above types of human error. In machine-to-machine communication, the nature of threats is different, and they are rarely addressed by adding more auth factors.

Carlgo11:
@tonyteddyid Yes I agree with you on that TeddyID is just like Authy or GA but the same was as GA and Authy by themselves are not two factor authentication but 1 factor to be added with another. **Even tho they provide PIN code features**
I'm not saying TeddyID isn't a TFA provider. I'm just arguing that @warlight's statement that PIN codes would be an extra factor is invalid.  

I disagree with you that 2FA is only to protect against human errors. It just adds another layer of security. It's not designed to just stop humans in the way Captchas are designed to stop machines.

I also disagree with you that `more layers of authentication doesn't prevent attacks in machine-to-machine communication`. Almost all attacks on the internet are intended for machine-to-machine communication. The NSA is eavesdropping on machine-to-machine communication and by adding TLS/SSL that can be prevented. TLS/SSL is a second form of authentication (Not to be mistaken for two factor authentication). Machines doesn't question things. They just do as they're told to (<a href="http://time.com/3614349/artificial-intelligence-singularity-stephen-hawking-elon-musk/">for now</a>). That's why we use encryption for so many things nowadays. If the machines could determine what is a reasonable command and what is not, most attacks would vanish.  
If you think that machine-to-machine communication doesn't need any layers of authentication I urge you to stop developing security related software.

P.S. `Humans use weak passwords` is a very broad statement. I, as a computer security researcher and privacy geek, use what ever is available. I let computers make up my passwords. The only thing hindering me from using really good passwords are sites limiting the password lengths. Since I only use computer generated passwords I have the same password strength as a computer would if it could do the same things (once again referring to A.I:s :wink:).
Passwords are an old form of authentication and much weaker than Private/Public keys for example but it doesn't mean humans have any weaker passwords than what a machine would. A machine simply can't have better passwords than what's allowed by length, variety etc by the service asking for the password.
That's another reason why TFA strengthens the security. And once again, not a human error but a policy related issue (Often the length policy is applied to stop DoS attacks because of machine limitations and not human errors).

tonyteddyid:
@Carlgo11 to make it clear, layers is not the same as factors of authentication (knowledge, possession, and inherence).
I was glad to talk to you but I'm afraid we lost focus on the original subject.
Is there anything else you need us to clarify before merging our pull request?

tonyteddyid:
@Carlgo11 ?

Carlgo11:
@tonyteddyid please have patience.
I'm personally against TeddyID being added to the site and so I won't merge this PR.
Instead I'll let another collaborator make the choice whether to merge this or not.
All collaborators are volunteers and do this at our spare time.

Also, please don't highlight anyone if not necessary. I see that as spam.

Thank you,
//Carlgo11

RichJeanes:
I agree that this thread wandered pretty far off the topic of "will this pull be merged?" and needs to be brought to a close.
I'm going to concur with @Carlgo11. You say that TeddyID doesn't meet the "academic definition" of TFA, but we need to have solid guidelines for people to understand what we're looking for.

tonyteddyid:
@RichJeanes I want to remind you that you can plug TeddyID as an additional (possession) factor on top of existing (usually knowledge) factors. In this use case, we are similar to GA, which is already listed on your site and which similarly doesn't provide the full set of two factors, just the 2nd factor.

This is not how we see our primary use case though.

Going slightly off topic, I want to say that often there are no clearly cut lines between different types of authentication factors. For example, passwords (which I hate) are intended to be knowledge factors, but if you look at how most people really use them, they save the passwords in their browsers or write them on post-it notes. Then your laptop or your post-it note becomes a possession factor: if somebody gets hold of it, he can use your password. Conversely, soft tokens, such as GA, are thought to be possession factors, but their security hinges on a key stored in their memory. If somebody gets knowledge of the key (it is easy if you use a rooted Android or jailbroken iPhone), he can produce a valid OTP. (Hardware tokens that have unextractable keys are better in this respect, they are true possession factors.) So this requirement that auth factors should be from different categories (knowledge, possession, inherence) has little practical value.

RichJeanes:
The purpose of the Providers page is to, in fact, list providers of the second factor; in that you are correct. But you destroy your own argument by saying that the intended use of your product is _not_ to be used that way, therefore I believe that it should not be listed.

And I agree that we do not live in a perfectly secure world. If we did, there would be no need for second factors or such. But as I have stated, I think we should be working off the "academic" definition of factors in order to provide clear cut guidelines to our contributors.

(I'll risk chasing you down the rabbit hole a bit here)<DISCLAIMED<
As far as software tokens on an Android phone go, yes having root gives full access to the file system, but only with access to the hardware. I am not as intimately familiar with Apple products, but I would imagine it is a very similar case. However, at this point I would say it is more secure than a dedicated hardware token. With my phone, "offline" access (through something like ADB or recovery, not through the OS) is (practically) impossible thanks to filesystem encryption (highly recommended, supported optionally as of Android 2.x (pretty sure, forgive me for not Googing an exact answer) and enabled permanently as of 5.0) and "online" access requires knowledge of my phone PIN/Password _and_ TFA app PIN (assuming they got the phone alive, kept it charged, and aren't dumb enough to reboot it, otherwise see "filesystem encryption"). And if you have a rooted phone, you're probably savvy enough that it's either encrypted or is used for development/testing and doesn't carry personal info.
With dedicated hardware it's "press button, get number". Though, yes they are sometimes protected by a PIN, therefore _could_ be considered "as safe as" an app with a PIN (or a phone with PIN/no app PIN, you get the idea). But tack on a phone PIN plus an app PIN and now there are two independent layers that the bad guy has to go through to get to my numbers (plus the possibility of encryption). My point being that the issue of possession of token vs knowledge of key becomes moot in any practical sense by making access to said token/key difficult, securing the factor, not only physically, but also behind other independent authentication factors. </DISCLAIMED<
(DISCLAIMER: Please understand that I make this statement as personal preference and opinion and has nothing to do with the goals or guidelines of the twofactorauth.org site, project, or any of its other contributing members.)

tonyteddyid:
Well, we never actually say "do not use TeddyID as a 2nd factor", in fact our product is already used as a second factor by a small bitcoin site and our current offer for banks is limited to providing a second factor again. But as I said, we see a broader application of our product as its primary use case, which I agree is beyond the current scope of Providers section of twofactorauth.org.

Anyway, it's up to you to decide or let the users make their choice.

Commenting on your disclaimed paragraphs, it's actually much easier to steal the keys from a rooted/jailbroken phone than you think. You don't need physical access to the phone. The new generation OSes like Android normally run all apps sandboxed so that every app doesn't have access to other apps private data including keys. By rooting your Android you effectively defeat the sandboxing security model and make the OS behave like Windows PC. After you enable root for a potentially malicious app, the app can read and write data of any other apps including security related apps like GA. So, to steal GA keys from a rooted phone you only need to make the user install and run an app that reads GA's files and sends them over to your server.

RichJeanes:
That is true, but that would mean that said savvy user would be lax in their security and not respecting the power of sudo. If you are diligent enough to protect your phone as I have described, you would be quite remiss to be granting root access to apps that you don't have reason to trust. Such a malicious app would be exploiting the user, not the system.

tonyteddyid:
The fact of life is, people often don't know what they are doing, even when you expect them to be savvy enough. Yes, such a malicious app would be exploiting the user, not the system, and it shouldn't be surprising as any reasonable attacker would attack the weakest link, that is the user (@Carlgo11 didn't agree with me on that earlier in this thread). That's why security depends so much on what people actually do, not what they are supposed to do, that's why user-friendliness matters so much for security, that's why we designed TeddyID around the single idea of security that is easy to use (sorry for unsolicited self-promotion).

RichJeanes:
And I think it's awesome that you guys have thought of something new and original to try and help solve that problem. Innovation can't happen without people trying new things and I appreciate those people. However, I still believe TeddyID doesn't quite line up with what we're doing here, mainly for the sake of clarity and consistency. But I absolutely think that you should continue working towards improving and expanding your platform. Thank you for your contribution to the online community and trying to improve user security (and for the discussion here). I believe that we share that same goal, but are working towards different objectives within it. (No need to apologize, business is business, software can't survive without users)

For everyone's sake, I think this thread needs to be put to rest.
If any of our other collaborators would like to give their input (@mxxcon @jamcat22 @ilyakatz @jdavis, sorry for spam), please do so, otherwise I will be closing this PR on @Carlgo11's and my judgement. I will leave it open until Friday afternoon, barring any dissent.

PR #891, opened 2014-11-30
Updated Tresorit

jamcat22:
They also have phone call support.

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -145,6 +145,7 @@ websites:
     img: tresorit.png
     tfa: Yes
     sms: Yes
+    phone: Yes
     software: Yes
     email: Yes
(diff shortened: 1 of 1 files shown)


mxxcon:
Their docs url says

> - by entering the code you receive via SMS
> - by entering the code you receive via e-mail
> - via a verification app

I see no mention of phone calls..

jamcat22:
2fatresorit

jamcat22:
Also, from their blog (https://tresorit.com/blog/introducing-2-step-verification-for-tresorit/), 

> When you set up two-step verification, you register one or more trusted devices. A trusted device is a device you control that can receive 4-digit verification codes using either SMS, voice call, a verification app or an email."

PR #900, opened 2014-12-05
Update Skype in communication.yml

### LINKED ISSUE #706: Skype & Microsoft Accounts
Skype, even with a Microsoft account paired, does not support two-factor auth.

1) Even after pairing MS account, your Skype password is still used.
2) Skype itself does not have 2FA.

Because the MS account is irrelevant to logging in to Skype, you do not have the 2FA protection with Skype under any situation. Thus, there are no options for effective 2FA for your Skype account.

I don't know how you want to document this, but it should be known that right now, the site is communicating a false sense of protection.

mxxcon:
Let me know if it can be phrased better.
Closes #706.

[Review on Reviewable](https://reviewable.io/reviews/jdavis/twofactorauth/900)

### DIFF
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -73,13 +73,15 @@ websites:
       software: Yes
       doc: http://app.sendloop.com/help/article/account-003/how-can-i-make-my-account-more-secure
 
-    - name: Skype (w/Microsoft Account)
+    - name: Skype
       url: https://www.skype.com
       img: skype.png
       tfa: Yes
       sms: Yes
       software: Yes
       doc: http://windows.microsoft.com/en-au/windows/two-step-verification-faq
+      exceptions:
+            text: "Two-factor authentication is used only when logging in using Microsoft Account, not plain Skype account."
 
     - name: Slack
       url: http://www.slack.com
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -73,7 +73,7 @@ websites:
       software: Yes
       doc: http://app.sendloop.com/help/article/account-003/how-can-i-make-my-account-more-secure
 
-    - name: Skype
+    - name: Skype (with Microsoft Account)
       url: https://www.skype.com
       img: skype.png
       tfa: Yes
@@ -81,7 +81,7 @@ websites:
       software: Yes
       doc: http://windows.microsoft.com/en-au/windows/two-step-verification-faq
       exceptions:
-            text: "Two-factor authentication is used only when logging in using Microsoft Account, not plain Skype account."
+            text: "Two-factor authentication is available only when logging in using Microsoft Accounts, not Skype's stand-alone accounts."
 
     - name: Slack
       url: http://www.slack.com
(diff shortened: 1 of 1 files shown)


Carlgo11:
I don't like that it shows that Skype has tfa when <b>they</b> don't. There's no way of moving a Skype account to MS so that it is compatible with TFA. Any ideas on how this could be dealt with in another way?

RichJeanes:
You could reverse the flag. Leave the name as `Skype (w/Microsoft Account)` and have the flag stat that 2FA is not available with a standard Skype account. That makes the distinction more apparent up front.

dsoegijono:
What if we flag Skype as not having TFA, but add an exception saying that Microsoft Accounts have TFA?

mxxcon:
@dsoegijono MS is moving away from skype-specific logins, so making the "future-proof" login type an exception seems counter-intuitive. :)

mxxcon:
How about this?

jamcat22:
Maybe say "when logging in using a Microsoft Account not linked to a stand-alone Skype account." or something similar?

This is because you can have
- A stand-alone Skype account (that doesn't have 2FA)
- A stand-alone Skype account linked to a Microsoft Account and login with either (allowing you to bypass the Microsoft 2FA)
- A Microsoft Account created Skype account (where you can only login with the Microsoft Account, allowing you to force 2FA)

Carlgo11:
the MS login feels kinda like when sites support Facebook logins. Sure facebook supports 2FA but that doesn't mean the site does.

jamcat22:
I agree, except, Skype is owned by Microsoft and Microsoft is trying to migrate from Skype accounts to all Microsoft Accounts. You can also have a Microsoft Account on Skype without a Skype account.

mxxcon:
> Maybe say "when logging in using a Microsoft Account not linked to a stand-alone Skype account." or something similar?

Well, that's not exactly accurate. I can login w/ 2fa using my MS account even though it's linked to a skype account..

jamcat22:
@mxxcon But that 2FA doesn't help because it can be bypassed by Skype. It's like me saying if you login with this password, you will be protected by 2FA, but if you login with a different password you are not. Because you can bypass 2FA entirely by using the second password, the 2FA does not protect you.

Like I said, if you have a stand-alone Skype account linked to a Microsoft Account, you can bypass the 2FA by logging in with your Skype password (which the attacker would most likely try first since your username is publicly available). If you have a Skype account created using a Microsoft Account (meaning you don't have a normal Skype username and password), you can force 2FA making your account secure.

jamcat22:
To clarify, this doesn't mean you shouldn't enable 2FA on your Microsoft Account, it just means depending on your circumstance it might protect you completely, or it might only protect you partially.

rugk:
Let's come back to topic... :smiley:
@mxxcon
You asked how to phrase this. So what about "via Microsoft Account"?
"[via](https://en.wiktionary.org/wiki/via)" is a international word and suggests that you have to use the Microsoft Account for logging in.

mxxcon:
I'm fine with this too.

jamcat22:
Agree

jamcat22:
Review status: all files reviewed, all discussions resolved, some commit checks pending.
Reviewed files:
- _data/communication.yml @ r3

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/jdavis/twofactorauth/900)

PR #901, opened 2014-12-06
Ensure consistent HTML and CSS practices

fpigerre:
Ensure that <i> tags are closed, remove unnecessary units of measure,
ensure consistent media tag spacing.
This pull request should allow HTML parser tests to complete successfully.


mxxcon:
`table.min.css` looks significantly different..is that intended?

fpigerre:
Yes, that is intended @mxxcon. I changed several inconsistencies in `table.css`, then re-minified it so that the `table.min.css` file would be lighter than it previously was (realistically, the size changes that occur in this pull request make very little difference to the page's loading time, as the file size only changes by about 0.04 kilobytes). I have tested the site with the committed changes, and it appears that the layout has not changed.
Furthermore, I realised that the `table.css` and `table.min.css` files are not actually referenced in the site.

mxxcon:
:+1:

Issue #878, opened 2014-11-23
Add Interactive Brokers to the list


jdavis:
I got an email asking to add  [Interactive Brokers](https://www.interactivebrokers.com/ind/en/main.php) to the list. The email claims they have 2FA using hardware tokens.

Carlgo11:
They didn't provide a doc link for that in the mail?

jamcat22:
Created pull request. Should we close this?

PR #903, opened 2014-12-06
Add Interactive Brokers

jamcat22:
As per #878


jdavis:
Thanks for this, @jamcat22. Great job.

jamcat22:
Should we delete the patch-interactivebrokers branch then? Doesn't look like anything was ever committed.

jdavis:
Yup, I just deleted it. Thanks for pointing that out.

PR #912, opened 2014-12-23
Added dwolla to banking list plus image.

j256:
Thanks so much for the site!!


mxxcon:
I don't think `banking` is a correct category for this. I think `payments` is a better fit.

j256:
Whooops.  Ok.  I'll move it.

PR #916, opened 2014-12-29
Add Cloud9

dsoegijono:
TFA available if you sign up using GitHub account, but not for email signups. Do we mark this as `yes` or `no`?


Carlgo11:
I'd say `no`

mxxcon:
Ya, I'd say it's a similar situation as with sites that support SSO.
Also maybe use this logo instead?
cloud9
Your current logo has white clouds on transparent background, which in our case will be white background. My picture above has a slight outline. :smile:

dsoegijono:
@mxxcon Thanks! Image changed. They also have a blue cloud logo:
cloud9-blue

Would that be better?

mxxcon:
yup, either one would be better than white on white :smile_cat:

jdavis:
Feel free to swap with the blue if you don't think it looks good; there's no need for a PR either: https://twofactorauth.org/#developer

Issue #914, opened 2014-12-23
Add a warning for Charles Schwab


somethingnew2-0:
As described in this [blog post](http://www.jeremytunnell.com/posts/swab-password-policies-and-two-factor-authentication-a-comedy-of-errors), Charles Schwab 2FA is not actually required when enabled and therefore doesn't actually work.  I would go so far as to suggest removing them altogether until this issue is resolved.

jamcat22:
That is not what the blog post is saying. The blog post says that sometimes, the token may not be enabled properly if you tried to set a password over 8 characters. If you activate the token correctly, the token is required.

jamcat22:
We could however add an exception saying that their website has a maximum length of 8 characters and only checks the first 8. So we could warn people that they can only type 8 characters then the token code.

somethingnew2-0:
Ok,  I see that now. Yes, I believe that would be fair. Thanks for this site btw!

Carlgo11:
Who wants to make the PR then?

PR #917, opened 2014-12-30
Add Charles Schwab exception

dsoegijono:
For #914. How's the wording?

### DIFF
--- a/_data/investing.yml
+++ b/_data/investing.yml
@@ -11,6 +11,8 @@ websites:
       tfa: Yes
       hardware: Yes
       doc: http://www.schwab.com/public/schwab/nn/legal_compliance/schwabsafe/protect_your_account
+      exceptions:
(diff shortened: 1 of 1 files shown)


mxxcon:
Not remembering what happened in 914, when I read that exception is sounded very strange and out of context. Perhaps expand/elaborate it a bit?

Carlgo11:
How about `Password length limited to 8 characters when for TFA to be available` instead of `Password length limited to 8 characters.`

jamcat22:
But the password length is limited to 8 characters even if you don't use 2FA.

Carlgo11:
Oh okay. Then yeah, it looks good

dsoegijono:
Sorry I was away for the holidays. If @mxxcon doesn't think the note is clear enough, maybe it should be reworded @Carlgo11 @jamcat22 ? Although at the moment, I can't think of what to put without writing a long paragraph of explanation. =/

PR #919, opened 2015-01-05
Added Visual Studio Online to the 2FA Developer section

edblankenship:
Hope this helps!


jamcat22:
Looks good! You can remove the twitter line if you would like though.

edblankenship:
Awesome! - just removed the twitter line.

Carlgo11:
@EdBlankenship does visualstudio provide anything else than Microsoft Account login? If not then you can remove `(w/Microsoft Account)`

edblankenship:
> @EdBlankenship does visualstudio provide anything else than Microsoft Account login? If not then you can remove ('w/Microsoft Account')
> -- @Carlgo11

It does... You have two options as an account owner:
- You can use a Microsoft Account-based approach by default.  
- You can alternately [setup your Visual Studio Online account to be connected to an Azure Active Directory tenant](http://www.visualstudio.com/get-started/manage-organization-access-for-your-account-vs) which includes a combination of Microsoft Accounts, Active Directory accounts synchronized from on-premises, and cloud-only Azure Active Directory User Accounts.

There is also more information available on [enabling multi-factor authentication for an Azure Active Directory tenant](http://technet.microsoft.com/en-us/library/dn249466.aspx).

mxxcon:
So then regardless which method you use, TFA is available either way?

edblankenship:
> So then regardless which method you use, TFA is available either way?
> -- @mxxcon

That's correct.  The difference is in the details. The documentation link would be different for each option and the Azure Active Directory approach supports authentication by phone whereas the Microsoft Account approach does not.  Is the suggestion to have two different entries in the list?

edblankenship:
Anything else you would like me to do for this PR to make it look good?

Carlgo11:
Okay well if TFA is supported either way then why not just using the doc for using a Visual Studio Online account instead of Microsoft Account?
Also remove `(w/Microsoft Account)` if now again it is supported either way. 

Then change the `url:` to http://www.visualstudio.com/. We like to link directly to the main page of the service supporting TFA.

And thank you for contributing Ed! Great to see that Microsoft reaches out to the internet communities.

jdavis:
:+1: to @Carlgo11's comments.

edblankenship:
Always happy to help!  Made some changes that are getting pushed as I add this comment.  I also wrote a quick blog post about the topic in case anyone is interested in more details:  [Securing Your Visual Studio Online Account with Multi-Factor Authentication](http://www.edsquared.com/2015/01/17/Securing+Your+Visual+Studio+Online+Account+With+MultiFactor+Authentication.aspx)

PR #921, opened 2015-01-08
Added Sinch for SMS

kwaimind:
Thought it would be useful to add Sinch as an SMS provider for 2FA solutions.


mxxcon:
Is that the best url to use for this?
And overall I'm leaning a bit towards not accepting this entry. Your service is used simply as an SMS delivery mechanism rather than a full 2FA provider. I would still need to have my own or 3rd party otp system to handle everything but the delivery of it...

RichJeanes:
Yeah, I'm with @mxxcon on this one. Sinch is just providing 2FA as an example of what you could use their platform for. You could do the same thing with Twilio, Callfire, etc.

PR #923, opened 2015-01-12
updated rackspace support for 2-factor auth

joesavak:


mxxcon:
Hmm..I wonder if it makes sense to have 2 separate entries for Rackspace, one under Cloud Providers, and one here under Hosting/VPS.
The link @joesavak provider relates to Cloud accounts.
However, Rackspace Hosting(physical/enterprise) also has http://www.rackspace.com/knowledge_center/article/myrackspace-two-factor-authentication
While theoretically they are both SMS-based TFA, instructions to set them up are different...

joesavak:
Yes - there are 2 portals (myRack for hosting customers and myCloud for cloud customers) - agree that Rackspace should be listed in both

Walkman100:
See the travis build, I think if you are adding rackspace to `cloud` then you also need to copy the image to the `cloud` images folder too

joesavak:
thanks @Walkman100 - I corrected. : )

Issue #521, opened 2014-05-27
Add Filtering


jdavis:
As talked about in #336 with @mxxcon and @dsoegijono, filtering could solve a lot of problems with organization. It could also address #241 as well as #410.

To do:
- [x] Filtering/searching by text. (See #1486)
- [x] Filtering by Category.
- [x] Filtering by method. (See #4584)
- [x] Filtering by Country? (depending on #241)
- [ ] Filter by date of last update.

jdavis:
I received an email regarding how it would be good to list the "last updated" date next to each entry. I'll add it to the checklist for now.

jdavis:
I just received another email in the last week asking for the "last updated" feature as well, in particular, they wanted to be able to receive updates to sites that didn't have 2FA that now do with the notifications sent through RSS/email.

Walkman100:
Since all changes to the site are managed here, maybe RSS updates should be based on GitHub commits/merges, so whenever a PR is merged, the RSS is updated. It could use WebHooks to receive the updates, which is the system GitHub has for CIs to receive updates

Walkman100:
And getting an email on an RSS update is easy with [FeedBurner](https://feedburner.google.com) or [UglyReader](http://reader.dfm.io/)

hoovercj:
~~I'm not sure how you plan on implementing filtering, but I have some experience with KnockoutJS which is capable of search and filter. See [here](http://opensoul.org/2011/06/23/live-search-with-knockoutjs/) for an example of a search implementation.~~

~~The downside of this, though, is that it does decrease the performance of the page -- it is being rendered and managed through JavaScript rather than lovely HTML. I can whip up a prototype though if you want to compare.~~

EDIT: Though on second thought, there must be a perfectly adequate jquery plugin that wouldn't require as much of a restructure.

hoovercj:
Though on second thought, there must be a perfectly adequate jquery plugin that wouldn't require as much of a restructure.

hoovercj:
Ok, so this might be a dumb question -- how do I initiate pull requests against this? I have to fork the repository first, create my own branch in that repo, push it up into github, and then make a pull request from there?

jdavis:
Yeah, click the "Fork" button and then you can create your new branch:

```
$ git checkout -b new-branch
```

Then if you push

```
$ git push origin new-branch
```

then you'll see the Fork button on the branch. This isn't required, but I like to add the original and call it `upstream` so I can pull in changes like so:

```
[...]

PR #926, opened 2015-01-14
Filter

hoovercj:
To start: I am sorry that I'm so bad with having a clean git history for these things. Also, should I be submitting pull requests against Master or Dev? Ok, now to the main message:

This work added filtering capabilities based on two main categories
1. TFA Methods -- sms, email, phone, hw, sw.
2. No TFA -- all non-tfa sites, non-tfa but in-progress

These filters play nicely with search. In fact, some of their functionality relies on the help of the search plugin. Currently everything works well EXCEPT for the checkboxes being disabled (as in: in doesn't make sense for the TFA method filters to be enabled if one of the Non-TFA filters is checked). Currently they go 'disabled' and look disabled, but you can still click on them. I'm not sure if there is an issue with this version of semantic-ui or not, but disabling the checkboxes does not seem to be preventing them from being clicked on as is suggested in the documentation. They also don't respond to the same events as described in the documentation. I suggest The site be updated to use the current version of semantic-ui and then these changes be tested and considered against that.

Also, somebody could pull this into dev and change how it looks or make suggestions on how to change the UI.

Lastly: It's obvious that my javascript conventions are less than stellar. I haven't picked up on how to properly encapsulate the javascript files. For example, the code block below is the project as I found it. I don't know if I'm supposed to be defining my functions inside one of those blocks or if they can just sit out in the root of the document. Thanks.

```
(function (root, $) {
    $('.menu .dropdown').dropdown();
    $('span.popup.exception').popup();
    $('a.popup.exception').popup();
    $('.ui.checkbox').checkbox();
}(window, jQuery));

$(document).ready(function() {
    $("img").unveil(50);
    updateFilterClasses();
});
```

Reference to Issue #521


jdavis:
This is cool; I'm looking through it now. And quick note, dev is an old branch. I don't even know what it was used for.

Carlgo11:
Uhm? Status?

fpigerre:
This issue has been resolved in issue #1368.

PR #927, opened 2015-01-14
Add Netim.com

c960657:
[Netim](http://netim.com) is a domain name registrar that specializes in European and generic domains. I would love to see them adopt 2FA, but so far they haven't.


Carlgo11:
Since they don't even have https I doubt they ever will... but of course it wouldn't hurt adding them.

c960657:
They do have HTTPS. Their sign-in page is on https://netim.com/direct/.

PR #928, opened 2015-01-15
Update Dynadot information

grawity:
- As far as I know, it has never supported hardware 2FA; the (now
  deprecated) "Dynadot Token" was just an app for slightly-nonstandard
  OATH-TOTP.
- The second factor is not needed to log in to the website, only for
  some actions (changing the password or exporting a domain).

### DIFF
--- a/_data/domains.yml
+++ b/_data/domains.yml
@@ -86,7 +86,8 @@ websites:
       tfa: Yes
       sms: Yes
       software: Yes
-      hardware: Yes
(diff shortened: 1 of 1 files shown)


mxxcon:
@grawity do you know if this information is still current or did they update their system to require 2fa for login?

grawity:
It has not changed as far as I see. I can still log in without being asked for an OTP.

PR #930, opened 2015-01-18
Update Zendesk - status

Carlgo11:
https://twitter.com/mariogama/status/556900730125451264

Sounds like they are actively working on adding TFA

### DIFF
--- a/_data/other.yml
+++ b/_data/other.yml
@@ -116,7 +116,7 @@ websites:
       twitter: Zendesk
       img: zendesk.png
       tfa: No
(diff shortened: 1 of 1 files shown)


mxxcon:
@Carlgo11 I think this link would've been better https://twitter.com/Zendesk/status/556903689022361600 since it points to Zendesk's twitter rather than the guy that asked a question.

jdavis:
Yeah, either link but I think @mxxcon's point is a good one as replies to the original link can mess with the replies.

Carlgo11:
done

Issue #580, opened 2014-06-15
Paypal doesn't seem to have a software implementation


bbigras:
![image](https://cloud.githubusercontent.com/assets/24027/3280046/178f61ce-f43b-11e3-8a5c-01528ab17c12.png)

![image](https://cloud.githubusercontent.com/assets/24027/3280047/378ef5f2-f43b-11e3-9f7d-f79ab4dc9922.png)

I don't see how to use software implementation like [Google Authenticator](https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2).

Am I missing something?

mxxcon:
Verisign VIP has software https://idprotect.verisign.com/desktop/home.v and mobile website https://vipmobile.verisign.com/home.v

dsoegijono:
Maybe a note should be added that software implementation is through Verisign VIP? The linked doc doesn't seem to say anything about it, and it might be confusing.

mxxcon:
@dsoegijono you don't think if somebody decides to active PayPal with verisign vip they wouldn't know what it supports?

dsoegijono:
@mxxcon I was thinking the note is more for PayPal users who don't use Verisign VIP. I'm guessing that's why @brunoqc doesn't think Software Implementation is supported.

mxxcon:
Hrm...but then we are getting back to the subject of listing every implementation for every site :sweat:

dsoegijono:
@mxxcon Was there another thread discussing this issue?

mxxcon:
@dsoegijono #208 among a few other.

dsoegijono:
@mxxcon I agree that we shouldn't list every implementation for every site, but I think for cases like this one, it should be noted. If I see that an implementation is supported, I would assume that I can set this up through the website's account settings, unless noted on the doc linked. In this case, the PayPal doc doesn't say anything about software implementation, or about Verisign VIP.

smholloway:
See previous pull requests discussing PayPal: https://github.com/jdavis/twofactorauth/search?q=paypal&ref=cmdform&type=Issues.

From your screenshot, I would say that the "Register your mobile phone" blurb indicates that they have a software implementation (a phone-based security key) even if they don't support Verisign, which they seem to.

bbigras:
> From your screenshot, I would say that the "Register your mobile phone" blurb indicates that they have a software implementation (a phone-based security key) even if they don't support Verisign, which they seem to.

[...]

PR #932, opened 2015-01-21
Add Google Domains, TIAA-CREF, and edit Paypal exceptions

stephengroat:


jamcat22:
PayPal 2FA does work with mobile apps, it's just a bit complicated. It only works when you use SMS 2FA. It might be better if we remove that part of the Pull Request for now, and I can explain it more (including make docs for it) in another PR.

stephengroat:
I agree that their is a Paypal workaround (disabling, re-enabling, etc.), but right now it's definitely not officially supported.

Since this is the message that people get when trying to login, maybe that detail should be included? I only say it because it's such a pain to get it to work and it still doesn't work inside the mobile eBay app, even with the work around.

screenshot_2015-01-20-23-22-15

Great website by the way, I send everyone there to look to see which services are supported! thanks for all the work

jamcat22:
I never said it was a workaround. The iPhone PayPal and eBay apps let you login with 2FA.

When you use hardware or software 2FA to login to the PayPal iPhone app (which the app doesn't support):
image

When you use SMS 2FA to login to the PayPal iPhone app (which the app does support):
image

When you use any type of 2FA to login to the eBay iPhone app:
image

The eBay Android app doesn't support any type of 2FA (even if you append the token).

It seems the PayPal Android app also doesn't support any type of 2FA.

I think we should not solve this in this PR. I would like to create a separate one to solve this and #580 (about how vague PayPal docs are).

This way, we will not mix up adding sites with making huge changes to existing sites.

Also, in the future please give each site its own PR.

stephengroat:
Sorry, didn't know. I'll get them separated out today. Thanks,

Stephen

jamcat22:
It's all good! Thanks!

stephengroat:
Is this going to be added or should i pull it? Thanks!

Stephen

Carlgo11:
@stephengroat Please have patience. We're all doing this on our spare time.

Issue #935, opened 2015-01-26
Tons of whitespace?


RichJeanes:
Anyone looked at the page source lately? There's some 80 lines of whitespace between each entry. What's up with that? It makes the source over 41,000 lines and 1.5 MB.

Also noticed in the source, but unrelated, there's two conflicting charset declarations... ? (Line 5 "meta charset" and line 8 "meta http-equiv content")

mxxcon:
I bought up empty lines a long time ago, but a don't remember the result of that conversation.
On a plus side, with compression, blank lines are irrelevant.
https://redbot.org/?uri=https%3A%2F%2Ftwofactorauth.org%2F

> HTTP supports compression of responses by negotiating for `Content-Encoding`. When RED asked for a compressed response, the resource provided one, saving 94% of its original size (from 1628123 to 82150 bytes).

Still, we should figure out how to fix this.

Also this looks pretty scary http://validator.w3.org/check?uri=https://twofactorauth.org/

Walkman100:
The majority of the W3 validator errors are `& did not start a character reference. (& probably should have been escaped as &amp;.)`
I'd say 99% of the 385 errors are.
Edit: And about 80% of the CSS validator warnings (link below) are `Property [property] is an unknown vendor extension`

jdavis:
I know a lot of the lines are from `{% ... %}` tags in Jekyll as it just leaves the whitespace. The big block after the `<tbody>` is a great example of that.

One thing we could do is just use [Cloudflare's optimizer](https://www.cloudflare.com/features-optimizer), but cleaning up the source itself might be good too.

RichJeanes:
The [CSS Validator](http://jigsaw.w3.org/css-validator/validator?uri=https%3A%2F%2Ftwofactorauth.org&profile=css3&usermedium=all&warning=1&vextwarning=&lang=en) is even worse.

Carlgo11:
Well that css validator is a bit of a overkill isn't it? I mean even google has a lot of errors http://jigsaw.w3.org/css-validator/validator?uri=https%3A%2F%2Fgoogle.com&profile=css3&usermedium=all&warning=1&vextwarning=&lang=en

mxxcon:
@Carlgo11 unfortunately, google is hardly the defacto when it comes to following internet standards.. :(

PR #936, opened 2015-01-26
Escaped &amp; characters

Walkman100:
#935

http://validator.w3.org/check?uri=https://twofactorauth.org/#error_loop

Fixed the W3 HTML validator errors that say:
`& did not start a character reference. (& probably should have been escaped as &amp;.)`


jdavis:
Interesting, I didn't know link ampersands needed to be escaped. I had to test it to find out too and it worked (of course): http://jsfiddle.net/rkddk5zs/

Cool.

Walkman100:
I wouldn'th've (is that a word) expected that to be required either, since the links are in double quotes...

PR #937, opened 2015-01-26
Update hosting.yml

matgalli:
Add Domainfactory


mxxcon:
Still missing an image..

PR #938, opened 2015-01-27
add domainFACTORY

matgalli:


Carlgo11:
Please don't make multiple pullrequests about the same website.

Carlgo11:
We have several sites like this who provide both domains and webhosting. One example is <a href="https://www.loopia.com/">Loopia AB</a> that we've put under the **Domains** category.
Considering the site's name is Domain Factory I think the same should be done with this site.

Please move the site to `_data/domains.yml` and while doing it try not to create another pull request.

Thank you,
<s>@psgs :palm_tree:</s>
I mean, @Carlgo11

matgalli:
ok, I will try

Carlgo11:
img/domains/domainfactory.png is not 32x32px. Please change the image to 32x32px

matgalli:
There is no icon bigger then 32x32px so I have it scaled up

mxxcon:
@matgalli :+1: Or you could do it the other way around and scale down a larger image :wink:
But good enough either way.

Carlgo11:
well it looks okay for the human eye. I'm giving it a go

PR #943, opened 2015-01-30
Update Stripe two factor auth doc URL

jordanmerrick:
Changed the doc from the blog post to dedicated support article with more information.

PR #944, opened 2015-01-30
Updated Stripe url to point to support doc

jordanmerrick:


mxxcon:
@jordanmerrick why this pullrequest in addition to #943?

PR #956, opened 2015-02-02
Add Esri Inc.

scw:
Mapping platform used by ~300,000 organizations.


mxxcon:
Should it be under `developers` or perhaps `other`?

scw:
Either one is fine -- the only other mapping related entry I found (Mapbox) was under 'developers', so I thought this should be in the same place. It may make sense in the long run to have a 'mapping' category, and include other entries like CartoDB and Google's offerings.

PR #957, opened 2015-02-03
Bitbucket status update

patryk:


Carlgo11:
```
2FA is unlikely to be part of the initial SSO roll-out since it would delay the release even further. But it is on the roadmap after the initial release.
```

Closing since they're not actively working on 2FA

PR #958, opened 2015-02-03
Added WebFaction and Raiffeisen Bank

iBobik:


mxxcon:
:+1: @iBobik could you please rename Raiffeisen's image to have a more descriptive filename. No point making it obscure to save a few chars. :smiley:

iBobik:
@mxxcon Done :-)

Carlgo11:
What about changing `raiffeisenbank_cz.png` to `raiffeisenbank.png`

iBobik:
I think this is not good idea, because RB is in many countries, usually technologicaly separated.

Carlgo11:
oh, very well then.

PR #959, opened 2015-02-03
Added bitbucket status

CodeFoodPixels:


mxxcon:
Similar to #957, in that unfortunately in that linked issue they stated that they will not be working on it until their overarching SSO system is implemented. So while Atlassian marked it as `critical` priority, they also stated

> the priority of issues is set by the reporter and does not necessarily reflect Atlassian's internal priority. 

So I feel that we still have no clear definitive answer from BitBucket. :disappointed:

Carlgo11:
I agree with @mxxcon. We shouldn't set a status before they've actually started working on it.

PR #963, opened 2015-02-05
New Cloudinary logo + fix link for onedrive

ranrub:


Carlgo11:
uhm cloudinary doesn't have anything to do with onedrive does it?
I'm aware of the false "failed" statement by travis on your last PR. Just ignore that.
When we've fixed that your PR will be looked at more carefully.

Thank you for your patience.

//Carlgo11

Issue #964, opened 2015-02-05
[URGENT] Errors on invalid urls


Carlgo11:
So right now travis returns false when ever a site has changed their url even if the PR haven't touched that url. We need to change that.
I recommend just outputting errors to the console. That way we just need to learn to check the travis output a little bit better.
Right now we can't merge PRs because some other sites have decided to change their urls.

jdavis:
Addressing this in #970.

PR #970, opened 2015-02-08
Update Travis builds

jdavis:
This addresses #964. It just catches HTML Proofer's exception and prints it instead.


jdavis:
And this is what it looks like now instead: https://travis-ci.org/jdavis/twofactorauth/builds/49967874

PR #974, opened 2015-02-12
Fix onedrive url & remove userify status url

Carlgo11:
Quick fix of current Travis CI errors.
I couldn't find any other status links for userify so I just removed it.

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -99,7 +99,7 @@ websites:
     tfa: No
 
   - name: OneDrive
-    url: https://www.onedrive.com
+    url: https://onedrive.live.com/about/en-us/
     img: onedrive.png
     tfa: Yes
     sms: Yes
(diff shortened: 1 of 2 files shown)


Carlgo11:
hm I actually don't know what the [onedrive error](https://travis-ci.org/jdavis/twofactorauth/builds/50495610#L208) referees to.
@jdavis is there no way to make it indicate what the url looked like when it was read?
Also, could the script possibly be blocked from the site because it's a script?
I mean, does the script listen to [robots.txt](https://en.wikipedia.org/wiki/Robots_exclusion_standard) or something similar?

jdavis:
Ahh, sorry @Carlgo11; I didn't see this.

Carlgo11:
@jdavis no worries. So I really can't find the ondrive link that's failing.
As the link contains /pagenotfounderror I'm guessing it's a redirected link that it's outputting. Any ideas on how to output the original one?

LeFouineur:
I think we can close this PR now since all the changes were resolved and merged

PR #977, opened 2015-02-20
no more TFA for OVH

pierreprinetti:
Following up on #870.

It really looks like OVH doesn't offer TFA anymore. The section on their website that used to advertise TFA has been wiped out some time ago.

I think the time has come to revert their state to `tfa: No`.

### DIFF
--- a/_data/hosting.yml
+++ b/_data/hosting.yml
@@ -93,7 +93,6 @@ websites:
       url: http://www.ovh.com/
       img: ovh.png
       tfa: No
-      sms: Yes
 
     - name: Rackspace
       url: https://www.rackspace.com/
--- a/_data/hosting.yml
+++ b/_data/hosting.yml
@@ -93,7 +93,7 @@ websites:
       url: http://www.ovh.com/
       img: ovh.png
       tfa: No
-      twitter: OVH
+      twitter: ovh_support_en
 
     - name: Rackspace
       url: https://www.rackspace.com/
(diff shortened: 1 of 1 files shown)


Carlgo11:
Have you actually contacted them and asked about this?

Carlgo11:
I just sent them a question asking if they still support TFA and in what countries.
I'm marking this PR as `pending` until they reply.

pierreprinetti:
I've got no news since #870.

Carlgo11:
Okay here's my response:

```
Hello,

Thank you for reaching OVH support team, regarding your 2FA question. My name is Rachid, I will be answering you today.

Unfortunately we do not suport 2FA, and I am not being aware that we are planning to in a near future, otherwise it would be advertised on our web page.

On the OVH North America we did not have the double authentication method it was only on Europe,.

I hope this answers your questions. Don't hesitate if you have further questions.


Thank you for contacting OVH. Have a great day!

Rachid Y.
Customer Advocate
```

So I guess you're good to go. Please wait while another collaborator confirms everything is okay.

mxxcon:
@qrawl perhaps add their twitter for public shaming.

pierreprinetti:
@mxxcon 
et voilà

Carlgo11:
Says the Italian ^_^

Carlgo11:
Change the Twitter handle to `ovh_support_en`

mxxcon:
@Carlgo11 Looks like regular `ovh` account has more tweets and followers than the support account, so it could be higher profile. However, support account does reply to people.
Can we list 2 accounts so that when you press that button it generates message with both?

Carlgo11:
@mxxcon we usually send to the twitter account that replies. Otherwise they probably don't see the tweets.

PR #979, opened 2015-02-20
Add phone to coinbase

Carlgo11:
Coinbase supports Authy which also means it supports phone calls


Carlgo11:
hm after reading through this more carefully I think they've disabled phone calls. never mind then

jamcat22:
When I login, it gives me the option:

coinbase2

Also, in their docs it says you can:

coinbase

Carlgo11:
Oh, goodie.

PR #982, opened 2015-02-22
Add 1Password to Providers

ahelpingchip:
Before anything else, yes, I know 1Password is on the Blacklist/Exclusion list. :)

I think this addition bears merit, however, since I am adding 1Password as a _2FA Provider_. 1Password recently added the ability to generate TOTP tokens — in other words, it can perform the same function that Google Authenticator provides.

https://blog.agilebits.com/2015/01/26/totp-for-1password-users/

Of course, worth noting is that AgileBits themselves say that using 1Password in this manner is arguably _not_ 2FA, since you are storing your TOTP token and password in the same place. That all said, I still believe it is worth adding:
- A user could conceivably generate his/her TOTP token in 1Password without also storing his/her password there. (this is admittedly a bit contrived.)
- In many cases, people who use password managers on their phone (1Password, Keepass, Dashlane) already have their TOTP tokens in another app on the same phone anyway (Google Authenticator, Authy). Putting them now in the same app is not significantly less secure.
- This might be a feature not too many people are aware of, so I think it's worth adding for awareness' sake.

My apologies if I end up starting another long debate. :)


Carlgo11:
If not even aglebits think it's a 2FA provider then I don't think it's worth adding to the site.

Carlgo11:
also, +1 to your funny site ^_^ (http://phi.ph/). The social media things are hilarious.

ahelpingchip:
Thanks! :) I actually based them off something else, but I don't recall where from right now.

As for the PR itself — I was originally writing this up as a suggestion or question under issues, but I thought: hey, it's only 5 lines of "code", and it's a simple yes or no decision in the end. (merge or close.) Might as well make it a PR.

Carlgo11:
Thank you for your PR. All PRs help a lot!
I myself think that 1Password shouldn't be added. We'll just have to wait and hear what the other collaborators say. (Most of them are US based and should wake up soon)

mxxcon:
Well, you could use 1password as 2fa if you don't store any passwords in it.
Caveat emptor, but it does provide 2fa functionality..

jamcat22:
I don't think it should be a provider because as I understand it it just allows you to add Google Authenticator tokens to the app. It doesn't look like it has its own token protocol.

I love that they have added the functionality, but I think the providers section is for different ways for websites to use 2FA, and Google Authenticator is already on there.

mxxcon:
@jamcat22 In that case we should remove "Authenticator Plus" and FreeOTP since they are the same thing, just different clients for Google Authenticator-style otp.

jamcat22:
Oh then I stand corrected. I think we should add it then.

Carlgo11:
I think the "provider" thing should be remade.
Some of them can be used almost everywhere while some of them, like transakt & SAT Mobile ID, can only be used on 1-2 sites.
It's a bit off topic so I'll address this further in another separate issue tho.

Carlgo11:
Anyways, so are you guys going to mark this as `looks-good` then?

jamcat22:
The providers page is being removed. You can read more about it on #1177.

PR #985, opened 2015-02-26
Adding Nexmo Verify to Providers

tjlytle:
Nexmo offers both lower level SMS/Voice APIs and a [high level 2FA API](https://www.nexmo.com/product/verify/).


smholloway:
No need to include the `No` entries--can you trim those, please?

tjlytle:
@smholloway pulled them out.

smholloway:
Right on. Looks good! After another approval we will merge.

mxxcon:
We created providers section for users to see what's available.
If this thing is only sms/voice, is there a point to list it here?
It requires no additional effort from the user to set it up(ie no client to install/configure)...

tjlytle:
If the list is limited to those that provide user facing systems, I understand. But with Authy, Telesign, and others listed that provide an API to add 2FA to applications, I thought this would fit.

Carlgo11:
totally agree with @mxxcon. Closing the PR

smholloway:
I think it's within the scope of providers. Nexmo offers 2FA as a service.

To me providers page is mostly for implementers. The thought is something like, "I want to implement 2FA--let's see what's available! Ah, this one has voice and SMS while this one has an app. I like the idea of an app, so I'll explore this more." As a secondary benefit, users can review the page to see what is available.

tjlytle:
That what I thought it was (a page helping those non-2FA enabled sites / services find a way to enable it). But obviously, I'm not a curator of the list so could only glean that from looking at the list (while something like Clef is very user facing, something like Authy is very developer / integrator facing).

Either way, the data is here if it's found useful.

mxxcon:
Then perhaps we should clarify that.. @jdavis?

RichJeanes:
FWIW, I agree with @smholloway.

jdavis:
I'm a little confused on what Nexmo does.

So it verifies phone numbers as well as functionality to send SMS messages? Where does the 2FA come into play? Does that mean the site/app is supposed to generate the OTP, send it out through Nexmo, and then do the verification themselves?

tjlytle:
@jdavis The verify API can be used by a service to implement 2FA (just like other 2FA providers). Nexmo also offers APIs for sending SMS messages or making phone calls. 

The site/app does not generate an OTP and is not responsible for validating that the user has the correct OTP. 

Like other 2FA APIs, the first call to the API initiates the process (Nexmo generates a OTP, then delivers it by SMS or Voice depending on the number type, falling back to another channel if needed), and the second call is made to the API with the user provided OTP (Nexmo confirms if the code is correct in the response).

jdavis:
Right. It _can_ be used to implement 2FA, but unless I'm missing something, we don't have any other providers on the /providers/ page that don't do the 2FA for you (generating OTP, sending the messages, actually being specifically for more secure auth).

Putting a service on that page that makes it easier to send SMS messages isn't a 2FA provider to me.

tjlytle:
@jdavis Maybe I'm missing something you're saying, but Nexmo does do the OTP generation and confirms the user provided the correct code. It's not _just_ sending an SMS (though there's another API that lets you just do that).

jdavis:
@tjlytle Ah, okay. Do you have better documentation? According to [this](https://www.nexmo.com/product/verify/), I'd expect the use case to be:
- User comes to the website and enters in account info.
- They sign up for SMS notifications.
- You confirm the phone number using Nexmo Verify (because the tagline is: "Verify any number, hassle-free").
- A code is sent to confirm the SMS was received.

Where does it mention the 2FA/OTP for when you aren't needing to verify a phone number and instead verifying authentication?

tjlytle:
@jdavis  We've found that the 'verify' process is heavily used both in a signup flow and for adding 2FA to logins, so we're not limited to a single use case (effectively they do the same thing prove you have something right now).

Here are the docs, more generic than the product page: https://docs.nexmo.com/index.php/verify

I can update the URL to that if it's better.

jdavis:
@tjlytle No, I think the link you have now is better. But you really should look into improving the main landing page if you are hoping to capture more users of the auth verification of the product. It isn't very clear at all.

If anyone else has objections, feel free to mention them. If not, add the looks-good label and it can be merged.

tjlytle:
@jdavis There's also [this page](https://www.nexmo.com/verify/), but doesn't really say anymore about 2FA specifically. Great point, I'll pass it on.

PR #1001, opened 2015-03-09
Add Vk (VKontakte) as a social website using 2FA

LeFouineur:
Here is the clean PR for VK (with the typo fix in the extension)


mxxcon:
@LeFouineur Not to nitpick, but that logo is 15kb. Use this one instead, it's 714b :)
vk

LeFouineur:
No problem, I love nitpicking, so fire away. I'm the same.

PR #1003, opened 2015-03-09
Update OneDrive

LeFouineur:


Carlgo11:
This has already been addressed in #974

LeFouineur:
Yes, I saw it after, but in #974 you use the link with English-US locale in hard.
It would be better to use the generic base url (https://onedrive.live.com) and then let the server deliver the correct localized version

Carlgo11:
That would make the validitor fail

mxxcon:
@Carlgo11 your pr also touches userify which was already addressed elsewhere. Plus @LeFouineur brings up valid point regarding language/locale.

LeFouineur:
@Carlgo11, then we could put the one OneDrive uses after a server redirection: https://onedrive.live.com/about/auth, less clean, but will redirect to the correct localized page (at least while Microsoft doesn't change the structure), while also checking if you're already logged.

Carlgo11:
@LeFouineur pretty sure redirects fail as awell

LeFouineur:
One thing is a little "funny": https://onedrive.live.com gives a 200 HTTP/1.1 answer, so there shouldn't be a problem normally.
Could it be because of the html-proofer version used in the tests (v1.6.0), since they changed few things in the way they managed urls in the next versions (https://github.com/gjtorikian/html-proofer/compare/v1.6.0...2f7f0bd)

Carlgo11:
@LeFouineur You're more than welcome to try and figure out what the problem is. https://travis-ci.org/jdavis/twofactorauth/builds/53698507#L208

LeFouineur:
By updating to html-proofer 2.1.0 (#1006), there isn't an error anymore.
www.onedrive.com and https://onedrive.live.com both passed.
So there shouldn't be a problem by using https://onedrive.live.com anymore

Carlgo11:
Will rerun Travis once #1006 is merged

LeFouineur:
@Carlgo11, you can relaunch the test now. Normally, there shouldn't be any errors anymore.

Carlgo11:
doing it now

LeFouineur:
Will need to restart the test: It was executed while Travis had a load issue and was failing most of them.

PR #1005, opened 2015-03-10
Add plex.tv and netfirms.com

Efreak:
Plex Media Server is an application that you install on your computer, however accounts go through their servers.


Carlgo11:
Hi @Efreak.
Thanks for contributing to the project! All contributions helps a lot! :smiley: 

I'm adding the `looks-good` label now.
To see the different steps in getting your contributions added to our site visit our [new wiki](https://github.com/jdavis/twofactorauth/wiki/FAQ_Merge)

mxxcon:
What damage can happen if somebody gets access to your plex account?

Carlgo11:
oh I just saw that netfirms accepts https. It's not added to https-everywhere and there for I didn't see it before I did the `looks-good`. @Efreak could you change the url?

Efreak:
There's an interesting post [here](https://www.reddit.com/r/PleX/comments/2ih85r/an_update_on_the_security_of_plex/) on plex. I'll fix the URL in a moment.

Efreak:
I should note that there's several other websites in this list that support https (netim, badger, domain, etc) and some that switch from https to http for some reason (dnsmadeeasy).

dsoegijono:
@mxxcon They seem to have paid premium features. I think this brings it to the same level as Netflix, Hulu, etc. The paid service can be used, although no other harm can be done? @Efreak what do you think?

Efreak:
Similar to netflix, except that Plex makes content on your own computer available remotely, not content that's off in the cloud somewhere. It's more annoying than netflix, since you can remotely delete content and eat bandwidth by downloading everything.

PR #1008, opened 2015-03-13
Added QuickBooks and Clio

plecavalier:


jamcat22:
Any docs link?
Also, could you separate each site into their own PR?
And could Clio either go in Other or another category since it is the only one in its category? Or could we add some sites to the Legal category?

plecavalier:
Hi Jacob.

On Mar 12, 2015 9:43 PM, "Jacob" notifications@github.com wrote:

> Any docs link?

No. When asked I was told to call support and they would explain it to
me...I do have written confirmation that tfa is in fact supported, however.

> Also, could you separate each site into their own PR?

I see. Didn't realize that was the etiquette. That was my first one ever.

> And could Clio either go in `Other` or another category since it is the
> only one in its category? Or could we add some sites to the `Legal`
> category?

I debated that for some time. I feel it's a matter if time before the other
players in that field will follow suit. At which point I will gladly
populate. I can change it if you want and recreate legal if and when...up
to you.

Phil

plecavalier:
It's been a while and I haven't heard back. @jamcat22 are you asking me to start over? Like I said I'm new at this so...

PR #1011, opened 2015-03-14
Add entry for BitGo

jlopp:
I'm an engineer at BitGo and recently came across your site. We require 2FA via Authy or SMS for anyone who uses our wallet service.


Carlgo11:
wow that guide makes it look unreasonable hard to use your service. Anyways that's not why we're here. So since you use Authy do you support phone calls as well?

jlopp:
The guide is for creating a wallet, which the user only needs to do once :-) We don't have an entire guide around 2FA. It is a bit complex because our #1 priority is security while ease of use is second. We don't support phone calls as far as I know; is there a compelling case to support phone calls in addition to SMS?

mxxcon:
Phone calls are for people who don't have or want to receive sms messages.

Carlgo11:
Okay looks good then.
To see what happens next, read [this FAQ post](https://github.com/jdavis/twofactorauth/wiki/FAQ_Merge)

PR #1012, opened 2015-03-15
Added Vultr

shouptech:
I've added the VPS/Hosting provider Vultr.


Carlgo11:
@shouptech what kind of hardware two factor authentication do they support?

shouptech:
They support Yubikey devices.

PR #1014, opened 2015-03-17
Adds link to @IFTTT tweet.

kevinebaugh:
Hey all, Kevin from IFTTT here. We're working on two-factor auth so we wanted to reflect that here.


jdavis:
@kevinebaugh :+1:

kevinebaugh:
Thanks all!

PR #1015, opened 2015-03-19
Update Dyn info to show they support TFA, and link to the docs

apple4ever:
Dyn now supports TFA, and I updated their info accordingly


Carlgo11:
seems like they also support SMS.

apple4ever:
Oh whoops I didn't realize there was a separate variable for sms. I'll update the commit.

Issue #966, opened 2015-02-05
Requests for the last month


jdavis:
Here are the various email requests that I've received recently:
- [x] It looks like WellsFargo now has 2FA: https://www.wellsfargo.com/privacy-security/advanced-access/
- [x] Someone requested CleverCoin to be added to the Cryptocurencies list: https://clevercoin.com/en/security
- [x] Another person requested Navy Federal Credit Union which apparently doesn't support 2FA: https://www.navyfederal.org/

PR #1017, opened 2015-03-19
Wells Fargo now support TFA via SMS

Carlgo11:
As per #966


jamcat22:
They support phone call too. They also support RSA SecurID if you have a dongle from them.
image

Carlgo11:
All good?

todd:
I just stumbled back across this after 2fa.org was featured on Hacker News again today. It looks like this PR is still assuming Advanced Access is 2FA, which it is not. Please see the discussion in #242.

PR #1021, opened 2015-03-20
Added iD Opener to 2FA providers list.

nelsonkram:


mxxcon:
@rpr4394 what kind of hardware does it offer?

nelsonkram:
Hi @mxxcon-
ID Opener offers hardware storage of the private key credentials on mobile devices that offer hardware storage - such as Nexus devices with at least Jelly Bean MR2. 
Does this answer the question?  Feel free to reach out to me at mark@hydrantid.com directly if you need more information!
Thank you!
Mark NelsonChief Security Architectmark@hydrantid.com

jamcat22:
If this is accessible with an app, it would be classified as software. Hardware is usually something like a physical token dedicated to the 2FA (like a RSA keychain).

nelsonkram:
I removed the checkbox for hardware - let me know if more is required!

mxxcon:
@rpr4394 All looks good, except could you please rename image file and reference to it in yml to something more descriptive than ido-32.png
Something like idopener.png would be good. Makes it easier to identify which entry this image belongs to.

nelsonkram:
Renamed ido-32.png as per suggestion.

rugk:
@rpr4394
Could you please explain why this is a **two**-factor-authentication?
I mean your product is a nice idea, but what is the second authentication factor? As far as I understand it you just have to click on a button or scan the QR code to login. So IMO this is more a password alternative than a 2FA.

PR #1023, opened 2015-03-20
Coinify patch

RichJeanes:
Looks like BIPS got bought out by Coinify, updating the site to match. Right now, looks like Coinify supports TFA only on their merchant accounts, not their trade accounts. I couldn't find any documentation, but I have an email off to them about both of those things (see image at bottom). I'm going to create this PR as a placeholder, but maybe we should hold off on merging it until I hear back from them?

Also, I listed hardware support because they support printed character matrix cards (which they call "Secure Cards") in addition to TOTP. I know it isn't _electronic_ hardware, but I didn't really know how else to categorize it.
securecard

Also alphabetized CEX.IO

coinify


mxxcon:
I consider printed grid as still "hardware".

RichJeanes:
Also, as a note, the grid looks like it's generated when you visit the page while you have the feature disabled on your account, but then the characters are not visible once the feature is enabled. I assume to get a new card, you just disable, print, then re-enable.

RichJeanes:
I haven't heard back from them yet, but I'm thinking we go ahead and merge it without a docs link. I'll update whenever I hear from them. Sound good?

Carlgo11:
sounds good

RichJeanes:
Update: I got a response and they said that they will add both documentation for their existing TFA and TFA for trade accounts to their to-do list. So nothing to update on the site entry for now. And then they included links to seven different versions of their logo... Helpful, I guess, though I was a little confused by the file extension on the first one. Haha. Apparently the svg version also exists if you just remove the .png extension on it.
(Sorry for super large image)
screenshot_2015-03-26-11-15-42

PR #761, opened 2014-09-08
Add carriers category + add AT&T

Carlgo11:
Had to add some website to the category so I added AT&T.
inb4 carriers spam.

I made this PR at request from @wilsonmar in #723.

PR #1024, opened 2015-03-20
Adding Utilities Category

RichJeanes:
This is something that has been sitting idle for a while now in #761 (patch-6)  so I thought I'd bring it up to speed and see if we can get it live.

It's not ready yet and I could use some help with it. I think what I have listed now is a good start for the category to get visitors thinking about who else to add and covers a lot of the major players. I'll take care of all the icons that are missing right now, but I only use two of the services listed, so any help there would be much appreciated. I can take care of Cox and Verizon Wireless, but I don't know about any of the others.

As a side note, these web sites are terrifying. A good number of them don't support HTTPS on their home pages (even if said page has login forms) and most of those that do are still using MD5 and RC4. WTF?! :scream:


Carlgo11:
You have to admit that our error script is neat :smile: 
I've never seen a such detailed report on travis-ci

Carlgo11:
So yeaaaah, I added the rest of the images and accidentally added a commit updating Circle's doc link. When I removed it from this PR I must have accidentally added a --force for some reason. Sorry about that :disappointed:

smholloway:
Looks good to me :+1: 

My only complaint is that to me "utilities" means electric, gas, water and wastewater, trash disposal, etc. Wikipedia says my definition is a [public utility](https://en.wikipedia.org/wiki/Public_utility). I agree that the companies in this PR deserve to be listed, but I don't want everyone to demand that their local energy provider be listed.

mxxcon:
@smholloway why not? Likelihood of that happening is relatively low, but I wouldn't be against PSEG or ConEd or whoever provides power/water to you to also support that.
Seth, how else would you describe the category of phones, ISPs, cable/satellite tv?

smholloway:
Defining this category is really hard. I'm fine with it going in as `utility` because its good enough and we can rename later. I didn't mean to stop the PR--only to express my concern. I'm 100% :shipit:

PR #1028, opened 2015-03-23
Remove bips

Carlgo11:
Hi y'all!
So travis outputs an error on bips.me. I looked a the link and saw this:

So I guess there's no need to list bips now.
It doesn't look like coinify is here tho so maybe that's a future site to add to the list.
I'll contact their support and ask about two factor auth.
For now though, here's a commit removing bips from the list.

Take care!

//Carl


RichJeanes:
#1023 Already gotchya :wink:

Carlgo11:
heh, okay then

PR #1029, opened 2015-03-24
Add new Hosting PlanetHoster.net

NormandLemay:


mxxcon:
@NormandLemay I think it's better suited for `Hosting/VPS` section rather than domain, don't you think so?

jamcat22:
I agree with @mxxcon.

NormandLemay:
I agree with you : @mxxcon , @jamcat22, I will move it.

mxxcon:
What kind of hardware 2fa does it support? I can't read any of that. :/

NormandLemay:
The hardware 2fa is usb keys but, I think his just for the back office. Not for everyone.

Carlgo11:
If the TFA feature isn't for everyone then we won't add it

NormandLemay:
No problem I will remove it

Carlgo11:
Oh that logo is quiet large.
Here's a compressed version of the logo you added:
planethoster

When you've replaced the current logo with that one I think we're done and ready to merge.

Carlgo11:
Nice. Thank you :smiley:

_data/hosting.yml
```diff
@@ -95,6 +95,14 @@ websites:
       tfa: No
       twitter: ovh_support_en
 
+    - name: PlanetHoster
+      url: https://planethoster.net
+      twitter: PlanetHoster
```
Carlgo11: One last thing, could you remove the twitter handle?
Twitter isn't needed for sites already supporting TFA.

PR #1036, opened 2015-03-27
Slack now support 2FA

didlix:
https://slack.zendesk.com/hc/en-us/articles/204509068-Enabling-two-factor-authentication

### DIFF
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -87,7 +87,7 @@ websites:
       url: http://www.slack.com
       twitter: SlackHQ
       img: slack.png
-      tfa: No
+      tfa: Yes
 
     - name: ToutApp
       url: http://www1.toutapp.com/


jdavis:
@didlix Would you be able to add a `doc:` value as well pointing to their documentation? I think this site would be best: https://slack.zendesk.com/hc/en-us/articles/204509068

RichJeanes:
Also, their Twitter tag can now be removed and `software: Yes` needs to be added.

mxxcon:
Does keeping twitter tag interfere with anything? If not, then I think we should keep them just in case for the future..

Carlgo11:
omg ;-; #ninjad

Carlgo11:
actually, I'm closing this PR since it needs more work and mine's complete. Hope that's okay with everyone.
It just feels a lot easier to do it like that.

See #1041

Issue #948, opened 2015-01-31
No two-factor on Lloyds Bank


Terrance:
The docs link on [enhanced authentication](http://www.lloydsbank.com/help-guidance/security/authentication-procedure.asp) is only for adding new payees on an account - there are currently no 2FA options when actually logging in.

I guess that is still a form of 2FA, but it seems a bit misleading in the list?

MethodGrab:
:+1: agreed!

Carlgo11:
I agree with you as well @OllieTerrance. Could you make a PR removing the doc link or replacing it if you find a better page to link to?

If not then one of us well probably do it though that might take a while.

jamcat22:
Only one problem, we have many sites (Wells Fargo, NatWest, and other sites) that work this way too (only use 2FA to protect certain things).

What should we do about those?

plecavalier:
Could add a column for partial / select areas. Or whole site versus
portions thereof...etc

RichJeanes:
I know we have at least one site listed (Domains/Dynadot) that uses an exception flag to point out that it only protects certain actions with TFA. If people would like to update sites to add exceptions, PRs are welcome.
Closing for house keeping.

mxxcon:
@RichJeanes But then we also have sites that are tagged as `no` because they support 2fa only for some functions.. In our own https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions very last section covers this.

jamcat22:
How about we tag them as no, but add an exception?

PR #1042, opened 2015-03-28
Add exception for Lloyds Bank

Terrance:
2FA is only used for confirming the addition of new payees, not for logging in to online banking (see issue #948).

The docs link stays as-is as it seems to be most descriptive of the behaviour.  This still shows Lloyds as supporting 2FA - perhaps it may be more suitable to consider it non-2FA altogether.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -173,6 +173,8 @@ websites:
       img: lloydsbank.png
       tfa: Yes
       phone: Yes
+      exceptions:
+            text: "Only for adding a new payee to your account, not for logging in to online banking."
       doc: http://www.lloydsbank.com/help-guidance/security/authentication-procedure.asp
 
     - name: Natwest UK


Carlgo11:
I think it's better just to set TFA to no.
What does the other collaborators say?

<i>We should really have a `votes` label for these kind of things. *cough* @jdavis *cough*</i>
Edit: I added the label myself. Didn't know Collaborators could do that :o

mxxcon:
I think exception the way @OllieTerrance added is good enough.

RichJeanes:
While I would of course love to see them support TFA at login, something is better than nothing. Perhaps we could add a Twitter button for sites with exceptions so people could Tweet at them to improve/expand their TFA support?
I vote the exception is added and `tfa` stays `Yes`.

mxxcon:
This could be opening a full can of worms, but maybe `tfa:` `yes`|`no`|`incomplete`? And `incomplete` would still include public shaming button?

RichJeanes:
Well I didn't say shaming. More of an encouragement. Like "Thanks for doing what you have. We would love to see more of it!" kind of tweet.

jamcat22:
I agree. ^

Carlgo11:
Oh look at this:

> As an example, a site that prompts you for an authentication token following user login would be considered Two Factor Auth. A site that does not prompt you for a token upon login, but prompts you for a token when you try to perform a sensitive action would not be considered Two Factor Auth.
> - [CONTRIBUTING.md](https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions)

So I guess that's a `tfa: No` then.

PR #1044, opened 2015-03-31
Add TIDAL to entertainment

Carlgo11:
So apparently there's this new music streaming service that everyone talks about.
Thought it would be a good idea to add it to the list.


smholloway:
It'll be cool when new sites are created with 2FA! Maybe the next big music streaming service will get it right...

Carlgo11:
Yup but security is often not the first priority for startups. It should be tho...

PR #1049, opened 2015-04-01
add kayako.com

jmedwards:
Two-factor authentication coming soon


mxxcon:
@jmedwards that url should be `status:` not doc.
Also that logo is 18kb, way too big than it can be. Please compress that png.

Carlgo11:
The site supports https. please change the url

_data/other.yml
```diff
@@ -79,6 +79,13 @@ websites:
       software: Yes
       doc: https://support.formstack.com/customer/portal/articles/1833071-two-factor-authentication
 
+    - name: Kayako
+      url: http://www.kayako.com
```
Carlgo11: http to https please.

PR #1050, opened 2015-04-01
Add Zocle

zhexiao:


mxxcon:
@zhexiao You need to add 32x32 png logo as well.
The `doc:` link seems to be a "print to pdf" file of what appears to be their docs site. Can't you use the original article instead of this?

Carlgo11:
I don't think that exception is needed.

zhexiao:
Thank you very much. I will fixed it ASAP.

apolakos:
Hi, the reason we submitted a print to pdf is because the tutorial for the doc is behind a login for users (Otherwise, we can leave the doc out.).  We'll submit the logo shortly.  

EDIT:  Nevermind, we will link to our announcement blog post regarding the availability of two-factor which is public accessible.

mxxcon:
@apolakos / @zhexiao Is it logical to hide documentation behind a login in the 1st place? Seems like unnecessary hoops to jump through for your users and harder to find relevant answers using regular search engines.

PR #1051, opened 2015-04-01
Loopia now supports TFA

Carlgo11:
@StefanWallin informed us earlier today that Loopia AB now support TFA in the form of BankID.
BankID & Mobilt BankID (Mobile BankID) is an app that's provided by most Swedish banks to it's customers and used to login to your online banking but also used by several governmental authorities, for example by the tax authorities when registering a company (personal experience :wink:).

Here's how it works:
1. You enter your unique birth number (similar to social security ID)
<br>
2. You open your BankID app and enter your PIN code <i>(that you chose when registering at your local bank office)</i>.
<br>
3. You're logged in.

Yes this is only available for Swedes but there are other sites that are region specific on our list.
PayPal for example doesn't provide TFA outside the U.S.


StefanWallin:
:+1:

jamcat22:
Can we put an exception that it only works in Sweden?

mxxcon:
Is it always the same pin number for the same bank?

StefanWallin:
It's Always the same pin for that instance of your personal identification certificate. You can have 3 desktop certificates and two mobile certificates that authorize the same identity(it authorizes your Swedish version of a "social security number" that all legal swedish citizens have and get assigned at birth). The multiple certificates is to allow installation on different devices.

So you could potentially have 5 different pins. Most people just use the same.

What the pin does is to decrypt a high grade client certificate which is used to generate an OTP for authorization

mxxcon:
Hmm..should this be under domains or hosting?

Carlgo11:
domains. That's their number 1 product and what they started as

Issue #1045, opened 2015-03-31
SpiderOak has 2FA


almafeta:
#211 reported SpiderOak had disabled 2FA permanently, but that was only a temporary bug with the SMS service, not a permanent shutdown.  backup.yml should be updated to the following:

```
  - name: SpiderOak
    url: https://spideroak.com/
    twitter: spideroak
    img: spideroak.png
    tfa: Yes
    sms: Yes
    exceptions:
        text: "US and Canada only.  Not available for free accounts."
    doc: https://blog.spideroak.com/20110620235134-2-factor-authentication-to-your-spideroak-account
```

I don't know where the token request URL https://spideroak.com/tokenrequest should go in this project's format.

mxxcon:
@almafeta would you like to make a pull request with these changes?

PR #1052, opened 2015-04-01
Change SpiderOak status

Carlgo11:
As brought up in #1045 SpiderOak is working on TFA and have a beta version available for NA customers. They also say that they'll make this available for all countries in a near future and so I thought an exception wasn't needed as it we'll only have to remove later.
After all, the `doc` link explains it all.

Tell me your thoughts on this!

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -133,7 +133,9 @@ websites:
     url: https://spideroak.com
     twitter: spideroak
     img: spideroak.png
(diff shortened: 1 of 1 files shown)


mxxcon:
That blog post is 4 years old...are they "still" working on 2fa in other regions? :-1:

Carlgo11:
Oh I didn't see that. Closing

PR #1053, opened 2015-04-01
Add TFA provider BankID

Carlgo11:
I think most of the stuff have been explained in #1051.
Except from adding BankID I also moved Azure since `U` comes before `Z` in the alphabet :wink:


mxxcon:
(compress this png. shouldn't be 3.8kb)

PR #1054, opened 2015-04-02
Add Emailcenter Maxemail to communication category

chrisminett:


mxxcon:
@chrisminett Does it use yubikey?

chrisminett:
@mxxcon Yup

PR #1073, opened 2015-04-16
Changed PayPal exception, added more countries

rugk:
As described in https://github.com/jdavis/twofactorauth/issues/1072 PayPal supports 2FA in more countries than US/Canada.
Because of this I added this information.

### DIFF
--- a/_data/payments.yml
+++ b/_data/payments.yml
@@ -58,7 +58,7 @@ websites:
       software: Yes
       hardware: No
       exceptions:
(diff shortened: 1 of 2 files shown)


jamcat22:
Anybody see any problems?

rugk:
I also removed the wrong information in the eBay hint about PayPal.
I'm not sure if the list of countries where PayPal supports 2FA is the same for eBay so I don't added this. We would need some verification about this.

Issue #1026, opened 2015-03-22
Let's Fix PayPal


jamcat22:
Get ready for the longest description of an issue yet. There are currently about 14 issues/PRs related to PayPal. Let's face it, their 2FA is broken. It's inconsistent, it's not user-friendly, there aren't adequate docs, and there are many loopholes (but I will give them a chance to fix them first). Let's fix it!
# Todo

(A bullet point means in progress.)
- [x] Determine country compatibility. (See #1073.)
- [x] Fix compatibility matrix on main page. (See #1074.)
- [x] Determine compatibility with desktop account page.
- [x] Determine compatibility with desktop checkout.
- [x] Determine compatibility with mobile account page.
- [x] Determine compatibility with mobile checkout.
- [ ] Determine compatibility with physical store checkout.
- [x] Determine compatibility with iOS apps.
- Determine compatibility with iOS APIs.
- Determine compatibility with Android apps.
- Determine compatibility with Android APIs.
- [x] Determine compatibility with Windows Phone app(s).
- [x] Determine compatibility with Windows 8/8.1/10 app(s).
- [ ] Redo exceptions and add more information on the main page.
- [ ] Write new doc page using information gathered and include instructions on how to use and activate 2FA for PayPal.
# Key
- [ ] Login=Can you even login with any 2FA enabled?
- [ ] SMS=Can you use SMS 2FA?
- [ ] Software/Hardware=Can you use Software/Hardware 2FA?

A dash through a property indicates that it is not applicable.
A question mark next to a property means the status is unknown.
A bullet point means the status depends on certain conditions.
When applicable, a screenshot will be provided below or next to the property.
# Status
### Account Page
- [x] Login
- [x] SMS
  ![smsyes](https://cloud.githubusercontent.com/assets/1897053/10724464/640e45b0-7b8a-11e5-99d4-81259e993c27.png)
- [x] Software/Hardware
  ![shyes](https://cloud.githubusercontent.com/assets/1897053/10724466/67db1a7e-7b8a-11e5-8a5e-2b88b09e9ad3.png)
### Checkout
- [x] Login
- [x] SMS
- [x] Software/Hardware
### Responsive Account Page

This is the new design. It is the same as the desktop account page (with the option to login with phone number and PIN added), and therefore inherits its login compatibility.
### Responsive Checkout

This is the new design. It is the same as the desktop checkout, and therefore inherits its login compatibility.
### Old Checkout (Typically used for donations.)
[...]

PR #1074, opened 2015-04-16
PayPal: hardware token supported

rugk:
The main issue of https://github.com/jdavis/twofactorauth/issues/1072 was something else, but PayPal also supports hardware token authentication.

### DIFF
--- a/_data/payments.yml
+++ b/_data/payments.yml
@@ -56,7 +56,7 @@ websites:
       tfa: Yes
       sms: Yes
       software: Yes
-      hardware: No
+      hardware: Yes
       exceptions:
           text: "2FA only available in the US and Canada. SMS only available in the US. Hardware-based 2FA is no longer available."
       doc: https://www.paypal.com/us/cgi-bin?cmd=xpt/Marketing_CommandDriven/securitycenter/PayPalSecurityKey-outside&bn_r=o


mxxcon:
Perhaps combine this with #1073?
Do you have confirmation that hardware token is still in use? https://www.paypal.com/us/webapps/mpp/security/security-protections has no reference to it.

rugk:
Yes you can combine the pull requests if you want. (I didn't knew how to do, so I just created a new one)
I already linked to the [German page](https://www.paypal.com/de/cgi-bin/webscr?cmd=xpt/Marketing/securitycenter/general/PPSecurityKey-outside) but by changing the URL a bit you can also get the English page with this information:

> We can send your security code to your mobile phone by SMS message or **you can order a credit-card size Security Key to carry with you**.

https://www.paypal.com/us/webapps/mpp/security/hardware-software-protection#logging_in-security_key

BTW here is even a huge FAQ about it: https://www.paypal.com/us/webapps/mpp/security/security-key

jamcat22:
One problem. In the US at least you cannot order a new physical security key, but you can register an old one you already have. We also need to create a new docs page for PayPal so that we can sort some of their other weirdness out. I vote for merging this but then adding more information in the docs and update the exception with more information after #1073 is merged.

See #1026 for more info.

rugk:
Yes I agree with this.
As for the doc I think it might be important to mention that PayPal offers a "backdoor" (or call it workaround :wink:) - and it's not the usually used one with recovery codes.
Look at these questions in their [FAQ](https://www.paypal.com/us/webapps/mpp/security/security-key):

> **My text messages don't always come through. Can I still log in without a code?**
> Yes, but we'll ask you a few security verification questions so we're sure it's you. And in the meantime, make sure we have your correct phone number on file, as this could be the reason for the problem.
> **Can I still log in to my PayPal account if I lose or break my key, or if I don't have my mobile phone with me?**
> Yes, as long as you can answer a few security questions correctly.

I don't know what questions they ask in this case, but personally I wouldn't trust some "verification questions" so much.

jamcat22:
Ok. All in agreement?

davehatz:
New Hardware tokens are not available from PayPal but that's not the end of the world. Paypal allows BYO VIP keys.
I don't like the security of the phone or desktop apps.  Kinda defeats the purpose of 2FA to have the secret key generated on the same device you're using to login.
You can buy New Symantec VIP hardware tokens with displays directly from Symantec or you can buy a New Yubikey VIP USB client device.

PR #1079, opened 2015-04-22
Update OVH

tdutrion:
As precised in the following tweet, OVH now supports TFA with both sms and software (mobile apps).

https://twitter.com/OVH/status/590450825425653760

### DIFF
--- a/_data/hosting.yml
+++ b/_data/hosting.yml
@@ -92,7 +92,9 @@ websites:
     - name: OVH
       url: http://www.ovh.com/
       img: ovh.png
-      tfa: No
+      tfa: Yes
+      sms: Yes
+      software: Yes
       twitter: ovh_support_en
 
     - name: Rackspace


Carlgo11:
@tdutrion any docs to prove this with? Or is the only documentation a badly-written tweet?

tdutrion:
I just asked @olesovhcom on twitter if any documentation exists, I haven't seen any so far. I tried it though and I can testify it works for both sms and Google Authenticator.

Carlgo11:
Okay. Let's wait and see if they reply. It's always better with a doc link.

tdutrion:
According to the [latest discussion on Twitter](https://twitter.com/ovh_uk/status/599145384565346305), there is apparently no plans of a proper documentation...

My experience with both SMS and Google Authenticator are successful and I can't really see what documentation could be required, but my view is biased here as I use 2 factors for a while on many providers.

PR #1080, opened 2015-04-22
ovh now has 2fa

speed47:
supported 2FA types :
- SMS
- TOTP (google authenticator / freeotp / etc.)
- Static OTP codes (not advertised on twofactorauth.org as I can't find a category for that)

also changed the OVH logo to the new one


Carlgo11:
Already been brought up in #1079 tho if you find an appropriate doc then we might use this one instead of that one.

speed47:
Didn't notice the other merge request indeed, sorry for the duplicate.
Both seems to be the same as far as the yml is concerned, except that I took the opportunity to also update the logo in this one.
As for the doc, you're right, there's none yet, which should hopefully change in the near future (I work at ovh and i'll try as much as I can to push that internally).
Also hardware tokens should also be supported at some point as we have started work on those (yubikey & u2f), so there should be another merge request in a couple weeks/months I guess, from me or any other member of the community !

PR #1083, opened 2015-04-23
Added the old school favicon for Yahoo! Japan

havill:
Yahoo! Japan never changed its logo compared with the main Yahoo!


havill:
Forget to commit to yaml to the first pull; the second request has it all.

Carlgo11:
bump

PR #1084, opened 2015-04-23
Add countermail

Carlgo11:


RichJeanes:
I don't know how I would categorize the use of a key file... it's not actually authentication hardware and can be relatively easily copied as opposed to dedicated hardware.

mxxcon:
Is it similar to Google's https://support.google.com/accounts/topic/6103521?hl=en or more like http://keepass.info/help/base/keys.html#keyfiles ?

Carlgo11:
It's like keepass' but requires a password as well

RichJeanes:
So we're going to say key files constitute hardware auth because you can use key files to create your own hardware keys?

mxxcon:
I'm personally still on the fence regarding key files...the same reason as `email` as one of the factors.

PR #1087, opened 2015-04-24
Added VerIDial provider

jzelos:
https://www.veridial.co.uk/


Carlgo11:
Hi Jason,
Thank you for contributing to twofactorauth!
While the provider just above you somehow slipped away with using a messed up ordering for the diffirent strings I'd like to see that everyone use this ordering:

```
- name: <site name>
 - url: <site link>
 - img: <site name .png> (We'll go back to this later)
 - hardware: Yes (If hardware authentication is available)
 - sms: Yes (If SMS authentication is available)
 - software: Yes (If software based authentication is available)
 - phone: Yes (If phone call based authentication is available)
```

If you don't support one of the tfa solutions then just don't add it instead of adding a `No`.
It looks like you used a jpeg picture and just changed the name to .png. Please convert the picture to png or find another picture that uses the png format.
Also, try to compress the picture. While this one is rather compressed it could be compressed more.
I recommend using [tinypng](https://tinypng.com/) if you don't have another compression tool.

Once again,
thanks for contributing!

jzelos:
Cheers for the info,  should email be just below software? 

Sent from my HTC

----- Reply message -----

Carlgo11:
above I'd say. Doesn't really matter as long as name is first, url is second and img is third.

_data/providers.yml
```diff
@@ -261,6 +261,15 @@ providers:
     sms: Yes
     software: Yes
     url: https://www.vasco.com/
+
```
mxxcon: There's tab character on blank line. In general our yaml doesn't allow tabs, only spaces.

jzelos:
I'll get these errors fixed on Monday. 

How we ended up renaming an image file instead of converting I don't know.....

jzelos:
Previous build failure was a build bug with Travis.

Carlgo11:
So email and software? I don't see anything about that on the site.

jzelos:
Email isnt advertised much caus its not something we like folks using, but is mentioned under pricing and a few other places. 

There is an iOS app in final testing and an android port in progress. The iOS app should be publicly available within a few weeks. 

Sent from my HTC

----- Reply message -----

jzelos:
Do you guys need any more changes?

Carlgo11:
I don't know about the app part. If it doesn't currently exist then we would lie if saying your product supports it. Maybe remove `software: Yes` for now?

jzelos:
Fair enough, can always be added in later. I was hoping we would have had it released by now but as always that last 10% never seems to end.

PR #1089, opened 2015-04-26
Finance: Check is now Mint Bills

mark-adams:
In December 2014, Check was acquired by Intuit and renamed Mint Bills. This PR updates the site to reflect that change:
- Check Inc. is now "Mint Bills (formerly Check)" and reflects the new Twitter and URL
- Logos for both "Mint" and "Mint Bills (formerly Check)" have been changed to reflect the new Mint logo after their recent rebranding


mxxcon:
It's not critical, but Mint and Mint Bills have different logos. :)
Mint's logo has blue and green stripes while Mint Bill's logo has blue and yellow stripe.
If you want to update that, that's cool.

PR #1091, opened 2015-04-27
Compressed all images

Carlgo11:
Hi everybody!
So I decided to look at how to automate compression of the images. It turns out not that many services provides an API and those who do want you to pay for it.
I found out that tinypng (that i've used before) provides an API for free for up to 500 pictures.

For those of you who are curious on how I did it, here's the php code I used:

```
<?php
ini_set('max_execution_time', 3000);

$key = ""; //tinypng api key.
$files = glob('img/*/*.{png}', GLOB_BRACE);
foreach($files as $input){
    echo $input."<br>";

$output = "comp-img/".$input;

$url = "https://api.tinypng.com/shrink";
$options = array(
  "http" => array(
    "method" => "POST",
    "header" => array(
      "Content-type: image/png",
      "Authorization: Basic " . base64_encode("api:$key")
    ),
    "content" => file_get_contents($input)
  ),
  "ssl" => array(
    /* Uncomment below if you have trouble validating our SSL certificate.
       Download cacert.pem from: http://curl.haxx.se/ca/cacert.pem */
     "cafile" => __DIR__ . "/cacert.pem",
    "verify_peer" => false
  )
);

$result = fopen($url, "r", false, stream_context_create($options));
if ($result) {
  /* Compression was successful, retrieve output from Location header. */
  foreach ($http_response_header as $header) {
    if (substr($header, 0, 10) === "Location: ") {
      file_put_contents($output, fopen(substr($header, 10), "rb", false));
    }
  }
} else {
  /* Something went wrong! */
  print("Compression failed");
}
}
```


Carlgo11:
Looks like GitHub didn't like that many changes so I guess we'll have to merge locally then.

mxxcon:
Tinypng is a LOSSY compressor.
The more times we'll use it, the higher image quality degradation.
I'm against using Tinypng on mass scale.

If need to, every once in a while I can re-compress all images using PngSlim and ScriptPNG. Most likely resulting images will be even smaller than tinypng and they will be 100% visually identical.

@Carlgo11 Your effort is well-meaning, but I'm :-1: on this PR.

Carlgo11:
@mxxcon [this](http://css-ig.net/pngslim)?
I just tried it but it doesn't look like it's working for me.
screenshot

mxxcon:
probably because of file path? I have no problem running it..

Carlgo11:
Ah right. I hope the one who came up with putting spaces in `Program Files (x86)` died a painful death.

mxxcon:
The original `/images` was 561868 bytes.
After running it through pngslim and fastio, I got 542144 bytes.
Saved ~19kb. I don't think it's worth making another commit and growing log even larger.

Carlgo11:
Okay closing this PR then. (although it would be nice with some kind of script to automate the compression of images)

PR #1094, opened 2015-04-28
add privacyidea

cornelinux:
privacyIDEA is an open source system to manage all kinds of tokens. 
You can run it on premise in your own data center/network.


Carlgo11:
@cornelinux look at our [wiki](https://github.com/jdavis/twofactorauth/wiki/I-want-to-add-a-TFA-provider-to-the-list) on what's needed in the .yml file for providers as you seem to have copied the requirements for a site rather than provider.

cornelinux:
seems i need to remove the tags "tfa" and "twitter".
OK will do so.

Carlgo11:
Yes and doc as well I think

cornelinux:
other providers also contain the tag "doc".
If it is not used, it can be removed...

Carlgo11:
Ah yes they do. May bad.

PR #1097, opened 2015-04-28
Add new provider DigiFlak

Driim:


mxxcon:
This service has invalid SSL cert and when I load the page all I see is:

> You were probably seeking something else, but this is all we have to offer at the moment. In all likelihood, the page that used to be here, has been moved or removed.

Carlgo11:
@mxxcon the site looks to be parked at a domain registrar called "zone15".
I seriously hope @Driim entered the wrong url...

Driim:
I tested site in http, https version doesn't work.

Carlgo11:
Pro tip: If you don't support https then don't use a https url :P

mxxcon:
I'm hesitant that a provider doesn't even have their own ssl... :/

Driim:
But you can buy them on Amazon.

Carlgo11:
so?

Driim:
I mean device must work even without ssl on their site :)

jamcat22:
The providers page is being removed. You can read more about it on #1177.

PR #1105, opened 2015-05-01
Update psn link

Carlgo11:
So playstation change their psn link. It looks like there's no en-us version so I guess we'll have to use en-gb then.


mxxcon:
Why not just link it to https://www.playstation.com?

Carlgo11:
sure, we could do that.

PR #1107, opened 2015-05-02
Update CloudFlare

ghost:
It asks for your phone number, but there is no way to receive tokens via text messages. Only Authy is supported.

### DIFF
--- a/_data/domains.yml
+++ b/_data/domains.yml
@@ -23,7 +23,6 @@ websites:
       url: https://cloudflare.com
       img: cloudflare.png
       tfa: Yes
-      sms: Yes
       software: Yes
       doc: https://support.cloudflare.com/hc/en-us/articles/200167866-What-is-Authy-2-Factor-Authentication-
 


mxxcon:
Doesn't Authy works through sms?

Carlgo11:
It does and they do.
@nyuszika7h I suggest you look at Authy's website.

Here's a screenshot:
authy supports software, sms, phone

ghost:
Would you care to explain how it works with CloudFlare? It provides you no way to proceed at all without installing the Authy app. Just because Authy supports SMS (unfortunately) doesn't mean every site that implements Authy also does.

I've only seen that work with Humble Bundle.

Carlgo11:
Hmm true.
I thought they supported that though I could see why they don't as SMS is more insecure than the Authy app.
Could someone who doesn't use Authy confirm that Cloudflare doesn't support SMS TFA?
I use Authy for everything :see_no_evil:

mxxcon:
CloudFlare requires Authy and step 2 asks for your phone number. I'm not sure if it's for sms or to identify you as a user in Authy's system.
clipboard01

(I can't enable 2fa on this account since it's shared at work.)

ghost:
Like I said, I already tried that, and I didn't get any text message. The next step told me to install the app.

(I really wish Authy and/or CloudFlare used TOTP. It's not just that Authy is proprietary, but it's also reinventing the wheel and using non-standard tokens for no good reason, when everyone else was happily using TOTP before. Sure, Authy **supports** TOTP tokens, but other apps don't support Authy tokens, which are a product of [NIH syndrome](https://en.wikipedia.org/wiki/Not_invented_here).)

PR #1118, opened 2015-05-06
Added Rockstar Games Social Club

UmbyUmbreon:
The Rockstar Games Social Club is a social network which hooks into games created by Rockstar Games. Users can purchase and permanently link new games to a Social Club account, and in the PC version of Grand Theft Auto V, users can purchase virtual currency with the option of saving credit card details and associating it with their Social Club account. Unauthorized access to a person's Social Club account can potentially result in unauthorized transactions.


mxxcon:
where can you add credit card to the account?

UmbyUmbreon:
That's done in-game in Grand Theft Auto V when you go to buy in-game currency. I can provide screenshots a little later if required.

mxxcon:
you enter your cc# directly from the game? :fearful: So do all microtransactions are always performed only in-game?

UmbyUmbreon:
To Rockstar's credit, I think the Social Club program uses a HTML wrapper in its in-game overlay to present the transaction screen, and that uses HTTPS, but don't quote me on that. Just found there is also the possibility to buy microtransactions using Rockstar's [online storefront](http://www.rockstarwarehouse.com/store/tk2rstar/en_US/pd/productID.317353200) as well (at least for the PC version), which is also tied to a Social Club account. Microtransactions for both Xbox and both PlayStation versions are, however, handled using the usual methods through the PlayStation Store, etc.

e: removed a redundant 'the', whoops!

_data/gaming.yml
```diff
@@ -104,6 +104,12 @@ websites:
       twitter: AskPlayStation
       img: psn.png
       tfa: No
+    
+    - name: Rockstar Games Social Club
+      url: http://socialclub.rockstargames.com
+      twitter: RockstarGames
```
Carlgo11: We usually use the support twitter account if available since most companies listen to that more frequently.
Please change the twitter handle to `RockstarSupport`

UmbyUmbreon: Aaaand done!

PR #1119, opened 2015-05-06
Updating OVH's status

valentindotxyz:
OVH now provides 2FA :) https://www.ovh.com/fr/apropos/securite.xml#double-authentification

### DIFF
--- a/_data/hosting.yml
+++ b/_data/hosting.yml
@@ -92,7 +92,9 @@ websites:
     - name: OVH
       url: http://www.ovh.com/
       img: ovh.png
-      tfa: No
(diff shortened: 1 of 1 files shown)


mxxcon:
2 things:
This is already addressed in #1079.
Your specific link seems to be a leftover page. OVH used to have that text on their UK, US and German sites, but then removed it. I think they just forgot about French version.
You can see archived version here https://web.archive.org/web/20140905101412/http://www.ovh.com/us/about-us/security.xml

mxxcon:
Closing in favor of #1079.

Issue #1112, opened 2015-05-05
Image size


Carlgo11:
Hi everyone!
Not too long ago I made a compression PR that we decided to close due to that the service I used did a lossy compression. Since then I've thought about new ways of doing this without affecting all images that doesn't need to be compressed.

I came up with a solution using travis-ci to output a list of images that are too big.
That way we could easily just tell people to compress, or maybe even make travis output a link to a compressed version of the images.
Thought about this?

So yeah, I don't know what you guys think about it but if we do decide to implement it, what size would be the maximum allowed for logos?

Carlgo11:
Closing since #1120 is up.

PR #1120, opened 2015-05-06
Add image size check

Carlgo11:
Hi,
I mentioned in #1112 that I were working on a image size checker for the verify.rb code.
Here it is.

Right now it exits with 1 (As in the build fails) when an image is larger than 2500 B.
I didn't know if we wanted it to fail and therefor I added the option to disable that by setting `$max_size_ignore` to true.
I also added an `$error_ignore` just in case we'd need that. (that would useful for debugging)

### DIFF
--- a/verify.rb
+++ b/verify.rb
@@ -2,45 +2,67 @@
 require 'yaml'
 require 'fastimage'
 $output=0;
+$image_error_output=0;
+$max_size = 2500 # Max file size (Bytes)
+$max_size_ignore = false #Ignore max file size errors
+$error_ignore = true  #Will output the error but won't make the build fail.
+
 begin
   def error(msg)
     $output=$output+1;
-    if($output == 1)
+    if ($output == 1)
       puts "<------------ ERROR ------------>\n"
     end
     puts "#{$output}. #{msg}"
 
   end
 
-  # Load each section, check for errors such as invalid syntax
-  # as well as if an image is missing
+  def image_size_error(msg)
+    if ($max_size_ignore == false)
+      $output=$output+1;
+    end
+      $image_error_output=$image_error_output+1;
+
+    puts "#{$image_error_output}. #{msg}"
+  end
+
+  def check_image(image)
+    unless File.exists?(image)
+      image_size_error("#{website['name']} image not found.")
+    else
+
+      image_dimensions = [32, 32]
+
+      unless FastImage.size(image) == image_dimensions
+        image_size_error("#{image} is not #{image_dimensions.join("x")}")
+      end
+
+      ext = "\.png"
+      unless File.extname(image) == ext
+        image_size_error("#{image} is not #{ext}")
+      end
+
+      file_size = File.size(image)
+      unless file_size < $max_size
+        image_size_error("#{image} is larger than #{$max_size} Bytes. (#{file_size} Bytes)")
+      end
+    end
+  end
+
+
+# Load each section, check for errors such as invalid syntax
+# as well as if an image is missing
   main = YAML.load_file('_data/main.yml')
   main["sections"].each do |section|
     data = YAML.load_file('_data/' + section["id"] + '.yml')
 
     data['websites'].each do |website|
       image = "img/#{section['id']}/#{website['img']}"
+      check_image(image)
 
-      unless File.exists?(image)
-        error("#{website['name']} image not found.")
-      end
-
-      image_dimensions = [32,32]
-
-      unless FastImage.size(image) == image_dimensions
-        error("#{image} is not #{image_dimensions.join("x")}")
-      end
-
-      ext = ".png"
-      unless File.extname(image) == ext
-        error("#{image} is not #{ext}")
-      end
     end
   end
 
-  if($output > 0 )
-    exit 1
-  end
 rescue Psych::SyntaxError => e
   puts 'Error in the YAML'
   puts e


RichJeanes:
I'm not sure about the best way to handle this. Setting a hard file size doesn't really seem to be a good measure. For example, running img/domains/netfirms.png through optipng still doesn't reduce it below 2500 B, even with -o9 it only tims 7 B bringing it down to 2710 B.

mxxcon:
Rich, I got it down to 2314 using http://css-ig.net/scriptpng

Carlgo11:
@RichJeanes do you see any noticeable difference?
old
new

RichJeanes:
Not at that size. Did you use lossy compression? I see you got it down to ~1600 B.

Carlgo11:
Yes I did. @mxxcon did a lossless though.

Carlgo11:
I can set the limit higher if that's what you want.

RichJeanes:
The dislike of having arbitrary hardcoded values is just something that's stuck with me from my comp sci classes. If lossy compression isn't going to cause any noticeable difference at these sizes and resolutions, then I don't really see a problem with it.

mxxcon:
As I chatted with Carl about it, I dislike automating lossy process since there's no easy way to review the results. I have no problem automating lossless re-compression. However, I don't know if existing solutions would do as extensive job as ScriptPNG does.
Perhaps we can hook some sort of other bot into this process beyond just Travis-CI to triage PRs? Such bot would look at a commit and comment on improvements, such as suggest user would recompress logo.
I dislike hardfail if an image is 2501 bytes...

Carlgo11:
@mxxcon I'm not talking about automating compression here. This is just to make it easier for us to review a PR.
Yes it would fail if it's 2501 bytes. Either we higher the max size or we set the script to ignore max size failures. It doesn't really matter if I'd add a code ignoring the file is it's close to the max size value. Then that would just fail when someone adds a picture 1 byte larger than the max size ignoring value.

One thing I could do is to add a file where we can put a list of images it should ignore. How about that?

jdavis:
Why are there so many dollar signs? What is this, PHP or Perl? =]

Carlgo11:
Haha no @jdavis. But I mostly code in PHP and Perl so... yeah... :wink:
I don't have that much knowledge on what $ signs does in Ruby. The only thing I noted is that without them functions can't use them.

Carlgo11:
@jdavis 
But except from the dollar signs, what do you think about the PR?

smholloway:
Currently, `verify.rb` ensures that we have a 32x32 PNG--do we need anything more? The big change here appears to be a file size check, but it's been a while so I've lost the context for the change.

Carlgo11:
@smholloway the context is that we receive quiet a few PRs containing logos that are unreasonably large. this is just a way to ease our work (and make sure we don't miss any) by automatically checking the size

smholloway:
What's the danger with logos being too large? GitHub allows us 1GB for the repo and `du -sh` shows that we're still under 20MB, so we have a while before we need to worry too much about size. 

The two largest directories appear come from images and the Semantic UI files:

```
1.7M    ./_site/img
2.2M    ./_site/lib
```

If we want to check image file size, what upper limit would appease the previous commenters?

mxxcon:
not the size of git repo but the webpage.
It seems like 2500 is a good size for 32x32 png with transparencies. But Travis can't do "soft" checks, like I wouldn't really want it to fail on 2501 bytes.

Maybe set it to like 10kb from ridiculously large images and then the rest will be taken care of by manual recompression process once every couple of month.

Carlgo11:
I could set an exception file for files like that.

smholloway:
Oh, yeah--we want the page to load quickly. Sounds like 5-10kb would be a fair compromise. 

I think we could reduce the size of the diff by adding only the image size check (and I'd base it off the current `master` branch where we have removed providers). I think the diff could be as simple as something like this:

```
      file_size = File.size(image)
      unless file_size < $max_size
        error("#{image} is larger than #{$max_size} Bytes. (#{file_size} Bytes)")
      end
```

Carlgo11:
Well I still think that it would be nice with at least some form of output when the files are large.
Just setting `error_ignore` to true would be good enough.

Carlgo11:
And the diff is larger because I've done some reformatting of the existing code so that it's easier to work with in the future.
Instead of doing the same thing twice for providers and non-providers I moved the code to a separate method that's run by both of the image verifications. (although we've removed providers I still think that proper code is a good thing)

I'll do some coding later tonight to fix the merge conflicts and remove the providers part.

Carlgo11:
There,
I did a `-f` push just to remove any unnecessary commits.

Now the code displays errors but won't fail the build because an image is too large.
I also removed the providers part of the code.
The diff is still a big as I cleaned up the existing code as mentioned earlier.

verify.rb
```diff
@@ -2,45 +2,67 @@
 require 'yaml'
 require 'fastimage'
 $output=0;
+$image_error_output=0;
+$max_size = 2500 # Max file size (Bytes)
+$max_size_ignore = false #Ignore max file size errors
+$error_ignore = true  #Will output the error but won't make the build fail.
```
smholloway, line 8: I don't see that this is used--can we remove it?

verify.rb
```diff
@@ -2,45 +2,67 @@
 require 'yaml'
 require 'fastimage'
 $output=0;
+$image_error_output=0;
+$max_size = 2500 # Max file size (Bytes)
+$max_size_ignore = false #Ignore max file size errors
```
smholloway, line 7: This value never changes, so I vote to remove it and the associated conditional on line 21.

verify.rb
```diff
@@ -2,45 +2,67 @@
...
     puts "#{$output}. #{msg}"
 
   end
 
-  # Load each section, check for errors such as invalid syntax
-  # as well as if an image is missing
-  main = YAML.load_file('_data/main.yml')
-  main["sections"].each do |section|
-    data = YAML.load_file('_data/' + section["id"] + '.yml')
+  def image_size_error(msg)
```
smholloway, line 20: How do you feel about making this into a more generic `warning` function?

PR #1121, opened 2015-05-07
Updated EVE Online

okand:


mxxcon:
@okand Looks good, but could you please rephrase "we were unable to fix this time around" so that it sound like this site says that..

okand:
Hopefully this looks better. I guess just copying exactly what the site said was a bit careless.

jamcat22:
Sounds much better.

PR #1122, opened 2015-05-10
Added Twitter's single account problem

joesondow:
Many Twitter users have multiple Twitter accounts, and other Twitter accounts are managed by multiple people. Twitter's 2FA is useless for all those people, because they incorrectly assume a one-to-one relationship between a phone number and a Twitter account.

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -103,7 +103,7 @@ websites:
       tfa: Yes
       sms: Yes
       exceptions:
-          text: "SMS only available on select providers, SMS required for 2FA."
+          text: "SMS only available on select providers, SMS required for 2FA, 1 account per phone number, 1 phone number per account"
       software: Yes
       doc: https://support.twitter.com/articles/20170388
 


mxxcon:
@joesondow But this is similar to having an issue that Twitter allows only 1 password per account. That sounds ridiculous, doesn't it? :) Why would a service allow multiple passwords per account. But phone number of 2fa sms is no different. 2nd factor should not be treated with a higher or lower sense of security.
Yes, you share your password with your coworker, but that is a very bad security practice. You can not do that with a phone number. So why should they lower their security practices for this use case.

jamcat22:
Plus, you can now use TweetDeck to share Twitter accounts without sharing passwords. Since you sign in to TweetDeck with a Twitter account, it uses 2FA.

joesondow:
Okay. I've read about social media teams who don't use Twitter's 2FA because they need two people to be able to tweet to the same brand account from different phones, but it's up to you whether you think that's a problem. I see it as a problem, but this is your site so your decision stands.

What about the other problem? A single user who has multiple Twitter accounts can't use Twitter 2FA. This one directly affects me, as I maintain more than one Twitter account.

jamcat22:
I guess the other problem makes sense. Most other services let you use one phone for multiple accounts (like Google, Apple, etc.). What do you think @mxxcon?

mxxcon:
@joesondow multiple people accessing the same twitter account which has 2fa enable is problematic. but i think this kind of nuance will be irrelevant for most users of this site.
I think 1 account per phone number is potentially more problematic. How does twitter complain when you try to add the same phone number to a 2nd twitter account?

joesondow:
I just tried putting 2FA on my Twitter accounts to see if they'd made any progress. They haven't. When I try to associate my phone number to a second Twitter account, this is Twitter's response:

> The phone number you're trying to use is already associated with another Twitter account.
> 
> To add your phone, please remove it from any other accounts. Learn more

So basically I can't use 2FA on Twitter unless I acquire a unique SMS-capable phone number for each Twitter account I maintain.

mxxcon:
Review status: :shipit: all files reviewed, all discussions resolved, all commit checks successful.
Reviewed files:
- _data/social.yml @ r1

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/jdavis/twofactorauth/1122)

PR #1128, opened 2015-05-16
Add Mega.co.nz to backup

rugk:
Kim Dotcom [claimed](http://techcrunch.com/2013/05/22/kim-dotcom-claims-he-invented-two-factor-authentication-has-a-patent-to-prove-it/) [he](http://www.techspot.com/news/52677-kim-dotcom-holds-two-factor-authentication-patent-wont-sue-tech-giants-in-exchange-for-legal-help.html) [invented](http://www.theregister.co.uk/2013/05/24/kim_dotcom_2fa_no/) [2FA](http://www.theguardian.com/technology/2013/may/23/kim-dotcom-authentication-patents) but there is still no 2FA in Mega. :expressionless:

About the pull request: I used _MEGAprivacy_ as their Twitter account, but they also have a seperate support account ([_MEGAprivacyHelp_](https://twitter.com/MEGAprivacyHelp)) so if you think this support account is more suitable then we can of course change it.


mxxcon:
I think keep current twitter.
megaprivacyhelp has only 22 followers, which megaprivacy has 91k. And it seems like _help_ account will reply to tweets to the main account anyway.

PR #1129, opened 2015-05-17
Add Digital Ocean Cloud service

dupuy:
Software 2FA supported

They also have docs about adding 2FA to your sites hosted in their cloud, which is great.
But there doesn't seem to be any field for that (I am already "stretching" the exception field
to document their use of SMS as an unlock fallback).


dupuy:
DigitalOcean already present under "Hosting"

PR #1130, opened 2015-05-17
Add Hurricane Electric to Domains section

dupuy:
They also act as an IPv6 tunnel broker and ISP, but no categories for that


mxxcon:
That doc link has a message that it can't be enabled..

dupuy:
Wasn't really documentation - just a forum post, so I have removed it.

mxxcon:
@dupuy can you confirm yourself that HE have 2fa?

dupuy:
Yes, I use it myself (why I added it).

gene1wood:
@dupuy Can you point out where to enable 2FA on he.net? I'm an existing he.net customer but am unable to find where in their UI 2FA is set?

Is the 2FA specific to their tunnelbroker.net site ( mentioned in https://forums.he.net/index.php?topic=3385.msg19501#msg19501 ) and not to Hurricane Electric in general?

PR #1138, opened 2015-05-22
Kualto doesn't support TFA

ciagar:


mxxcon:
hmm..travis fails even after i kicked it again

Carlgo11:
@mxxcon works now.

PR #1139, opened 2015-05-22
Add Messente as 2FA provider

jbaanus:
Add new provider to the 2FA providers list - Messente


Carlgo11:
compressed logo
Here's a compressed version of the logo. Please use that instead.

jbaanus:
Swapped out the logo with compressed version. Thank you for the updated version!

Issue #920, opened 2015-01-08
Exostar Provider


jdavis:
I received an email from an individual that mentioned [Exostar](http://www.exostar.com/) and said that it wasn't on the list.

mxxcon:
according to http://www.exostar.com/Problems-We-Solve/Products/Second_Factor_Credential_Authentication/ they support
Hardware tokens
SMS (text message)
IVR (voice call)

RichJeanes:
Handled in #1146.

PR #1146, opened 2015-05-26
Exostar - as per #920

RichJeanes:
Did some clean-up of unnecessary tags in providers.yml as well.


mxxcon:
Wait, I think `tfa: yes` is a required tag, isn't it?

Carlgo11:
@mxxcon not for providers is it?

mxxcon:
:+1:

PR #1147, opened 2015-05-26
Fixed odd image permissions

RichJeanes:
These images have execute permissions... That isn't intentional, is it?


Carlgo11:
Well you know, how else are we supposed to sudo rm -rf --no-preserve-root / than not from a .png

RichJeanes:
That's actually along the lines I was thinking. I'd hate to see some malicious party slip something into a commit that gets distributed to our community. A little paranoid? Maybe. Too paranoid? I don't think so. Haha

Issue #824, opened 2014-10-18
Request for Treasury Direct


jdavis:
I received the request to add [Treasury Direct](https://www.treasurydirect.gov/) in an email:

> This is a US government website for purshasing and holding Treasury bills, notes, bonds, TIPs, I bonds, EE savings bonds, etc...

Apparently they have OTP via email but that's it.

PR #1149, opened 2015-05-26
Add Treasury Direct as per #824

RichJeanes:


mxxcon:
do you have a link to docs?

RichJeanes:
I do not. I have an account and as far as I remember it's actually a required part of sign-up (though it has been a while since I set up my account, so possibly not).

RichJeanes:
My previous statement seems to be supported by what I can find on the site. Even their login demo includes the OTP step, making no mention of having to set it up.

PR #1155, opened 2015-05-30
Add EuroDNS.com

c960657:
EuroDNS.com offers TFA using TOTP, e.g. Google Authenticator.

I have included the same logo as they use in the favicon (16x16 pixel) on their site. It is not their actual company logo - that is hard to squeeze into a 32x32 image.


_data/domains.yml
```diff
@@ -111,6 +111,14 @@ websites:
       tfa: No
       status: https://twitter.com/enom/status/447074738033008640
 
+    - name: EuroDNS
+      url: https://www.eurodns.com
+      twitter: EuroDNS
```
Carlgo11: The twitter handle isn't needed when the site already supports TFA. Please remove it

c960657: @Carlgo11 Done.

PR #1156, opened 2015-05-30
Add Frankdomains

c960657:
Frankdomains offers TFA using a hardware token.

They charge 19,99 GBP/year for this.


mxxcon:
Wow. It's so pathetic of them to put a price on safety of their users... What's next, they will charge money for SSL encryption? :-1:

Carlgo11:
Add an exception saying that you have to pay extra for the TFA service.

mxxcon:
Carl, do we usually do that? They are pretty up-front with the cost.

Carlgo11:
We do.

c960657:
@Carlgo11 Done.

Issue #819, opened 2014-10-16
Hi, Hoping you would add a category Graphic Arts, Photography


eossipov:
Adobe http://www.adobe.com (no)
Behance http://www.behance.net (no)
TypeKit (Adobe) (no)
Creative Market (no)
Deposit Photos (no)
Big Stock (no)
iStock Photo  http://istockphoto.com  No.

I took a shot of the front page and will provide you with a spread sheet later you can simply upload if that would help?

Carlgo11:
Does the sites fit the <a href="https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria">site criteria</a>?
I can see why Adobe should be added but I don't know about the others.

jamcat22:
Agree with @Carlgo11 so far on this one. We should probably add Adobe and maybe the others, but I don't think we need a new category for them.

mxxcon:
What negative sid-effect would there be (beside account compromise) if somebody gets access to stock imagery site?

snaray:
Well, any stock photo site that charges for access to content can incur false charges if an account is compromised (and if credit cards\* are kept on file), just like any other retailer. Though I don't see why they wouldn't just be listed under retailers.
- even if it's just the token & not the actual card number, a bad actor can charge downloads to it.

Adobe should be added though - customers can buy direct from them, lease cloud services (including online storage), they have an Enterprise licensing portal, and they're a major DRM platform for 3rd-party content providers.

eossipov:
Hi, Folks, Sorry I don't know why I didn't receive any notification on this issue, but At least I'm seeing it now.

Regarding negative side-effects: Image sites can do a large volume especially sites like depositphotos where the originating business is East Asian, and have many people using their api, in addition to corporate clients. You are Required to have a subscription  + CC ON FILE, for that ever useful monthly transaction or a person can opt to purchase tokens, but many business accounts wouldn't do that. 

So, as in any other site that keeps CC on file, who knows if it's encrypted or not? Stored with the client info or not?  We don't, not until you look at a database. I've had people who wanted re-designs or consulting who have emailed me their database with all their clients CC's in there, unencrypted, right in the same row as a persons username and unencrypted password, and address, and zip code, etc. 

I'm sure everyone has horror stories like that, else, 2FA wouldn't be needed. 

sorry for the incorrect wording of "category" I believe I mean "section" (of the page, just a section for "graphic sites"

jdavis:
[...]

PR #1167, opened 2015-06-08
Add Adobe ID

therealmarv:
Creative Cloud offers syncing of images (unlimited) and also files.
I would use their service it if they would offer 2FA.
Images (private or commercial) and creative creations need a good protection.


mxxcon:
Hmm, i'm not sure if Backup is the correct category for it. While Creative Cloud does offer image syncing, I don't think it's the primary product.
Also see #819.

therealmarv:
Well it is a multipurpose product. You can put binary files into it (like Dropbox/GDrive), you can work collaborative on files there (covering most of Adobe products), you can sync your Lightroom image sets to it (which is not a 100% backup but a sync which is two way and in a very good quality). Same login is also used for downloading their products. So it is most similar to a Google Account with various sync products underneath. In my opinion every service which syncs your images in one way or another needs a good protection.

mxxcon:
@therealmarv Yes, every service should have 2fa for user accounts. However, this site is not intended to list every single possible site out there.

When people think about Creative Cloud, image syncing is not the 1st thing that comes to mind. I think it's all of the Adobe applications.

If this site is going to be added, I think it should not go into backup category, but rather into "other" category.

therealmarv:
I'm fine with moving it to another category. No problem with that.

jamcat22:
I say it `looks-good`.

Carlgo11:
Yeah, like #819 I don't think this should be added to the list.

therealmarv:
I see that on #819 everybody is ok with Adobe. I do not care about downloading their products. The main reason for this PR is beeing able to use the Creative Cloud with your professional or private images together with iPad/Android Creative Cloud apps and 2FA protected image syncing.

jamcat22:
Why don't we just change it to Adobe then? They also have document cloud which would be important to protect. All of their products use the same account system, so why don't we just add Adobe itself instead of their individual products?

therealmarv:
@jamcat22 good point. I'm not into every corner of their products but you are right. They all use the same  so called **Adobe ID** 
image
 for Creative Cloud and DC (Document Cloud) and maybe also other things...

I will correct that as soon as I can...

jamcat22:
@therealmarv Are you still going to update this?

Carlgo11:
Just added the "Inactive" label to make it easier to close in case the contributor doesn't reply within a reasonable amount of time.

therealmarv:
@jamcat22 @Carlgo11 updated it now. Thanks for reminding me.

jamcat22:
@therealmarv Thanks! See the [wiki](https://github.com/jdavis/twofactorauth/wiki/FAQ_Merge) for more info on what happens after a request is merged.

therealmarv:
:+1: :)

PR #1191, opened 2015-06-24
Add Webmin

tswestendorp:
Not sure if category is correct, might also be 'Hosting/VPS' or 'Developer'


_data/other.yml
```diff
@@ -147,6 +147,15 @@ websites:
       sms: Yes
       doc: https://secure.ssa.gov/RIL/HpsView.do#question3
 
+    - name: Webmin
+      url: http://www.webmin.com/
+      img: webmin.png
+      tfa: Yes
+      software: Yes
+      sms: Yes
+      phone: Yes
```
mxxcon: When you say `phone` here, it implies there's a phone call with voice or keypad prompt...I don't think that's accurate here.

tswestendorp:
Webmin can be setup with Google Authenticator or Authy. Authy allows 2FA using phone call I think

Carlgo11:
Google Auth or Authy?
For Authy implementation the site must use Authy's native OTP API. Even though the user **could** use Authy with Google Auth it isn't required that the user does that and therefore we don't mark sites using Google Auth with Authy specific features (such as SMS and Phone calls.)

_with reservation for errors. I'm writing using my phone on a moving train_

tswestendorp:
Ok, makes sense, I'll change the addition later on

jamcat22:
Hm. It supports native Authy integration (with the Authy api), but I can't tell if it forces you to use the app.

tswestendorp:
You're not forced to use Authy, Google Auth is also an option. When setting up 2fa Webmin gives you the choice between one of both

jamcat22:
I meant that if you use Authy, I am wondering if they support SMS or if they make you use the Authy app (if you select Authy instead of Google Auth).

Sorry that I wasn't very clear.

I think it's good to merge unless someone says it does support SMS.

tswestendorp:
When you choose to use Authy, you'll be asked to supply an API key, which is to be created in the Authy dashboard by creating an application. After doing that I think you can setup your Authy application to support SMS and or phone.

PR #1196, opened 2015-06-25
Update CAcert status

BenBE:
This includes some information regarding 2FA support by CAcert.

2FA is not currently enforced on accounts but optional to be used. To be used a valid client certificate issued to that account must be available for which the "allow certificate login" flag is set.

### DIFF
--- a/_data/security.yml
+++ b/_data/security.yml
@@ -1,12 +1,12 @@
 websites:
     - name: CAcert
-      url: https://www.cacert.org/
+      url: http://www.cacert.org/
       twitter: CAcert_News
       img: cacert.png
       tfa: Yes
       hardware: Yes
       exceptions:
-          text: "Requires X.509-capable cryptotoken or smardcard."
+          text: "Requires X.509-capable cryptotoken or smartcard."
 
     - name: DigiCert
       url: http://www.digicert.com/
--- a/_data/security.yml
+++ b/_data/security.yml
@@ -3,7 +3,10 @@ websites:
       url: https://www.cacert.org/
       twitter: CAcert_News
       img: cacert.png
-      tfa: No
+      tfa: Yes
+      hardware: Yes
+      exceptions:
+          text: "Requires X.509-capable cryptotoken or smardcard."
 
     - name: DigiCert
       url: http://www.digicert.com/


mxxcon:
Any doc links to confirm this?

BenBE:
Currently there are no really good links to confirm this, apart from the Git repository with the application source in it (cf. https://github.com/CAcertOrg/cacert-devel/). There are plans to extend the 2FA part for user accounts, but that's still future developement (part of which is documented on the CAcert blog regarding the plans for the new software, cf. https://blog.cacert.org/2015/02/new-software-rewriting-the-software-driving-our-site/ and follow-up posts).

For the currently running software there was TOTP support (disabled due to an issue we found), leaving client certificate login, which can be considered 2FA when properly used with a password-protected crypto token/smartcard. Some documentation on systems that allow client cert login can be found at https://wiki.cacert.org/Technology/KnowledgeBase/ClientCerts#Client-cert-enabled_Systems

I hope this helps to assess the pull request.

P.S.: I'm one of the Software Assessors in the development team. Feel free to contact me in case of further questions; contact info @ https://wiki.cacert.org/BenBE

Carlgo11:
@BenBE do you know that the certificate is invalid for https://www.cacert.org/?
Either fix the certificate (The preferred solution) or change the url to http://www.cacert.org/.
Also, if you now use TFA then please remove the Twitter line.

P.S. I must say that I find it very amusing that a Certificate Authority can't provide a valid certificate :smile: 

(Edit: It seems that https://www.cacert.org/ uses a signing algorithm that's insecure)

RichJeanes:
It seems that the issue isn't that their cert is bad or insecure, but rather that their root cert is not included with standard browsers. It looks like their inclusion in the pre-installed Debian CA Root Certs has even been unsure recently.

And as I'm typing this I discovered that their root cert uses MD5... Maybe that's why it isn't included anywhere... So that's what your edit was talking about, @Carlgo11.

Carlgo11:
MD5 o_o
WAAAAT

I made a sarcastic tweet about this a few days ago... I didn't think anyone did that anymore *facepalm*
<br>
https://twitter.com/carlgo11/status/621513074185515008

RichJeanes:
Yeah, go to the https site and check out the root cert their site cert is signed with. They also don't use an intermediary cert, as is best practice.

BenBE:
Hi @carlgo11 and @RichJeanes,

I find it quite amusing to see people without proper background knowledge present their ignorance on such a great stage. Lets start with your points:

@carlgo11: The certificate is not invalid and is properly signed using our root certificate. If you need private lessons on how to do this properly I'm sure you may find someone to help you out. The root certificate has been created quite a while ago where MD5 sure was not the best option available, but also not the worst. Using MD5 with intermediate and subscriber certificates is a bad idea (and nothing we issue: All current certs use SHA2-256 or higher), but does not change the security attributes when used as a trust anchor: Either your trust anchor is correct (correct public key, attributes associated, ... - which you'd ensure in different way like burning them in) or you are screwed regardless of the algorithm used to sign the root anchor. It's kinda funny that most software assumes to gain any trust-related information off of a self-certification. Give me a CRC32-self-signed root and I've no issue accepting it, IFF the information contained in it are ensured to be the correct ones. Also there's quite some certificates using MD2 out there (by RSA Security Inc., Verisign Inc. and others), which are still trusted[1]. I won't necessarily trust those to vouch for anything unless the trustworthiness of the holder and his methods are checked, but this brings us to the next common fallacy.

@RichJeanes: What's much more important is that you trust the provider of a signature, which Debian and some other distributions chose to not do by default (don't get me started on the regarding Debian BTS issue). BUT: Several distributions DO trust CAcert.org by default as can be seen in [2]. And even looking at Debian there is our root certificate available for easy installation [3][4]. It's a quite interesting way to implement trust YOU have in something by hearsay of some third party you barely know or at best assume is saying the truth. Unfortunately even getting the approval to be secure doesn't mean you are[5] - quite to the opposite effect: Some of the well-known trusted CAs repeatedly failed without consequences. I wonder but if a person I trust would fail as catastrophically as did some of those listed I'd have only little more trust to spare for them than I have for fairy-tales.

/rant

Let's get back to constructive work:
- @carlgo11: In your message you mentioned a twitter line? Do you mean the twitter contact in line 4? If so, shouldn't be an issue to be removed
- The Website link: As the certificate is valid and properly issued there's no reason to not use HTTPS. But given your audience might not necessarily have the root installed (and we are working to change this) I'm fine with both; though with a strong preference on HTTPS ;-)
- Use of TFA: Actually TFA had been implemented and deployed years ago (both cert login and OTP), we just chose to remove OTP recently.

[1] https://ssltest.security.fail/trust
[2] https://wiki.cacert.org/InclusionStatus
[3] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=785423
[4] https://packages.debian.org/sid/ca-cacert
[5] https://wiki.cacert.org/Risk/History

RichJeanes:
Let me just say, @BenBE, that insulting people is not a very good way to build your reputation, let alone CACert.org's. Ignorance is not something to insult, belittle, or attack someone for. We obviously have a personal interest in these things if we are willing to spend our free time on projects such as this, so we are more than likely to welcome correct knowledge and information if we are presented with it. If this is the kind of representation CACert.org has in the the community at large, I would rather not have them represented at all on TFA.org.

That said;

I simply stated that the inclusion of the CACert.org Root Cert in Debian (a rather widely used and trusted distro) seemed unsure, something that would appear to be supported by the fact that it is not pre-installed. Is that fact inaccurate?
Yes, it is readily available for install through a debian package, but any CA cert is readily available for use as a trusted root in 3 easy steps.
1. Download cert
2. Place cert in /usr/local/share/ca-certificates
3. Execute update-ca-certificates

Viola. The given certificate is now a trusted CA cert on your system. Being trusted (or not) by the community that develops and maintains something like Debian (widely regarded as one of the most reliable distros) seems to be a factor to consider. In fact, the whole point of certificates is to create a web of trust that ultimately falls on humans validating the claims that certificates make. There is no way for a computer to completely validate a claim to real-world identity.

You say:

> Using MD5 ... does not change the security attributes when used as a trust anchor: Either your trust anchor is correct (correct public key, attributes associated, ... - which you'd ensure in different way like burning them in) or you are screwed regardless of the algorithm used to sign the root anchor

I thought that the point of using a strong signing algorithm is to prevent fraudulent certificates? Doesn't the algorithm used have less to do with trusting the organization claiming ownership of the cert and more to do with it's cryptographic trust? If someone can replicate a cert due to it's weak signing algorithm, then doesn't that defeat it's purpose as a root anchor? Isn't using a weak algorithm on a CA cert worse than on a site cert since it validates more than just one site's trust?

In fact, you seem to degrade your own argument. You cite [5] https://wiki.cacert.org/Risk/History

On this page, I found: https://wiki.cacert.org/Risk/History#MD5-Root

Which states:

> 2008.2. Weak root. An academic group succeeded in attacking a CA with weak cryptographic protections in its certificates ([Sotirov et al](http://www.win.tue.nl/hashclash/rogue-ca/)). This resulted in the attackers acquiring a signed certificate over two keys, one normal and one that acted as a sub-root. This gave them the ability to sign new certificates that would be accepted by major vendors. Consequences: as the root that was attacked was slated to be removed within the month, consequences were limited. Faster rollout of the new root, perhaps a few certificate re-issuances and reputation damage.

On the linked source, they state they are able to create a rogue CA certificate using an MD5 collision. They even go so far as to state:

> Our work shows that known weaknesses in the MD5 hash function can be exploited in realistic attack, due to the fact that even after years of warnings about the lack of security of MD5, some root CAs are still using this broken hash function.

And that was written in 2008! Yet you say that it is okay that CACert.org's root cert uses MD5 because CACert.org is trustworthy. It is not the trustworthiness of CACert.org that is called into question here, but rather the security of it's certificate.

If I am incorrect or inaccurate in anything I have said, please let me know so that we can have civil discussion over it. I am always happy to learn new things.

However. Do not waste our time by continuing to insult the member of this community. If you do, it will be considered harassment and do nothing but degrade your reputation and that of the organization that you represent. I believe I speak for all of our contributors and collaborators when I say this.

P.S. There is a typo in your commit. Smartcard is misspelled.

BenBE:
Hi @RichJeanes,

sorry if the previous comment was kinda grunty.

> Let me just say, @BenBE, that insulting people is not a very good way to build your reputation, let alone CACert.org's. Ignorance is not something to insult, belittle, or attack someone for.

That's also nothing I intended to do. Yet what maybe all involved parties should take to heart is that investigating your statements before writing them might help avoid some pitfalls. Be it classifying a valid certificate incorrectly (I know this IS hard) or rating of certain events. Let's all take it with a grain of salt here ;-)

> We obviously have a personal interest in these things if we are willing to spend our free time on projects such as this, so we are more than likely to welcome correct knowledge and information if we are presented with it. If this is the kind of representation CACert.org has in the the community at large, I would rather not have them represented at all on TFA.org.

As one of the goals of CAcert - apart from providing its community with X.509 certificates and other cryptographic services - is education I think we can agree, that sharing knowledge on cryptography (and other topics) benefits all of us. Thus even though my previous note was written a bit harsh (and you get annoyed when people don't read and especially their programs are written in a way that they actually cant't get it right[1]) most points remain in their essence.

> That said;
> 
> I simply stated that the inclusion of the CACert.org Root Cert in Debian (a rather widely used and trusted distro) seemed unsure, something that would appear to be supported by the fact that it is not pre-installed. Is that fact inaccurate?

Mostly. Even though CAcert (please take care of proper caps :)) Root Certificates are not currently installed in Debian this doesn't make them any less trustworthy. The original inclusion stemmed from a time where there didn't exist any fixed rules on whom to include and whom not. In the mean time many things changed and one of the rules established (CAcert did participate in this process) was the requirement for regular audits.

Now this audit requirement is the point where it gets hard for community projects like CAcert with only a small set of people doing most of the day-to-day work. When you further know that the core team currently works on major overhauls of the current system's implementation and fixing of some mistakes of the past you can hopefully guess there's little time to spare for the extra work the external audit brings (especially when there's an internal todo list with enough work left).

The main complaint on the Debian bug tracker was argued "CAcert is not trustworthy because they don't have an audit", when the only correct way would have been "We choose to disable them by default because some requirements are not met". Interestingly SPI PKI that also fails the audit requirement is still in Debian ;-) Same goes for some CAs that repeatedly failed horribly with proper issuance of certificates only to legit owners.

> Yes, it is readily available for install through a Debian package, but any CA cert is readily available for use as a trusted root in 3 easy steps.
> - Download cert
> - Place cert in /usr/local/share/ca-certificates
> - Execute update-ca-certificates
> 
> Viola. The given certificate is now a trusted CA cert on your system.

Not quite. As indicated in my previous post, it's not only about downloading the root certificate file - even as this make some key trusted with the steps - but rather also having a trusted way to know you've the right one. BTW: This is why we recently chose to publish the fingerprints via DNSSEC in our DNS zone: It opens a second channel to verify integrity. That's the background about why I emphasized there to be a Debian package available to install our root certificate (apart from various other ways to do so).

> Being trusted (or not) by the community that develops and maintains something like Debian (widely regarded as one of the most reliable distros) seems to be a factor to consider. In fact, the whole point of certificates is to create a web of trust that ultimately falls on humans validating the claims that certificates make. There is no way for a computer to completely validate a claim to real-world identity.

Sure. Though please note that X.509 does not span a Web of Trust, but a set of hierarchies, beginning at the various trust anchors. With most other CAs this applies to both the Registration Authority and the Certification Authority parts. The difference with CAcert - and historically also Thawte - lies in the different approach to the Registration Authority. For X.509 there's not much other chance than to cook with water on the certificate issuance part otherwise.

And while reality and its computer representation usually differ quite a lot you can take precautions to keep this human element to a minimum by keeping rules to adhere to as simple as possible.

> You say:
> 
> > Using MD5 ... does not change the security attributes when used as a trust anchor: Either your trust anchor is correct (correct public key, attributes associated, ... - which you'd ensure in different way like burning them in) or you are screwed regardless of the algorithm used to sign the root anchor
> > Let's take on those questions one by one, as it's the details here that count.
> > I thought that the point of using a strong signing algorithm is to prevent fraudulent certificates?

Yes. That is IFF the signature is what makes you transfer trust into the information being signed.

Thus the strong hash algorithm helps with intermediate and subscriber certificates, but not with anything self-signed (like a root) as nobody prevents me to create a different key and attach the information contained in the proper CA's root certificate. It's just like with OpenPGP here: Anybody can claim to be Phillip R. Zimmermann ;-) Yet only if you checked that the guy you hold for PRZ actually uses that key you will gain the certainty this key belongs to that person.

> Doesn't the algorithm used have less to do with trusting the organization claiming ownership of the cert and more to do with it's cryptographic trust?

I'm not sure how to parse this question, but if I parse it correctly as "You use X because it provides N security." then this covers the aspect of building a trust chain that is not tampered with. What this does not cover is "Because X is used the signer is trustworthy with amount T." The security N provided by a strong algorithm allows to infer correctness of a given chain of signatures to be genuine, but does not guarantee that all intermediates in the chain play by the rules.

Yet a strong algorithm DOES protect the legitimate owner from being framed by making it hard for an attacker to forge signatures in a given chain.

> If someone can replicate a cert due to it's weak signing algorithm, then doesn't that defeat it's purpose as a root anchor?

No. It's not the signature on the root certificate of interest but the actual key material. Validation of the key material always needs a second channel for bootstrapping (e.g. like the distribution you trust which hopefully verified that they got the correct key of the CA they pre-install for their users).

> Isn't using a weak algorithm on a CA cert worse than on a site cert since it validates more than just one site's trust?

No. For two reasons:
1. For intermediate certificates that you can issue, you can use a birthday attack for the collision (which IS feasible for MD5 and in near reach for SHA1), but for the root you need to collide the actual hash of the TBSCertificate (The part that the signature's hash is calculated on).
2. With a trust chain containing different algorithms of various trust levels the overall chain can only be as strong as the weakest link. If the weakest link is the signature on a subscriber or intermediate certificate this also presents the weakest link. Yet with the weak signature on the root you already have this part of the chain firmly gripped (because of your trust anchor) and the weak signature is only a pro-forma attachment used to avoid special casing the handling of root certificates. A quite good explanation among some other aspects can be found in https://www.cs.auckland.ac.nz/~pgut001/pubs/x509guide.txt if you want to read more.

> In fact, you seem to degrade your own argument. You cite [5] https://wiki.cacert.org/Risk/History
> 
> On this page, I found: https://wiki.cacert.org/Risk/History#MD5-Root
> 
> Which states:
> 
> > 2008.2. Weak root. An academic group succeeded in attacking a CA with weak cryptographic protections in its certificates (Sotirov et al). This resulted in the attackers acquiring a signed certificate over two keys, one normal and one that acted as a sub-root. This gave them the ability to sign new certificates that would be accepted by major vendors. Consequences: as the root that was attacked was slated to be removed within the month, consequences were limited. Faster rollout of the new root, perhaps a few certificate re-issuances and reputation damage.
> > On the linked source, they state they are able to create a rogue CA certificate using an MD5 collision. They even go so far as to state:
> > 
> > Our work shows that known weaknesses in the MD5 hash function can be exploited in realistic attack, due to the fact that even after years of warnings about the lack of security of MD5, some root CAs are still using this broken hash function.

The cited attack refers to - if I'm not mistaken here - a birthday attack on a subscriber certificate and does not apply to root certificates.

Let's assume you have CA_legit installed in your system and CA_legit works properly. But due to old age its root certificate is signed with H_weak. While you CAN try to collide your intermediate certificate with the root certificate so both have the same signature, this requires that H_weak allows you to determine second preimages for the existing hash. Even IF you found such a pre-image you won't be able to use that certificate since everybody would be using either the already installed root certificate OR reject the intermediate for its weak hash function used. The security here rests on the trust you can put in the validation/verification of the correctness/authenticity of the root certificate/trust anchor and not its actual signature. If that root issues certificates with MD5 this becomes a different story though.

> And that was written in 2008! Yet you say that it is okay that CACert.org's root cert uses MD5 because CACert.org is trustworthy.

No. My point was that I don't need to care about the root certificate's signature as that signature is not what I put my trust in with that certificate: It's the authenticity of the key material which can be validated with additional channels independent of what the certificate says itself.

> It is not the trustworthiness of CACert.org that is called into question here, but rather the security of it's certificate.

You find both schools of though here. As outlined above, I'm on the side of "if I know the trust anchor I don't care about its self-signature" while other people put the self-signature into the validation for both code simplicity and the hope to gain the additional information that the signer knows the private key of the certified public key.

> If I am incorrect or inaccurate in anything I have said, please let me know so that we can have civil discussion over it. I am always happy to learn new things.

Sure. I hope the above elaboration helps a bit getting into the details. Feel free to ask if there's anything I should explain any further.

> However. Do not waste our time by continuing to insult the member of this community. If you do, it will be considered harassment and do nothing but degrade your reputation and that of the organization that you represent. I believe I speak for all of our contributors and collaborators when I say this.

That was not my intention.

> P.S. There is a typo in your commit. Smartcard is misspelled.

k, will change it. Anything else to take care of? What about the Twitter line that was mentioned in one of the previous comments?

Kind regards,
BenBE.

[1] Modern browsers try to hide most of the essential information necessary to assess if any given SSL-related error is legit or whether it was just the inability of the computer program lacking information to do the check for you. Distinguishing unknown issuers and invalid/broken chains is nearly impossible with recent GUIs in browsers as the error message is essentially the same. To some extend this IS intentional, but it is much more likely for users to ignore those messages completely than to vet them properly as they don't have a chance without expert knowledge.

PR #1199, opened 2015-06-26
Add Gilt to retail.yml

hongchangwu:


Carlgo11:
While our goal is to list as many sites as possible, at the same time we can't list all sites since that would make the list enormous and unbearable to walk through.
I don't know whether this is one of those sites where there's just not enough traffic to be listed on this site.

I think we should put it to a vote.
Let's wait for some other collaborators to give their feedback.

mxxcon:
I think Gilt is a pretty legit site. They are rather big in US. :+1:

jamcat22:
^

hongchangwu:
@Carlgo11 any futher thoughts on this? As of 6/29/2015 Gilt is ranked 1604th globally and 654th in US according to http://www.alexa.com/siteinfo/www.gilt.com, which is higher than Comixology and Everlane in the Retail list.

PR #1200, opened 2015-06-26
Add Foursquare to social.yml

hongchangwu:


Carlgo11:
From what I can see Foursquare doesn't fit the social media category. Maybe another one?
Also, I don't think the site fits the [site criteria](https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria).

hongchangwu:
I think this fits criteria 1 since it contains personal image and detailed location information

smholloway:
Foursquare has a lot of location data about its users, which could be bad if leaked. I agree it is worth adding to the site and the PR looks good.

mxxcon:
Seth, but that location data is public, isn't it?

@hongchangwu "personal image" in a sense of reputation, not your avatar...What's the worst that can happen, somebody will steal your account and become a major of a strip club? :wink:

hongchangwu:
@mxxcon location is private data. people can see who have checked in at the same place as you, but they cannot see the detailed history of the places you have been (unless they are your friends). Suppose if your friend's account is breached and then someone figure out your check-in pattern and then break in when you are not home.

mxxcon:
I disagree with this risk assessment, just like I wouldn't consider Yelp qualified for this site, but I won't stop others if they think it's a valid entry.

Carlgo11:
We should list the NSA on our site. That's what I call private data :eyes:

Jokes aside, 2 yeas and 2 nays. Need more voters (@jdavis @jamcat22 ?)

Carlgo11:
Oh and social category?
Let's at least agree on that that's wrong.

jamcat22:
I think the social category is right.

Here's what most people think of as Foursquare: https://www.swarmapp.com/

About 6 months ago, Foursquare split into two different apps. One was called Swarm (basically what Foursquare used to be) and the other was Foursquare.

Foursquare used to be an app that you would check into places with your phone (shared either publicly or privately) so that your friends could see where you were. It was basically the thing before Facebook Places (which actually stole Foursquare's old logo). Then, all of those features were moved to an app called Swarm by Foursquare. Lots and lots of people were mad (about how mad people were when Facebook forced people to download Facebook Messenger), but Foursquare stood their ground. Eventually, Foursquare turned into an app like Yelp, and Swarm was like the old Foursquare.

Both use the same account system though so it makes sense to list Foursquare under social, as when most people think of Foursquare, they are actually thinking of Swarm.

jamcat22:
It makes sense to add Foursquare, as their app (Swarm) tracks your location and you can post almost everywhere you go. Plus, there is a private option, as most people don't want everyone knowing where they have been.

Basically, Foursquare is Facebook, but only for location. (So if we can justify adding Facebook because of their Places feature, we can justify adding Foursquare.)

jamcat22:
https://www.swarmapp.com/ will make adding this as an entry justifiable. You should look at it. (But again, Facebook can do the same thing.)

jamcat22:
Also, this:
image

hongchangwu:
@Carlgo11 I'm fine with changing if there is a more suitable category (FWIW the main Foursquare app is listed in the "Travel" category in the Apple Store as of now while Swarm is listed in the "Social" category)

@jamcat22 I agree with you on Swarm. But problem is that even after one year of launching, Swam simply doesn't have the same recognition as Foursquare. When most people talk about Foursquare, they still more of less think of it as a check-in service. Also I don't think https://www.swarmapp.com/ has a separate login.

jamcat22:
@hongchangwu That's my exactly the point I was trying to make. (Sorry that it got a bit long.)

smholloway:
Any resolution? I'm a weak yes. The risk I see: it would be bad if, for example, a public figure was hacked then checked in at every strip club in their city during work hours.

Carlgo11:
Okay yes it is then.

Carlgo11:
Anyone want to add the looks-good?
I'm on my phone 🙈

PR #1206, opened 2015-07-03
Update eNom

davisonio:


_data/domains.yml
```diff
@@ -112,10 +112,10 @@ websites:
 
     - name: eNom
       url: http://enom.com/
-      twitter: enom
       img: enom.png
-      tfa: No
-      status: https://twitter.com/enom/status/447074738033008640
+      tfa: Yes
+      software: Yes
+      doc: https://www.enom.com/news/978/protect-account-two-step-verification.html
```
Carlgo11: that url redirects to http://www.enom.com/news/article.aspx?p=978& (which does **not** support https).
Change the url maybe?

davisonio: Updated, although when I visit https://www.enom.com/news/978/protect-account-two-step-verification.html it doesn't redirect.

PR #1208, opened 2015-07-06
Add Braintree to payments section

jtdowney:


mxxcon:
If the build passes, +looks-good.

Source code of the main page is pretty funny.

RichJeanes:
That's awesome.
@mxxcon So do you always check the source code of websites?

mxxcon:
I usually check sourcecode to see what kind of favicon a site uses and just get a general feel on site's competency. :)
Here their 32x32 favicon is different from what @jtdowney included. Not a big deal since both are pretty bland and I don't think warrant updating.

PR #1211, opened 2015-07-11
Add Globat to Hosting section

ld3300:


Carlgo11:
The logo you provided is too big.
Here's a compressed version that you can use.
globat

ld3300:
Thank you.  I have submitted the change.

PR #1214, opened 2015-07-12
Add TD Waterhouse to Investing

ESWAT:


Carlgo11:
> For example, you will not be able to access any of our secured sites as cookies are used for security and multi-factor authentication [...]
> - http://www.tdwaterhouse.ca/popup/privategiving/privacypolicy.jsp (Our use of Cookies)

Something you know + something you have. I'm guessing that's two factors then (even though it's not a very good method...)

RichJeanes:
I'm going with @ESWAT on this one. Cookies are not TFA. Cookies should not be used as part of a TFA solution since they can be modified by the client. So someone could bypass the TFA on my account by just crafting a cookie saying they'd already provided the second factor. If they have an actual TFA implementation, I'd want to see proper documentation, not a reference to it in the site's privacy policy.

Carlgo11:
But that's kinda like most other TFA providers solutions. If one sent a maliciously crafted post request to some provider's API, let's say Transakt, they would go ahead and accept that user.
It's something you have isn't it? Just like MFA schemes and usb keys. Just because it isn't a good solution doesn't mean it isn't a solution. MFA schemes can be modified by the client. Apps can be modified by the client. REST-like http requests can be modified by the client.

I don't think they use it to let the user bypass logins but use it to store a 'secret' key to use when logging in.

mxxcon:
So what happens if I don't have those cookies? How would I get those cookies in the 1st place?

Perhaps we should not accept "isn't a good solution" as 2fa. Do it right or don't project false sense of security.

Carlgo11:
As they say in that post, you wouldn't get in to your online banking page.
Okay I'm down for adding cookies to a blacklist.

mxxcon:
I still don't understand. I have a brand new computer, it doesn't have any cookies, how would I login then?

Carlgo11:
Yeah I don't get it either.

ESWAT:
I think their text is misleading. All you need to log into their "secured sites" is a username and password, and I guess the cookie they’re referring to is generated after that since to perform specific actions may require another factor (or I don’t understand how generating a cookie before authentication would help).

ESWAT:
They won’t let you log in if you have cookies disabled (just tried this).

jamcat22:
Most people (including Google) use cookies and local storage to authenticate computers that are "trusted", but it doesn't make sense if TD doesn't have any second factor required in the first place.

PR #1215, opened 2015-07-12
Add Zag Bank to Banking

ESWAT:


Carlgo11:
> Zag Bank uses multi-factor authentication to confirm your identity when you login to Zag Online or Zag Mobile. Specifically, you will be required to set up a series of security questions and answers when you first login. Afterwards, the system will prompt you to answer one of your personalized security questions to login, in addition to your user ID and password.
> - https://www.zagbank.ca/Security/OnlineMobileSecurity

It looks like they just don't understand what multi-factor authentication is.
My guess is that by just sending them a note that they're actually just using 1 factor they'll add some sort of TFA.

Carlgo11:
None the less it looks good.

PR #1216, opened 2015-07-13
Registro.br Domains 2FA

rubenskuhl:
Added TFA information for Registro.br., a domain registrar for .br domains.


mxxcon:
@rubenskuhl I'm not exactly clear on that exception.
Does it support more than .br domains but 2fa works only on .br?
I don't read Portuguese, can you confirm that their 2fa actually makes phone calls or offer physical hardware tokens or smart cards?

rubenskuhl:
On Mon, Jul 13, 2015 at 10:36 PM, mxxcon notifications@github.com wrote:

> @rubenskuhl https://github.com/rubenskuhl I'm not exactly clear on that
> exception.
> Does it support more than .br domains but 2fa works only on .br?

Registro.br only supports .br domains, whether with or without 2FA. You
can't register any other TLD in Registro.br. All accounts can be secured by
2FA, no matter the role of that account in domains (registrant, admin,
tech, billing), or even if doesn't have any role in domains (yet).

> I don't read Portuguese, can you confirm that their 2fa actually makes
> phone calls or offer physical hardware tokens or smart cards?
> 
> Nope, they were both my interpretation of the criteria that after your
> question I now know to be wrong. It's a HOTP+TOTP solution that either
> require the customer to print or carry disposable codes (HOTP) or use an
> Smartphone app like Google Authenticator to use TOTP, or any other
> RFC-compliant TOTP implementation that someone might prefer.

Rubens

mxxcon:
Then perhaps remove `phone:` category and exception?

Carlgo11:
@rubenskuhl may I ask where you learnt how to add a site to twofactorauth.org?
(As in where you learnt what needed to be added in the commits in order for your site to be listed)
I'm trying to map out where people get their information.

rubenskuhl:
A mention of the existence of twofactorauth.org was published on an icann.org blog ; how to add and what to do was based on information available on twofactorauth.org.

Carlgo11:
Right but you didn't look at the wiki or CONTRIBUTING.md to see how to add the `registro.br` to the list?

rubenskuhl:
contribution.md - yes
wiki - nope

PR #1222, opened 2015-07-14
Add Arduino to Social section

ld3300:


Carlgo11:
How is a mini computer a social network?

ld3300:
I primarily use the forum side, but I suppose it does have a store as well.

mxxcon:
If you want to add this site because of its store, I don't think social is the correct category. I think retail or other or even developers would be a better match.

PR #1225, opened 2015-07-15
Making more clear for Coinkite

nvk:


Carlgo11:
Well the documentation link doesn't work anymore. Have any proof of Coinkite supporting sms & hardware?

nvk:
We even support Paper OTP :)

Here https://coinkite.com/settings/2fa

settings coinkite 2015-07-15 17-05-11

Carlgo11:
Okay so I see Software, SMS and paper there. But hardware?

nvk:
You can add smart cards to be Mutlsig Co-Signers http://blog.coinkite.com/post/116127008376

mxxcon:
Carl, paper is "hardware"☺
@nvk is there docs link we can use?

nvk:
The docs are on the 2FA setup internal page https://coinkite.com/settings/2fa but some info here too https://coinkite.com/iforget 

@mxxcon  that's true!

Carlgo11:
Paper is considered hardware? That's odd...
Okay then

nvk:
Even though we have smart cards as co-signing for transactions which is essentially 2 factor authentication because you are using a secondary system to authenticate an action, the idea of hardware 2FA is that is a device (chip, paper cards, etc) that provide offline form of OTP/TOTP

mxxcon:
Carl, would you say it's "software"? :) Paper codes are a legitimate 2nd factor...

Carlgo11:
I would say it's a category we haven't yet added.

RichJeanes:
We've been calling it hardware. At least, that's what we went with when I ran into it with Coinify (#1023), though I did specify using the exception flag.

Carlgo11:
Okay

PR #1229, opened 2015-07-17
Add Sling TV to entertainment section

hongchangwu:
Sling TV is an American live streaming TV provider that is one of the most promising to replace the traditional bundled cable TV model. It offers some popular channels (e.g. CNN, ESPN, TNT, AMC) for $20 a month with the option to add additional packages and HBO. (http://www.gizmag.com/sling-tv-review/37185/)

It was launched in January 2015 and as of June 2015, it has more than 250,000 subscribers (http://recode.net/2015/06/05/sling-tvs-web-tv-subscriber-numbers-keep-growing-now-around-250000/)


Carlgo11:
well... I can't really confirm anything :/
sling

Carlgo11:
Okay back in US it looks fine.
There's just one thing I'm thinking about;
GitHub says the png is an "executable File".
I've never seen that before. Is it just a GitHub bug?

mxxcon:
I think it's permissions on the file set to executable.
fwiu, `git config core.filemode false` will tell your git to ignore permissions. I'm not sure how to tell that to the upstream.

Carlgo11:
well it's not really a problem then. If we want to change it later we can do so I guess.

hongchangwu:
@mxxcon ah yes my bad will fix that in a separate PR

PR #1232, opened 2015-07-19
Reddit does not seem to be working on 2fa any more

captn3m0:
I asked on [reddit itself](https://www.reddit.com/r/ideasfortheadmins/comments/3b8b3e/twofactor_authentication/) recently, and the comments made me realize that they are no longer working on it.

The last update was 5 months back, and was quite negative. The person who was [working on implementing it](https://twitter.com/hueypriest/status/538396874232070144) does not work there any more.

Thus, this proposal.

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -87,8 +87,7 @@ websites:
       twitter: reddit
       img: reddit.png
       tfa: No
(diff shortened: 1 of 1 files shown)


Carlgo11:
Yup doesn't look like they care that much for TFA at all.

PR #1234, opened 2015-07-20
Fix file permission

hongchangwu:
png file should not be executable :/


mxxcon:
@hongchangwu while you are doing that, perhaps look though other png files and make sure they don't have the same?

RichJeanes:
Yeah, I did that a little while ago. There were two or three that I fixed back then. Don't know how it happens.

hongchangwu:
@mxxcon yeah will do a quick scan

hongchangwu:
Seems like this file is the only offender 

```
$ find ./img -executable -name '*.png'
./img/entertainment/sling.png
```

PR #1236, opened 2015-07-21
Add Square 2FA info

jtamplin:


Carlgo11:
Nice. #TFAEveryWhere :wink:

mxxcon:
The doc link talks about logging into a dashboard. Is there some other place where one would login that is not covered by 2fa?

jtamplin:
@mxxcon To take payments, you log into the native app on your phone/tablet, and that does not require 2FA.  The primary goal at this point is to protect any way that could be used to move money out of your account (such as changing the bank account or adding employees), which can only be done via the dashboard.

Technically, the implementation is that logging in via the web always forces 2FA if you have it enabled and the sensitive endpoints that the clients use fail if 2FA isn't provided for the logged-in session (say if an attacker copied a session token from a mobile device etc).  That way, existing mobile clients (some of which are very old and we can't force people to upgrade) continue to work as they always have, yet the sensitive operations are protected.  Eventually, we will support in-app 2FA and make all endpoints 2FA-protected, but it will take time to get there.

jamcat22:
How about also including that in the exception?

jtamplin:
@jamcat22 That seems verbose and overly technical for a general audience (similar to simplifying the 2FA type presented in the table rather than distinguishing between hardware token types or software implementations) - what wording would you suggest?

jamcat22:
Sorry! Didn't mean that exact message. :laughing:

Maybe just add something like "2FA only protects web dashboard."

jtamplin:
Updated the exception text - PTAL

jtamplin:
The build timed out, and I don't see a way to make it retry - what do I need to do?

jamcat22:
@jtamplin It's ok. Travis has been timing out recently. We're looking for a fix.

jtamplin:
@jamcat22 So can we merge this with a failed CI build, or what do we need to do to land this?

Carlgo11:
Nothing

jtamplin:
Thanks.  Is there a regular deployment schedule for this to go live?

Carlgo11:
See https://github.com/jdavis/twofactorauth/wiki/FAQ_Merge

PR #1237, opened 2015-07-21
Switch TravisCI to use containers

mxxcon:
According to http://docs.travis-ci.com/user/migrating-from-legacy/ it looks like this is all that we need to do to use Travis' new container based infrastructure.
I have not actually tested this(don't know how to).

[Review on Reviewable](https://reviewable.io/reviews/jdavis/twofactorauth/1237)


Carlgo11:
well let's see how this goes then.
I've been thinking of doing the same but clearly you finished first :wink:

Carlgo11:
Okay so it exits with 0 but the bundle cache isn't working atm.
I Googled the problem and it looks like every single build running on the new containers are experiencing the same problem so I vote to just merge this and fix the error later.
After all, it doesn't do anything (good or bad) at the moment so why not just leave it there?
Maybe we don't have to do anything when they fix it.

jamcat22:
Ugh. Why did you have to beat me to it @mxxcon? That's why I was wondering if we were using sudo. :P

Carlgo11:
His new name is @ninjacon

jamcat22:
Ha. But I merged it so :P

jamcat22:
Wait, don't we overwrite the install step?

PR #1242, opened 2015-07-23
Added Ping Identity to Identity Management

allynf:


RichJeanes:
It seems that Ping is an SSO and TFA provider, correct? This site is for listing consumer websites, not providers. Unless you think that Ping is widely used enough to merit its listing for itself as a website that people will be logging in to.

Carlgo11:
Yup. Provider.

allynf:
Yes, Ping does SSO and TFA and competes with Okta and OneLogin, among others, which are also listed in the Identity Management section.

RichJeanes:
Then perhaps we should look into removing them if those are the only services they provide. Thank you for pointing that out.

PR #1245, opened 2015-07-28
Change Mint Bills icon

hongchangwu:
The Mint Bills logo is actually different from Mint.com :/


hongchangwu:
Seems like the link check timeouted?

Carlgo11:
Yeah don't worry about it. It happens to everyone

mxxcon:
i originally didn't want to use 2 different logos since they are so similar, it's kind of irrelevant. ;)

PR #1250, opened 2015-08-01
Add exception for Facebook

### LINKED ISSUE #1181: Facebook 2FA is also one account per phone number
I just witnessed the following on Facebook:
- user has two accounts on Facebook: one for a business and another for personal use (I'll call these `account1` and `account2`.)
- user enables 2FA for `account1` using phone number and Google Authenticator
- user enables 2FA for `account2` using same phone number and Google Authenticator
- Facebook sends the following email to the address associated with `account1`:

> "You recently removed your primary phone from your account.  Because of this, we turned off login approvals on your account to ensure that you don't get locked out when using an unrecognized computer or mobile device to login."
- user is now able to login to `account1` on a new device with just their password and no second factor
- checking the 2FA settings on `account1` reveals 2FA is indeed turned off for that account
### The Facebook Entry Should have a Hazard Symbol like Twitter

image

_À la_ the twitter entry, the Facebook entry should have a hazard symbol that states something like the following:

> "SMS required for 2FA, 1 account per phone"

jamcat22:
Adds an exception stating how Facebook requires SMS and only supports one phone per account.
Also corrected punctuation for Twitter exception.
Closes #1181.

[Review on Reviewable](https://reviewable.io/reviews/jdavis/twofactorauth/1250)

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -43,6 +43,8 @@ websites:
       tfa: Yes
       sms: Yes
       software: Yes
+      exceptions:
+          text: "SMS required for 2FA. 1 account per phone number, 1 phone number per account."
       doc: https://www.facebook.com/help/148233965247823
 
     - name: Foursquare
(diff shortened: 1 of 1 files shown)


mxxcon:
+looks-good 

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/jdavis/twofactorauth/1250)

PR #1060, opened 2015-04-05
Add StartJOIN

juanda-097:
StartJOIN is a crowdfunding platform that accepts as payment crypto currencies.

I have a problem to upload startjoin logo. If someone help me upload it, I would be very grateful https://pbs.twimg.com/profile_images/378800000584095301/fd897157c1d35f17e9aba63ba2dbb7d2_400x400.png

PR #1254, opened 2015-08-01
Added StartJOIN

jamcat22:
Added StartJOIN to Other.
I thought that category suited it better than the original since it is a
crowdfunding platform, not a Bitcoin wallet.
See #1060.

[Review on Reviewable](https://reviewable.io/reviews/jdavis/twofactorauth/1254)


Carlgo11:
So it's kickstarter with crypto currencies.
Isn't that a bit like finance then?

mxxcon:
This is a bit of a grey area between finance and something else

jamcat22:
Well Kickstarter is in Other, so I thought it would fit there.
It isn't really a Bitcoin wallet because you have to convert whatever it is to StartCOIN. It's also not really designed to hold money.

Carlgo11:
Oh well is kickstarter is in `other` then I guess `other` is the right category.

PR #1262, opened 2015-08-04
Updated Information for Hetzner online

doppelgrau:
http://forum.hetzner.de/wbb2/thread.php?threadid=22478 states yubikey
and RFC 6238 support. RFC 6238 tested


mxxcon:
Any `doc` urls to include?

doppelgrau:
At the moment only the customer forum which requires a login.
I guess in the near future it will appear in the wiki

PR #1268, opened 2015-08-07
Origin now supports software based 2FA through Google Authenticator

virtualmarc:
Origin added Support for Google Authenticator together with renaming the Origin Accounts to EA Accounts.


Carlgo11:
Can confirm that Origin indeed does use Google Auth now.

PR #1276, opened 2015-08-12
Dropbox has added U2F support

brycied00d:
Announced today, Dropbox has added support for U2F security keys. https://blogs.dropbox.com/dropbox/2015/08/u2f-security-keys/


RichJeanes:
Can confirm. They've even updated their help doc already. :+1:

PR #1281, opened 2015-08-14
Added ZenBanx

ESWAT:
Yeah… another new Canadian bank that starts with a Z, has the word Bank in it – well, almost – and doesn’t have 2FA.


mxxcon:
How big/popular is it?

ESWAT:
It has a good following already because:
1. It’s started by the founder of ING Direct
2. They’ve done some promos to grow their customer base in the last few months

PR #1287, opened 2015-08-19
Add pairNIC domain registrar

fbonzon:
About the doc link, I'm unable to find a URL that links directly to the "What is Two Factor Authentication?" question and answer. Is better to leave the doc URL blank instead?


Carlgo11:
wow they made the FAQ page's topics 100% JS without an option for non-JS users.
It's okay. Just leave the URL there. Let's not do any extra work because of their stupidity...

PR #1293, opened 2015-08-27
Added SMS support to Usabilla entry

josja:
Initially I forgot to add the SMS support to the Usabilla entry :/

### DIFF
--- a/_data/other.yml
+++ b/_data/other.yml
@@ -217,5 +217,6 @@ websites:
       img: usabilla.png
       tfa: Yes
       software: Yes
+      sms: Yes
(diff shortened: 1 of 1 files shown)


jamcat22:
Do they support SMS only? Or do they require Google Authenticator?

josja:
We use SMS as fall back mechanism in case you loose your Google
Authenticator.

Hope this answers your question?

Regards, Joshua

Carlgo11:
sounds like sms: No then

josja:
Ah in that case I'll retract my pull request of course

Verstuurd vanaf mijn iPad

Op 28 aug. 2015 om 07:54 heeft Carlgo11 notifications@github.com het
volgende geschreven:

sounds like sms: No then

PR #1297, opened 2015-08-30
Update Charles Schwab

mlissner:


Carlgo11:
proof of this?

mlissner:
Have an account, made long password, tested it with long password minus
last letter?

mxxcon:
The existing doc url is 404. @mlissner can you find a more current version and include with this pull request?

mlissner:
The only external link I can find is this one: https://www.schwab.com/public/schwab/banking_lending/bank_online_security.html

But it says that:

> Your password must meet the following criteria:
> - 6-8 characters long
> - Include both letters and numbers
> - Include at least one number between the first and last character

Meanwhile, if you get their forgot password page (which can only be reached after they send you an email), you get:

> Create a new password that follows the Schwab Password rules.
> -  Password must contain at least one letter
> -  Password must contain at least one number
> -  Password must be 6 characters or more in length
> -  Both new passwords entered must match
> -  Your new Password:
> -  Can contain commonly used symbols (!@#$%^&, etc.)
> -  Is case-sensitive (exAMp1e)
> -  Cannot match or contain your Login ID, SSN or Tax ID
> -  Can have internal spaces. Cannot have leading or trailing spaces.

So...not sure what link to put here. There's no good option; not sure what convention is for this project.

fpigerre:
Thanks for the pull request @mlissner! :+1:

PR #1306, opened 2015-09-04
Added Todoist

aidanknowles:
Added productivity web app Todoist to 'Other' list.

Todoist does not currently support 2FA natively.


Carlgo11:
I'm thinking about a productivity category. I'll put this PR as pending until we decide whether to add that or not. If we do then I think this site definitely fits there.

aidanknowles:
Productivity category sounds like a good idea.

mxxcon:
We can add this site now and move to another category if/when that happens, no point hanging this PR..

PR #1308, opened 2015-09-07
Turbotax has 2FA + SMS. 

mikeatlas:
doc link here: https://ttlc.intuit.com/questions/2903072

Two-Step Verification is a growing financial industry standard in which
a customer, after entering their User ID and Password, is required to
provide an additional point of verification. This is often done via a
SMS or text message to your phone or in some cases an authentication
app that provides a series of numbers that changes at regular intervals.

We continue to enhance the security of our systems and provide
additional options for our users to have control over their own
security. We introduced Two-Step verification, sometimes called
Multi-Factor Authentication or MFA, to our full customer base earlier
this year which would occur on your first sign in.

How do I use the Google Authenticator App?

First you must have Two-Step Authentication Active; follow the steps
below to activate two-step verification for your phone. Once this is
active you will have the option to use the Google Authenticator App.

Sign into your TurboTax Account.
Click MyAccount, in the upper right corner.
Click the Account Settings option.
Scroll down to Two-step Verification.
Click Turn On.
Enter your Password.
Click Turn on two-step verification.
Where do I get the Google Authenticator App?

The App is available through the AppStore on iPhones or PlayStore on
Android based phones. This feature is not available on Windows Phones.

Using and Deactivating the Google App on an iPhone

Using and Deactivating the Google App on an Android

I lost my device - What do I do?

You have a few options available to you if you lose the device you had
the Google App on. If you indicate you have no access to the App you
will default back to the SMS option. If that also fails you could try
the following:

Choose the Confirm my account a different way (takes longer) option and
follow the on screen prompts
Check your cell carrier's website. Many carriers allow you to check
your messages online.
Contact us and one of our support representatives will be able to
assist you. You will be asked to verify your identity as part of our
security to ensure your account remains safe.

### DIFF
--- a/_data/finance.yml
+++ b/_data/finance.yml
@@ -40,8 +40,9 @@ websites:
       twitter: turbotax
       img: turbotax.png
       tfa: Yes
+      phone: No
       sms: Yes
-        sms: Back-up 
+        text: Back-up 
       software: yes
       doc: https://ttlc.intuit.com/questions/2903072
 
--- a/_data/finance.yml
+++ b/_data/finance.yml
@@ -39,7 +39,11 @@ websites:
       url: https://turbotax.intuit.com/
       twitter: turbotax
       img: turbotax.png
-      tfa: No
+      tfa: Yes
+      sms: Yes
+        sms: Back-up 
+      software: yes
+      doc: https://ttlc.intuit.com/questions/2903072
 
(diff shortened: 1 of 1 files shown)


mikeatlas:
messy. doing this over 1 more time.

PR #1319, opened 2015-09-11
Added Domain provider NearlyFreeSpeech.net

steeef:
Added this domain provider (though they also do limited hosting). I grabbed their favicon.ico and had to blow it up a bit (original was 16x16), so it's not that pretty.


fpigerre:
Thanks for the pull request @steeef! :+1:

Carlgo11:
Maybe this one is better? It's shrunken down from 215x215px 
nearlyfreespeechnet_215 1

Carlgo11:
And while you're at it you can remove the twitter handle as that's not needed for sites already supporting tfa.

PR #1309, opened 2015-09-07
TurboTax has 2FA.

mikeatlas:

Issue #1141, opened 2015-05-22
Intuit products do support MFA


stacis254:
Intuit Turbo Tax - https://ttlc.intuit.com/questions/2805648-how-can-i-use-2-step-security-for-login
https://turbotax.intuit.com/best-tax-software/secure-online-tax-prep/

Mint Bills- http://support.quickbooks.intuit.com/support/articles/how18938

QuickBooks Online- https://community.intuit.com/questions/1136616-can-i-add-extra-security-to-my-qb-login-like-mfa-or-can-you-point-me-to-any-documentation-you-have-that-explains-your-login-security

http://payroll.intuit.com/support/kb/1003046.html

http://payroll.intuit.com/support/kb/2001797.html

mxxcon:
Regarding QuickBooks Online, "Security Questions" are not an acceptable 2nd factor. It is still "something you know" just like username/password.

Other than that, all of them use email as 2nd factor and only when logging in from devices w/o cookies or making account changes, not every single time during login.

p4gs:
TurboTax supports SMS and email for 2FA. However, it looks like they use both cookies and IP address history (maybe?) to allow username/password 1FA. I just used 2FA to authenticate from a private browsing window in Chrome for iOS. Then I authenticated from a private browsing window in Safari and it didn't prompt me for 2FA. Very odd.

An "Exception and Restriction" should be added to TurboTax stating this caveat. It does support 2FA via SMS, though, and its record on the website should be updated accordingly.

![2015-06-07 12 54 44](https://cloud.githubusercontent.com/assets/10093271/8025049/7ffe1148-0d16-11e5-907c-e91aa98fcc5f.png)
![2015-06-07 12 54 14](https://cloud.githubusercontent.com/assets/10093271/8025050/7fff004e-0d16-11e5-8830-3e7784258e38.png)
![2015-06-07 12 54 27](https://cloud.githubusercontent.com/assets/10093271/8025048/7ffd13e2-0d16-11e5-8fd3-6cd2ed4aa190.png)

mikeatlas:
![screen shot 2015-09-07 at 5 46 54 pm](https://cloud.githubusercontent.com/assets/12704/9723142/e5144b3c-5588-11e5-962f-4808d4b6db46.png)
bump

mxxcon:
@mikeatlas would you like to submit a pull request for this?

mikeatlas:
[...]

PR #1325, opened 2015-09-14
Patch turbotax

Carlgo11:
As per #1141, #1308 and #1309


Carlgo11:
> it was discussed before and I think it was decided that it must be every time, otherwise it doesn't count
> or at the very least an exception 
> - @mxxcon 

So should I add a PR setting TurboTax to `tfa: No`?

jamcat22:
Why don't we ask @jdavis first?
Just asked in Slack.

PR #1329, opened 2015-09-16
Add Fasthosts to domains

edmorley:
They only have username+password and a security answer for "contacting support".
No MFA options provided.

Also removes some trailing whitespace.


edmorley:
Sorry for the churn - have just updated the PR to add the new entry to "domains" instead of "hosting". Fasthosts offers both - and looking at the other sites listed that offer both, the precedence seems to be to add them to domains.

mxxcon:
I doesn't seem like they give any more weight to domains vs hosting, but considering their name is "fasthosts", i imagine they were a hosting company first.. then again, it doesn't really matter. :smile:

edmorley:
I wonder if in the future the two categories should be combined? :-)

mxxcon:
Or just be a bit more selective on what gets added..At the end of the day this site is not really intended every single possible site out there..

Carlgo11:
The logo is a bit weird. I think it's placed out of the image borders. 
Here's a better picture which also is compressed. 
d7h9susi copy 1

edmorley:
PR updated with new logo.

PR #1331, opened 2015-09-16
Add InMotion Hosting to hosting

edmorley:
Also remove trailing whitespace.


mxxcon:
How big is this host? How many users does it have? I'm a little bit hesitant adding every hosting company because there are so many of them out there..

edmorley:
I don't have any specific numbers. I thought they were fairly well known, but then you've not heard of them, and I'm a sample size of one. Some of the figures on http://www.inmotionhosting.com/meet-us and the fact they appear in the list of companies being compared in this PC mag review http://www.pcmag.com/article2/0,2817,2424725,00.asp , make me think they can't be that small.
Totally understand you don't want to add everything - up to you :-)

PR #1334, opened 2015-09-16
Add PyPI to developer

edmorley:
PyPI is the main Python package repository. Access to it allows people to upload/edit packages used by many other sites and tools, which has serious security implications, yet they do not offer 2FA.


fpigerre:
Thanks for the pull request @edmorley! :+1:

PR #1335, opened 2015-09-16
Add SourceForge to developer

edmorley:
SourceForge hosts the source code repositories and issue trackers for many open source projects, but does not use 2FA. This is particularly risky given that many projects use SourceForge to host the compiled binaries (and link to them from their websites) - and so a compromise could result in malware being unknowingly installed by end users.


edmorley:
Thanks :-)

PR #1342, opened 2015-09-29
Update TurboTax

veratu:
Fixed TurboTax URL / auth types.

### DIFF
--- a/_data/finance.yml
+++ b/_data/finance.yml
@@ -42,7 +42,7 @@ websites:
       sms: Yes
       email: Yes
       software: Yes
-      doc: https://turbotax.intuit.com/best-tax-software/secure-online-tax-prep/
+      doc: https://ttlc.intuit.com/questions/2902682-what-is-two-step-verification
 
     - name: Kiva
       url: https://www.kiva.org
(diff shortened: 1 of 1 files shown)


mxxcon:
Don't you think the original URL is more useful to users how to enable 2fa vs marketing fluff on the new URL?

veratu:
The original link was specific to Google Authenticator, which is only 1 of several 2fa's the product supports, and I'm working to influence the marketing fluff people to have links underneath each topic to individual URLs with more info/instructions on all of the 2fa capabilities instead of just the one.

jamcat22:
How about [this](https://ttlc.intuit.com/questions/2902682-what-is-two-step-verification)? It links to both Google Auth, and SMS.

Do you have anything to show email auth?

veratu:
You mean a screenshot of the email auth?  Or a link on their site?  You can see their email auth experience by just going to turbotax.com (do it in an incognito window if you're an existing user, that will force their 2fa process), and when you sign in you'll see a screen that offers SMS or Email 2fa.   Once you are inside, you can set additional items on your account to be permanent 2fa via SMS or Google Authenticator.

veratu:
I updated the URL to the what is two factor auth link.  Let's go with that one.

veratu:
Are we good to merge this now?

PR #1343, opened 2015-10-01
Remove travis_wait from .travis.yml

Carlgo11:
Removed the `travis_wait` arg since that isn't really needed anyways.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: ruby
+install: bundle install
 cache: bundler
-install: travis_wait bundle install
 rvm:
     - 2.1.1
 notifications:


mxxcon:
Is there some drawback in having `travis_wait` there?

Carlgo11:
The travis_wait arg was a thing we tried when Travis failed all the time. It never worked since it apparently only applies to the bundle command.

RichJeanes:
But it was timing out on the bundle command. I thought it did help? We're just timing out on a different part now.

Carlgo11:
No the bundle command never timed out. It was the html-proofer that timed out.

RichJeanes:
You sure? I seem to remember it dying sporadically...

PR #1346, opened 2015-10-02
GitHub added U2F support.

### LINKED ISSUE #1345: GitHub U2F
Just saw an article that GitHub is adding/has added U2F support. But I'm on my phone, so I can't PR.

RichJeanes:
Resolves #1345


RichJeanes:
They even have their docs up to date, so the doc link is still good. Don't you just love it when a site has their crap together?

PR #1347, opened 2015-10-02
note that MailChimp supports SMS TFA

stembrain:


RichJeanes:
@stembrain:
So do they require a text code every time you log in? Because their documentation on it [HERE](http://kb.mailchimp.com/accounts/management/best-practices-for-account-security) says they only require it "If we see patterns that seem unusual for your account".

Also, as a heads up, we recommend making changes on a new branch, not master, to avoid merge conflicts when your fork gets out of sync with upstream. It just makes it easier for everyone.

@ the team:
It seems like our current doc link needs to be updated as well, since the blog post says you have to set up AlterEgo, but their link to it just redirects to the [Google Authenticator Setup page](http://kb.mailchimp.com/integrations/other-integrations/set-up-google-authenticator). It looks like they've just discontinued their [AlterEgo product](https://alteregoapp.com/).
Which means `hardware` needs to be a `No`.
I'll take care of these updates myself.

jamcat22:
@RichJeanes You are correct. They only request SMS sometimes, and AlterEgo was discontinued a few months ago.

PR #1348, opened 2015-10-03
Add a column for account recovery documentation

runasand:
Two-factor authentication is great until you lose the device. I wanted to help users get through the recovery process faster by specifically linking to the recovery documentation for the different services.

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -5,6 +5,7 @@ websites:
     software: Yes
     img: aerofs.png
     doc: https://blog.aerofs.com/two-factor-authentication-for-hybrid-and-private-cloud/
+    recovery: https://support.aerofs.com/hc/en-us/articles/202645974-How-Can-I-Recover-Access-To-My-Two-Factor-Authentication-Enabled-Account-
 
   - name: Apple iCloud
     url: https://www.icloud.com
@@ -15,6 +16,7 @@ websites:
     exceptions:
         text: "See http://support.apple.com/kb/HT5593 for a list of supported SMS carriers."
     doc: http://support.apple.com/kb/ht5570
+    recovery: https://support.apple.com/en-us/HT202649
 
   - name: Backblaze
     url: http://www.backblaze.com/
@@ -63,6 +65,7 @@ websites:
     software: Yes
     hardware: Yes
     doc: https://www.dropbox.com/help/363/en
+    recovery: https://www.dropbox.com/en/help/364
 
   - name: Evernote
     url: https://evernote.com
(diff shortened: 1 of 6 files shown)


mxxcon:
I can't tell from the code what it'll look like. Is it another checkmark?
Also do you think users would come to this site to look for a way how to recover their account?
It's a useful information, but I'm not sure users will refer to this site when they lose their phone..I just don't have a sense this is the type of user-workflow we have(not that I know what kind of flow we do have :smile:)

runasand:
A new column is added on the right and the icon used is the same as for "Docs."

I think the same users who use the site now would use the site to find account recovery information. It seemed silly to create a new site just for that purpose, when twofactorauth.org is already a well-known page. I know I, personally, would use it if I ever lost my phone and had to go through the process for all the services I use.

gcochard:
This is really useful. Would have saved me hours when getting a replacement phone. If someone comes to the site to learn how to add 2fa, they will know where to go when (not if) they need to recover. :+1: :clap:

thorsheim:
I doubt normal users will visit this site to find links to relevant account recovery docs, but for anyone developing or comparing services, this would be a great addition to the site. Also for all of us who nags service providers to improve their security, this addition will also be useful.

Carlgo11:
We (the collaborators) will discuss whether or not to add this, but right now we're focusing on updating the site.

Carlgo11:
Hi again Runa,
thank you for your patience and dedication towards making the site better.

We've now discussed this and come to the conclusion that this wouldn't look good for mobile users.
The current layout is a mess for mobile users and this would unfortunately make it look worse for them.

While we all think the idea behind your PR is a great idea we'll have to pass at this time.
I apologize for not being able to merge this after all the work you've done.

<i>//Carl</i> :snowman: 
<br>
<i>P.S. I like your name. It's very old school Nordic</i> :wink:

jamcat22:
@carlgo11 I think we should merge it, because we would need the extra column to determine how to make responsive tables work.

Carlgo11:
What, you said you agreed with Rich yesterday o.o

jamcat22:
I said "I agree with Rich. I love the idea, but we'll need to fix responsive tables if we add it."
I meant we should add it, and then fix responsive tables.

RichJeanes:
Also, to be clear and not give the wrong impression to the public community, I said I really like the idea, but think the site is too cumbersome on mobile to be adding more to it. If we improve the mobile performance of the site (and other usability improvements along the way would not be a bad thing), I would love to see this feature.

jamcat22:
Hello @runasand!

We're in the process of redesigning the overall look of the site (currently the `alternate-layout` branch), so I'm sorry that there hasn't been any activity on this PR in awhile.
With the new design, we should be able to include account recovery documentation without it being cumbersome on mobile device layouts.
There's still a bit of debate on whether or not we should add this, but if we choose to, it could take anywhere from a couple weeks to a month.

Thanks!

gingerbeardman (approved):

PR #1359, opened 2015-10-07
Reword column header

RichJeanes:
I don't think it's really necessary to use "Implementation" instead of "Token" (b/c it is still a token) and it makes the column unnecessarily wide and ugly.

Opinions?


mxxcon:
I guess so... :santa:

PR #1362, opened 2015-10-10
Remove non-available options from Paypal

alerque:
As best I can find, PayPal only supports SMS based MFA. The wording all over their site makes it sound like they are geared up for other options such as an OTP app or hardware keys, but nowhere can you actually setup anything but SMS based authentication.


mxxcon:
I'm sorry, but that is incorrect. Paypal uses VeriSign Identity Protection, which comes in a form of hardware tokens or desktop/mobile apps.
paypal

PR #1364, opened 2015-10-11
Add Paysafe

Carlgo11:


RichJeanes:
Wait, where would the TFA come in here? If the whole point is that it spends online like cash, then what is there to protect? They even say that the whole point is for no personal info to be involved.

jamcat22:
True. I'm not sure how they would even authenticate with 2FA since they don't have a username, password or any other information about you.

PR #1370, opened 2015-10-14
Added some major UK services

groenroos:
- [EE](http://www.ee.co.uk/)
- [Hailo](https://www.hailoapp.com/)
- [Three](http://www.three.co.uk/)
- [Transport for London](https://tfl.gov.uk/)

All of these deal with money (saved payment methods), and most include personal information, including telephone numbers and home addresses.


Carlgo11:
Great work @groenroos!
I actually have nothing to complain on (which is very very unusual for PRs).

Carlgo11:
Wait, actually I do.
You added Three. Three is operating in several countries and their twitter is only for UK customers.
Maybe you could change the `name:` to Three (UK) or something similar.

groenroos:
Right you are! Went ahead and fixed that to "Three UK" which is the way they style it on their official communications.

PR #1371, opened 2015-10-14
Add note of Blizzard's "hardware token" support

smiley:
Blizzard doesn't support the new U2F standard, or any other hardware device other than their own physical authenticator. With the new standard, it's bound to confuse some people (like me), so perhaps a small note can be added to clarify this.

### DIFF
--- a/_data/gaming.yml
+++ b/_data/gaming.yml
@@ -16,7 +16,7 @@ websites:
       hardware: Yes
       doc: http://us.battle.net/en/security/
       exceptions:
-          text: "Hardware token refers only to Blizzard's own physical \"Battle.net Authenticator\" device. They do not support U2F or any other physical tokens."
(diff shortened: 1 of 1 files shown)


RichJeanes:
I don't think this is necessary. We have had people suggest separating standard/proprietary hardware/software solutions multiple times and we think it would just add unneeded clutter and noise to the site. If you are really interested in the specific implementation of something, that is what the doc links are for.

smiley:
That's why I added it as a tooltip note, rather than as a different column:

image

RichJeanes:
I didn't say anything about columns, I just said that this kind of differentiation adds clutter. We are just trying to give people a general idea of the security of a site. The doc link is for details if people are interested using a site.

jamcat22:
I agree with @RichJeanes on this one. Adding this to Blizzard would justify adding it to every site with its own authenticator. At some point it might begin to look like the site before Awesome 1.0 and our first major redesign.

smiley:
Okay, I see your point now. I guess it's not that big of a deal.

PR #1383, opened 2015-10-23
Move the Zocle listing to preserve alphabetical ordering

michel-slm:
Zocle was listed at the top of social.yml, it should be at the bottom.


_data/social.yml
```diff
@@ -34,7 +27,7 @@ websites:
       img: bufferapp.png
       tfa: Yes
       sms: Yes
-      software: Yes
+     software: Yes
```
mxxcon: Indentations are important. Please preserve them otherwise it's not a valid yaml file.

michel-slm: yup, sorry. just fixed it.

PR #1386, opened 2015-10-24
Add Taiga.io

michel-slm:


fpigerre:
Thanks for the pull request @michel-slm :+1:

PR #1387, opened 2015-10-26
Updated CIBC

strattondev:
Updated CIBC in Banking

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -103,11 +103,7 @@ websites:
       url: http://www.cibc.ca
       twitter: cibc
       img: cibc.png
-      tfa: Yes
-      sms: Yes
-      email: Yes
-      phone: Yes 
-      doc: https://www.cibc.com/ca/features/banking-enhancements.html?sf3204759=1
+      tfa: No 
 
     - name: Citibank
       url: https://online.citibank.com/US/Welcome.c
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -103,8 +103,11 @@ websites:
       url: http://www.cibc.ca
       twitter: cibc
       img: cibc.png
-      tfa: No
-      status: https://www.cibc.com/ca/features/banking-enhancements.html?sf3204759=1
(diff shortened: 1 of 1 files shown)


mxxcon:
Your PR looks fine, however on their site it states:

> Before you complete a sensitive transaction that requires greater protection, you will be prompted to enter a one-time verification code. 

Does it mean it's not prompted every time, and would also does not qualify as `tfa:yes`?

strattondev:
No, you're not prompted every time. I figured it would be valid as there are other banks with the same set-up of not entering 2fa all requests; Bank of America, UW Credit Union and Discover. That is what I gathered from my reading of other banks before updating CIBC to `tfa: Yes`.

jamcat22:
The biggest difference is what 2FA is required to do. For example, Bank of America and Discover require 2FA before you can sign in, while it looks like CIBC only prompts for 2FA when performing certain transactions.

Right now our stance on this is that we only list websites that require 2FA for login, not just for certain transactions.

Carlgo11:
> a site that prompts you for an authentication token following user login would be considered Two Factor Auth. A site that does not prompt you for a token upon login, but prompts you for a token when you try to perform a sensitive action would not be considered Two Factor Authentication

See [CONTRIBUTING.md](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions)

strattondev:
@Carlgo11 that makes sense. 

I'm guessing if that is the case then CIBC should be updated to just `tfa: No` without the `status: status: https://www.cibc.com/ca/features/banking-enhancements.html?sf3204759=1` as their status page, or any supporting documentation that I can currently find, mentions 2fa for login? If so I'll update the commit, otherwise I can close the request.

Carlgo11:
@strattonw that's right. `tfa: No` without a `status:` line.

strattondev:
Updated as such.

Carlgo11:
Great!

PR #1389, opened 2015-10-29
Add CM Telecom

mikerynart:


RichJeanes:
The doc link doesn't seem to link to any documentation on their site login and authentication, rather showing documentation on their services. It also seems to only be referring to one of CM's services.

mikerynart:
Hi RichJeanes, 

Thanks for your reply. We do provide two-factor-authentication and I will double check the documentation for this service.

Best,

Mitch

mikerynart:
Hi Rich, we've updated the doc link, this information should be sufficient for all.

RichJeanes:
So I understand that you can use CMT's SMS service to send OTPs for authentication, but I don't see anything talking about TFA on CM Telecom accounts. This site is meant to list sites that use TFA to protect their end users, not sites that provide TFA services for developers. So I'm looking for documentation on TFA for CM Telecom accounts, though despite my best efforts, I am unable to determine whether CMT even has accounts to log in to. All I see are "Contact Us" buttons everywhere I look.

mikerynart:
Hi Rich, Thanks for your feedback. We use the following domain https://dashboard.onlinesmsgateway.com/en/login for accounts to log in to via TFA to protect our users. A login button is currently active on our Dutch domain https://www.cm.nl/ and the button will be enabled for all other domains next week.

mikerynart:
Hey Rich, we have created a page with documentation on TFA for CM Telecom acocunts and updated the docs url in the pull request. Could you please check if the information is sufficient.

https://help.onlinesmsgateway.com/en/accounts-and-payments/how-do-i-enable-two-factor-authentication

Thanks,

Mitch

RichJeanes:
That works perfectly.

PR #1392, opened 2015-10-31
Image files optimization

mxxcon:
Re-compressed most images, replaced a couple with newer versions.
Before: 643440 bytes
After: 578334 bytes


smholloway:
Wow! That's a labor of love. Thanks for keeping the images small and up-to-date.

:+1: :shipit:

Issue #1393, opened 2015-11-02
Amazon retail now supports 2fa.


mxxcon:
As @clex25 mentioned, Amazon store now supports 2fa. Here are screenshots.
![clipboard01](https://cloud.githubusercontent.com/assets/847140/10893364/647b4a24-8176-11e5-87b7-14cb1cc9de70.png)
![clipboard02](https://cloud.githubusercontent.com/assets/847140/10893365/647b7c92-8176-11e5-800a-a010886d1add.png)
![clipboard03](https://cloud.githubusercontent.com/assets/847140/10893366/647eb6d2-8176-11e5-8062-a7b875d783b6.png)

clex25:
It also supports voice calls and TOTP application. [documentation](https://www.amazon.com/gp/help/customer/display.html/ref=help_search_1-5?ie=UTF8&nodeId=201596330&qid=1446498215&sr=1-5)

Carlgo11:
@clex25 see #1394.
I can't find anything about phone calls.

Kallb123:
Seems to be limited to Amazon.com, or at least isn't enabled on Amazon.co.uk.

Customer services said

> Amazon.com and Amazon.co.uk are independent website and we do business independently. So we have different security policies. So there is nothing to worry about it, your account is safe with us.

Carlgo11:
@Kallb123 #1395

Kallb123:
@Carlgo11 thanks! Apologies for my failed search attempts.

Carlgo11:
No worries.

PR #1394, opened 2015-11-02
Update Amazon

Carlgo11:
As per #1393 
Couldn't find any docs.

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -1,9 +1,11 @@
 websites:
     - name: Amazon
-      url: https://amazon.com
-      twitter: AmazonHelp
+      url: https://www.amazon.com
       img: amazon.png
-      tfa: No
+      tfa: Yes
+      software: Yes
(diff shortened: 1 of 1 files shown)


mxxcon:
Carl, this one is cleaner `https://www.amazon.com/gp/help/customer/display.html?nodeId=201596330`
And perhaps squish your commits into 1?

Carlgo11:
@mxxcon done.

jdavis:
:fireworks:

clex25:
There is a phone call option. From the page linked above:

>  You can receive this security code in a variety of ways depending on the option you select during sign up, including text message, voice call, or authenticator app.

Carlgo11:
yup but I can't see any way to do that.
Just tried to login but it won't give me any options to get a phone call.

jamcat22:
You can use it as a backup option, but not as the main option.
capture

PR #1395, opened 2015-11-04
Add exception for amazon retail (amazon uk does not support)

rdark:
As per title - amazon.co.uk does not support 2fa (I suspect only amazon.com/US does)

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -5,6 +5,8 @@ websites:
       tfa: Yes
       software: Yes
       sms: Yes
+      exceptions:
+          text: "Amazon retail only supports 2FA in some countries."
       doc: https://www.amazon.com/gp/help/customer/display.html?nodeId=201596330
 
     - name: Apple


Carlgo11:
Amazon.co.uk does support TFA just like amazon.com.
If you can't find the setting I suggest you login via amazon.com and activate TFA, then continue to use Amazon.co.uk.
amazon

rdark:
@Carlgo11 - you are totally right - it needs to be enabled whilst logged into amazon.com (only then does the 'advanced authentication' link appear in your account settings). When enabled, any future logins direct to amazon.co.uk are via 2fa.

Cheers for the pointer!w

mxxcon:
Actually maybe it's worth mentioning this nuance or raise this issue with Amazon's support?

Carlgo11:
Let's give them a couple of weeks to see if they roll it out to all their sites first.

PR #1397, opened 2015-11-08
Update SpiderOak

Carlgo11:
Spideroak supports TFA for US & Canadian customers.

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -142,7 +142,10 @@ websites:
     url: https://spideroak.com
     twitter: spideroak
     img: spideroak.png
-    tfa: No
+    tfa: Yes
+    doc: https://spideroak.com/articles/2factor-authentication-to-your-spideroak-account
+    exceptions:
+        text: "TFA only available for US and Canadian customers"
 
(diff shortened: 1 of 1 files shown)


mxxcon:
SpiderOak have been claiming to offer beta version of 2fa only in US and Canada for the last 5 years. International support is still on their roadmap...On their roadmap FOR 5 YEARS. This is ridiculous and disgraceful of them! A company that has "Privacy Matters" slogan in giant letters on their home page can't get their shit together _FOR 5 YEARS_ to allow everybody use 2fa? Obviously privacy of their users doesn't matter to SpiderOak.

jamcat22:
@mxxcon At least they have some form of 2FA. I do understand their weird form of not prompting for it and instead requesting it through a separate form.

@Carlgo11 Could you update the exception to say that a paid account is required?

If 2FA doesn't protect the desktop app's first login, then I think we should add an exception or set it to `no`.

Carlgo11:
@jamcat22 They provide no form of TFA for desktop applications.
Closing this PR.

PR #1398, opened 2015-11-08
Remove Frankdomains

jamcat22:
Frankdomains has been taken over by Netim and now uses their account system.


jamcat22:
Ugh. GitHub isn't being nice.

PR #1401, opened 2015-11-09
Twitch.tv now supports two factor authentication

Carlgo11:
Finally! Twitch.tv now supports TFA for all users (not just Twitch-partners).
They use Authy but with phone calls turned off.

Now let's try and get them to use https... :unamused:

### DIFF
--- a/_data/entertainment.yml
+++ b/_data/entertainment.yml
@@ -55,9 +55,11 @@ websites:
 
     - name: Twitch
       url: http://www.twitch.tv/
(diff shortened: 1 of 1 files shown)


Carlgo11:
That was quick...
@mxxcon pretty much added the `looks-good` label the second I made the PR :yum:

RichJeanes:
Well it was pretty clear cut. Not to mention there weren't any problems with your PR (as I would hope).

Issue #1404, opened 2015-11-12
Passpack Supports 2 Factor Auth


ericcub7:
The identity management program Passpack supports email and hard tokens as a method of two-factor authentication

Carlgo11:
@ericcub7 see [Contributing.md](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md) to learn how to add a site to our list.

PR #1420, opened 2015-11-20
Add Passpack

Carlgo11:
As per #1404


jamcat22:
Maybe we could use this link (https://help.passpack.com/knowledgebase/idx.php/44/0/) as the doc, as it refers to all of their Two Factor Authentication articles.

Carlgo11:
@jamcat22 done.

PR #1429, opened 2015-11-23
facebook: Fix exception help text

tvierling:
Facebook allows multiple mobile phones per account (and will indeed SMS-text all of them simultaneously upon login). User can switch to TOTP OATH after initial setup using any GAuth app, or via the Facebook mobile app which has an OATH generator built-in.

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -36,7 +36,7 @@ websites:
       sms: Yes
       software: Yes
       exceptions:
-          text: "SMS required for 2FA. 1 account per phone number, 1 phone number per account."
+          text: "SMS required for 2FA setup, but OATH generator/Facebook mobile app can be added afterwards."
       doc: https://www.facebook.com/help/148233965247823
 
     - name: Foursquare


tvierling:
Wasn't in the commit comment, but Facebook also allows the same phone number to exist on multiple accounts, hence the removal of both statements about 1:1 phone:account mapping.

mxxcon:
Is this documented somewhere on facebook's site?

jamcat22:
This is only partially true. I just tried to add my phone number to another Facebook account to enable login approvals. Upon verifying the phone, it deleted the phone from my other account. Removing a phone from your account also disables Login Approvals, making SMS required for 2FA even if you are using the mobile app.

I received these emails:
screenshot_112315_070555_pm

> Hi Jacob,
> 
> The phone number _______ has been deleted from your profile due to the following reason: the mobile number is registered and verified by another Facebook user.
> If you changed your mobile number or switched mobile service providers, please visit Facebook to add your number again.
> Thanks,
> 
> The Facebook Team

screenshot_112315_070651_pm

> Hi Jacob,
> 
> You recently removed your primary phone from your account. Because of this, we turned off login approvals on your account to ensure that you don't get locked out when using an unrecognized computer or mobile device to log in.
> If you want to use login approvals again, you can add a mobile phone and then turn on login approvals again.
> 
> Thanks, The Facebook Team

However, it is possible to add multiple numbers to an account and have Facebook text both.
screenshot_112315_073640_pm

_data/social.yml
```diff
@@ -36,7 +36,7 @@ websites:
       sms: Yes
       software: Yes
       exceptions:
-          text: "SMS required for 2FA. 1 account per phone number, 1 phone number per account."
+          text: "SMS required for 2FA setup, but OATH generator/Facebook mobile app can be added afterwards."
```
jamcat22, line 39: Instead of changing this to `SMS required for 2FA setup, but OATH generator/Facebook mobile app can be added afterwards.`, simply remove the message about one phone number per account.

The line should look like this:
`text: "SMS required for 2FA. 1 account per phone number."`

tvierling:
OK, the no-sharing-of-number behavior is newer; I had two accounts set up with the same number for some time, but deleting and re-adding a number to one account prevents me from doing so.

Closing this pull request, will submit a simpler one as requested above.

PR #1430, opened 2015-11-23
twitter: Include note about mobile app

tvierling:
Twitter's mobile app can function as the primary 2FA via push-based approvals, though to set it up, it's still necessary to set up a mobile number.

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -108,7 +108,7 @@ websites:
       tfa: Yes
       sms: Yes
       exceptions:
-          text: "SMS only available on select providers. SMS required for 2FA. 1 account per phone number, 1 phone number per account."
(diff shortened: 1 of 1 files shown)


mxxcon:
How do you know about this?

tvierling:
Because I use it daily and it's described at the Twitter documentation link right in that stanza:

https://support.twitter.com/articles/20170388

jamcat22:
This is already implied when we say it supports `software`. Our "SMS required for 2FA" message means you must have a phone number on the account to use 2FA. If you remove your phone number from your account, it disables all forms of 2FA (including app-based 2FA). See #1122 and [#627](https://github.com/2factorauth/twofactorauth/pull/627#issuecomment-48353655)

PR #1432, opened 2015-11-24
Facebook: more than one phone number permitted per account

tvierling:


mxxcon:
wait, you can now add more than 1 phone number to an account?

tvierling:
Yes, you can (and this has been true for a long time).

jamcat22:
Yep. Even before Login Approvals were a thing. I remember when they just had Login Notifications and not Login Approvals.

mxxcon:
@jamcat22 in #1250 you added 1 number per account. was that info wrong?

jamcat22:
Yes. I was thinking about Twitter when I made that PR. :laughing:
Mistake on my part. They do only allow 1 account per phone number though.

fpigerre:
Thanks for the pull request @tvierling! :+1:

PR #1433, opened 2015-11-24
Add note about Amazon

jomo:
I think this is relevant because many people prefer software tokens because they **don't** have to share their phone number with the vendor.
Most of (all?) the other sites do not require this extra step and provide a set of backup tokens instead.

screenshot

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -5,6 +5,8 @@ websites:
       tfa: Yes
       software: Yes
       sms: Yes
+      exceptions:
+          text: "Software token requires backup phone number for voice call or SMS"
       doc: https://www.amazon.com/gp/help/customer/display.html?nodeId=201596330
 
     - name: Apple
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -6,7 +6,7 @@ websites:
       software: Yes
       sms: Yes
       exceptions:
-          text: "Software token requires backup phone number for voice call or SMS"
+          text: "SMS/Phone Call required for 2FA"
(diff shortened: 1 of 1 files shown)


mxxcon:
At the bottom of your screenshot it shows "you can opt out at any time", so it seems like it's not required..at least for continual use, you can opt out from it.

Also aren't you required to provide a phone number when signing up for amazon account?

jomo:
Well you can't actually get past that screen without entering a phone number and entering the received the code.
You can cancel it, but you can't even disable 2FA anymore because that requires completing this step.
IIRC the software token was not required for sign in at this point, but it was definitely required to change (or disable) 2FA.

After completing setup you can either disable 2FA completely:
amazon 2fa overview

Or you can change the backup method from SMS to voice call, but you can not opt out:
change backup method

> Also aren't you required to provide a phone number when signing up for amazon account?

Possible, I don't remember. However you're not required to use your actual phone number.
Of course the same applies here, but it's an additional hurdle which other 2FA providers don't have.

jomo:
Yep, just tried again. They have removed the message about opting out now.
After setting up the software token and logging out, log in **does not require** the token.
They acknowledge you've set that up but you still need to "Get Started":

screenshot get started

And as said before, this ironically **does require** the token:

screenshot

jamcat22:
Thanks @jomo!
Maybe change the message to "SMS/Phone Call required for Software 2FA."?

jomo:
done @jamcat22!

jamcat22:
Awesome! `looks-good`

Issue #1423, opened 2015-11-20
Please add Site5.com


kraymitchell:
For Web Hosting, please add Site5.com

mxxcon:
If you'd like to, you can submit a pull request. It's a simple yaml block with 32x32 png logo.

jamcat22:
Closed via #1439.

PR #1439, opened 2015-11-28
Add Site5 Web Hosting

treed593:
#1423


jamcat22:
Maybe use their "5" logo? 
site5 1

And you can use their main page (http://www.site5.com/) instead of their login page.

jamcat22:
Awesome! Thanks @treed593!

PR #1452, opened 2015-12-11
remove SMS-only exception message for Amazon.com, no longer true

fallenpegasus:


mxxcon:
At least for amazon.com you no longer need sms. Can somebody with not-.com account confirm this?

jamcat22:
I do not see how this isn't true anymore. On the Amazon website, you must have a backup method, and if you have an authenticator app listed, that backup method must be a phone number.

Their docs also say:

> Note: You won't be able to turn on Two-Step Verification without adding a back-up phone number. This is so that you have a back-up option to receive a security code if you no longer have access to your primary mobile device.

mxxcon:
But that's a backup phone number, different from your own..

jamcat22:
But that also makes SMS/Phone Call required for 2FA, as you need to have a phone number added.

PR #1468, opened 2015-12-19
Remove Mitro since it has shut down

jzbz:
Mitro has ceased operating https://www.mitro.co/shutdown.html


mxxcon:
@jzbz also please delete image file.

jzbz:
Done.

Carlgo11:
@jzbz you have to do it in the same PR.

PR #1470, opened 2015-12-20
Update GOG.com icon

Eitot:
Optimised with `optipng -o7`.


mxxcon:
@Eitot if you are using mac, https://imageoptim.com. if you are using windows, http://css-ig.net/pngslim

PR #1471, opened 2015-12-20
Add Medium to Social

dsoegijono:


jamcat22:
I'm not totally sure how they would incorporate this, as they only use social login or they send you a one time email. They got rid of passwords awhile ago.

dsoegijono:
@jamcat22 hmm true. The one time email counts as a second factor right? But since they got rid of passwords, they don't really have the first factor, except for email?

jamcat22:
Yeah. I've been wondering about that. If we count it as a second factor, then Shazam would also have 2FA. If we don't add Medium, it would sort of advocate for them to add Software, Hardware, Phone Call, or SMS for 2FA, even though that would be two possession factors. Maybe it would tell them to add an optional password in addition to their email system, like Telegram.

mxxcon:
Maybe add it as tfa: email, but also add exception indicating lack of password?
Alternatively add medium and similar type of sites to https://github.com/2factorauth/twofactorauth/blob/master/EXCLUSION.md ?

dsoegijono:
@mxxcon Both options make sense, but I'm leaning towards the former (email tfa with exception). I feel that sites should only be added to `EXCLUSION.md` if they requested it, or as in the case with 1Password they just have a different definition of 2fa.

Does anyone have an opinion on this?

dsoegijono:
Changed the tfa to email, and added exception. @mxxcon @jamcat22

PR #1475, opened 2015-12-27
Twitter now allows 10 accounts per phone number

joesondow:
Twitter has updated their two-factor authentication policies to allow up to ten Twitter accounts to use a single device for 2FA.

Reference doc for new policy mentioning the newly raised account limit: https://support.twitter.com/articles/110250

I now have several of my Twitter accounts successfully using a single iPhone for 2FA.

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -108,7 +108,7 @@ websites:
       tfa: Yes
       sms: Yes
       exceptions:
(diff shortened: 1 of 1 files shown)


jamcat22:
Thanks for the PR @joesondow! Another contributor will review it before it's merged.

After a PR is merged, it can take a little while before the site is updated. See our [wiki](https://github.com/jdavis/twofactorauth/wiki/FAQ_Merge) for more details.

PR #1476, opened 2015-12-29
Add SigFig to investment category

e-shreve:
Added SigFig.com to investment sites, including image. SigFig does not have 2FA support (they support a PIN for their mobile app, but I wouldn't consider that 2FA.)


jamcat22:
Thanks @e-shreve for your PR!

It may take some time before your edit appears on the site. For more info, check out our [wiki](https://github.com/jdavis/twofactorauth/wiki/FAQ_Merge).

PR #1480, opened 2016-01-01
Add tables to alternate layout

fpigerre:
Hello!

This pull request starts to make the alternate layout usable.
At the moment, categories can be clicked and opened, as well as closed, and tables appear when the categories are opened. The tables still need some styling to appear smoothly, however, and the mobile layout is virtually non-existent, however, it still seems to be usable.

Thankyou,
psgs :palm_tree:


jamcat22:
So. Many. Commits.

fpigerre:
Yes, it seems `alternate-layout` hadn't been updated from master for a while :stuck_out_tongue:

jamcat22:
I was waiting for something to happen to `alternate-layout`!! :smile:

Carlgo11:
how about we just do a force push with the commits from `master` before we merge any PR?
Just so that we clean up the PR:s.

fpigerre:
Ok, I've manually merged the commits from `master` into `alternate-layout` @Carlgo11.

mxxcon:
@psgs Unless i'm mistaken, I don't think your current `_data/sections.yml` is a valid YAML file. Afaiu each block must be prefixed by `-`, ie

```
- backup:
    id: backup
    title: Backup and Sync
    icon: disk outline

- banking:
    id: banking
    title: Banking
    icon: dollar
```

fpigerre:
@mxxcon: According to [Ansible's documentation](http://docs.ansible.com/ansible/YAMLSyntax.html), and [YAMLlint](http://www.yamllint.com/), the YAML contained in `_data/sections.yml` appears valid. It seems that Key/Value pairs can simply be created within each file without the need for a list, meaning that in `sections.yml`, a key/value pair is established for the backup section; `backup` being the key, and the underlying elements being the value. This way, a structure is established in a similar fashion to how an object might be declared in a language such as JavaScript.

The reason I primarily chose to do without a list in the `sections.yml` file is because of the way section names are passed as parameters to the `_includes/table.html` file when displaying websites that support TFA. In order to correctly access a particular section of the file without iterating through the entire file, individual sections needed to be accessed through Jekyll's `site.data` variable.
This is shown through how variables are passed to the `table.html` file:

```
{% include table.html variable-param=site.data.sections.finance %}
```

Unfortunately, when using a list, I wasn't able to simply reference a particular element in the list, as is shown above.

Issue #1320, opened 2015-09-12
Fidelity 401k now supports 2FA


j2xl:
If you have two-factor authentication enabled for a Fidelity account, logging into retirement accounts at 401k.com (Fidelity's retirement benefits site) now prompts you for your authentication code.

Please remove the note at the bottom of the "How to Setup 2FA for Fidelity" section on /notes/fidelity/.

jamcat22:
Do they use the same login and password system?

j2xl:
Yes. I have the same account for my Fidelity investment accounts (which I've had for 10+ years) and for retirement benefits (401(k) established 1 year ago). I enabled 2FA last October and in the last couple of months I've noticed that it's enforced whether I log in on fidelity.com or 401k.com.

In the past, if I logged in to 401k.com I wasn't prompted for 2FA until I clicked a link that took me to an investment account (link from 401k.com to fidelity.com). Now I'm prompted for 2FA at logon on either site.

j2xl:
In addition, the number to call is now 800-544-6666:

![image](https://cloud.githubusercontent.com/assets/3752878/9834326/33af1d1e-596a-11e5-917c-4eaa88469293.png)

jamcat22:
Great! Thanks for telling us. I'll make a PR in a couple of minutes.

j2xl:
Thank you! I appreciate the quick response.

PR #1485, opened 2016-01-02
Add NetBenefits Exception to Fidelity

kubla36:
NetBenefits users do not get to use the two factor auth (tried to set up this week and was told no -- worse I have to send land mail to complain).

### DIFF
--- a/_data/investing.yml
+++ b/_data/investing.yml
@@ -25,6 +25,8 @@ websites:
       tfa: Yes
       software: Yes
       hardware: Yes
+      exceptions:
+        text: "NetBenefits not supported (e.g. employer sponsored 401k/403b)."
       doc: /notes/fidelity/
 
     - name: HealthEquity


mxxcon:
@kubla36 are there different login locations for NetBenefits accounts vs other accounts?
edit:
So it seems like there's www.netbenefits.com which redirects to https://nb.fidelity.com/public/nb/default/home
Then there's also https://login.fidelity.com/ftgw/Fas/Fidelity/IspCust/Login/Init and https://login.fidelity.com/ftgw/Fas/Fidelity/NBPart/Login/Init?ISPBypass=true

kubla36:
I only have netbenfits through my company, and can only login through the
netbenefits web page.  When I called Fidelity, they told me that two factor
was not available for netbenefits (would not provide a token or validate
the VIP software code I downloaded).

Their FAQ does not state clearly but (
https://www.fidelity.com/security/soft-tokens/faqs) if you read between the
lines indicates that netbenefits login page does not support.

I've read other postings on the web (but cannot confirm since I cannot get
a code validated), that netbenefits sites never requires a second factor
even when they have one for the fidelity site.  Defeating the purpose
(hacker goes through the netbenefits site).

jamcat22:
Hm. According to #1320, NetBenefits should support 2FA now. I wonder if they changed their system again.

kubla36:
When I call they say two factor is not offered for netbenefits.  It could
be my particular company's arrangement?  I don't have a regular fidelity
account as mentioned by the person in #1320.

I'd note the security page for netbenfits does not mention two factor (
https://nb.fidelity.com/public/nb/default/resourceslibrary/articles/onlinesecurityatfidelity),
whereas the page for the fidelity site does (
https://www.fidelity.com/security/our-security-measures).

Carlgo11:
I feel like we need some more information about Netbenefits' and Fidelity's TFA arrangements before we can do anything with this PR.
Maybe someone could contact them and ask?

stephengroat:
The document provided by @kubla36 (https://www.fidelity.com/security/soft-tokens/faqs) now shows that netbenefits is supported by "VIP Access" aka 2FA (great rebranding name for 2FA too).

This should probably be closed

kubla36:
I followed the instructions and they did not work.  I called Netbenefits
and they said they were rolling out a new service client by client
(July-August).  However, the service was _not_ "VIP Access", despite what
their FAQ at the Fidelity (not Netbenefits) site says online.  I requested
they update that in the FAQ.

Rather, they said it was going to be text message or email based for "high
risk transactions".

So depending on who you are as a Netbenefits customer, you may or may not
have access.  But soon there will be some form of 2FA.

Carlgo11:
so no 2FA then. And the "high risk transactions" thing you mention is two factor **authorization** and not **authentication**.
Either set `tfa: No` or close the PR.

Issue #1368, opened 2015-10-14
Alternate Design and Layout


fpigerre:
At the moment, [TwoFactorAuth](https://twofactorauth.org) displays all added websites on the front page inside a number of tables, ordered by category. This creates a number of issues, notably the rendering of the site for mobile users, speed of loading with regards to users who have access to lower bandwidth speeds and scaling the number of websites listed at TwoFactorAuth.org.
Furthermore, adding columns for more data, or sections for peripheral data seems to be quite cumbersome at the moment.

In order to solve these issues, a non-tabular layout could be implemented.
Some suggestions up to this point include the implementation of a carousel-type layout:

> How about a change to something like a carousel where each category would get a separate view?
>    @mxxcon 

as well as the implementation of an icon-based category selection layout:

> And maybe when you go to the site, it shows you a grid of icons. You can choose which category to view, and there's a transition to that category.
>    @jamcat22 

It is possible, however, that moving away from a tabular layout could reduce the simplicity of twofactorauth.org, in turn reducing user engagement.
Would moving away from a tabular layout be beneficial in the long term?

This issue may involve discussion or information relating to pull request #1348.

groenroos:
Personally I'd go with the icon grid as per @jamcat22.

![Grid of icons](http://i.imgur.com/LZmyqWv.jpg)

When you click on one, the data could be appended below, so that the icon-based navigation is retained (no back button ping pong).

![Data](http://i.imgur.com/biXtbAq.jpg)

This would ease navigation on desktop, especially since on the current site, the category nav is in the header, which isn't fixed (so you have to scroll up to even see it).

This should also improve mobile performance, since the whole mess isn't rendered in one go. As for the mobile-friendliness of tables, the major problem is that the majority of the table cells are just empty space. It's quite easy to eliminate this space;

![Mobile layout](http://i.imgur.com/lm5In1m.png)

Not sure how to achieve this in terms of table-based markup. This layout wouldn't look too amazing on a desktop resolution -- I think the table layout works pretty well when you have the space for it.

Also, not really sure if this is part of this conversation, but I would like to suggest adding a local search bar, that would filter the table, much like the [FontAwesome icon search](https://fortawesome.github.io/Font-Awesome/icons/). There's always Ctrl-F, but it'll pick up on any words on the site (in lieu of just site names), and it's a somewhat obscure feature for the average non-techy user.

RichJeanes:
[...]

PR #1486, opened 2016-01-04
Add filtering capabilities

fpigerre:
Hello!

This pull request adds filtering capabilities and a search bar using the [Jets library](http://NeXTs.github.com/Jets.js/).
This pull request also fixes issues #926 and #521.

More information, including screenshots, can be found in issue #1368.

Thankyou,
psgs :palm_tree:


stacksjb:
Hello,

This works wonderfully, but is not intuitive at all - I came to Github to request adding filtering because I didn't see any checkboxes or drop-downs or other similar filtering.   Is there a way to make this more intuitive, by doing one of any of the following:
- adding checkboxes for filtering? 
- Adding a selectable dropdown filter at the top of the columns:
- Adding text to the search pop-up with a "Common searches" listing (i.e. "Try "U2F", or "Phone")?
- Or making it so that if you click on the top of the column, it auto-adds a search for it?

PR #1491, opened 2016-01-07
Add CVS/Caremark to Health

elliotkendall:


mxxcon:
That twitter account doesn't seem to be managed by humans. Just useless rss feed spam.
Maybe use main CVSHealth account?

Carlgo11:
The "@CVS_CaremarkNws" twitter account is made by a Dutch company called Newslocker B.V. and not CVS

elliotkendall:
Whoops. Thanks, folks. I've updated it to use @cvshealth which is probably the closest option available.

PR #1494, opened 2016-01-09
Update Wildstar docs and icon

atomicptr:
Old doc link wasn't working also changed the icon to the "new" one.


Carlgo11:
Hello @kasoki,
The new logo you provided is larger than the size we use for images on [twofactorauth.org](https://twofactorauth.org).
Here's a compressed version of the same logo. I suggest you use that one instead.
wildstar logo

atomicptr:
done

jamcat22:
How about using this support link? https://support.wildstar-online.com/hc/en-us/articles/203982369-Two-Step-Verification-FAQ

PR #1496, opened 2016-01-09
Adding Nest To 'Other' List

robbettis:
Adding Nest, as the DVR service at home.nest.com often contains personal data (Criteria 1) and Nest does not yet offer 2FA.


Carlgo11:
Hello @robbettis,
Thank you for your contribution.

There's just one little thing I'd like you to change with this PR. The `status:` tag should only be used when the company is actively working to implement TFA. Since Nest isn't doing that, it would be awesome if you could remove the status tag.

Also, do you think it would be more appropriate to change the url from [home.nest.com](https://home.nest.com) to [nest.com](https://nest.com)?

Thank you,
//@Carlgo11

robbettis:
Hey @Carlgo11,

Thanks for the input. I’ll update the status tag. I was hoping to include a link to the support page where folks could vote for the feature (vs just tweeting them), but it doesn’t seem to fit into the existing format, which is fine. 

As for the URL, I was trying to differentiate between the Nest store. `home.nest.com` is the area that contains the camera DVR feeds and thermostat controls, which seems to be a better fit for the criteria, in my opinion. But I’m happy to change it, if you think that’s best. 

Thanks!

fpigerre:
Hello @robbettis,

It seems that [home.nest.com](https://home.nest.com) displays a page which prompts the user for access credentials. Although redirecting users to the page which displays the thermostat controls would be greatly beneficial for logged-in users of Nest, linking to the Nest front page instead may improve the experience for viewers of [twofactorauth.org](https://twofactorauth.org) that are not currently registered with Nest, or users of Nest that are not logged in.

Kind regards,
psgs :palm_tree:

robbettis:
@psgs @Carlgo11 Good feedback. Thanks. My PR has been updated accordingly.

PR #1501, opened 2016-01-10
Fix ID Linkage Collision with header. Add Titles to Icons. Remove Border Radius from Nav.

jaxgeller:
First off, thank you all for making this site. I've been using it to influence my banking choices. 

Anyways I've pushed three commits here.
- Anytime you link to a section via an id, like https://twofactorauth.org/#banking, the nav is cutting off the table header. I've adding some padding equal to the height of the nav to remedy this. I'm happy to make this a round value, or use em or something. Let me know.
- When far down a list you can sometimes lose track of what column is which. I've added titles to each icon that represents each column which should be useful.
- There was some border radius on the menu that doesn't look too good for a full width fixed nav, so I removed it.

I'm happy to remove/add/tweak any of the commits just let me know. And once again, thank you all for making this, it's awesome and I'm glad something like this exists. 

Thanks!

### DIFF
--- a/css/base.css
+++ b/css/base.css
@@ -26,6 +26,10 @@ p {
   z-index: 1;
 }
 
+.ui.inverted.teal.menu {
+  border-radius: 0;
+}
+
 .ui.dropdown .menu.visible {
   max-height: 90vh;
   overflow-y: scroll;
--- a/css/base.css
+++ b/css/base.css
@@ -4,6 +4,10 @@ img { border: none; }
 ul { list-style: none; }
 a { color: #009FDA; text-decoration: none; }
 
+.section {
+  padding-top: 52px;
+}
+
 body {
   font-family: "Open Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
   color: #505050;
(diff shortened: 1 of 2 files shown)


smholloway:
Thanks for the PR, @jaxgeller! Can you include a screenshot or two, please?

jaxgeller:
### Border Radius

screen shot 2016-01-10 at 6 35 48 pm

vs

image

### Titles

screen shot 2016-01-10 at 6 38 50 pm

### ID Linking

image

Carlgo11:
Hello @jaxgeller,
thank you for your PR.

We're in the middle of changing website layout so I would recommend you check if the problems persist on the `alternate-layout` branch.

fpigerre:
Hello @jaxgeller,

This pull request looks awesome!
I really like the way in which you have added tool-tips to each icon.
At the moment, we are working on implementing a new layout for the website under the `alternate-layout` branch. I will definitely attempt to incorporate the changes you have made in the new layout.

Thankyou,
psgs :palm_tree:

PR #1512, opened 2016-01-12
Removes "working on it" from Payoneer.

mxxcon:
It's been close to 2 years now since they claimed that and it's still not released.


Carlgo11:
https://twitter.com/2faorg/status/687024611641421824

Carlgo11:
As they haven't replied yet I'm okay with removing their status tag.

PR #1519, opened 2016-01-15
Remove Hardware Token from Fidelity Investments

ghost:
I was just on the phone with a rep and he mentioned that they don't support hardware tokens, just the Symantec VIP shovelware.

### DIFF
--- a/_data/investing.yml
+++ b/_data/investing.yml
@@ -24,7 +24,6 @@ websites:
       img: fidelity.png
       tfa: Yes
       software: Yes
-      hardware: Yes
       exceptions:
           text: "Only available for some account types"
       doc: /notes/fidelity/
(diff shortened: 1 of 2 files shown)


jamcat22:
Hello @ainola!

Did they say they wouldn't send a hardware token, or they wouldn't add an existing one?
Credential IDs for the software token and the hardware token are the same format, so even if they don't sell the hardware tokens anymore, you should be able to add one to your account.

ghost:
They wouldn't send one. They said they only support the apps. But I guess it sounds like one could just call them up and give the serial to them anyway.

Carlgo11:
Yup I think an exception isn't needed as it seems like one could just buy a hardware token from another place and give them the serial number.

ghost:
Feel free to close, then!

PR #1526, opened 2016-01-16
Added Onshape

vonProteus:
added Onshape


_data/other.yml
```diff
@@ -244,3 +244,10 @@ websites:
       tfa: Yes
       sms: Yes
       doc: http://support.cosmolex.com/customer/portal/articles/2158628
+
+    - name: Onshape
+      url: https://cad.onshape.com/
```
Carlgo11: That url is login protected. I recommend that you use `https://www.onshape.com/` instead.

Carlgo11:
Second confirmation on this?

PR #1528, opened 2016-01-17
Set table column widths for desktop

dsoegijono:
Set table widths so that they're fixed, not dependent on table content.

Before:
screen shot 2016-01-17 at 1 33 32 pm

After:
screen shot 2016-01-17 at 1 26 20 pm

I'm using Semantics' 16 grid system. I can change them to `px` widths to be more flexible if you guys prefer that.

### DIFF
--- a/_includes/desktop-table.html
+++ b/_includes/desktop-table.html
@@ -4,13 +4,13 @@
   <table class="ui celled table content-wrapper">
     <thead>
     <tr>
-      <th class="single line"><h3>{{ section.title }}</h3></th>
-      <th>Docs</th>
-      <th>SMS</th>
-      <th>Phone Call</th>
-      <th>Email</th>
-      <th>Hardware Token</th>
-      <th>Software Token</th>
+      <th class="single line four wide"><h3>{{ section.title }}</h3></th>
+      <th class="two wide">Docs</th>
+      <th class="two wide">SMS</th>
+      <th class="two wide">Phone Call</th>
+      <th class="two wide">Email</th>
+      <th class="two wide">Hardware Token</th>
+      <th class="two wide">Software Token</th>
     </tr>
     </thead>
     <tbody class="jets-content">


mxxcon:
Is it possible to keep cells the same height even if site name wraps to 2nd line?

fpigerre:
Hello @dsoegijono,

Do you think it might be possible to ensure that the section title column uses eight semantic units of width in order to maintain the current cell grouping system?

``` html
<th class="single line eight wide"><h3>{{ section.title }}</h3></th>
```

Thankyou,
psgs :palm_tree:

dsoegijono:
@mxxcon I'd think so. So do you want all rows to be able to fit 2 lines, or squeeze the extra line into the 1 line height? For me personally, different row heights don't bother me as much as different column widths :smirk:

dsoegijono:
Hey @psgs, sure, I'll make the changes in a bit

mxxcon:
"squeeze extra line into 1 line height". I think keeping things visually uniform is more pleasant, but I won't fight it if other folks object.

dsoegijono:
Hey @psgs, so the reason why I chose those widths (`four wide` for section title, and `two wide` for others) is because they add up to 16. If we do `eight wide` for the section title, it means half of the the table is taken up by section title/website names, and it leaves eight column widths for the six remaining columns. Unless you want a couple of those columns to be wider than the rest, I'm not sure how to make it work. :confused:

dsoegijono:
And, @mxxcon, it's proving to be more difficult than I thought to control row heights, so I'm not sure I can do that today.

fpigerre:
@dsoegijono, I see. In that case, leaving the section title column width at four semantic units would be best :smiley: 
Perhaps we might be able to change the default column count in the future in order to create a better fit, as is said to be possible in [Semantic UI's documentation](http://semantic-ui.com/collections/grid.html#columns).

Carlgo11:
So what's the status on this PR?
Merge? Pending? Won't fix?

dsoegijono:
@psgs and I are in agreement on the column width issue, so I think that one's fine.

For the height issue that @mxxcon brought up, not sure when I can attempt to do it. So if someone would like to help, I appreciate it. Otherwise, maybe we can merge this first so it doesn't hang around too long, and then fix the height issue in a different PR?

Carlgo11:
Merging this one first sounds good.
I'm not a design wizard and know barely anything about CSS so I'll let someone else do the `looks-good`and merge.

PR #1531, opened 2016-01-18
Add some more transportation companies

Carlgo11:
First off, I know I shouldn't add multiple sites in 1 PR but I felt like it would just be a waste of everyone's time if I made multiple PRs for these sites that are so similar.

This PR adds SAS, Norwegian, American Airlines, Amtrak and British Airways.

PR #1532, opened 2016-01-18
Test more tags in the YAML files

Carlgo11:
Hello,
this is a small PR to add a test of the `tfa:` tag in the verify.rb script.

It checks whether the tfa tag is true/false which yes/no is transformed to.
This means that `tfa: true` also would be okay but I opted for this as I did the pros and cons of the extra work of verifying it is in fact Yes/No and we haven't had any cases yet where people have written true/false instead of Yes/No. (To my knowledge at least)

Most of all this is to stop me from doing the error I did in #1531 again.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -46,7 +46,7 @@ websites:
       hardware: Yes
       doc: https://www.bankofamerica.com/privacy/online-mobile-banking-privacy/safepass.go
 
-    - name: Barclays
+    - name: Barclays UK
       url: http://www.barclays.co.uk/
       img: barclays.png
       tfa: Yes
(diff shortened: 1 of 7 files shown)


jamcat22:
Can you have it check the other booleans too if they are included in the file?

Carlgo11:
@jamcat22 I'm not very good at Ruby yet but I'll see what I can do.
#Got2LookNice

jamcat22:
^ Same. That's why I made you do it. :stuck_out_tongue:

Carlgo11:
There we go. It now also checks that `software`,  `email`, `phone`, `sms` is set to `Yes` if set and that it isn't set when `tfa` is set to `No`

PR #1537, opened 2016-01-20
Add VentraIP Australia

ghost:


Carlgo11:
After some digging it looks like VentraIP only offers 2FA for their 'Wholesale users'.
The Wholesale page links to another domain and company called "Synergy Wholesale" so if anything Synergy should be listed as supporting TFA and not VentraIP.

Feel free to reopen this PR if I'm wrong in any way.

PR #1542, opened 2016-01-21
Rename Garena Online to Garena Offline

Carlgo11:
It looks like "Garena Online" no longer is "Online". _*Pun intended*_
Their domain garena.com seems to have been bought up by another company from Singapore (doing god knows what. But it isn't gaming related)

And no, this PR doesn't rename Garena Online. It removes the site completely. Just having fun with the header here at 6am.

//@Carlgo11 :tay:


jamcat22:
Login link at the top still works...

jamcat22:
Their [business page](http://corporate.garena.com/ourbusiness.php) still mentions gaming too, and their Garena+ desktop client still works.

mxxcon:
According to Wikipedia it was founded in Singapore..

jamcat22:
See #1543.

PR #1551, opened 2016-01-28
Update to Snapchat for Software Auth

kylekorth:
Snapchat now supports software based 2fa. 
https://support.snapchat.com/a/enable-authentication-app

I believe we should update the list with this information.


fpigerre:
Thanks for the pull request @kylekorth! :+1:

PR #1554, opened 2016-01-29
Added Pocket to Social

whatinthenameofgod:


whatinthenameofgod:
Looks like I accidentally added a second site to this request.

How can I keep these PR separate? Do I have to fork for every single PR?

mxxcon:
If you make separate pull requests for each site, you'll need to create separate branches in your repo, make changes there and then create pull requests from those branches.

jamcat22:
You might want to use [branches](https://help.github.com/articles/creating-and-deleting-branches-within-your-repository/) so that you can propose multiple changes

whatinthenameofgod:
Got it. I'm learning! Thank you! Is this PR okay to leave even though I accidentally tossed two in or should I redo it?

dsoegijono:
I don't know if this has been discussed before, but is Pocket really a `Social` site? I would say it fits better in `Backup & Sync`? Or even `Other`...

whatinthenameofgod:
Yeah, backup and sync also makes sense. Although it seems they're trying to get more and more into the social arena which I don't care for at all. Stick to what you do best!

mxxcon:
I'd think Pocket should be in `other`, and even then I'm not sure if it makes sense to list it...It's essentially a fancy bookmark site.

jamcat22:
I'd say it should be in `Other`, as Pinboard is there also.

Carlgo11:
Size reduced from 16kb to 654b.
pocket 1

jamcat22:
Thanks @Carlgo11. When I resized the new image, I forgot to upload the compressed version. :laughing:

whatinthenameofgod:
Going to kill these and resubmit as individual.

PR #1555, opened 2016-01-29
Add Xbox Live to Games

whatinthenameofgod:


_data/gaming.yml
```diff
@@ -200,4 +200,14 @@ websites:
       img: wildstar.png
       tfa: Yes
       software: Yes
-      doc: https://support.wildstar-online.com/hc/en-us/articles/203982369-Two-Step-Verification-FAQ
+      doc: https://support.wildstar-online.com/hc/en-us/articles/203982369-Two-Step-Verification-FAQ'
```
mxxcon: no need for `'` at the end of url here..

whatinthenameofgod: Yeah, that was a mistype. I added a second fix to remove it.

mxxcon:
Hmm, I can't confirm phone calls...
As a matter of fact I can't confirm email or sms either.
Just their authenticator app..

whatinthenameofgod:
Yeah, I'm surprised because I thought they had more than just the app since they say so on the main two-factor page, but I just tried it out and could only see the app as well. Just scrubbed those and kept the app only.

jamcat22:
They do support all of those methods. You just need to not have an authenticator app enabled for it to work. It uses the same methods your Microsoft Account uses.

whatinthenameofgod:
In that case, I'll fix it back!

mxxcon:
@jamcat22 ya, it's for any microsoft-related services, not just xbox.
ok, i kinda figured out how to get it to prompt for login code, but i get options only for sms or email, not phone call...

whatinthenameofgod:
Removed the phone!

fpigerre:
Thanks for the pull request @whatinthenameofgod! :+1:

PR #1562, opened 2016-01-30
Update Userify - 2FA enabled

perpetual-hydrofoil:


mxxcon:
There's no need for `sms: No` and `hardware: No`. If you don't support it, just omit it.

perpetual-hydrofoil:
Sorry about that.

PR #1564, opened 2016-01-31
Fix image size

Carlgo11:
This PR compresses 2 logos that are larger than 2500 bytes.


mxxcon:
maybe while you are here, recompress all the other images?

Carlgo11:
@mxxcon I use a lossy compression :/
If I did that I'd ruin some.

Issue #975, opened 2015-02-12
Apple iMessage/Facetime


mxxcon:
http://techcrunch.com/2015/02/12/apple-adds-more-security-to-imessage-and-facetime-with-two-factor-authentication/

Btw, should these services be added separately from Apple's "retail" entry?

RichJeanes:
If they're separate from your retail account, then yes. I'm not an Apple user, but I'm pretty sure they are. Aren't iMessage/FT part of iCloud? I thought that already had TFA?

Carlgo11:
@RichJeanes iMessage & FaceTime are separate from iCloud.

mxxcon:
Iirc it's all the same account, but apparently different products have different authentication mechanisms.

jamcat22:
Apple has two different systems for additional verification.
The first is 2FA (two-factor authentication), and the second is 2SV (two-step verification).
Both are separate mechanisms and how they work varies greatly.

| 2FA (two-factor authentication) | 2SV (two-step verification) |
| --- | --- |
| \- **Protects all parts of your account (except those listed towards the end of the comment)** | \- **Protects certain parts of your account** |
| \- Code entered in special box (for software that supports it), [app-specific passwords](https://support.apple.com/en-us/HT204397) (for certain software), added to end of password (for software without built-in support), or **no code required at all** (for software that isn't protected by 2FA; see list below) | \- Code entered in special box (for software that supports it), [app-specific passwords](https://support.apple.com/en-us/HT204397) (for certain software), or **no code required at all** (for software that isn't protected by 2SV) |
| \- Requires a device running at least OS X El Capitan or iOS 9 to enable and receive codes | \- Can be enabled on any Apple ID using the [Manage Your Apple ID page](https://appleid.apple.com/account/manage/security) |
| \- After enabling on a macOS/OS X or iOS device, SMS or Phone Call can be used to receive codes in addition to push notifications | \- SMS or a push notification to an iOS device can be used to receive codes |
| \- At least one SMS number is required | \- At least one SMS number is required |
| \- Codes are 6 digits | \- Codes are 4 digits |
| \- [Recovery is dynamic and **does not require a recovery key**](https://support.apple.com/en-us/HT204921) | \- [Recovery is static and **requires a recovery key**](https://support.apple.com/en-us/HT202649) |

[...]

PR #1579, opened 2016-02-03
Add Google Fit to Health

whatinthenameofgod:


mxxcon:
@whatinthenameofgod let's hold off on adding every possible Google service out there. Since they all use the same account, that's a lot of redundant data.
It's better to come up with some sort of unified and all-encompasing Google listing to address them all.

whatinthenameofgod:
It would be helpful to have one Google line item, but at the same time, that account encompasses a lot of products that I think should be listed categorically where relevant.

Is there a way to have a higher-level, all-encompassing Google account and then have the Play Store, Fit, Fiber, etc listed in their relevant categories and just referencing the Google account?

This kind of model could be used for all-encompasses services like Microsoft, Apple.

RichJeanes:
Well, unfortunately, that isn't a thing for Apple. See #975 about that. And I think there's a thing or two with Microsoft that isn't protected, too, isn't there?

jamcat22:
@RichJeanes Microsoft protects everything that uses a Microsoft account. The only two products that are possibly not protected by Microsoft 2FA are Skype (when using a Skype account to login or when using Facebook sign-in) and HealthVault (when using Facebook sign-in or when using an OpenID to login).

@whatinthenameofgod We're looking for a way to do this without switching from our YAML structure.

I'll merge this PR right now, until we can come up with a better solution in the future.

PR #1585, opened 2016-02-09
Add Liquid Web to Hosting

gregdotca:
Adding Liquid Web to Hosting section


mxxcon:
perhaps make URLs https?

gregdotca:
URLs updated to be secure

PR #1594, opened 2016-02-15
Removed Amazon's exception

Zian:
No need for a phone call any more. Per the documentation, "You can receive this security code in a variety of ways depending on the option you select during sign up, including text message, voice call, or authenticator app."

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -5,8 +5,6 @@ websites:
       tfa: Yes
       software: Yes
       sms: Yes
-      exceptions:
-          text: "SMS/Phone Call required for 2FA"
       doc: https://www.amazon.com/gp/help/customer/display.html?nodeId=201596330
 
     - name: Apple


mxxcon:
Is the info in #1433 not accurate?

Carlgo11:
I just went through the TFA setup on Amazon and #1433 very much still is accurate.

Zian:
The discussion around #1433 isn't wrong but it's misleading to say that 2FA requires an SMS or phone call because you don't have to use a telephone after initially setting it up. I just set up 2FA on Amazon and am able to use codes generated by the Google Authenticator application without requiring an additional phone call or SMS from Amazon.

mxxcon:
So then maybe rephrase the exception to more accurately represent the situation rather than remove it outright?

RichJeanes:
To me, your comment doesn't make sense.

> ...it's misleading to say that 2FA requires an SMS or phone call because you don't have to use a telephone after initially setting it up

If a text message or phone call is required to set up 2FA, then is it not required to use the service? Is that not the literal definition of a prerequisite? If I do not have a phone, then I cannot add 2FA to my account. If I cannot add 2FA to my account, then I cannot use 2FA for my account.
So how is it misleading to say that you must have a phone to use 2FA for this account?
Also the fact that if you are using Google Auth (or just about any other TOTP generator), it's probably on a mobile phone.

But for the sake of complete accuracy, yes, it could be re-worded to "SMS or phone call required to set up 2FA".

Issue #1582, opened 2016-02-06
Wells Fargo


sanjay:
Wells Fargo does support two factor authentication but apparently it isn't public yet.  I stumbled on this talking to them.  They call it Advanced Access Code and it works by doing phone calls or text messages to a phone or (apparently) using a RSA SecurID hardware token.  I don't know the details on the hardware token though.  To enable you have to call them and get them to enable this - there isn't any documentation online.

jamcat22:
Do you mean "Advanced Access"? They have a help document about it [here](https://www.wellsfargo.com/privacy-security/advanced-access).

That doesn't fit our definition of two factor authentication (or 2FA), as it only requires a security code during certain transactions. This is commonly referred to as two factor authorization, not authentication.

Please see our [note on definitions](/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions) in the CONTRIBUTING.MD file, issue #1114, and PR #1115.

I'm going to close this issue for now, unless they now have an option to require codes on login.

sanjay:
That is what I mean but this doc doesn't reflect what I've got.  I asked for this to be enabled for my account and now EVERY time I login, I have to enter the code they send in order to login.  That is 2FA in my book.

jamcat22:
Ah. That definitely changes things! Do you have a screenshot of Wells Fargo requiring 2FA during login?

sanjay:
Attached with the last 4 digits of my cell phone masked.  Note the tab for the RSA SecurID token.  I don't have one and I'm not sure how you add it to your account.  No support for Google Auth/Authy type codes unfortunately.

![screenshot at 2016-02-06 18 03 05](https://cloud.githubusercontent.com/assets/20825/12869943/1474371c-ccfc-11e5-9c02-c1937ef85490.png)

f1r-CTLF:
I'll add some additional info to this, because I have personally gone through the steps to setup true 2FA on Wells Fargo.

Steps:
1) Login
2) Navigate to the 'Account Services' tab.
![001_account_services_main_clean](https://cloud.githubusercontent.com/assets/15699955/13035856/11e1c560-d30e-11e5-90cd-f7daae456a90.png)
3) Under 'My Profile', select 'Manage Advanced Access'.
4) Setup mobile phone number or RSA SecureID Device.
[...]

PR #1595, opened 2016-02-15
Update Wells Fargo status

Carlgo11:
As per #1582.

It looks like they don't have any docs that are up to date so I left the doc tag out.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -432,6 +432,7 @@ websites:
       img: wellsfargo.png
       tfa: Yes
       sms: Yes
+      software: Yes
       phone: Yes
       hardware: Yes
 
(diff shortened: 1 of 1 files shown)


mxxcon:
But RSA SecureID has software tokens, can't those be used as well?

Carlgo11:
@mxxcon updated the PR.

mxxcon:
:+1: 
Maybe add info from PR1582 into a separate notes page the way we have for Chase, Fidelity and a couple of other sites?

Carlgo11:
Is that really necessary?

jamcat22:
I'm going to merge this, and create a doc page for it later on.

PR #1598, opened 2016-02-17
Update social.yml to include Instagram 2FA pending

stephengroat:


mxxcon:
Let's wait a day or two for it to be actually released so that we can add all the relevant info.
No point adding this url only to change it a few days later.

stephengroat:
No problem

PR #1601, opened 2016-02-17
Clarify Amazon exception

RichJeanes:
Addresses concerns mentioned in #1594


Eitot:
Can you put “Phone Call” in lowercase and end the sentence with a full stop?

Carlgo11:
I think it looks good. It's easier to read as it is.

PR #1605, opened 2016-02-22
Updated Guild Wars 2

Kezxo:
Guild Wars 2 supports both SMS and phone calls in addition to TOTP. I also changed the link to a support page that explains all methods, not just TOTP.

### DIFF
--- a/_data/gaming.yml
+++ b/_data/gaming.yml
@@ -81,10 +81,12 @@ websites:
     - name: Guild Wars 2
       url: https://www.guildwars2.com
       img: guildwars2.png
+      sms: Yes
       tfa: Yes
       email: Yes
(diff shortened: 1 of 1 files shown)


Carlgo11:
I don't see anything about phone calls.

Kezxo:
It's not very obvious but it is there, they just have it under SMS authentication for some reason. I guess they would prefer people using SMS?

> What types of authentication are available?
> 
> SMS Authentication
> 
> This type of authentication sends your security code as a text **or voice message to a cell phone or a land line**, but alternative services like Skype or Google Voice won’t work. See our article on [SMS Authentication](http://help.guildwars2.com/entries/95445077) for more information

PR #1606, opened 2016-02-22
remove Copy.com

tvierling:
Their service is shutting down on 1 May 2016.
https://techlib.barracuda.com/COPY/FAQ


RichJeanes:
Might I suggest an exception notifying people of the impending shut-down and waiting until May 1 to completely remove the listing?

tvierling:
If a service doesn't have 2FA today, and has announced a complete shutdown in the near future, what's the point of listing the service here anymore? It's a lame duck at this point, so I don't see what we gain by encouraging users to beg a soon-dead service for 2FA support.

Carlgo11:
You're very much right.
I don't think Rich noted that they didn't have TFA.

mxxcon:
Ya, no point keeping a soon to be dead entry.

RichJeanes:
Ah, that I did not. I guess I assumed they did for some reason.

PR #1608, opened 2016-02-22
Add LeaseWeb

tvierling:


mxxcon:
Um.. https://kb.leaseweb.com/display/KB/Two+factor+authentication%3A+Cyber+Security

tvierling:
Looks like that's available only for CDN customers; it's not yet available for leased virtual servers, at least. Leaving this req open pending a response from LeaseWeb Support and will force-push the branch with the correct answer (tooltip clarification if necessary). Should hear back within 24h.

tvierling:
Pull branch force-updated to include the 2FA available for some customers, with exception text.

mxxcon:
Perhaps instead of "_Currently_" put month/year so that people can judge how recent that exception is.

tvierling:
Done.

PR #1610, opened 2016-02-24
Add CallCentric, VoIP provider for home and business telephony.

tvierling:


Carlgo11:
They haven't updated their iPhone app since 2011 and haven't updated the android app since 2012.
That service doesn't look very active.

tvierling:
They support standard SIP clients, and SIP hard phones (and standard SIP clients are their real bread and butter, not their rather hokey "callback" mobile apps). I can assure you that Callcentric is very active; I have a half dozen DIDs through them.

mxxcon:
> To complement our "IP Whitelist" and "Allowed Calling Areas" security features we have now added a Two-Point Authentication system to help protect your account against unauthorized web login access.
> In addition to your standard login password, you can now choose to configure Challenge Questions or for even better protection activate our Security Key Authentication System.

"_Challenge questions_" are not considered another authentication factor. It is still "something you know" just like username/password.
What is that "_Security Key Authentication System_"?

tvierling:
The "Security Key" (I've already submitted a ticket to note that it's very badly worded) is a hashed numeric code (7-digit) e-mailed at login to the main address and/or an auth-specific e-mail address. The configuration page looks like this: http://i.imgur.com/I1w6IrI.png

jamcat22:
@tvierling That logo is a bit large. Could you use this one instead?
callcentric
For future reference, a simple tool to compress images without extra software is [TinyPNG](https://tinypng.com).

tvierling:
Wow... 2145 octets is large? Heh. I was trying to do a normal lossless transform (gimp export), but sure, whatever you like. Re-pushing. :)

Carlgo11:
@tvierling everything between 2-2.5KB is a bit of a gray area but we generally try and keep the size of the logos as small as possible.
We do this because loading 1500 images can be quite a heavy load for a user on a mobile network and the time difference between loading 1500 images that are between 4-5 KB vs 1-2 KB can be astoundingly big for users with bad internet connection.

tvierling:
Nah, I get the point, I'm just mildly amused. If the site grows enough, it might be time to consider sprite-izing all the images.

mxxcon:
With lazy-loading and http2 spriting is an anti-pattern. :)

tvierling:
@mxxcon Show me wide deployment of http2 on mobile (cf. @Carlgo11's previous comment ;) and I'll be all over that $#!+. heh.

mxxcon:
@tvierling https://blog.cloudflare.com/introducing-http2/ and http://caniuse.com/#feat=http2

tvierling:
I know what HTTP/2 is.

I was mainly pointing out that there's a sizable installed base of mobile devices running, for instance, Android with too little memory to run Firefox/Chrome for mobile (both of which are continuing to creep up even further in memory footprint), forced to stick with the AOSP webkit based browser. So if the goal is to "be nice" to low end network connections and low end mobile devices, HTTP/2 is not a panacea.

[edit: ..._yet_.]

PR #1621, opened 2016-03-04
Nexmo

AlessioCasco:
developer --> added nexmo


mxxcon:
How does this site fit in https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria ?

AlessioCasco:
Nexmo is one of the first operators for sms and voice over API, it deals with data, you use the 2fa to log in into your dashboard where you can find all the messages and calls you sent and received

Carlgo11:
No doc link?

AlessioCasco:
I had a look around but I don't see any doc or info specifically related to their dashboard & 2fa.
I can load a screenshot of the log-in page if you need a prove.

Carlgo11:
Yes a screenshot would be great.
(Blur out ant identifying information that might be visible in the screenshot)

AlessioCasco:
Screenshot attached

2fa_nexmo

_data/developer.yml
```diff
@@ -200,6 +200,12 @@ websites:
       img: newrelic.png
       tfa: No
 
+    - name: Nexmo
+      url: https://www.nexmo.com/
+      img: nexmo.png
+      tfa: Yes
+      sms: Yes
```
jamcat22, line 207: @AlessioCasco From the screenshot, it looks like a phone call is an available method of 2FA. Can you please add the line `phone: Yes` if this is the case?

AlessioCasco:
@jamcat22 You're right, the phone call is a method they do offer.
I updated the profile, 

Thanks

PR #1622, opened 2016-03-04
Add Cedexis

AlessioCasco:
domains --> added cedexis


mxxcon:
Where does this site fit with https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria ?
And regardless of that, I don't think Cedexis fits into domains registrars category.

AlessioCasco:
I put it there since is where other CDN's providers are.
Cedexis is not a proper CDN, it's more likely a load balance between them

Let me know if you want me to move it

mxxcon:
At first glace i see just CloudFlare listed there.
I think `developer` or `other` categories would be a better fit for cedexis.

AlessioCasco:
Ok will update it and move to Developer

PR #1623, opened 2016-03-04
Adds Nutmeg to finance.yml

AlessioCasco:
finance --> added Nutmeg


AlessioCasco:
Since it's a investment platform, would you like me to move this under Investing?

Issue #1534, opened 2016-01-18
Transportion companies


Carlgo11:
This is more a list for myself than anything but if anyone wants to add a company then feel free to edit this doc.

Transportation-related companies to add:
- [x] American Airlines
- [x] Amtrak
- [x] British Airlines
- [x] Norwegian
- [x] SAS
- [x] Air France
- [x] EasyJet
- [x] RyanAir
- [x] Lufthansa
- [x] DB

jamcat22:
Not all of these companies have been added. I'm going to reopen the issue, until all of the specified companies are added.

Carlgo11:
Thank you very much Jacob! :dollar:

Carlgo11:
All requests done! :smile:

PR #1628, opened 2016-03-07
Added easyjet

AlessioCasco:
Added easyjet as requested on issue #1534


_data/transport.yml
```diff
@@ -17,6 +17,12 @@ websites:
       img: ba.png
       tfa: No
 
+    - name: EasyJet
```
Carlgo11: easyJet seem to use a lowercase `e` on all platforms. Perhaps you could change the name to meet their standards?

PR #1633, opened 2016-03-08
added changetip

brandoncurtis:
added changetip to cryptocurrency provider list


Carlgo11:
I don't think cryptocurrencies is the best fitting category for changetip. How about finance?

brandoncurtis:
If you'd prefer!  With the imminent launch of ChangeTip Wallet for on-blockchain transactions I thought cryptocurrencies was slightly more appropriate, but I could see it either way.  Let me know.

To be honest I find the 'banking/finance/investing/payments/cryptocurrencies' divisions to be unhelpful—several times I was expecting to see something under one category, only to head back up to the filter bar to search manually and find it under some other category.

Carlgo11:
Hi again @brandoncurtis,
thank you for your feedback!
I do agree with you that there's sometimes a very thin line between the categories you mentioned.

I'll bring this up with the others and see what we can do to eliminate this inconvenience.

I also merged this PR after reading about the blockchain transactions so there's no need to update the category.

Thank you for your contribution!
Your changes will be live as soon as one of the core-members have validated that the recent changes looks good. You can read more about this process [here](https://github.com/2factorauth/twofactorauth/wiki/FAQ_Merge).

PR #1636, opened 2016-03-10
Add ICUK to hosting category

simonp22:


mxxcon:
Can you please explain why this qualifies as a utility?

simonp22:
ICUK is a provider of telecoms and broadband utilities, as well as hosting and domains.
If you think this would fit better in another category, let me know and I will correct this.

mxxcon:
I don't think this what people would expect from `utility` category. I think `hosting` or maybe even `other` would be a better fit for this.

simonp22:
Now moved ICUK to the hosting category

PR #1640, opened 2016-03-11
Cut out the white space from the favicon

Carlgo11:
Title says it all ^


mxxcon:
Changing file permissions?

Carlgo11:
@mxxcon Whops, I didn't look at the permissions before I pushed. It's fixed now in 000792e.

Carlgo11:
@psgs the reason I assigned you to the PR is because I'm not quite sure how the site loads the favicon but it seems not to care about any changes of the logo so I assume it somehow loads semantic-ui's default favicon directly from the css or something magical like that?

Anyways, I ran this code locally and couldn't find any errors after adding the icon line in the header so...  `¯\_(ツ)_/¯`

fpigerre:
Hello @Carlgo11 :smile: 

At the moment, the website's favicon is loaded from the [`favicon.ico`](https://github.com/2factorauth/twofactorauth/blob/master/favicon.ico) file in the project's root folder. Using a PNG file instead would probably be better, however, as it would allow for more flexibility. :+1:

PR #1646, opened 2016-03-13
Add Xapo

jamcat22:
Xapo is a Bitcoin wallet/debit card provider. They use SMS, Phone Call,
and Software 2FA.


RichJeanes:
I'm getting TLS certificate errors on their support page. Looks like missing chain certs. Anyone else see this?

mxxcon:
Rich, yes, incomplete chain https://www.ssllabs.com/ssltest/analyze.html?d=support.xapo.com

Carlgo11:
Well there's nothing we can do about that. The http page redirects to https.

PR #1647, opened 2016-03-13
Update LogMeIn

jamcat22:
LogMeIn also supports hardware and SMS as 2FA methods. Also updated the doc link and made TeamViewer links HTTPS.

### DIFF
--- a/_data/remote.yml
+++ b/_data/remote.yml
@@ -1,5 +1,4 @@
 websites:
-
     - name: Amazon AWS WorkSpaces
       url: https://aws.amazon.com/workspaces/
       doc: https://aws.amazon.com/blogs/aws/multi-factor-auth-for-workspaces/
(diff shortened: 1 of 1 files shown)


mxxcon:
I don't see where logmein talks about hardware..?

jamcat22:
I classify paper codes as hardware, because they're being used as the main second factor and not just as a backup.

RichJeanes:
Yeah, I would agree in this instance. If they allow you to use paper codes as your only form of 2FA, then I would list it. This is in contrast to something like Google's paper backup codes which cannot be used without having another form of 2FA enabled first. It certainly wouldn't be the first site marked for hardware because of paper codes.

mxxcon:
For some reason i read paper codes as recovery, not day-to-day use.

PR #1654, opened 2016-03-15
Add Virgin Mobile

adeishs:
Adding Virgin Mobile Australia.


mxxcon:
It seems like none of virgin mobile services offer tfa, not just Australian version. I think you might as well just make it a global entry with https://www.virgin.com/gateways/mobile url?

adeishs:
OK, but what should I use for the twitter account name? There’s [virgin](http://twitter.com/virgin), but it’s not specific for their mobile services. They have country-specific ones for mobile services: [virginmobileusa](http://twitter.com/virginmobileusa), [virginmobileaus](http://twitter.com/virginmobileaus), [VirginMobileFr](http://twitter.com/VirginMobileFr), etc.

mxxcon:
I think use virgin.. it's higher profile, so hopefully more people will see it.

adeishs:
OK. I take your suggestion, @mxxcon. Done.

PR #1662, opened 2016-03-18
Add Cloudytics.com

NotAmaan:


Carlgo11:
I don't think this site is popular enough to be listed here.
Furthermore it looks like they use cPanel and no own control panel.

NotAmaan:
@Carlgo11 It is indeed not so popular, wasn't sure if that was an requirement but its fine. Also, I'm their designer & its control panel so its under works, but yeah, the current stage is a MVP.

jamcat22:
@NotAmaan We usually only take into account popularity when adding a site that doesn't have 2FA, because if we didn't, we probably would need to list almost every website in existence. If the new control panel you're building will have 2FA, feel free to make another PR after it's released.

NotAmaan:
@jamcat22 Will sure do. Thanks!

Issue #1489, opened 2016-01-05
Add incapsula.com


TraderStf:
incapsula.com
Email, GA, SMS

jamcat22:
Hello @TraderStf! If you'd like to, you can make a pull request (PR) for this website. Just follow the instructions on the [wiki](https://github.com/2factorauth/twofactorauth/wiki/FAQ_New_TFA_SITE).

For future reference, it looks like the best `doc` link is: https://incapsula.zendesk.com/hc/en-us/articles/204795664-Incapsula-s-Password-Policy

TraderStf:
nope :) not enough time, sorry,don't blame me, it's my wife
few more
http://help.seagm.com/customer/portal/articles/571328-what-is-2fa-pin-and-how-to-set-it-

jamcat22:
That's ok @TraderStf. At some point in the future, another collaborator will make a PR for Incapsula.

SEAGM's implementation looks to be just a second password, not another factor of authentication. If this site is added in the future, it will be marked as not supporting 2FA.

TraderStf:
zonomi.com free dns host
TOPT (google auth)
https://zonomi.com/app/cp/twofactor.jsp

Zonomi lets you enable two-factor authentication on your account by requiring a password on login, as normal, plus an RFC6238 compliant Time-based One Time Password (TOTP) token.

mxxcon:
@TraderStf please open a separate issue for each site otherwise things will get lost.

PR #1663, opened 2016-03-19
Add Incapsula

Carlgo11:
As per #1489 
I decided domains was the best category as Cloudflare is in that category and their services are pretty much identical.


jamcat22:
@Carlgo11: That doc link talks about their ability to provide 2FA to websites, not how to enable 2FA for logging in to an account. This doc link explains the process of enabling 2FA for logging in to an account: https://incapsula.zendesk.com/hc/en-us/articles/204795664-Incapsula-s-Password-Policy

Carlgo11:
@jamcat22 updated

PR #1667, opened 2016-03-20
Update Code Climate doc URL

StuBez:


Carlgo11:
They seem to support SMS as well. Could you add that?

StuBez:
@Carlgo11 Good spot, I hadn't event noticed.

PR #1675, opened 2016-03-24
Added Thexyz, now with logo!

ptoone:


RichJeanes:
Looks like the image got put into a folder called "["img", "email"]" in the root of the project. Try uploading it to the proper directory in your for with the proper branch selected and deleting that new folder.

PR #1676, opened 2016-03-26
Add language tag handling for tweets

fpigerre:
Hello!

This pull request allows default tweet text to be altered based on a language tag, configurable per website. Language tags are placed under each website within the appropriate data file, and language codes use the [ISO 639-1 standard](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) in lower-case.

Thankyou,
psgs :palm_tree:

### DIFF
--- a/_data/languages.yml
+++ b/_data/languages.yml
@@ -1,0 +1,5 @@
+# Languages should be categorized according to ISO 639-1
+# Language codes (including within data files) should be in lowercase
+en:
+  work_tweet: Security is important, @TWITTERHANDLE. We'd like it if you supported two factor auth.
+  progress_tweet: Thanks for working on support for two factor auth, @TWITTERHANDLE!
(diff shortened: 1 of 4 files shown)


Carlgo11:
Is ISO 639-1 or ISO 3166-2 better?

fpigerre:
In this case, ISO 639-1 is probably more appropriate, as the websites listed on twofactorauth.org, as well as any future translations of twofactorauth.org itself are fairly limited in linguistic scope. Furthermore, for some websites, it could be difficult to distinguish and apply a particular region to the form of language they are using, taking into account that the type of language being used may not necessarily correspond with the region in which the website is based.
Also, ISO 639-1 generally corresponds with the abbreviations for languages commonly used by many people in daily life, making it easier for contributors to use the correct language code.

jamcat22:
@psgs, the default appears to be English, but could you give an example of how to specify a site's language in a data file?

Carlgo11:
@jamcat22 

```
- name:
   url: example.com
   img: example.png
   tfa: No
   twitter: example
   lang: de
```

Changes the tweet language to German

jamcat22:
Awesome! Just wanted to be sure, @carlgo11!

PR #1678, opened 2016-03-27
Added Thexyz

ptoone:
Took some time but I think (I hope) I have it right this time.


Carlgo11:
@thexyz you've created 5 Pull Requests in the last 3 days. We'd appreciate if you keep it to one PR in the future.

ptoone:
Okay thanks. Still learning GitHub.

Carlgo11:
You're welcome. Now to the actual PR.
You say Thexyz supports TFA in the form of Software, SMS and Hardware but the doc link only mentions the Yubikey which is a Hardware solution.

ptoone:
Yes, it does actually support SMS and authenticator apps. We will the other options to Yubikey page.

ptoone:
Only the password reset supports SMS so I set that to No. I also added links to use software apps for authentication instead of YubiKey.

RichJeanes:
In regards to the PR:
1. Tags should not be set to no, they should just be removed.
2. Instead of linking to pages discussing what 2FA is or how it works, we are looking for direct links to documentation of how to enable it. So I would recommend using https://www.thexyz.com/forum/discussion/2219/how-to-add-yubikey-authentication-to-thexyz Since that is what is linked to for instructions. Unless there is a different page that is on the support site, instead of the forums, that also has instructions for software 2FA setup?

In regards to your YubiKey support, is there a specific reason that you host your own YubiKey authentication server instead of using theirs? It would make setup easier for the user since they wouldn't have to go out of their way to generate an API key. Also, looking into switching to U2F instead of YubiKey Auth?

ptoone:
Those changes have been made. We do not host our own authentication server, it is done through Yubico. The API key is for first time users of YubiKey.

RichJeanes:
YubiKeys come with YubiKey Auth preprogrammed at the factory. I didn't realize you had to generate a key for first-time setup. I played with it so much when I first got mine that I don't remember all I had to do to actually use it in a normal fashion.

Carlgo11:
While Rich's complaint on the doc link is valid I'd like to see a doc page that takes up all the TFA factors you provide.

P.S. The old doc link says "No hardware or software needed" while Yubikey very much is hardware. You might want to change that for your customers. I don't know who's written that page but they clearly don't know much about the Yubikey or Yubico company.

RichJeanes:
^^

> Unless there is a different page that is on the support site... that also has instructions for software 2FA setup?

ptoone:
I just checked a new YubiKey and you can start using it right away. That step has been removed. 

The line that said "hardware" has been removed. On the YubiKey page there is an FAQ listing the different methods. I also merged the guide into one page on our knowledgebase - https://www.thexyz.com/account/knowledgebase/15/Enabling-two-factor-authentication.html

Carlgo11:
@thexyz Please use https://www.thexyz.com/account/knowledgebase/15/Enabling-two-factor-authentication.html as the doc link

ptoone:
That has been changed.

Carlgo11:
_(Note to other 2factorauth members: Remember to use the squash feature when merging)_

ptoone:
Thanks for help your help everyone. Looking forward to seeing Thexyz added. Thank you.

Carlgo11:
You're welcome @thexyz.
Your contribution is now live on twofactorauth.org. Congrats :smile:

PR #1685, opened 2016-03-31
add Mail.Ru

hepler:
Add Mail.Ru to list of email providers.


RichJeanes:
Might I suggest linking to their help page that gives instructions to enable 2FA instead of their general info about it?
I found this link: https://help.mail.ru/enmail-help/settings/2auth

hepler:
Ah, good call! Updated.

PR #1689, opened 2016-04-03
add Pinterest

hepler:
Add Pinterest to social section


fpigerre:
Thanks for the pull request @hepler! :+1:

Carlgo11:
while we usually don't like to add a twitter account that isn't monitored by the company there doesn't seem to be any other twitter account that they monitor.
Thanks for the PR @hepler!

PR #1690, opened 2016-04-04
Add an index page for notes

fpigerre:
Hello!

This pull request adds an index page to the notes section of twofactorauth.org.
When the site is run, the index page would appear at `twofactorauth.org/notes/` and looks similar to the following screenshot:

notes

Thankyou,
psgs :palm_tree:


jamcat22:
:shipit:

PR #1693, opened 2016-04-07
adding charles schwab bank

ashtonian:


jamcat22:
Do they use the same login system as their investment branch? We already have them listed under "Investing".

ashtonian:
Ahh yes - duh and they are more of a investing company then a banking company. So I apologize. 

Probably remove this branch and I'll make another PR to remove the broken link and add an exception note.

jamcat22:
It's ok. I'll close this PR.

PR #1704, opened 2016-04-20
Playstation Network: In Progress

TheSeg:
Official representative[ confirming in-progress development for TFA](http://www.polygon.com/2016/4/20/11470660/psn-two-factor-authentication-playstation-network) on PlayStation Network.

### DIFF
--- a/_data/gaming.yml
+++ b/_data/gaming.yml
@@ -142,6 +142,7 @@ websites:
       twitter: AskPlayStation
       img: psn.png
       tfa: No
+      status: http://www.polygon.com/2016/4/20/11470660/psn-two-factor-authentication-playstation-network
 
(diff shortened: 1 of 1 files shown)


RichJeanes:
> more details will be shared at a later date

We prefer to have something of a timeline for the feature before marking as In Progress. We've seen a lot of sites that say they are "working on it" and it never comes to fruition.

jamcat22:
I think this is a different situation @richjeanes, especially since there's proof already present in their software. https://mobile.twitter.com/tontsa/status/722664713113485312/photo/1

RichJeanes:
Okay. If others feel there is enough justification for updating the entry, then I certainly won't fight it.

PR #1734, opened 2016-05-09
Paypal exception

Carlgo11:
Instead of #1696


jamcat22:
Cross referencing with #1026.

PR #1758, opened 2016-05-26
Add United Airlines

stephengroat:
While United "plans" to add tfa (https://www.united.com/web/en-US/content/help/PIN-password.aspx), they have not yet


Carlgo11:
How about this one instead? It's 61% smaller.
united

stephengroat:
perfect. what tool are you using? i was using the compress in gimp, but i
can use tinypng or something else if it gives a lot better results

Carlgo11:
I use [tinypng](https://tinypng.com/).

fpigerre:
Thanks for the pull request @stephengroat! :+1:

PR #1763, opened 2016-05-28
Add Swedish language

Carlgo11:
Added option for Swedish `lang` tag.
Also changed the site url to https instead of http.


jamcat22:
If we change the site URL to HTTPS, can it still be served over HTTP? Right now, we only force HTTPS on browsers that support HSTS, because there are many browsers that can't handle our CloudFlare SNI certificate. I didn't think it was right to block people with older OSes (like Windows XP) from trying to secure their online accounts.

Carlgo11:
@jamcat22 Removed that line.

PR #1773, opened 2016-05-31
Add Jet

stephengroat:


_data/retail.yml
```diff
@@ -54,6 +54,12 @@ websites:
       img: gilt.png
       tfa: No
 
+    - name: Jet
+      url: https://www.jet.com
```
Carlgo11: Looks like they actually don't use the www suffix. Please change the url to `https://jet.com`

stephengroat:
Fixed

PR #1774, opened 2016-05-31
Add Massdrop

stephengroat:


Carlgo11:
I see you're on a PR streak today. Keep up the good work! :smile:

So would you care to explain a little to me what Massdrop is?
It doesn't have to be an essay. Just a line or two to explain the core function of the site and why it needs TFA/2FA.

//Carl

stephengroat:
Massdrop is a retail site that gathers a large group of buyers to offer volume discount on specific products voted on by its members. They store credit cards & addresses for shipment that can be re-used for purchases.

PR #1776, opened 2016-05-31
Ensure tweets only use available languages

fpigerre:
Hello!

This pull request fixes an issue where using a lang tag that wasn't present in the `languages.yml` file
would result in a partially empty tweet.

Thankyou,
psgs :palm_tree:


Carlgo11:
Just tested it locally. Works good! :smiley:

PR #1783, opened 2016-06-01
Add Dominion

stephengroat:


jamcat22:
Their logo is almost impossible to crop down to a square, but try using this one instead:
dom

Also, while it seems like both their @DomVAPower and @DomEastOhio accounts reply to questions, do you think that their Virginia division is the best one to ask?

stephengroat:
Logo updated.

I've from VA, so my bias is pretty obvious here. I'm willing to change it either way!

jamcat22:
I think the VA account is fine. Thanks for changing the image!

fpigerre:
Thanks for the pull request @stephengroat! :+1:

PR #1785, opened 2016-06-01
Add ConEd

stephengroat:


jamcat22:
Hello @stephengroat!
It looks like all of their branding materials use the name Con Edison, so please change the name line to that.
Also, please use the image below, which crops out the name so that the logo is a bit more readable. 
coned
Thanks!

stephengroat:
Updated both name and image

jamcat22:
Great! Thanks!

Issue #1744, opened 2016-05-17
New add suggestions


ptretter:
| Name | URL | Category | TFA | Twitter | Image | Added |
| --- | --- | --- | --- | --- | --- | --- |
| MedicAlert Canada | http://www.medicalert.ca | Health | No | @medicalertCA | ![icon](https://cloud.githubusercontent.com/assets/3535780/15413384/016d78d8-1e30-11e6-98de-eb65847f8db8.png) | Yes |
| Meridian Credit Union | http://www.meridiancu.ca | Banking | No | @meridiancu | ![icon](https://cloud.githubusercontent.com/assets/3535780/15413444/8827aa2e-1e30-11e6-8637-07a687a4d240.png) | Yes |
| President's Choice Financial | http://www.pcfinancial.ca/ | Finance? | No | @pcfinancial | ![3apxgwso](https://cloud.githubusercontent.com/assets/3535780/15413681/21617718-1e33-11e6-943b-d0c71298928e.png) | Yes |
| Royal Roads University | http://www.royalroads.ca/ | Education | No | @royalroads | ![5357801541](https://cloud.githubusercontent.com/assets/3535780/15413644/b7b635ba-1e32-11e6-8ae3-e49f6f194d2e.png) | Yes |
| BlueFur Hosting | https://www.bluefur.com/ | Hosting | No | @bluefur | ![icon](https://cloud.githubusercontent.com/assets/3535780/15413542/9339b4ec-1e31-11e6-8200-0ab7303fc81b.png) | Yes |
| State Farm Canada | http://www.statefarm.ca | Insurance Company? | No | @StateFarmCanada | ![ezlke1j3](https://cloud.githubusercontent.com/assets/3535780/15413599/358878dc-1e32-11e6-95b9-b34162f91c3b.png) | N/A |
| Hudson's Bay Company | http://www.thebay.com | Retail | No | @hudsonsbay | ![lcb6mcn7 copy 1](https://cloud.githubusercontent.com/assets/3535780/15413739/b5f8f5c2-1e33-11e6-9067-839f0290c7ac.png) | Yes |
| Flipboard | http://www.flipboard.com | Social | No | @flipboard | ![osigabvn](https://cloud.githubusercontent.com/assets/3535780/15413490/f392ad90-1e30-11e6-9cc5-7b9f81df7833.png) | Yes |
[...]

PR #1788, opened 2016-06-02
Add BlueFur

stephengroat:
#1744


stephengroat:
Gotta show the Canadians some TFA love too

PR #1790, opened 2016-06-03
Updated Discord

RichJeanes:
Discord has added TOTP 2FA. They also have added the option to require all server admins to have 2FA enabled if the server owner chooses. Awesome!

https://blog.discordapp.com/keeping-discord-safe-and-sound/

### DIFF
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -21,9 +21,10 @@ websites:
 
     - name: Discord
       url: https://discordapp.com/
-      twitter: discordapp
       img: discord.png
-      tfa: No
+      tfa: Yes
+      software: Yes
+      doc: https://support.discordapp.com/hc/en-us/articles/219576828
 
     - name: Disqus
       url: https://disqus.com


RichJeanes:
Also, this is why I love Discord:
[emphasis theirs]

> You'll see a new pop-up with the option to verify via phone call or text message. Internal testing has yielded results that claim that the most recent smartphones **are in fact** capable of making and receiving phone calls, despite how rare this phenomenon appears.

They have such a great sense of humor. Really gives the app that "By gamers, for gamers" feel that Razer tries to claim.

b1naryth1ef:
Thanks for PRing this so quick! You beat even us to it :)

RichJeanes:
You're welcome! Thanks for providing such an awesome service! I used to run my own Mumble server at home, but have since ditched it for Discord! I enabled 2FA as soon as I got the notification that it was available.

jhgg:
Just out of curiousity, what do you consider for "email" second factor? We also do an e-mail auth check if you've logged on from an IP we don't know of before (by default).

jamcat22:
@jhgg Is it possible to turn the feature off, and does it still work even after enabling software 2FA?

In addition, does the feature protect every access point to an account?

jhgg:
It's turned off when you enable software 2fa. You cannot log in from an IP that the system hasn't seen you log in or connect from. Meaning, you can't do anything on the account.

PR #1801, opened 2016-06-06
Adding Square Cash

p4gs:
Square Cash currently doesn't have proper two-factor authentication. By default single-factor authentication occurs using a 6 digit code that's sent via SMS or email to the user. An additional knowledge factor, debit card CVV code, is required for some transactions (when I tested three transactions, the first didn't ask for CVV, the second did, and the third didn't). Users can optionally create a 4-digit passcode, which is only required when a user sends money with their account or change security options on their account - it's not required to sign in. Also TouchID can be used on the iOS app instead of this passcode.

### DIFF
--- a/_data/payments.yml
+++ b/_data/payments.yml
@@ -120,6 +120,11 @@ websites:
       exceptions:
         text: "Currently only US merchants, protects only Dashboard access"
       doc: https://squareup.com/help/us/en/article/5593-2-step-verification
+      
+    - name: Square Cash
+      url: https://cash.me
+      twitter: SquareCash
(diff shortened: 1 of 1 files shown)


RichJeanes:
All sites require a logo. Can you add one, please? Just place the optimized .png in the appropriate `img/category` folder, then add the file name to the site entry with the `img:` tag.

You can check out the [Wiki](https://github.com/2factorauth/twofactorauth/wiki/FAQ_New_SITE) if you need more information.

stephengroat:
Here's a properly sized image to help with!

squarecash

PR #1802, opened 2016-06-06
Adding LabTech

guygregory:


mxxcon:
From the documentation link, it seems like this is a self-hosted software product. Am I correct? Generally this site does not list such.

guygregory:
It's offered as either self-hosted (LabTech) or SaaS (LabTech Cloud). The documentation is the same in both cases.

guygregory:
Either way, the actual software used (LabTech Control Center) is installed on a local PC. If you need any more info to help you decide suitability, please let me know.

_data/remote.yml
```diff
@@ -15,6 +15,14 @@ websites:
       software: Yes
       doc: https://secure.logmein.com/welcome/webhelp/EN/Pro/LogMeIn/c_common_Security_TwoStep.html
 
+    - name: LabTech Control Center
+      url: http://www.labtechsoftware.com/
```
Carlgo11: Seems like they support https. Please change the URL to `https://www.labtechsoftware.com/` :wink:

_data/remote.yml
```diff
@@ -15,6 +15,14 @@ websites:
...
+      img: labtech.png
+      tfa: Yes
+      software: Yes
+      hardware: Yes
```
Carlgo11, line 23: I can't find anything relating to hardware based authentication on the doc page or on AuthAnvil's official page.
Maybe you could point me in the right direction?

guygregory, line 23: https://www.scorpionsoft.com/tokens

RichJeanes:
Is there any documentation for end users to enable 2FA on their accounts and/or a page that actually lists the available types of 2FA? 

The currently linked documentation says

> Your employees will input their chosen PIN and then a one-time is generated from an AuthAnvil app on their smartphone

Is there any documentation of hardware 2FA support?

PR #1805, opened 2016-06-06
Add weibo

stephengroat:
Could you some other eyes on this. My chinese reading is subpar

The accounts have the idea of a "Bound phone," but I can't tell if it's two factor or single factor

No twitter account as they are a twitter competitor


Carlgo11:
Can't say I'm much better at Chinese. Are you aware of any support email address or a facebook page?
I added `email_address` and `facebook` tags not too long ago for sites that don't have a twitter account to turn to (or don't respond on twitter).

See of it as a last alternative to the `Twitter` tag.

stephengroat:
Do you have an example of the facebook? their page is https://www.facebook.com/chinatweet. A little bit of twitter competition

Carlgo11:
That looks like a spam account. Facebook handles are added like this:
`facebook: <Facebook_Handle>`
So for example, if there was a facebook page for TwoFactorAuth.org  the URL would probably be `https://facebook.com/twofactorauth/` and so the facebook tag would be `facebook: twofactorauth`

PR #1810, opened 2016-06-11
Change http URLs to https

Carlgo11:
Hello y'all!
I built a very simple PHP script to check for http links found in data.json. It then replaced `http://` with `https://` and checked if the site returned a `200` response.
That method has a lot of problems when it comes to redirects and non-working URLs but these are the sites that responded with the exact 200 code.

**TL;DR** I did a strict check on all http links to see if the https version existed.

All links updated have been double checked by me.

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -17,14 +17,14 @@ websites:
     doc: http://support.apple.com/kb/ht5570
 
   - name: Backblaze
-    url: http://www.backblaze.com
+    url: https://www.backblaze.com
     img: backblaze.png
     tfa: Yes
     sms: Yes
     doc: https://www.backblaze.com/blog/two-factor-verification-for-backblaze/
 
   - name: Bitcasa
-    url: http://www.bitcasa.com/
+    url: https://www.bitcasa.com/
     twitter: Bitcasa
     img: bitcasa.png
     tfa: No
@@ -37,7 +37,7 @@ websites:
     doc: https://support.box.com/hc/en-us/articles/200526658-Can-I-enable-2-step-verification-for-my-account-
 
   - name: CloudApp
-    url: http://www.getcloudapp.com/
+    url: https://www.getcloudapp.com/
     twitter: cloudapp
     img: cloudapp.png
     tfa: No
@@ -63,7 +63,7 @@ websites:
     tfa: Yes
     sms: Yes
     software: Yes
-    doc: http://blog.evernote.com/blog/2013/10/04/two-step-verification-available-to-all-users/
+    doc: https://blog.evernote.com/blog/2013/10/04/two-step-verification-available-to-all-users/
 
   - name: FileThis
     url: https://filethis.com
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -138,7 +138,7 @@ websites:
       tfa: Yes
       software: Yes
       sms: Yes
-      doc: http://www.citibank.com.au/otp/
+      doc: https://www.citibank.com.au/otp/
 
     - name: Citizens Bank
       url: https://www.citizensbank.com
@@ -201,7 +201,7 @@ websites:
       tfa: Yes
       hardware: Yes
       software: Yes
-      doc: http://www1.firstdirect.com/1/2/securekey/
+      doc: https://www1.firstdirect.com/1/2/securekey/
 
     - name: First Republic Bank
       url: https://www.firstrepublic.com/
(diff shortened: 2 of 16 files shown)


fpigerre:
This is awesome! :+1:

Carlgo11:
Thank you @psgs :smile: 
I've thought about doing it for a while now. Hopefully I'll get around to converting it to a ruby script so that we can implement it in the travis-ci checks.

mxxcon:
I have a concern. Just because it gave you `200` status code, doesn't mean the returned content was the same as plain http site.. Some sites can (incompetently) serve wrong/different content on https url...

Carlgo11:
@mxxcon, true but that's why I double checked the sites

jamcat22:
@mxxcon: I rechecked all the urls for both their HTTP and HTTPS versions, and almost all of them returned the same content for each page. 
I also double checked that none of the sites had sketchy certificates or mixed-content errors. The results are shown below, in addition to some general mistakes I found along the way. @Carlgo11: I rechecked every URL, and wrote each issue I encountered below. 
How should we deal with mixed-content errors, even if the site loads without needing to create a security exception or manually allow the mixed-content to load?

### Issues found with websites in this PR, that don't necessarily need to be fixed:
- [ ] Dyn's HTTPS URL for their main page returns a mixed-content error, but the website still loads correctly without further user input.
- [ ] TIDAL's HTTPS URL for their main page returns a mixed-content error, but the website still loads correctly without further user input. Disabling mixed-content protection causes the website to redirect to HTTP.
- [ ] Nintendo's HTTPS URL for their main page returns many mixed-content errors, but the website still loads correctly without further user input.
- [ ] Misfit's HTTPS URL for their main page returns many mixed-content errors, but the website still loads correctly without further user input. Disabling mixed-content protection yields many `GET` requests for various images over an insecure connection.
- MyFitnessPal's HTTPS URL for their main page returns a mixed-content error, however their website automatically redirects to HTTPS, so this is a non-issue. Disabling mixed-content protection yields many `GET` requests to AWS CloudFront.
- [ ] Softlayer's HTTPS URL for their doc page returns a mixed-content error, but the website still loads correctly without further user input.
- [ ] Pacific Gas and Electric Company's HTTPS URL for their main page redirects to their HTTP URL. 

### Issues found with websites in this PR, that should be fixed in this PR:
- [x] Fabulous's HTTPS URL for their main page returns a mixed-content error and an SHA1 certificate, but the website still loads correctly without further user input.
- [x] Desura's HTTPS URL for their main page returns a mixed-content error and a completely different website when the error is bypassed.
- [x] Trion Worlds' HTTPS URL for their main page returns a mixed-content error, rendering the site almost unusable without creating an exception. In addition, the page tries to redirect to a non-HTTPS version on first load. The non-WWW version of their HTTPS page returns an invalid certificate error. The HTTPS version of this website is completely unusable on browsers (such as iOS Safari) where bypassing mixed-content protection is not allowed.
- [x] Dome9 doesn't use `www.` in either its main or doc URLs, including for the HTTPS variants.

### Issues to fix in 7-8 other PRs, after this PR is merged, but before `gh-pages` is merged with `master`:
- [ ] The Evernote doc link should be changed to https://help.evernote.com/hc/en-us/articles/208314238.
- [x] SEB's doc link is broken.
- [x] TBC's doc link is broken. It should be changed to http://www.tbcbank.ge/web/en/web/guest/tbc-digipass.
- [ ] Sendloop's URL for their main page should be changed to https://sendloop.com.
- [x] Trion Worlds supports email 2FA by default. A doc link for it is available here.
- [x] Trion Worlds' doc URL should either be changed to https://support.trionworlds.com/hc/en-us/sections/200487368-General-Information as a general link, or https://support.trionworlds.com/hc/en-us/articles/204183357-Adding-a-Mobile-Authenticator as a link specifically for their software authenticator.
- [ ] Dome9's doc URL should be changed to http://support.dome9.com/knowledgebase/articles/399400-totp-mfa-google-authenticator-7-29-2014.
- [ ] Pacific Gas and Electric Company should have the ending `PG&E` removed from the name line, as this is redundant.

Carlgo11:
@jamcat22 when it comes to mixed content I think we should use the https URL as long as the website works somewhat like the http version. I think the maintainers of the websites will eventually come around to fix the mixed content issues when enough people use https (As in people don't use IE anymore.)

If you find a better suited URL than the one already listed I suggest we make another separate PR for that.
I've marked the subjects you brought up that I think should be brought up in another PR.

The SHA-1 certificate for Fabulous makes me think we shouldn't use their HTTPS link as it will be blocked by some browsers.

Will push a new commit soon.

jamcat22:
@Carlgo11 I've updated my comment with different sections to address issues that don't necessarily need to be fixed, should be fixed in this PR, or should be fixed in another PR.

jamcat22:
Thanks for making those changes @Carlgo11! :smile:

Issue #1813, opened 2016-06-14
Make "Tell them to support 2FA" tweets more constructive


troy:
The current "Tell them to support 2FA" button, which sends a tweet, has a few problems. It:
- implicitly assumes that Web apps have either never heard of 2FA or would benefit from one more otherwise-unrecognized person saying it would be valuable. For nearly all Web apps, that's not the case. Either (a) they've considered 2FA and prioritized it (possibly based on information that the user isn't aware of, like how hard it is to handle edge cases like lost 2nd factor devices), or (b) they're not familiar with 2FA and are so clueless as to be a lost cause (and the person reading their tweets isn't going to change that).
- encourages thinking of the Web app operators as enemies (a "them") to be talked at, rather than as collaborators who are equally concerned about security. The current flow encourages fire-and-forget one-way tweets, not useful two-way conversations.
- puts quantity above quality, because people click many of the buttons - including many services they don't personally use. Example screenshot below. If getting 2FA was as simple as mass-sending the same tweet to every Web service, it would already be implemented 😄  A user sending 25 or 50 canned tweets is the advocacy equivalent of a stereotypically rant-y comment on a local newspaper's comment section: it's someone making themselves feel better.
- omits what would actually be useful: a user's own personal story why 2FA in that app would be particularly helpful to them. In any even halfway-clued company, someone has considered or is considering 2FA. However, it's possible that the product manager is underestimating the impact or overestimating the work, and personal stories, not form-letter tweets, are what would reveal miscalculations ("Oh, I never considered that use case").

So, proposed improvements:
- change the button label and default tweet text to something which invites a two-way dialog. For example, a "Learn more about their plans" button which drafts "Hey, I noticed Service doesn't support 2FA. Could you tell me a little about your thoughts on 2FA?"
- discourage clicking all the buttons. Possible implementations: cap the # of uses of the button per browsing session and/or require the user to type a portion of each tweet (like finishing a sentence) and/or change the button to say "I use Service" so it's clear that the button is meant to be used a few times, not 25 times
- encourage personal anecdotes. This could be an empty portion of a sentence that the user is expected to complete: "I use Service and.." Of the different ways to trade less quantity for more quality, this is my favorite because anecdotes (a) have more impact than canned tweets ever can, and (b) share information (use cases) which might make the Web app re-examine how they prioritized 2FA.

Here's an example of how the current button gets used. This is one of many pages of tweets from the same person:

[...]

Issue #1212, opened 2015-07-11
Different means of contacting the sites not supporting TFA


Carlgo11:
Right now we only support Twitter to ask companies to support TFA. Maybe we could add Facebook & Email as well since some sites don't use Twitter or don't listen to tweets mentioning them.

mxxcon:
Is there a way to create ready-to-use facebook post?

Carlgo11:
Nope. But we could link them to the FB page and then they could tell them themselves.

dullroar:
I would add to this that I actually think using social media to do this is wrong, wrong, wrong from a security point of view. Think about it - you are basically telling the world (unless these things are sent as private messages, but it doesn't look that way) that you have an account at a site that doesn't support 2FA. So if the bad guys can figure out who you are from your Twitter handle (usually not hard) they're already have another interesting piece of info about you - that you bank at Poorly Protected Bank & Trust, which opens up things to social engineering as well as just trying to figure out your account and password.

RichJeanes:
That's a bit of a fallacy. You run that same risk every time you hand your debit or credit card to a waiter/waitress/cashier. More so, in fact, because they could then skim your card and get more than just your name and bank.

dullroar:
No, that argument is a red herring. Because I am handing it off to one person at a time, not the whole world with a searchable hash tag. The point is to not needlessly increase the attack surface, as opposed to getting bogged down in the whole, "Well, you're exposed over here anyway" argument. Because if that's the case, why use 2FA at all, if we're f----ed already?

dullroar:
To avoid getting into an argument, I would rather that notifications to orgs that don't support 2FA be made by private IM, not public tweet.

RichJeanes:
If you are concerned about that, then feel free to message them privately. While using a public medium may increase your attack surface (though to what degree is arguable), the point of making the statement publicly is to draw attention to the issue and spread awareness. A lot more people use Twitter than may necessarily know about twofactorauth.org which means the message can reach a grater audience and get more people contacting the service about providing TFA. How anyone wishes to display their support, whether publicly or privately, is completely up to them. We don't force anyone to use Twitter, though, yes, we do encourage it through including the link to try to spread awareness.

mxxcon:
@dullroar Part of it is intentional public shaming. To make a public statement that everybody can see that such-and-such doesn't support 2FA.
[...]

PR #1816, opened 2016-06-17
Update tweet link text and tweet (#1813)

troy:
See #1813. This:
- makes the tweet more constructive, and updates the link label to reflect that
- lets the user finish the tweet with why they want it or how they'd use it
- removes the `@TWITTERHANDLE` fake personalization, which is (a) cheesy when repeated in many tweets (looks like a spambot) and (b) a waste of characters now, since the changes above make it much more personal
- updates the translations

### DIFF
--- a/_data/languages.yml
+++ b/_data/languages.yml
@@ -1,11 +1,11 @@
 # Languages should be categorized according to ISO 639-1
 # Language codes (including within data files) should be in lowercase
 de:
-  work_tweet: Hallo! Was sind Ihre Gedanken auf Zwei-Faktor-Authentifizierung? Ich würde es gerne, weil ich
+  work_tweet: Hallo @TWITTERHANDLE! Was sind Ihre Gedanken auf Zwei-Faktor-Authentifizierung? Ich würde es gerne, weil ich
   progress_tweet: Vielen Dank für die Fortschritte bei der Zwei-Faktor-Authentifizierung, @TWITTERHANDLE!
 en:
-  work_tweet: Hi! What are your thoughts on two factor auth? I'd like it because I
+  work_tweet: Hi @TWITTERHANDLE! What are your thoughts on two factor auth? I'd like it because I
   progress_tweet: Thanks for working on support for two factor auth, @TWITTERHANDLE!
 sv:
-  work_tweet: Hej! Vad är dina tankar om två faktor auth? Jag skulle vilja det eftersom jag
+  work_tweet: Hej @TWITTERHANDLE! Vad är dina tankar om två faktor auth? Jag skulle vilja det eftersom jag
   progress_tweet: Tack för att ni arbetar på two factor authentication, @TWITTERHANDLE!
--- a/_data/languages.yml
+++ b/_data/languages.yml
@@ -1,12 +1,11 @@
 # Languages should be categorized according to ISO 639-1
 # Language codes (including within data files) should be in lowercase
 de:
-  work_tweet: Sicherheit ist wichtig, @TWITTERHANDLE. Bitte aktiviert Zwei-Faktor-Authentifizierung auf eurer Webseite.
+  work_tweet: Hallo! Was sind Ihre Gedanken auf Zwei-Faktor-Authentifizierung? Ich würde es gerne, weil ich
   progress_tweet: Vielen Dank für die Fortschritte bei der Zwei-Faktor-Authentifizierung, @TWITTERHANDLE!
 en:
-  work_tweet: Security is important, @TWITTERHANDLE. We'd like it if you supported two factor auth.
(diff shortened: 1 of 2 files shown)


Carlgo11:
Just out of curiosity, how are the users able to send the message to the correct place without a twitter handle?

troy:
@Carlgo11: good point. My intent was only to modify the tweet body, but I was acting like these were @ reply tweets. I think there's a strong argument that they should be reply tweets rather than mentions like they are now. That said, I'll retain the current behavior of mentions for now. Re-adding that now.

troy:
@Carlgo11: added at least enough to preserve the current mention behavior.

IMO, these should be @ reply tweets, like:

```
@blah Hi! ..
```

.. so they're only visible to people who care about both the sender and the recipient ([more](https://support.twitter.com/articles/14023)). That seems more fitting, since this tweet is fairly specific to the recipient - unlike the "Thanks!" tweet, it's hard to see this as general-interest. I can make that change if anyone wants.

In the interest of making this PR as narrow as possible, I didn't make that change. AFAIK this PR should be good as-is.

Carlgo11:
as of right now we provide a template for people to easily contact a company and request 2FA.
The template can easily be changed by the user to personalize the message better.
While I like the idea of not making tweets look like they're being sent from a spambot I dislike the idea of providing an unfinished template.
That will just result in unfinished tweets.

Rather than enable the users to speak freely it gives the impression of one of those standardized tests where you're supposed to finish the sentence correctly.

I think your goal would easier be achieved by giving the users an empty template.
However our intension with the template is to provide an easy alternative to writing a new tweet to the company in question.

I have seen lots of people costomize their #SupportTwoFactorAuth tweets. (Even though it's a very small minority of the tweets)

Again, I like your idea but I think this is the wrong way to achieve it.

//Carl

troy:
> However our intension with the template is to provide an easy alternative to writing a new tweet to the company in question.

I think that's a mediocre or even counterproductive goal. #1813 explains why, but the gist is:
- at best, it makes zero impact except making the user feel better (placates them)
- at worst, it wastes an opportunity for a useful conversation, inaccurately makes the user think they've contributed, and makes the user look bad in the process

This really comes down to whether the project is actually trying to increase the adoption of 2FA, or just encourage venting. The current tweet is venting, but it's not going to affect 2FA adoption.

The threshold for action needs to be higher than clicking "Tweet this!" because the threshold for making an impact is higher than that. If someone isn't willing to write 5-10 words, then they're tweeting to placate themselves, not to encourage 2FA adoption.

RichJeanes:
To be fair, I don't think it's necessarily wrong to play off the expected laziness of users. We don't expect people to actually modify the tweet before sending it, so I'd rather see a complete sentence sent out than an unfinished prompt. If someone wants to edit the tweet before sending it, they're more than welcome to. 

Also, another part of it is just to generate usage of the hashtag. If the hashtag gets used enough, it has the possibility of garnering interest from Twitter visitors, not just 2fa.org visitors through trending and searches.

Also, just fyi, Twitter recently changed their policy and image links and @ mentions don't count toward the 140 character limit any more, IIRC.

troy:
> To be fair, I don't think it's necessarily wrong to play off the expected laziness of users.

I think that comes down to the project goal. If you "don't expect people to actually modify the tweet before sending it," it doesn't encourage 2FA, it just recreates the tunnel-vision perspectives that local newspaper comment sections are known for. It's talking at someone, not talking with them.

If the goal is lots of rant-y tweets, it'll accomplish that, but that won't affect 2FA adoption. I want to see 2FA adopted, which is why I'm arguing for quality over quantity. Lots of inane tweets aren't the barrier to adoption.

If you see awareness of 2FA among less- or non-technical users as a problem, then encourage sending a single tweet in support of the concept or this site/project (without any mentions other than maybe this site). For example: "I'd love to have two factor authentication everywhere, so I #SupportTwoFactorAuth. Here's how you can too: http://.."

Basically, split out the case where talking "at" people is fine (general advocacy) from the case where talking "at" people (current tweet directed at Web apps) is painfully counterproductive.

RichJeanes:
It may be cynical of me, but if someone isn't curious enough to get interested in 2FA from a tweet, I don't think the wording in that tweet is going to change that. 

I also find it interesting that we are discussing the practical opposite of someone else who argued that someone tweeting at just the specific sites they use was a problem (information leak). Now we face the argument that spamming tweets is a problem instead of cherry picking. I love how much variety in discussion this project generates. (I believe that point was made [here](https://github.com/2factorauth/twofactorauth/issues/1212), but the user seems to have deleted their comments for some reason...)

troy:
So, is this dead? If so, the goal of at least the tweet feature, if not the whole project, looks much more about throwing stones than actually increasing 2FA adoption.

PR #1820, opened 2016-06-21
Update Uplay in gaming gaming.yml

berserkovich:
Uplay now supports 2FA

### DIFF
--- a/_data/gaming.yml
+++ b/_data/gaming.yml
@@ -203,7 +203,9 @@ websites:
       url: http://uplay.ubi.com/
       twitter: UplayUbisoft
       img: uplay.png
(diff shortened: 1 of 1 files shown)


Carlgo11:
The doc link you provided isn't a documentation link.
Can you try and find one that is a valid documentation page?

berserkovich:
http://forums.ubi.com/showthread.php/1463703-Secure-your-account-with-2-Step-Verification
is this one good enough?

RichJeanes:
If there's no official Support site article, then I'd say that link works.

berserkovich:
Official support gave me that link :)

PR #1823, opened 2016-06-22
Add hardware to Bitbucket in developer.yml

aladh:


jamcat22:
This has already been addressed in #1821, but thanks for contributing!

PR #1828, opened 2016-06-23
Added Rakuten to retail

kojaxs:


mxxcon:
Actually, maybe it's better to use @AskRakutenUS twitter account?
on the main one it says "Tweet @AskRakutenUS for customer support."

kojaxs:
Wasn't sure which one to use. I've switched it to AskRakutenUS.

PR #1830, opened 2016-06-23
Add DTE Energy

stephengroat:


Carlgo11:
Note to collaborators: _Even though the URL redirects to a sub-domain I think it's better to leave it as it is as the sub-domain contains `newlook`. They will probably remove that sub-domain soon and then that link won't work._

**TL;DR:** _Leave the URL as it is._

PR #1834, opened 2016-06-24
Added Libsyn to hosting

kojaxs:
Wasn't 100% sure if hosting was the right category.


mxxcon:
Can you explain how does it fit into our site criteria?

Carlgo11:
Hello @kojaxs,
would you please answer @mxxcon's question?
If you don't reply within a couple of days I'll mark this PR as inactive.

Issue #1839, opened 2016-06-27
Show exceptions even if no 2FA is available


RichyHBM:
If exceptions exists (i.e. ability to login via other methods that do provide 2FA) these should be displayed.

RichJeanes:
We have discussed the use of OAuth logins that provide 2FA, however there are many sites that support OAuth but also force you to have a native username/password on their site that doesn't support 2FA. While there are also sites that don't force a native login, we don't want users to think that this is the norm and fall into that false sense of security

fpigerre:
Hello @RichyHBM :smiley: 

Thanks for the issues and pull requests you have opened.
At present, no websites listed on [twofactorauth.org](https://twofactorauth.org) that do not support two-factor authentication, apart from [TransferWise](https://transferwise.com/), currently have any exceptions listed. For the most part, if an exception applied to a website that did not support two factor authentication, it would stand to reason that this would result in the website supporting two factor authentication (to some extent). An exception to this rule does include a situation where two factor authentication is supplied through OAuth, as with TransferWise.

Unfortunately, the vast majority of websites supporting authentication through OAuth also allow concurrent authentication without two factor authentication forcibly enabled. This results in a situation where two factor authentication is never necessary to login, a situation which, for the purposes of [twofactorauth.org](https://twofactorauth.org), is not considered true "two-factor authentication".

At the moment, I can't think of any further situations where an exception would need to be listed for a website that does not support two factor authentication. However, if you feel that such functionality would be necessary, please don't hesitate to comment.

Thankyou,
psgs :palm_tree:

RichyHBM:
Hey @psgs,

Sure completely understand, to be honest I can't think of a situation in which an exception could be needed other than if a site provided OAUTH.

I though it could be useful to list that a site allows a method of logging in that does support 2FA even if the site its self doesn't, though I do agree that doesn't really mean the site its self allows 2FA.

I would however say the mobile view already displayed the exception when in progress, I don't know if maybe it was originally a thought to have it displayed, but then removed due to not being needed at the time?

Thanks,
Richy

RichJeanes:
@RichyHBM If you would like examples of where exceptions are currently used, you can look on the site and see them. For some examples, see Apple and Steam, among others. While listing if a site supports OAuth logins might be nice for people looking for that feature, that is not the goal of 2FA.org. Perhaps, if you think it useful, you could start a site to list sites and services that support OAuth logins.

[...]

PR #1841, opened 2016-06-27
Show exceptions

RichyHBM:
This addresses issue #1839 displaying the exception section even when 2FA is not available


RichyHBM:
I just realised this depends on a separate commit, branching error. Fixed pull available at #1843

PR #1843, opened 2016-06-27
Add missing exception section

RichyHBM:
Resubmission of #1841. This addresses issue #1839 displaying the exception section even when 2FA is not available


fpigerre:
Discussion regarding this pull request is currently taking place in issue #1839 :mag_right:

PR #1847, opened 2016-06-28
Added uberpeople.net.

awelch83:


_data/social.yml
```diff
@@ -175,6 +175,17 @@ websites:
       software: Yes
       doc: https://support.twitter.com/articles/20170388
 
+    - name: Uberpeople.net
+      url: http://uberpeople.net
+      img: site.png
```
dsoegijono: This is incorrect image name, should be `uberpeople.png`

Carlgo11:
Hello @awelch83,
I'd just like to point out that the build log of your PR, along with error messages, can be found over at [Travis-CI](https://travis-ci.org/2factorauth/twofactorauth/builds/141126965).

The current build log gives these error messages:

```

1. Uberpeople.net should not contain a 'hardware' tag when it's value isn't 'YES'.

2. Uberpeople.net should not contain a 'sms' tag when it's value isn't 'YES'.

3. Uberpeople.net should not contain a 'phone' tag when it's value isn't 'YES'.

4. img/social/uberpeople.png is not 32x32

5. img/social/uberpeople.png should not be larger than 2500 bytes. It is currently 19313 bytes
```

Along with that I'd also like to point out that the `doc:` tag isn't needed if you don't have any link to redirect to. :smile: 

//Carl

awelch83:
@Carlgo11 Thanks for taking the time to get me straightened out. I look forward to making additional contributions

Carlgo11:
Great! Here's a smaller image. Even though the one you used works this one is 40% smaller. :wink: 
uberpeople

PR #1849, opened 2016-06-29
Added the 2500-byte *.png limitation to contributing.md.

awelch83:


RichJeanes:
That's kind of the point of saying you should run the png through an optimizer. Odds are, that will get it well below 2.5KB. 2500 B is just an indicator that it may not have been optimized and that we should double check the file.

awelch83:
I had a pull request fail checks because my png exceeded 2500 bytes

Sent from my iPhone

On Jun 29, 2016, at 17:05, Rich Jeanes notifications@github.com wrote:

That's kind of the point of saying you should run the png through an optimizer. Odds are, that will get it well below 2.5KB. 2500 B is just an indicator that it may not have been optimized and that we should double check the file.

RichJeanes:
Yes, I know. Uberpeople.
But, as I said, this is just an indicator to signal that there may be something that can be done with the image. If the image cannot be optimized and a suitable alternative cannot be found, the >2500B image can still be used. However, much more often than not, it just indicates that the image was not run through an optimizer.
Also, note that the image you used was double the resolution (quadruple the number of pixels) it was supposed to be, compounding the file size issue.

awelch83:
It needs to be explicitly stated. It makes sense to do so. It'll save new contributors from having their pull requests fail if they can get it right the first time. Come on..

Sent from my iPhone

On Jun 29, 2016, at 17:24, Rich Jeanes notifications@github.com wrote:

Yes, I know. Uberpeople.
But, as I said, this is just an indicator to signal that there may be something that can be done with the image. If the image cannot be optimized and a suitable alternative cannot be found, the >2500B image can still be used. However, much more often than not, it just indicates that the image was not run through an optimizer.
Also, note that the image you used was double the resolution (quadruple the number of pixels) it was supposed to be, compounding the file size issue.

awelch83:
I should've created a new branch for this pull request.

PR #1854, opened 2016-07-01
Fixed alphabetization for utilities.yml.

awelch83:


_data/utilities.yml
```diff
@@ -208,4 +208,4 @@ websites:
       url: https://voip.ms/
       img: voipms.png
       tfa: No
-      twitter: voipms
+      twitter: voipms
```
mxxcon, line 211: As per our `.editorconfig`, it contains `insert_final_newline = true`.
Please respect that.

PR #1860, opened 2016-07-01
Fixed alphabetization for investing.yml.

awelch83:


_data/investing.yml
```diff
@@ -180,4 +180,4 @@ websites:
       url: https://www.wisebanyan.com/
       twitter: WiseBanyan
       img: wisebanyan.png
-      tfa: No
+      tfa: No
```
mxxcon, line 183: As per our `.editorconfig`, it contains `insert_final_newline = true`.
Please respect that.

PR #1861, opened 2016-07-01
Fixed alphabetization for other.yml.

awelch83:


_data/other.yml
```diff
@@ -303,4 +303,4 @@ websites:
       tfa: Yes
       software: Yes
       sms: Yes
-      doc: https://support.zendesk.com/hc/en-us/articles/203824246
+      doc: https://support.zendesk.com/hc/en-us/articles/203824246
```
mxxcon, line 306: As per our `.editorconfig`, it contains `insert_final_newline = true`.
Please respect that.

PR #1864, opened 2016-07-01
Added entry for Private Internet Access VPN service.

awelch83:


Carlgo11:
The logo is a little too large for my taste. Here's a slightly more compressed version. (- 27% of the original)
pia logo

awelch83:
With all due respect, why do I (or any other contributor) need to concern ourselves with your "taste"? I based everything that I did off of the instructions set forth in the md files. If additions need to be made to those files to eliminate this sort of back and forth, let's do that - in accordance with your taste, of course.

Carlgo11:
To answer your question, as I'm the chief maintainer of this project I am one of the few who have the final say on what gets published or not. That's pretty much the only reason you need to comply with my requests regarding commits.

We've seen multiple reworks of this project that better fitted the authors needs and we are fine with that as long as they comply with our license.

As for why I asked you to use a slightly more compressed version is because we try to scale down the images the best we possibly can. The 2500 Bytes is mostly just a maximum limit. If we see something that can be compressed even further we'll ask the author of the PR to do so.

mxxcon:
@awelch83 The reason why we are a bit nitpicky about logo images is because on our site some categories serve a bunch of them at once, often a few dozen at a time. Needlessly large images make end-user experience worse and wastes bandwidth. This is especially noticeable on mobile devices..

awelch83:
Thanks mxxcon. In reality though, it's 2016, not 1996. A 2499 byte file will load in mere milliseconds even on a 3G connection. Let's be reasonable here :)

mxxcon:
We are being reasonable based on previous performance issues with the site and number and size of served images. Just because it's 2016 doesn't mean you have to be wasteful with end-users' resources.

@awelch83 We are spending too much discussing this rather than you simply re-compressing the logo or using the one @Carlgo11 provided.

awelch83:
@mxxcon I have a habit of questioning things that seem pointless. Society would be better off if more people did this.

PR #1867, opened 2016-07-01
Fixed alphabetization for transport.yml.

awelch83:


_data/transport.yml
```diff
@@ -177,4 +177,4 @@ websites:
       url: https://www.westjet.com/
       twitter: WestJet
       img: westjet.png
-      tfa: No
+      tfa: No
```
mxxcon, line 180: As per our `.editorconfig`, it contains `insert_final_newline = true`.
Please respect that.

PR #1869, opened 2016-07-04
Add S-ID-Check

pluscard:


RichJeanes:
I'm not sure that this should be listed here. It seems like this site is just describing the 2FA available for Sparkasse credit cards when making a purchase. I guess it is 2FA in a way, for real-world transactions, but you could say the same of every bank that supports chip-and-PIN transactions. While I think this is a fantastic idea, extending the second factor to more transactions that aren't traditionally protected...

I guess this boils down to a question of the scope of this site. Do we want to list every system that has some form of MFA or do we want to focus on online accounts that have MFA?

Carlgo11:
Rich certainly has a point there. Closing this PR.

pluscard:
@RichJeanes @Carlgo11 ,
thank you very much for reviewing the commit.

I'm writing on behalf of the company that developed the application and we'd really love to see it listed here - if possible, I'd like to quickly explain the scope of the application as I think there might have been a misunderstanding about what it actually does (granted, the page is in German).

S-ID-Check is an implementation of two-factor authentication for _online_ credit card transactions. It is not used for any kind of point-of-sale (or, as you put it, "real-world") transactions.
While it's correct that it technically doesn't protect the _login_ process of an online service, I'd argue that in Banking and Payment applications the _actual transaction_ is the interaction that requires the strongest protection as it carries the greatest financial risk (contrary to services like Facebook or Twitter that need to protect their _login_ process because their _actual service_ is fully accessible after logging in).
Restricting the site to applications that merely protect the process of logging in works well in most categories, but narrows the scope quite noticably in the Banking and Payment sections.
There are a few other online banking applications listed on the site that don't use MFA (primarily text messages) for the login processes but before executing transactions and I think that our application is comparable to those - at least in this region only protecting money transfers or changes to the account using MFA is quite common.
S-ID-Check is developed to comply with recent regulation by the German Federal Financial Supervisory Authority that specifically requires MFA for online transactions.

I'd be glad if you reconsidered, but thank you in advance regardless.

PR #1871, opened 2016-07-05
Fixed alphabetization in the files that needed it.

awelch83:
Here you go, @Carlgo11. I think all is in order now.


Carlgo11:
Thank you :smile:

awelch83:
@Carlgo11 This is just a point of curiosity for me, but why did you want me to mash all of these pull requests into one? And why didn't the person who merged some of my other alphabetization pull requests ask for the same?

PR #1872, opened 2016-07-06
Add freenet.de

guerda:


guerda:
I corrected the metadata and the logo according to the CI build.

RichJeanes:
Does Freenet have a Twitter account that could be added to the listing? If they have a Twitter dedicated to support, that is preferable.

guerda:
Freenet does not have a twitter account for support. They only have one-directional "news twitter feeds". So no contact with these feeds.

Carlgo11:
@guerda Do they have some email address to turn to for support? That or Facebook page?

guerda:
@Carlgo11 They have facebook. Is there an attribute for that? I haven't found any facebook attributes in the files.
They do not provide any support contact mail address.

Carlgo11:
Hello again @guerda,
sorry for the late reply.

Facebook pages can be added like this:
`facebook: <facebook_page>`
so for example if the facebook page is `https://facebook.com/2factorauth/` the facebook tag would be `facebook: 2factorauth`.
Hope that helps! :smile:

guerda:
Hi @Carlgo11 !
Thanks for the help! I added the section accordingly.

Carlgo11:
@2factorauth/core

PR #1874, opened 2016-07-07
Update banking.yml

urlicht:
Adding email to Bank of America authentication method.
Bank of America (BofA) does support two factor authentication through registered email addresses.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -49,6 +49,7 @@ websites:
       img: bankofamerica.png
       tfa: Yes
       sms: Yes
+      email: Yes
       hardware: Yes
       doc: https://www.bankofamerica.com/privacy/online-mobile-banking-privacy/safepass.go
 


Carlgo11:
How does that work? Is it required every time you log in like SMS and Hardware is or is it just Two Factor Authorization?

urlicht:
@Carlgo11 every and each time you attempt to login, it gives you options: pre-registered and verified email address(es) or phone number(s). It only gives you partial of the numbers/addresses (e.g. **_-**_-1234, a*****3@gmail.com). It is not fixed, so you can selection a different option every time. The authorization code is then sent to the selected destination, and you need to enter the code, which expires in 10 min.

Carlgo11:
Okay thank you for clearing that up :+1:

fpigerre:
Thanks for the pull request @urlicht! :+1:

Issue #1484, opened 2016-01-02
Please add MercadoLivre.com.br


UnderEu:
Add http://mercadolivre.com.br bidding website, it belongs to 'Retail' category.

Supported: Software token (via Authy app)

stephengroat:
https://www.mercadolibre.com/ shows a list of a bunch of seperate mercadolivre sites for different countries.

Do all countries share accounts and have the same software support?

sailenicolas:
Yes @stephengroat, is all the same site. Only shows different articles from country to country. But is all the same in the security section... authy.

PR #1876, opened 2016-07-08
Add mercado libre

stephengroat:
#1484


RichJeanes:
Just a little thing, but can you update the doc link to use HTTPS? Thank you!

stephengroat:
Fixed. Good catch!

RichJeanes:
Hmm, I also just noticed that they way the site is divided, each country they serve has a different TLD and this extends to the documentation sites. Does anyone think this is an issue since we can only link to one specific doc link? (Currently Argentina's `.com.ar`)

stephengroat:
I chose `ar` because it was the first country listed on the www.mercadolibre.com

sailenicolas:
Well, is .com.ar because the site is from Argentina. Check [Wikipedia](https://en.wikipedia.org/wiki/MercadoLibre.com) They also have [MercadoPagos.com](MercadoPagos.com) which is equivalent to "paypal"...

stephengroat:
MercadoPago is owned by MercadoLibre, so at least by including MercadoLibre the parent company is included

Carlgo11:
bump @2factorauth/core

PR #1895, opened 2016-07-14
Adds Promoter.io

avelis:
Promoter.io currently does not have 2FA. Listed in other as a reference.


mxxcon:
Please run this logo through png compressor such ImageOptim or Trimage or FileOptimizer. I was able to shrink this file down to 1.2kb

avelis:
@mxxcon You got it. :-)

PR #1898, opened 2016-07-15
Updated MailChimp details

aaronarich:
Updated the docs to a new Knowledge Base article to reflect SMS 2-FA ability

### DIFF
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -63,8 +63,9 @@ websites:
       url: https://mailchimp.com/
       img: mailchimp.png
       tfa: Yes
+      sms: Yes
       software: Yes
-      doc: http://kb.mailchimp.com/integrations/other-integrations/set-up-google-authenticator
+      doc: http://kb.mailchimp.com/accounts/login/set-up-two-factor-authentication-with-sms
 
     - name: Mailgun
       url: https://mailgun.com


mxxcon:
Is there a more generic docs url that directs to both, goog auth and sms?
edit: maybe this url? http://kb.mailchimp.com/accounts/management/best-practices-for-account-security

aaronarich:
@mxxcon hah, I can make that recommendation to our content team... but I updated to the SMS article because it has those directions and links to Google Auth as well

RichJeanes:
According the new article:

> If we see patterns that seem unusual for your account, we'll send you a text message with a verification code

Unless there is an option for a code to be required on every login, I believe this does not fit our criteria to qualify as Two Factor _Authentication_ via SMS. @Carlgo11?

Carlgo11:
Yes unless it's requested every time you log in it's not Two Factor Authentication.

aaronarich:
@RichJeanes @Carlgo11 this is the Knowledge Base article that I linked to in my PR. When 2-factor authentication is turned on, you will receive a 1-time use password via SMS every time you log in. The alternative (which was our initial implementation) would be to use Google Authenticator.

[Set Up Two-Factor Authentication with SMS: http://eepurl.com/b9a03L](http://eepurl.com/b9a03L)

This article is not only the documentation for how to setup SMS 2-FA, but also links to documentation on how to implement Google Authenticator. That's why I thought it was a better choice, since we don't have a unified article for both.

Carlgo11:
@aaronarich So the SMS Text Message Verification that Rich quoted does not work the same way as SMS 2FA?

aaronarich:
@Carlgo11 that's correct. We do send out SMS account notifications (for security alerts), but that is a different feature than SMS 2-FA

RichJeanes:
My apologies! I think I was looking at the page that mxx linked instead of the one that you added, @aaronarich . Thanks for the clarification!

Issue #1893, opened 2016-07-13
Open Classifieds


neo22s:
Using google authenticator.

Carlgo11:
Hello!
Thank you for your interest in twofactorauth.org.
While we'd love to add all sites to the list we simply don't have enough time to do so.
Therefor we rely on people's pull requests(PR). The requirements for pull requests can be found [here](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md).
If you need help making your PR or have any questions regards the PR requirements don't be afraid to ask us!

If you do not wish to make a PR please tell us so and someone will come back to this ticket in the future and make a PR. However please note that this can take a long time as we're all volunteers and do this on our spare time.

stephengroat:
should be closed, already includes yclas. See #1899 for details

neo22s:
Hello,

Open Classifieds is a FOSS php script that you need to host yourself, like wordpress.org. Where Yclas its a SaaS that has 2FA for customers to login to use their sites (that also have 2FA), llike wordpress.com.

Not sure if I explained myself properly xD, so yes 2 different stuff ;)

regards

PR #1897, opened 2016-07-15
Add Yclas

alanbares:
Adds yclas site and logo. Hope the png is ok. I ran it through through optipng and it took me a bit to even get the dev environment setup on windows for testing everything.  Closes #1894

PR #1899, opened 2016-07-15
Add open classifieds

stephengroat:
#1893


mxxcon:
is there docs link how people can enable 2fa?

stephengroat:
Added a link

mxxcon:
what's the difference between this and #1897?

stephengroat:
From what I can tell, yclas is a hosted solution, while open classifieds is the underlying project.

Which do you think would be better to keep?

mxxcon:
We generally don't list self-hosted projects...

stephengroat:
No problemo

neo22s:
Hello,

Open Classifieds is a FOSS php script that you need to host yourself, like wordpress.org. Where Yclas its a SaaS that has 2FA for customers to login to use their sites (that also have 2FA), llike wordpress.com.

Not sure if I explained myself properly xD

regards

Carlgo11:
Thank you for clarifying that @neo22s. By the sound of it Open Classifieds does not fit our list as we don't list self hosted sites (like wordpress.org) here.

neo22s:
aha! so do not add it heheh thanks!

Issue #1882, opened 2016-07-12
Wells Fargo does not support software tokens


chittti:
I am aware of #1582 . Wells Fargo does support Text, Call and RSA SecurID hardware token. But it **does not** support software tokens. I am pretty sure. Anyone who knows how to get 2FA codes using a software token, please show us how to do so. If you look closely at #1582, nowhere does it show an option for software token. 

Thanks.

mxxcon:
Addressed by #1900

PR #1809, opened 2016-06-08
Remove software token support from Wells Fargo

gcochard:
In their FAQ, they only mention an RSA token or sms/calls. There's no mention
of Google authenticator, Symantec vip, or other software tokens.

https://www.wellsfargo.com/help/faqs/advanced-access

PR #1900, opened 2016-07-17
Updated Wells Fargo 

chittti:
Please see #1882. Wells Fargo does not support software tokens. I have waited for someone to come forward with comment regarding availability.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -517,7 +517,6 @@ websites:
       img: wellsfargo.png
       tfa: Yes
       sms: Yes
-      software: Yes
       phone: Yes
       hardware: Yes
       doc: https://www.wellsfargo.com/privacy-security/advanced-access


mxxcon:
We already had #1809...And as I said there, there's https://www.rsa.com/en-us/products-services/identity-access-management/securid/software-tokens
Please explain why that does not qualify as software?

chittti:
Well the link you provided does qualify as software but Wells Fargo does not use that. 

Please see this - https://www.wellsfargo.com/biz/jump/securid . This is linked to 'RSA SecurID **device**' hyperlink [here](https://cloud.githubusercontent.com/assets/15699955/13035892/e93b2b82-d30e-11e5-8157-377a234f6eeb.png) which is shown in #1582 . So basically although they are using RSA SecurID software, you cannot use the applications that are quoted in the link you posted. There is no way to get the QR code for authentication in the app without getting the RSA SecurID device which costs $24.95 and following whatever the activation protocol is. 

And I am curious as to how you got that link. From Wells Fargo or googling? If Wells Fargo somehow does support that, then they are doing a very bad job or I'm overlooking something since I looked into software tokens since they are much more secure than OTP via SMS/call IMO.

will-in-wi:
I just ran across this, and I think the PR is correct. AFAIK, there are three options for 2FA: Call, Text, Hardware Token. Software Tokens are not available or supported.

👍

mxxcon:
@chittti i found that link from previously looking into secureid issues.

I don't use secureid or wellsfargo, so I can't comment if it's _impossible_ to activate the software solution.

PR #1903, opened 2016-07-18
Add CommuterBenefits

stephengroat:
WiredCommute is a product that allows direct withdrawals from paychecks to cover employee transportation expenses, needs 2FA to protect


mxxcon:
Hmm..should that be transport category or `other`?

stephengroat:
Sounds good. moved to other

RichJeanes:
What makes WiredCommute different from Commuter Benefits as a whole? Does it have a separate login system? Is there a difference, account security wise, between the employee and employer logins that they make available?

stephengroat:
From being a (forced) user, WiredCommute is a product of the company CommuterBenefits. In talking with our IT department, they seem not to other 3rd party auth, so CommuterBenefits users are forced to use their authentication system.

I don't know about the employee and employer login differences, don't know of any easy way to confirm that.

RichJeanes:
So do you log in on this website to manage your benefit?

stephengroat:
yup. the true joy is that it can deduct from your paycheck and the WiredCommute site lets you choose an address to send your benefits to (i.e. Metro card, etc.) So an attacker could login to a person who doesn't use the benefit, start it and get it sent to him.

RichJeanes:
So since this site login (presumably) applies to their other products and services, perhaps changing the URL to just their homepage would be in order?

stephengroat:
I can't confirm that. Do you want me to convert it to CommuterBenefits from WiredCommute?

RichJeanes:
If you use the login links at the top of that page to log in, I think it is reasonable to assume that they use the same login system for their other products since that login bar pervades the entire site. Please, do make the conversion. Thoughts from others are welcome.

mxxcon:
For lack of a better logo, perhaps instead of that squished red ball use icon from their android app? :) https://lh3.googleusercontent.com/L5CXl--FzYbxu2qADmY113SOYR9e1aPpE47s5jZ99F6CZZAgP3mgPV7cBbvXVyqVxk0

That reminds me, we should also add WageWorks.com. They are similar to commuterbenefits(plus more)

Carlgo11:
Doesn't look like they monitor their twitter account. Any other social media accounts?

stephengroat:
There's a facebook page: https://www.facebook.com/CommuterBenefits/

Want me to add that?

Carlgo11:
That link doesn't work for me. Let's skip it then.

Issue #1880, opened 2016-07-09
Sonic Has Two-Factor Authentication


matthewmspace:
Sent in a pull request a few days ago, but thought I'd send it in as an issue as well. Sonic, the ISP/utility, has two-factor authentication in the style of a code generator, sms, and voice calls: https://wiki.sonic.net/wiki/2-Step_Verfication.

mxxcon:
@matthewmspace You sent a pull request in your own repo rather than this one.. :)

matthewmspace:
Oh. Sorry, I have no idea how to use Github, as I can't code.

PR #1905, opened 2016-07-19
Update sonic in utilities

stephengroat:
#1880


RichJeanes:
It seems that `sonic.net` now redirects to `sonic.com`, but just hangs if you try to connect securely. Could you update the URL to point directly to `https://sonic.com` for us? The doc link seems fine, though. Thanks!

stephengroat:
Fixed

PR #1908, opened 2016-07-19
Add Northwestern Mutual

kennethhlee:


RichJeanes:
I'm going to be a little picky on this, but can I ask you to expand the name of the image to the full business name? Just for the sake of clarity. It seems like they don't really refer to themselves as just "NM" except in a couple of social media usernames.

Also, do they actively communicate with customers via the nm_news Twitter account or is it only used for announcements? (I would check myself, but I do not have access to Twitter at the moment) If they don't interact, do they have a different account that they do use for customer support and interaction?

kennethhlee:
@RichJeanes The only twitter account that I could find was 'nm_news'.  From what I can see, it is used only for announcements.  I can remove the line for Twitter if you think it is appropriate.

Carlgo11:
I found a facebook page that they interact with people on!
`facebook: northwesternmutual`

kennethhlee:
I've removed the twitter: field and added the facebook: field and added a new commit.  I did not see anything in the WIKI that talked about a facebook: field.

Carlgo11:
Hello again,
no the facebook tag isn't being advertised yet as we haven't added it yet.
When we do it will be documented on both the wiki and contributing.md.

The tag can however already be used. It just won't be displayed on the site yet.

PR #1910, opened 2016-07-20
Add Adirondack Insurance Exchange

kennethhlee:


mxxcon:
This is an insurance company. Is `finance` a right category for it?
also https://www.aie-ny.com/ doesn't work, but http:// does. Please correct that.

kennethhlee:
I've corrected the URL

kennethhlee:
Regarding the comment about which category this site belongs under. When I looked for categories that AIE fit under, the first one I looked for was insurance which did not exist. I then reviewed related categories and found that finance appeared to contain the insurance companies, e.g. Progressive and Geico.  This is how I decided on finance for AIE and for Northwestern Mutual which appears to have been recently merged.

Perhaps it may be time to create a category for insurance and reorganize.

RichJeanes:
A new category may be something to consider. Thank you for your contribution!

PR #1911, opened 2016-07-21
BofI Federal Bank

brianvargo:
bofifederalbank
Added BofI Federal Bank as a TFA bank

PR #1912, opened 2016-07-21
Add BofI Federal Bank to banking.yml

brianvargo:
BofI Federal Bank is a 2FA bank


RichJeanes:
Closing in favor of #1911

PR #1915, opened 2016-07-22
Add doodle.com

grahamlicence:
I've added this under other as there's not an obvious organiser category, although it could go under social..


Hillside502:
Suggest "Communication" to be the best category --- as sending/receiving invitations to (calendar) events is Doodle's purpose.

Hope this helps!

grahamlicence:
It sure does, have moved, thanks

mxxcon:
Wow. It's crazy that doodle provides SSL only to premium plans. :confused:

RichJeanes:
This is also the same site that redirects away from HTTPS on their home page, but then redirects _to_ HTTPS for every other page on the site. Wut?

Also, I can't say that I've ever heard of Doodle. How big is the user base?

grahamlicence:
Yeah their security approach is odd, especially when they're aiming themselves at businesses.

It's pretty big, they claim 20 million users (https://en.blog.doodle.com/2014/01/29/doodle-crosses-the-20m-user-mark/)

Hillside502:
Discovered several security issues with Doodle, but before I fill in the details, the first question is:-

Other than through third-party Google or Facebook 2FA, what gives the impression that Doodle is 2FA-enabled?

grahamlicence:
Sorry, not sure I understand?

Hillside502:
Hi Graham,

Why do we think Doodle has 2FA?

mxxcon:
@Hillside502 It doesn't. Hence `tfa: No`.

grahamlicence:
It doesn't have 2FA! 

If you mean why should it have 2FA I think it should as it stores a lot of personal information, such as event details, e-mail address, residential address or phone number. They also seem to say that having SSL is enough for protection, see https://doodle.com/premium

Hillside502:
@mxxcon @grahamlicence 
Silly me!
As I'm sure you've guessed, I'm new to https://twofactorauth.org/ and overlooked the fact that sites are listed for BOTH with and without 2FA.

Fully agree that Doodle should have 2FA for accounts signed up via email --- but a workaround is to sign up for an account with a 2FA-enabled Google or (God forbid) Facebook ID.

@RichJeanes 
With the following browser addon --- and after logging in to Doodle --- the user is (mostly) kept on a HTTPS course:-

HTTPS Everywhere | Electronic Frontier Foundation
https://www.eff.org/https-everywhere

(I say mostly as the addon's Doodle implementation is currently just "partial")

Having said this, Doodle says:-

> End-to-end SSL encryption
> Protect sensitive internal data with end-to-end SSL encryption - all access to your Premium Doodle polls is SSL-encrypted for both you and your poll participants.

As we move towards an HTTPS-only internet --- https://letsencrypt.org/ --- Doodle is out-of-step and out-of-order to downgrade the security of lower-tier users.

Moving on to Doodle at Qualys --- we have:-

> Warning: Inconsistent server configuration

67% of Doodle servers are grade A ---------- 33% grade F

SSL Server Test: doodle.com (Powered by Qualys SSL Labs)
https://www.ssllabs.com/ssltest/analyze.html?d=doodle.com

This last point is an overlooked (by Doodle) misconfiguration error --- a dealbreaker for the time being, as far as I'm concerned!

mxxcon:
While logging in with your google or facebook would provide 2fa, if somebody manages to get your regular username/password, your account still won't be secure. As such, we don't consider sites that use federated logins with 2fa, but w/o native username/pass/2fa to be `tfa: yes`.

Hillside502:
@mxxcon 
Thanks for the clarification but if, for example, you mean someone logged in via a 2FA Google account --- that account would need to be 2FA-authenticated in the browser session in advance of logging in to Doodle.

Using the example of Doodle, I fully agree that it should be listed as `tfa: no` but don't see how the Doodle account can be accessed otherwise.

RichJeanes:
@Hillside502 Yes, I use HTTPS Everywhere, but the vast majority of web users don't so that should be the main concern. I comment on their configuration because that must be very deliberate to redirect away from a secure connection for _only_ the home page, then back to a secure connection for everything else. 

As far as federated (OAuth) logins, many sites that allow this also require you to set up a username and password for their site specifically, logins which are not protected by 2FA. While some sites do allow you to have your OAuth login as your only login method, it would be more work to figure out exactly which ones (since it is not a very commonly documented detail) and would be more information to somehow relate to the end-users of 2FA.org.

Also, OAuth is technically authorization, not authentication, so it also doesn't fit the bill since we list sites with two factor authorization as `tfa: No`.

Hillside502:
@RichJeanes 

> ...the vast majority of web users don't so that should be the main concern

Fully agree!

As you say, when returning to the Doodle home page from other Doodle HTTPS pages, the user is deliberated redirected to HTTP.

This will be because --- on the home page ONLY --- ads are served via the third parties, AppNexus and Google AdServices.

_data/communication.yml
```diff
@@ -33,6 +33,12 @@ websites:
       img: disqus.png
       tfa: No
 
+    - name: Doodle
+      url: https://www.doodle.com
```
Carlgo11: Change the url to `http://doodle.com`

PR #1918, opened 2016-07-25
Add ING

nicolasleger:


Carlgo11:
Looks like they also support 2FA through some kind of smartphone app. Is that correct?
If so you can add `software: Yes` as well :wink:

nicolasleger:
I can not confirm the smartphone app, but I'm sure for the token generator key.

Carlgo11:
Could you add it anyway?

mxxcon:
Perhaps this should also be renamed to ING(Luxembourg)? Or somehow else differentiate it from ing.com proper? Unless they all share the same login system then perhaps list the main .com site?

_data/banking.yml
```diff
@@ -255,6 +255,13 @@ websites:
       img: huntingtonnationalbank.png
       tfa: No
 
+    - name: ING
```
Carlgo11: As @mxxcon requested earlier, please change the name to ING (Luxembourg)

nicolasleger:
Renaming done

PR #1922, opened 2016-07-26
Add DocuSign

stephengroat:


mxxcon:
And again, their official logo is at https://www.docusign.com/sites/all/themes/custom/docusign/favicons/favicon-32x32.png

What about https://support.docusign.com/en/guides/cdse-user-guide-sending-authentication-option-descriptions  and https://www.docusign.com/partner/docusign-phone-authentication-powered-by-authentify ?

stephengroat:
The DocuSign Phone Authentication and Authentication Options only apply to the document signing process, not to the account which store documents (potentially containing addresses, phone #, SSNs, and all of the other PII that signed documents have).

mxxcon:
Weird that they would have 2fa for documents but not the account :/

Issue #1924, opened 2016-07-26
Using verify.rb to check section file alphabetization


stephengroat:
I was thinking of something like this to prevent errors in section file alphabetization:

``` ruby
sections = YAML.load_file('_data/sections.yml')
sections.each do |section|

  data = YAML.load_file('_data/' + section['id'] + '.yml')

  if data['websites'] != data['websites'].sort_by { |h| h['name'].downcase }
    error("#{section['id']} is not alphabetized by name")
  end

  data['websites'].each do |website|

    validate_tags(website)
    validate_image("img/#{section['id']}/#{website['img']}", website['name'])

end
```

Any other comments or ideas?

Carlgo11:
Closing as #1925 now exists.

PR #1925, opened 2016-07-26
Using verify.rb to check section file alphabetization

stephengroat:
#1924 Had time on a plane to try it


mxxcon:
Do we want to be that strict about enforcing alphabetizing `yml` files? :) I think it might be a barrier to entry and not that critical to the end result. Manually re-sorting these files every few months is not that big of a deal?

Carlgo11:
Maybe we could separate it so that it isn't in the travis build script but in a separate that could be run manually?

stephengroat:
I was thinking that the skill required to resize an image down to 32x32 has to be harder than re-alphabetizing a file.

What about adding a check that if only 1 entry is out of order in a file, recommending the correct place for that entry?

mxxcon:
Once again comes up my desire for Travis to have something like a _soft fail_ state for "nice to have" things 😃  That way if folks want to, they can fix it up, otherwise we can merge anyway.

Stephen, proper state of logo images are a bit more critical because they are being directly exposed to the end user. yml files are behind the scenes kind of thing.

I'm not against your change, just brainstorming.

stephengroat:
Easy enough to add a `@output_warn` that simply print out warnings of out of order files. travis will pass, but the order warning will be recorded

I understand about the importance of the images, and maybe it's just me, but it seems harder to do the image manipulation than alphabetization. I also understand trying to lower the bar for contributions **as much as possible** as well

mxxcon:
will `@output_warn` show anything here before we merge or will we have to dig into  raw travis logs?

stephengroat:
I don't see any way in the travis api to return any status strings that show up on github PR page, only success or failure

stephengroat:
Let me bite my tounge, I found [this](http://stackoverflow.com/questions/35714151/how-does-travis-ci-reply-back-to-github-that-tests-are-passed) on a flight. I'll try it tonight

edit: no dice, no success

mxxcon:
Maybe there's some other service we could utilize for "nice to have" checks? And that service could perhaps even add a regular comment to a PR..

Carlgo11:
```
Maybe we could separate it so that it isn't in the travis build script but in a separate that could be run manually?
```

\- [@Carlgo11](https://github.com/2factorauth/twofactorauth/pull/1925#issuecomment-235615572)

stephengroat:
I have to admit, I'm a big fan of automating myself out of existence (so that I can focus on something else).

What about putting it in as a warning in travis to start and looking at the logs (everyone once in awhile) to see how disorganized the files are getting. If it's not too much of a problem, it could then be enabled as enforcing after a while?

I'm looking into the rest of [the integrated CI system for Github](https://github.com/integrations) to see if any others allow for custom commit status messages

stephengroat:
I do see an issue of having the warning when merging with `master` if there's already an warning on the branch. The new branch picks up the error but can't do anything about it unless the merge in any changes and then fix them.

Leaving the current branch un-merged to display the problem

```
$ bundle exec rake
/home/travis/.rvm/rubies/ruby-2.3.0/bin/ruby ./verify.rb
<----------- WARNING ----------->
1. _data/retail.yml is not alphabetized by name
No errors. You're good to go!
```

mxxcon:
@Carlgo11 yes, you said it, but I don't want to run anything manual here 😄 If i'm doing that, since I'm there I might as well just re-organize files myself 😄

stephengroat:
Added in some functionality to detect unused image files. can be broken off into another commit if necessary. See the last 2 logs from travis to see how it works

Also, in thinking about it, I think it's a mistake to have travis only warn on alphabetization and not error out. I've been automating a bunch of stuff recently and have found that while stuff works, people rarely look under the hood at logs. When they do, it's usually just a "i'll fix that later" for any warning. Since code coverage tools won't really work on testing for this, I think it's all or none

stephengroat:
Any more comments on this? Merging master is a pain :-)

_data/gaming.yml
```diff
@@ -181,20 +181,20 @@ websites:
...
           text: "Software implementation requires you to have purchased either Final Fantasy XI or Final Fantasy XIV and registered the game."
 
+    - name: "Star Wars: The Old Republic"		
+      url: http://www.swtor.com/		
+      img: swtor.png		
+      tfa: Yes		
+      software: Yes		
+      hardware: Yes		
+      doc: http://www.swtor.com/info/security-key		
+
```
mxxcon: You have tabs at the end of every line here.

stephengroat: Can and will clean.

I'm also looking into using rubocop or some other sort of lint tool to
enforce the editorconfig rules (so that I don't make mistakes like this)

mxxcon:
:lgtm: Except for the ruby code. Not in a position to comment on it.

Carlgo11:
I will go over it tomorrow morning. I don't currently have access to any platform to run the script on.

mxxcon:
Reviewed 7 of 16 files at r1, 4 of 10 files at r3, 4 of 6 files at r4, 3 of 3 files at r6, 1 of 1 files at r7.
Review status: 19 of 20 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/2factorauth/twofactorauth/1925)*

Carlgo11:
Thanks for the improvements on verify.rb @stephengroat!
It looks really good :smile:

PR #1939, opened 2016-07-30
Add Samsung [Retail]

Jawshy:
#### Changes:
- Added: [Samsung](https://account.samsung.com/) under retail. 
  - No TFA. [One of Samsung's verified Twitter handles](https://twitter.com/SamsungMobile). 

I'm not entirely satisfied with the category as Samsung has products and services outside of retail.


mxxcon:
perhaps change url to samsung's main page?

Jawshy:
Okay, but note that their main page is not secured through HTTPS.

PR #1940, opened 2016-07-30
Add Dreamstime [Other]

Jawshy:
#### Changes:
- Added: [Dreamstime](https://www.dreamstime.com/) under other. 
  - No TFA. [Twitter handle](https://twitter.com/Dreamstime).


mxxcon:
How would this fit our criteria to be listed?

Jawshy:
The site uses and stores personally identifiable information including, but not limited to, a user's full name, physical address, and phone number for the site's paid services. Billing information is also stored if a payment method other than PayPal is used. I believe that these would fit [site criteria](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria) 1 and 2.

PR #1946, opened 2016-07-30
Add Computer Alliance [Retail]

Jawshy:
#### Changes:
- Added: [Computer Alliance](http://www.computeralliance.com.au/) under retail. 
  - No TFA. No Twitter handle.


_data/retail.yml
```diff
@@ -31,6 +31,11 @@ websites:
       tfa: No
       twitter: comixology
 
+    - name: Computer Alliance
+      url: http://www.computeralliance.com.au/
+      img: computeralliance.png
+      tfa: No
+
```
Carlgo11, line 45: I suggest `facebook: ComputerAllianceAus` as their Twitter doesn't seem to be monitored by them.

Jawshy, line 45: Added.

Carlgo11:
Again, by the looks of it I wouldn't classify it as a significant site.
Plus, they could start with supporting TLS before 2FA 😜

Jawshy:
> Plus, they could start with supporting TLS before 2FA :stuck_out_tongue_winking_eye:

Damn right! 
[DAMN RIGHT !](https://www.youtube.com/watch?v=gHQhlhz-_K0)

PR #1950, opened 2016-07-31
Added Evolution Host [Hosting]

WatakiWatako:


Carlgo11:
By the looks of it I'd say this site does not qualify to being listed here as it seems to be a really small hosting provider.
We have to pick which sites are significant enough as we can't list all sites on the entire internet here.
(Except from the lag that would create I'm pretty sure @GitHub wouldn't appreciate us filling up their server space for that 😉)

PR #1957, opened 2016-08-03
Added site: bluecross.ca

tfriesen:


tfriesen:
I'm going to withdraw this pull request because it turns out to be a bit more complicated than I originally though. Blue Cross Canada is a major provider of health insurance in Canada, and allows users to submit claims, view old claims, and use direct deposit, I believe it meets all of the site criteria.

However, one does not log into bluecross.ca per se, but instead each province has its own website and log in forms. It's not immediately clear what the relationship between Blue Cross Canada and its member orgs is, or how much independence they have.

PR #1961, opened 2016-08-05
Add visa checkout

stephengroat:


Carlgo11:
I'm unfamiliar with this service but isn't it somehow connected to your bank?
As they don't list any EU countries in the registration form that would suggest that they indeed don't have 2FA (I'm pretty sure they're obligated by law to support hardware 2FA in any EU country) but if they now support, let's say, bank of 'murica's cards then they might support that bank's 2FA system?

stephengroat:
Not connected to a bank, just acts as a mobile wallet for credit cards.
Similar to Google Wallet or any of the other 3rd party mobile wallet
systems that are popping up

PR #1969, opened 2016-08-13
Add dell

stephengroat:


_data/retail.yml
```diff
@@ -31,6 +31,12 @@ websites:
       tfa: No
       twitter: comixology
 
+    - name: Dell
+      url: http://www.dell.com
```
Jawshy, line 35: Their site appears to support HTTPS, but note it also includes HTTP resources.

stephengroat, line 35: A raw pull to https://www.dell.com redirects to http://www.dell.com.

PR #1971, opened 2016-08-13
Add home depot

stephengroat:


_data/retail.yml
```diff
@@ -72,6 +72,12 @@ websites:
       img: groupon.png
       tfa: No
 
+    - name: Home Depot
+      url: http://www.homedepot.com
```
Jawshy, line 76: When using HTTPS I receive a certificate error saying: 

> There are issues with the site's certificate chain (net::ERR_CERT_COMMON_NAME_INVALID).

Is anyone else experiencing this?

RichJeanes, line 76: No, it's because it's just using their CDN's default stuff. The CN listed on the cert is `a248.e.akamai.net`. In other words, their home page is not properly configured for HTTPS.

Their secure pages are actually on a separate subdomain that is properly configured, `secure2.homedepot.com`. Even then they use mixed content so the URL bar isn't green.

Carlgo11:
Could you add `facebook: homedepot` as well?
Just to give options for the users.

stephengroat:
facebook added

stephengroat:
wow. that was fast

PR #1975, opened 2016-08-14
Add kohls

stephengroat:


Carlgo11:
Interesting. Most companies move away from the `'` nowadays but this company still uses it.

Carlgo11:
@2factorauth/collaborators

PR #1983, opened 2016-08-17
Add VyStar

kevindb:


Carlgo11:
The documentation page you linked to is a bit fuzzy when it comes to describing the actual system.
Is a user required to get a code from their phone on every login attempt or just on some that the system classifies as extra suspicious?

kevindb:
I, too, am dismayed at VyStar's documentation. Great interest rates, poor security.

Prompted for SMS or telephone code based on suspicious activity, including first-time login from new browser. Subsequent logins from the same browser require only the user/password.

Carlgo11:
Okay then it's not two factor authentication by our definition.
https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions

So either set it to `tfa: No` or close the PR.

kevindb:
Thank you for correcting me. I misunderstood the definition that not requiring TFA _every_ time is not TFA (under your definition).
I've updated my PR to match.

P.S. Thank you for this tool. It's incredibly helpful.

stephengroat:
I don't want to cause problems, but a LOT of the sites that are currently listed as `tfa: Yes` follow the same scheme. Dropbox, Google, and Github all use a cookie-based system for first time login from new browsers and then redo password (1factor auth) for higher sensitivity operations

RichJeanes:
However, a lot of those sites provide user choice in the matter. For example, Google requires 2FA _at least_ every 30 days, but the user can opt to be prompted every time by simply unchecking the box at login.

Though, I will agree, the list could use an audit. That's just a rather daunting task...

kevindb:
To clarify, in the case of VyStar, there is no user control for 2FA. Once you have logged in, you are _not_ prompted for 2FA using that browser again (likely using cookie).

I am in favor of the strictest definition of '2FA'. In fact, I support dropping SMS and automated phone calls from the definition, especially since NIST has depricated SMS:
https://pages.nist.gov/800-63-3/sp800-63b.html#out-of-band-verifiers

stephengroat:
NIST is beginning the phase-out of SMS and phone calls, but has not officially dropped them yet. From the draft release:

```
OOB (out of band verification) using SMS is deprecated, and may no longer be allowed in future releases of this guidance.
```

Carlgo11:
@stephengroat Yes but that doesn't have anything to do with this PR.
Try and keep off topic discussion to a minimum in PRs and Issues. 😉

PR #1989, opened 2016-08-24
Webde

luh2:
add web.de mailer


Carlgo11:
I suggest you use this image instead.
It's a bit less pixely 😉 
fsie74o2

luh2:
done

stephengroat:
Do we want to add the `lang: de` tag to this as well?

Carlgo11:
Yeah that sounds like a good idea

Carlgo11:
Since this is the #1989:th PR here's an obligatory Tay Tay gif 😄

Jawshy:
~~~~

PR #1993, opened 2016-08-25
Add Sentry

stephengroat:


_data/developer.yml
```diff
@@ -279,6 +279,15 @@ websites:
       doc: https://semaphoreci.com/docs/two-step-verification.html
       software: Yes
 
+    - name: Sentry
+      url: https://getsentry.com
+      img: sentry.png
+      tfa: Yes
+      sms: Yes
+      hardware: Yes
+      software: Yes
```
Carlgo11, line 288: I can't find anything about software or SMS on that doc page.

stephengroat: In the blog post, first line:
    We think security is important so we’ve rolled out two-factor authentication for Sentry. You can use U2F devices (like yubikeys), Google Authenticator (or similar apps supporting TOTP), or SMS as second factor now.

Carlgo11: oh! 😄

stephengroat: I also signed up for an account and confirmed all 3 options are valid

PR #2002, opened 2016-08-27
Add monster

stephengroat:


_data/other.yml
```diff
@@ -220,6 +220,13 @@ websites:
       sms: Yes
       software: Yes
 
+    - name: Monster
+      url: http://www.monster.com
```
fpigerre: @stephengroat, It looks like Monster supports https. Would it be possible to update the URL to reflect that?

fpigerre (changes requested):

stephengroat:
@psgs updated

PR #2006, opened 2016-08-29
Add 123 Reg

AndyMDoyle:


Jawshy:
@AndyMDoyle, it's difficult to read the icon's text at that size. Would you like to use the following icon instead? 
123reg 
123 Reg also has a Facebook page at [123-reg](https://www.facebook.com/123regfans). Would you like to add that as well? 

``` ruby
      facebook: 123regfans
```

AndyMDoyle:
That's a much better icon. I toyed with making a modified version myself but yours does the job great.

Thanks.

stephengroat (approved):

PR #2007, opened 2016-08-30
Add ikea

stephengroat:
in honor of college starting in the US, let's secure the college furniture store of the US


Carlgo11:
*cough\* but IKEA is Swedish 😉

stephengroat:
touché

IKEA is really a celebration of cheap people everywhere, not just US college students. Obviously a nationalistic blunder on my part :)

_data/retail.yml
```diff
@@ -103,6 +103,13 @@ websites:
       twitter: hypop
       tfa: No
 
+    - name: IKEA
+      url: http://www.ikea.com
+      img: ikea.png
+      tfa: No
+      twitter: IKEA
```
Carlgo11: How about IKEAUSA here as well? They seem to actually use that account.

Carlgo11:
I feel like it's my duty as a Swedish citizen to provide a better image to use.
Here have this one: 
IKEA  
<br>  
But, well... this image also works I suppose...
För kung och fosterland

stephengroat:
Good catch, I just checked to make sure that the account existed.

And if you can shrink it down, maybe it should be made the tooltip for ikea, just to remind everyone about sweeden's greatness

Carlgo11:
That could work. Or we could change the entire page's background to the great Swedish flag and an `<audio>` tag to play the national anthem.

PR #2010, opened 2016-08-30
Update backup with facebooks

stephengroat:


fpigerre (approved):

_data/backup.yml
```diff
@@ -41,12 +41,14 @@ websites:
     twitter: cloudapp
     img: cloudapp.png
     tfa: No
+    facebook: getcloudapp
 
   - name: CrashPlan
     url: http://www.crashplan.com
```
Jawshy: They are using HTTPS now. @stephengroat, would you like update the URL here or should it be done in a separate pull request?

Jawshy:
Second `looks-good`[†](https://github.com/2factorauth/twofactorauth/pull/2010#discussion_r83551307) unless we are going to be strict enough to have the `twitter` and `facebook` fields ordered directly above/below each other.

PR #2013, opened 2016-08-31
Add ping identity

radeksimko:
https://www.pingidentity.com/en/products/pingid.html


Carlgo11:
How is this not a two factor authentication provider?

radeksimko:
@Carlgo11 I'm not sure I follow? It does support MFA as described in the link.

Jawshy:
@radeksimko, it's difficult to read words when the logo is shrunk down to that size. Would you like to use this image instead? 
pingidentity

stephengroat:
@radeksimko This is a provider of MFA SW, not a site that supports MFA itself. I made an account at https://www.pingidentity.com/en/account/sign-on.html and there was no MFA options.

The 2FA site is focused on listing sites that provide MFA options (and shaming those that don't). For example, do you know any sites that use Ping Identity? Those may be valid options

radeksimko:
@stephengroat I see what do you mean. PI provides couple of products and indeed the PingID is the part of PI portfolio which seems to be purely responsible for the MFA bit.

That said the Ping One offering seems to be integrated w/ PingID - hence it should support MFA in that sense. https://www.pingidentity.com/en/products/pingone.html

I personally have experience only with their on-premise offering (which does support MFA) so I'd need to sign up for the Ping One and see how that works in reality - according to the product page it _should_ have the MFA option though.

> The 2FA site is focused on listing sites that provide MFA options (and shaming those that don't). For example, do you know any sites that use Ping Identity? Those may be valid options

I'm happy to accept this and close the PR if this is really the case, but first I'd like to understand why OneLogin and Okta are in the list because they work pretty much the same way - i.e. they don't provide any other service except providing the auth bit + MFA for other services - i.e. they act as identity providers.

If there is no place for identity providers in the list, then I assume Okta + OneLogin should be removed. 😉

stephengroat:
@radeksimko I completely missed the PingOne stuff. Let me look into when I get a chance. PingOne is a "cloud-based" product, right (not on premise)?

stephengroat (changes requested):
@radeksimko any comments on whether PingOne is a SaaS product? This project tries not to focus on self-hosted products, but SaaS products instead.

Jawshy:
Any updates on this pull request, @radeksimko and @stephengroat?

Carlgo11:
Closing due to inactivity by the author.

PR #2014, opened 2016-08-31
Add Andrews Federal Credit Union [Banking]

umeditor:


stephengroat:
Try using https://tinypng.com to compress the image.

Jawshy:
@umeditor, I think this image would be better because it lacks text which is hard to read at such low resolutions. 
andrewsfederalcu

They also have a [Facebook page](https://www.facebook.com/AndrewsFederal). Would you like to add that as well? 

``` ruby
      facebook: AndrewsFederal
```

umeditor:
I added the Facebook info and ran the image through https://tinypng.com.

umeditor:
…and add the image without text.

fpigerre:
Thanks for the pull request @umeditor! :+1:

PR #2016, opened 2016-09-03
Added Poloniex [Cryptocurrencies]

MickD:


stephengroat:
Thanks for the pr @MickD . Could you push this to `master` instead of `gh-pages`? we merge from `master` to `gh-pages`

MickD:
Oops, will do

PR #2018, opened 2016-09-04
Add BofI Federal Bank [Banking]

Jawshy:
#### Changes:
- Added: [BofI Federal Bank](https://www.bofifederalbank.com/) under banking. 
  - No TFA. [Twitter handle](https://twitter.com/BofIFederalBank). [Facebook page](https://www.facebook.com/BofiFederalBank). 

The [previous pull request for this site](https://github.com/2factorauth/twofactorauth/pull/1911) said that the site uses two-factor authentication, specifically email and hardware options. However, the documentation that was linked did not specify the aforementioned types of 2FA and said that MFA was only for first login. 

> NOTE: Protecting your online identity is extremely important to us. When logging in for the first time you will be establishing your credentials for the multi-factor authentication security processes we use to identify users and companies. 

\- [bofi-federal-bank-online-banking-instructions.pdf](https://www.bofifederalbank.com/Portals/0/Documents/BofIFederalBank/bofi-federal-bank-online-banking-instructions.pdf)


fpigerre:
Thanks for the pull request @Jawshy! :+1:

Carlgo11:
Looks like they don't monitor their Twitter and Facebook.
And with only 700 likes & 400 followers I begin to question if it's really worth adding them to the list.
Plus, from looking at their Facebook account it looks like 2FA isn't their only problem.

Carlgo11:
But if we are going to add them, this is their support email: `customerservice@bofifederalbank.com`

PR #2020, opened 2016-09-05
Add Cloudways [Hosting]

Jawshy:
Closes https://github.com/2factorauth/twofactorauth/issues/2017. 
#### Changes:
- Added: [Cloudways](https://www.cloudways.com/) under hosting. 
  - [Software token TFA](https://support.cloudways.com/enabling-two-factor-authentication-for-your-cloudways-account/) (Google Authenticator and Authy).


Carlgo11:
So it's a RFC 2289 system and no native Authy support?

Jawshy:
As far as I can tell from the doc, yes.

PR #2021, opened 2016-09-08
Add GeForce (Nvidia) [Gaming]

Jawshy:
Changes: 
- Added: [Nvidia](http://www.nvidia.com/) under gaming. 
  - No TFA. [Account page](https://account.nvidia.com/). [Twitter handle](https://twitter.com/nvidia) and [Facebook page](https://www.facebook.com/NVIDIA).


stephengroat:
@Jawshy are you referring to the authentication for their shop?

Jawshy:
@stephengroat, no, I'm only referring to the account which can be used to log into their [GeForce Experience](http://www.geforce.com/geforce-experience) application. Their [online store](http://store.nvidia.com/store/nvidia/DisplayEditProfilePage) must require different credentials as I can't log into it when using the same credentials I would use for their application.

stephengroat:
Maybe change the name to `GeForce` and the website url to `http://www.geforce.com/`?

Nvidia just looks like the shop

Jawshy:
Sure, I'll change it over to `GeForce`, but could `Nvidia` (in parenthesis) also be appended at the end as people may search using "Nvidia" as the keyword?

Jawshy:
Shall I also change the social media links from `NVIDIA` to `NVIDIAGeForce`?

stephengroat:
Yes to both

Jawshy:
Done.

PR #2030, opened 2016-09-13
Add Mercado Livre, Brazilian branch of Mercado Libre

jrbenito:
-Add Mercado Livre, a Brazilian branch of Mercado Libre
-Also add remark to Mercado Libre regarding it only supporting 2FA using
Authy app and no others

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -139,7 +139,19 @@ websites:
       tfa: Yes
       software: Yes
       sms: Yes
+      exceptions:
+          text: "Mercado Libre only supports 2FA with specifc app"
       doc: https://ayuda.mercadolibre.com.ar/ayuda/Consejos-para-fortalecer-la-se_961
+
+    - name: Mercado Livre
+      url: https://www.mercadolivre.com.br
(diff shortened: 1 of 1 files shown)


RichJeanes:
Does Mercado Libre have separate accounts for each country's site? We specifically made the URL point to their country selection page since they operate so many separate sites. It would be quite a mess to list them all individually. The documentation just happens to point to the Argentinian page only because it's the first on the list (alphabetical) and they don't have a country agnostic support site.

As far as making an exception for their use of Authy, we usually only single sites out if they use a proprietary method, such as Steam or Blizzard since it means having an extra solution (aka an inconvenience) just for one service.

jrbenito:
Yes, the account for MercadoLivre does not work on MercadoLibre. However, they do have almost the same page flow while Brazilian page is in Portuguese and  and Argentinian's is in Spanish.

Although I understand your point regarding exceptions, I personally consider being forced to use one single app beside any other compatible of my choice an inconvenience. But this is personal taste right? Also I do agree that Steam method is worst.

Therefore, I think you can drop this PR.

PR #2035, opened 2016-09-15
Add isaca

stephengroat:
Don't love the image, didn't see any better options

All of these security cert companies deserve extra special love


Carlgo11:
And with love you mean hate?

stephengroat:
not love or hate, just attention. extra hugs, if you will

Jawshy:
@stephengroat, try this icon. It's essentially their favicon, but at 32² pixels. 
isaca

stephengroat:
@Jawshy usually your images are the best, but i can't see anything with that one (it looks too dark). is that my monitor or can you see any detail there?

Jawshy:
I agree, but that's their icon for you; poor contrast between the symbol and background. Here's it with a transparent background. 
isaca-no-bkg-s

stephengroat:
thanks for that, but i think i'm going to stick with what i have (even though it's hard to read). i looked through their website and never saw one instance of isaca representing themselves using that symbol as their logo, plus it's chopped off (quite a bit) if you compare it to the full logo.
thanks though!

fpigerre (approved):

_data/education.yml
```diff
@@ -72,7 +72,14 @@ websites:
       tfa: No
       exceptions:
           text: "Illinois Tech uses Google Apps for Eductation for their email; so they could very easily add tfa to their email, but at the time of this commit, they have it disabled and do not plan to implement it any time soon."
-       
+
+    - name: ISACA
```
Jawshy, line 76: According to [sources on Wikipedia](https://en.wikipedia.org/w/index.php?title=ISACA&oldid=744441453#cite_note-isaca1-1), **ISACA** was previously known as **Information Systems Audit and Control Association**. For this reason I ask if that name should be included alongside its current name or is that unnecessary?

stephengroat: It's just so long it's going to look horrible on the site for small pages, don't you think?

Jawshy: Most likely, yes.

PR #2043, opened 2016-09-16
Cleanup verify

stephengroat:
This is an idea.

I enabled rubocop, which is an **aggressive** ruby linting tool. you can see that i disable a few of the check for the of the more complex functions, but this does force uniformity in the ruby code.

adding a .rubocop.yml is also an option (instead of the inline `# rubocop:disable` comments)

general comments or hatred welcome

next goal is to get/make a linting tool that tools at the editorconfig for all files

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,4 +9,5 @@ gem 'kramdown', '~> 1.10.0'
 group :test do
   gem 'rake'
   gem 'fastimage'
+  gem 'rubocop'
 end
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,9 @@
-task :default => [:verify]
+require 'rubocop/rake_task'
+
+RuboCop::RakeTask.new
+
+task default: %w(verify rubocop)
 
 task :verify do
-    ruby "./verify.rb"
+  ruby './verify.rb'
 end
--- a/verify.rb
+++ b/verify.rb
@@ -17,122 +17,111 @@ require 'fastimage'
 # TFA forms
 @tfa_forms = %w(email hardware software sms phone)
 
-# Should the script ignore checking the image size?
-@ignore_image_size = false
-
 # Image max size (in bytes)
-@image_max_size = 2500
+@img_max_size = 2500
+
+# Image dimensions
+@img_dimensions = [32, 32]
 
 # Image format used for all images in the 'img/' directories.
-@image_extension = ".png"
+@img_extension = '.png'
+
+# Send error message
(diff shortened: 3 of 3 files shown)


mxxcon:
So what's the advantage of this over what we already have?

Carlgo11:
@mxxcon it's just a cleanup. No advantages more than that the code looks nicer

mxxcon:
But we now have a new dependency on `rubocop` gem, don't we?

stephengroat:
adds a gem dependency only for test. my goal is to add a lot more to the test
1. figure out more extensive tests to run just on sections updated by a git pull (url checking etc)
2. get an internal section that uses apis (twitter, facebook, domaintally). since the only PRs that are made by members can use API keys, maybe start creating an internal PR from master to gh-pages to trigger travis to do those

i think it would be a a great goal to automate ourselves out of existence

Carlgo11:
psst, rebase when you're done 😉

stephengroat:
will do. geez, it's like the guy trying to commit stuff for code cleanliness should have clean commits. seems like a lot of work ...

Carlgo11 (commented):
The code looks nice but I don't know how good of an idea it is to introduce yet another way the build can fail.
My view on the build statuses is that checks should be run for things essentials for the project but not for cosmetic purposes only.

Maybe instead of adding the rubocop gem you could use this PR to tidy up the verify.rb script?

stephengroat:
I agree it's dependency/another way for the build to fail, but I'm not sure I agree with:

> My view on the build statuses is that checks should be run for things essentials for the project but not for cosmetic purposes only.

Why not check whatever pieces can be automatically checked? I would think that the eventual goal of this project would be that every piece of a PR is checked during the CI process and only minimal human review is needed (i.e. i saw that previous URL checking produced timeouts during the build process, but what if only the URLs affected by the PR were checked (reducing the URL checking surface area)).

P.S. i don't mean to be an ass

Carlgo11:
With cosmetic purposes I mean things like formatting and white space. URL checking is not something cosmetic.

Making the code look nice is a good thing but the build shouldn't fail just because of formatting.

Carlgo11:
(Accidentally closed the PR. Opened it again :sweat_smile:)

stephengroat:
I we don't allow PRs if they remove trailing whitespace, shouldn't that be part of the automated checks?

fpigerre (approved):

PR #2046, opened 2016-09-19
Add carilion clinic

stephengroat:
specifically targeting the https://mychart.carilionclinic.org product, but that's the only obviously available publically facing login from the homepage


fpigerre (approved):

Jawshy:
They have a contact email address at the bottom of their website. Should that be added? 

``` ruby
      email_address: direct@carilionclinic.org
```

Issue #585, opened 2014-06-16
Reanalyze 1Password (again)


jdavis:
1Password is incredibly responsive on Twitter and I've been watching their responses and Sam tweeted me this conversation as well: https://twitter.com/SamuelDebruyn/status/478550817432039424

Basically 1Password still believes they dont' belong on the list. They have a paper explaining it: http://cl.ly/text/302c041B1M0c

What does everyone think? I know @mxxcon was the one that argued that it should be re-added.

mxxcon:
@jdavis the way I see that paper is they decided not to implement tfa/mfa because it's too difficult for users, not because they are immune to the problems tfa/mfa solves.
Paragraphs 7 and 8 in their own article says that they could implement mfa using various approaches but decided not to.
That twitter conversation is once again focusing too much on _tfa for a service_ point of view. It is a common use-case but not an exclusive one. tfa can be implemented in stand-alone desktop applications as well.

OneJaeAtATime:
I agree with 1Password on this. The services you sync through are the ones
that use 2FA. 1Password doesn't use their own servers to sync anything.
Nothing is seen by them and nothing is authenticated by them. The sync
services look at the files where you sync them from for changes and other
things.

mxxcon:
@jonathanbloom Two/Multi-factor authentication can exist not only in online services world.
Previous example that I used is KeePass. See http://keepass.info/help/base/keys.html#keyfiles
- One master password decrypts the complete database.
- Alternatively you can use key files. Key files provide better security than master passwords in most cases. You only have to carry the key file with you, for example on a floppy disk, USB stick, or you can burn it onto a CD. Of course, you shouldn't lose this disk then.
- For even more security you can combine the above two methods: the database then requires the key file and the password in order to be unlocked. Even if you lose your key file, the database would remain secure.

Without a password(something you know) and a key file on some sort of removable storage(something you have) you can't access the database.
1Password can implement such feature as well.

@jdavis If it's a really contentious issue, perhaps could we compromise on showing they don't have any tfa, but also an exception saying that they made a business decision not to implement it and link to that article.

jpgoldberg:
I work for AgileBits, the makers of 1Password.

Two factor authentication may be a very fine hammer, but 1Password is not a nail.

[...]

Issue #422, opened 2014-04-09
1Password's response


jdavis:
1Password isn't a hosted solution. It probably doesn't belong on the list as they pointed out: https://twitter.com/1Password/status/453706163402444801

mxxcon:
I disagree with this.
Two factor authentication is considered to be: something you know(your username/passwords) _and_ something you have.
It seems like we slowly transitioned into thinking that "something you have" means only OTP for online services. But this not the case.
For example take KeePass(http://keepass.info/features.html) password manager.
It supports encrypting your database with your master password AND/OR a key file.

To quote their site:
- One master password decrypts the complete database.
- Alternatively you can use key files. Key files provide better security than master passwords in most cases. You only have to carry the key file with you, for example on a floppy disk, USB stick, or you can burn it onto a CD. Of course, you shouldn't lose this disk then.
- For even more security you can combine the above two methods: the database then requires the key file and the password in order to be unlocked. Even if you lose your key file, the database would remain secure.

So two-factor authentication is not only for online services. It can be OTP solutions, but it can also be encryption key files, solutions like YubiKey or even nfc and lower-power bluetooth.

jdavis:
Okay, I am not opposed :+1:

How should we handle 1Password? Should there be a section for apps? What do you think, @mxxcon?

I'd also love input from some of the other active people lately: @jamcat22, @ilyakatz, @smholloway

mxxcon:
I don't think a separate section is necessary. 1Password fits into our existing sections along with LastPass and KeePass.
1Password would marked as "no" and if we were to add KeePass it would be marked as `software: yes`

But yes, more input from other folks would be nice.

ilyakatz:
I vote to keep separation by use case (the sections that exist now) and not implementation (cloud vs app). The line between an offline app and cloud service is kind of blurry in many cases

jdavis:
Okay, the reason I asked about the separate section was, as trivial as it sounds, because the description of the site is 'List of sites...'

My brain just didn't like the mismatch.

jdavis:
I'm totally cool with adding it as is to Identity section.

jamcat22:
I agree with @ilyakatz. We should keep it separated by use case.

jdavis:
I love Git. This is how easy it was to re-add 1Password:

``` bash
$ g cp b6223d85c516b0b173ba32c8c5b49cc04ff1f5ad
```

[...]

PR #2048, opened 2016-09-20
Readd 1Password

Carlgo11:
### Rework exceptions.link YAML tag

The old way the `exceptions.link` tag worked was by redirecting to the `restrictions` page however since that page no longer exists as per #1637 I thought I could change what the tag links to.
This PR changes it to work as a documentation link for non 2FA  compliant sites.
For an example see bc56ae9.
### Add 1Password again

In #422 and #585 the 1Password authors asked us to remove them from our list as their views were that, since the 1Password client only used a decryption password, 2FA wasn't needed for 1Password.
The @2factorauth members have discussed this a lot and come to the conclusion that we disagree with their stand on 2FA. Especially since they now have a second authentication step (Don't really understand this but @RichJeanes apparently does so I'll let him explain this more in depth.)

I've also discussed this with some other colleges in the 2FA industry, including an employee of AgileBits, whom have agreed with me that they should be using 2FA.
I therefore made this proposal to add 1Password again.


RichJeanes:
EDIT:
Doing some closer reading, they say you need your Account Key the first time you sign on with a new browser, which makes the statement that you can see it if you sign in confusing at face value. It seems with "viewable by signing in to your account on 1Password.com", they should have appended "on a previously authorized device". This leaves the Account Key as just a split key, just without the usual hassle of creating, storing, and transporting one. While I appreciate the step forward in making stronger security more user-friendly, there is still an online portion of the account, even if used for nothing but billing the annual subscription fee, that seems to be left unprotected. While it seems overkill to have 2FA for something that you only need to use once per year, it is something to consider with 1Password's shift from Software to Service.

Now that it has been around for a while and I have spent more time with the documentation, outside of the one point I make above, I actually have to conclude that the situation with 1Password has not changed and that the core of their functionality is still reliant on local decryption, not remote authentication, therefore 2FA still does not seem to apply.

@jpgoldberg, perhaps your input could be useful here if I am still mistaken.
/EDIT

1Password uses what they are calling an "Account Key", claiming it is "Better than two-factor™" (trademark theirs). It is a 128-bit string that is used for encryption (basically as key splitting in conjunction with your password). ~~I'm of mixed feelings about it, but~~ I have to agree with them that it is, by definition, not Two Factor authentication. 

In the most basic terms, the Account Key is a secondary password. It is static, therefore once compromised, must be changed manually (if possible? Requires more research). They claim "your Account Key is never sent over the Internet" but contradict themselves with "it's... viewable by signing in to your account on 1Password.com". So while it is never transmitted as a form of authentication, only actively used locally for decryption, I find the claim that it is "never sent over the internet" ~~and "unlike 2FA, it cannot be reset, intercepted, or evaded"~~ to be very misleading (at best). (EDIT: With all authentication methods that require something generated by the other party, it has to be transmitted over the internet at some point to get to you, eg. the seed for a TOTP token, so to say the Account Key is "never" sent is still misleading. The "cannot be reset, intercepted, or evaded" does hold true since it is used as a split keyfor local encryption)

While they now provide a hosted password service, I imagine that it is simply file storage for your encrypted password vault that is then downloaded and decrypted by a local client (even when using the web client). With this in mind, I'm sure we will get the same discussion as last time where someone will claim that this means 2FA doesn't apply since authentication is not taking place. 

~~I STRONGLY disagree with that. With stating that your account key is "viewable by signing in to your account on 1Password.com", they are saying that you can view the account key with just your password. So while your password vault may be protected by key splitting (Master Key + Account Key), your Account Key is available to anyone with your Master Password, completely negating it's benefit. This could be mitigated by providing 2FA for online authentication for account management while still providing the Account Key for the key splitting and local decryption of your vault (without authentication, as argued by the developers).~~

~~What I have to conclude from this is only that your 1Password vault is now more vulnerable than before. Before 1Password was a hosted service, you did not have to authenticate with your Master Password to a website, therefore your password (and it's hash) never existed anywhere you didn't put it. Now the same password you use to encrypt your vault is living as a hash on 1Password's servers and has to be used for online authentication. In addition, your vault is now being transmitted in the same channel as your password authentication, whereas before moving your vault was your responsibility and your password was never sent to a remote server.~~ I feel this shift from Software to Service is a big step backwards for the ~~security (not to mention~~ flexibility~~)~~ that 1Password has always lauded as  (EDIT: one of) their ~~biggest~~ selling point(s).

All quotes were taken from https://support.1password.com/understanding-account-key/

mxxcon:
@RichJeanes Once again i disagree with you on 1pass :) and even more so with the current state.
I'm using "1pass for teams" at work and their online component is more than just for billing. You perform all account management tasks online, you add/remove users online, you give users access to vaults online, you can access the actual vaults online too.

It is possible to regenerate Account Key. It is also fully visible in plain text from your own account. It's possible that's not transmitted over the wire but retrieved from local storage/cookies.

And funny enough, they now have this:
screen shot 2016-09-20 at 4 40 24 pm

stephengroat:
Correct me if I'm wrong, but whether the Account Key is a split key that is or isn't transmitted in whole, isn't it still just knowledge (something known), not possession(something they have) or inherence(something they are)? Does the definition of 2FA for the site require more than one category?

Carlgo11:
@stephengroat you are right. It's still knowledge.

RichJeanes:
@mxxcon I was not looking at 1Password for Teams. If it applies there, then I would argue that 1PfT should be listed separately.
It does not appear (from the documentation that I have seen) that a personal 1Password account can have the Account Key reset without losing your vault since it is part of the encryption key. If it is different for 1PfT, then I believe that is more reason that it should be listed separately.
Also, if you have to provide your Account Key to log in and the device stores the Key for later use, then I would image that, yes, it is just displaying a local copy.

@stephengroat The three factors of authentication are:
- Something you know
- Something you have
- Something you are

Two/Multi Factor Authentication is the practice of requiring more than one of these factors to authenticate someone. Duplicating a single factor does not count as MFA (eg. two knowledge factors).

The issue with 1Password is more the difference between authentication and de/encryption. One of their developers made the argument that (old) 1Passord was only performing de/encryption, therefore Two Factor _Authentication_ doesn't apply. I would define authentication as "proving one's identity to another party". Since there is no other party involved in de/encryption, I tend to agree with said 1Password developer and I believe that still applies for the new hosted version of 1Password (Personal).
(See #585 for the whole discussion. Lots of interesting details. Worth the read if you have the time.)

jamcat22:
I'd just like to point out for clarification that 1Password's personal version is hosted using the exact same architecture as 1Password for Teams. They each use the same encryption and authentication methods and practices. In fact, when using 1Password's app, the app prompts you to sign in to "a team or family account" even when using an individual account, as the only difference is that the subdomain "my.1password.comhttp://my.1password.com" is used instead of a team or family specific subdomain.

RichJeanes:
Then why do they advertise that admins on Team accounts can reset Account Keys and then advertise that Account Keys cannot be reset on personal accounts? The way that Teams operates seems to be very different, architecturally, to allow for things like vault sharing and administration. Just because the app can handle both types of accounts doesn't mean that they are handled the same in the background. That's like saying "My image viewer can open both of these image files, therefore they must be the same format." Do you have any documentation to verify that they are the same?

jamcat22:
I have reached this conclusion upon reading how recovery groups work in the 1Password White Paper (available [here](https://1password.com/files/1Password%20for%20Teams%20White%20Paper.pdf)). One part of the paper mentions,

> If there are no members of a Recovery Group, the capacity to recover
> data is lost to the team.

which to me suggests that for personal accounts, either a recovery group isn't created, or one is created with no members. The rest of the architecture seems to be the same based on the details given in the white paper and on their support site.

One of their [articles](https://support.1password.com/sync-options-security/#1password-account-no-fuss-security-and-convenience) says,

> Most website have a password reset feature which relies on a master key in the possession of the service provider. Your 1Password account has no such key, so it can only be recovered by people you entrust. They alone can help you regain access if you forget your Master Password or lose your Account Key.

and, 

> Your 1Password account uses the SRP protocol to authenticate your login details without sending your Master Password over the Internet, so it can’t be stolen while it’s in transit.

_data/identity.yml
```diff
@@ -1,4 +1,15 @@
 websites:
+    - name: 1Password
+      url: https://
+      img: 1password.png
+      tfa: No
+      email_address: support@1password.com
+      facebook: 1Password
+      twitter: 1Password
+      exceptions:
+        text: "Although AgileBits, the company behind 1Password, feel that two factor authentication doesn't really fit 1Password, we disagree. Because of the way in which 1Password handles authentication and password data."
```
jamcat22: "Because of the way in which 1Password handles authentication and password data." sounds a bit like a sentence fragment.  It could be rephrase the sentence as, `"AgileBits, the company behind 1Password, feels that two factor authentication doesn't really fit into 1Password's offering. We disagree because of the way in which 1Password handles authentication and password data."`

_data/identity.yml
```diff
@@ -1,4 +1,15 @@
 websites:
+    - name: 1Password
+      url: https://
```
jamcat22: This URL should be set to `https://1password.com/`

jamcat22 (changes requested):
Found one incomplete URL and gave suggestions for rephrasing the exception text.

_data/identity.yml
RichJeanes: Copy paste error on your first example? That's still two sentences...
In the second example, sentence two, I would remove the comma. Unnecessary pause.

jpgoldberg:
I'm Jeffrey Goldberg from AgileBits. I would not consider our Account Key to be 2FA. It has very different security properties and is used to solve a different problem than 2FA is used for.

_data/identity.yml
jamcat22: Ah thanks. The form didn't save correctly. I'll fix that. Feel free to suggest a better version. I'll remove the comma.

jamcat22:
@jpgoldberg We currently don't consider Account Key to be a form of 2FA, which is why we were going to list 1Password as not supporting 2FA. How do you feel about this decision?

jpgoldberg:
Hi @jamcat22 

> We currently don't consider Account Key to be a form of 2FA, which is why we were going to list 1Password as not supporting 2FA. How do you feel about this decision?

I fully concur that that is the correct decision.

_data/identity.yml
```diff
@@ -1,4 +1,15 @@
 websites:
+    - name: 1Password
+      url: https://1password.com/
+      img: 1password.png
+      tfa: No
+      email_address: support@1password.com
+      facebook: 1Password
+      twitter: 1Password
+      exceptions:
+        text: "AgileBits, the company behind 1Password, feels that two factor authentication doesn't really fit into 1Password's offering. We disagree because of the way in which 1Password handles authentication and password data."
```
mxxcon: I feel this is an unnecessary qualification. Ultimately the reasons behind why 2fa is not implemented are irrelevant. Especially now that 1pass functions similarly to other online password managers. Your password database is no longer kept offline-only.

mxxcon:
@Carlgo11 Should `CONTRIBUTING.md` change be part of this PR?

_data/identity.yml
thegeekkid: I agree; however, since Agilebits has been very responsive, it's probably only fair that they are given a chance to defend themselves.

mxxcon: I don't see anything that they should defend considering they already have 2fa in the beta form.
Right now it's `tfa:no`, when, if ever, it gets released, it'll be `tfa:yes`
Or I guess at best it's `in progress`..

thegeekkid: True.  My main issue with not acknowledging the argument that they made is that there was obviously a difference of opinion as to whether or not tfa was important for them.  I disagreed with their viewpoint; but I think that's something that people should have been able to decide for themselves, and acknowledging that there was a different viewpoint would have done the trick.  I guess since they are implementing it now, they've pretty much conceded their previous stance.

mxxcon: They could argue that when 1pass was purely an offline solution. Now it's no longer that.

jamcat22:
@mxxcon I'd think so since we're changing how the `link:` tag works.

_data/identity.yml
jamcat22: According to @jpgoldberg, they agree with our decision. I believe this exception text is no longer necessary.

_data/identity.yml
jamcat22: I now believe this line should be removed.

Carlgo11:
@mxxcon @jamcat22 I removed the exceptions tag and therefore also the exception-rework from this PR.
See new commit history.

jamcat22 (approved):

PR #2057, opened 2016-09-24
Added exception to Blizzard in gaming.yaml

fdeitylink:
For software tokens, Blizzard requires that you use their own Battle.net Authenticator mobile app

### DIFF
--- a/_data/gaming.yml
+++ b/_data/gaming.yml
@@ -27,6 +27,8 @@ websites:
       sms: Yes
       software: Yes
       hardware: Yes
+      exceptions:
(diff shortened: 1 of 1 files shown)


RichJeanes:
@2factorauth/core Didn't we already have this at some point?

Carlgo11:
@RichJeanes Is [**this**](https://github.com/2factorauth/twofactorauth/pull/75) what you're thinking about?

RichJeanes:
No, I thought we had callouts for both Steam and Blizzard using proprietary apps. Mostly because it's an inconvenience for the end-user to have to have an app installed just for one site. Not saying that it should deter anyone from using it, just a nice heads up to give, I think.

PR #2060, opened 2016-09-25
Added exception to Humble Bundle in gaming.yaml

fdeitylink:
For software tokens, Humble Bundle requires that you use the Authy service


Carlgo11:
Same as #2061

PR #2061, opened 2016-09-25
Added exception to Twitch in entertainment.yaml

fdeitylink:
For software tokens, Twitch requires that you use the Authy service

### DIFF
--- a/_data/entertainment.yml
+++ b/_data/entertainment.yml
@@ -101,7 +101,7 @@ websites:
       software: Yes
       exceptions:
           text: "Software implementation requires use of the Authy service."
-      doc: http://blog.twitch.tv/2015/11/two-factor-authentication-now-available-on-your-twitch-account/
+      doc: https://blog.twitch.tv/two-factor-authentication-now-available-on-your-twitch-account-b03300862ba7#.o39docmjl
 
     - name: Ustream
       url: http://www.ustream.tv/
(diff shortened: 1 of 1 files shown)


_data/entertainment.yml
```diff
@@ -99,6 +99,8 @@ websites:
       tfa: Yes
       sms: Yes
       software: Yes
+      exceptions:
+          text: "Software implementation requires use of the Authy service."
       doc: http://blog.twitch.tv/2015/11/two-factor-authentication-now-available-on-your-twitch-account/
```
Jawshy: Their doc is now available over `HTTPS`. Would you like to add that change to your PR?

fdeitylink: Done. There was also a redirect setup to the new URL "format", which now doesn't contain the date of the post

Carlgo11:
@FDeityLink I don't see why we would add an exception for this. We generally don't make exceptions for sites using one specific implementation as it doesn't make it less of a TFA implementation than any other site listed on here.

fdeitylink:
Well there was an exception already in about Steam requiring that you use their own app, so I figured we were documenting all cases of 2FA that requires a certain service/app.

RichJeanes:
But the difference is that Authy can be your token for many sites, whereas Steam and Blizzard use proprietary apps that only work for their site.

Carlgo11:
Closing as we don't add exceptions for these kinds of things.
Sorry @FDeityLink 😦 
Hopefully you'll find something else to contribute with.

PR #2063, opened 2016-09-26
Add Carbon Black Community Exchange

thegeekkid:
Forum/user support for Carbon Black product administrators.  Often contains somewhat sensitive data about our network environments and in some cases there are discussions on privilege escalation tactics that we have used to bypass Carbon Black products during pen-testing.  As of yet, there is no support for 2fa.


stephengroat (dismissed):

stephengroat:
@2factorauth/collaborators

_data/communication.yml
```diff
@@ -13,6 +13,13 @@ websites:
       sms: Yes
       doc: https://help.37signals.com/sign-in/questions/460-how-do-i-set-up-and-use-phone-verification-for-logging-in
 
+    - name: Carbon Black Community Exchange
+      url: https://community.carbonblack.com/
```
fpigerre, line 17: @thegeekkid, Do you think it would be possible to change this link to https://www.carbonblack.com, since the community link redirects to a login page?

_data/communication.yml
```diff
@@ -13,6 +13,13 @@ websites:
...
+      img: cb.png
```
fpigerre: Also, would it be possible to rename the image to "carbonblack.png", so that it is clear which site the image is used for?

fpigerre (changes requested):

_data/communication.yml
thegeekkid: Sorry for the late response.  It's been a pretty crazy few weeks for me.
I certainly could; however, that is a completely different platform, and I don't know that there is anything on their main site that should require a login.  The community site however does have information that would cause some pretty significant security concerns if it were breached (which is why I submitted it).  Thoughts?

_data/communication.yml
thegeekkid: Yep.  Working on it now.

_data/communication.yml
stephengroat: with the name as `Carbon Black Community Exchange` and not `Carbon Black`, I'd keep the `url` pointing to the community site

fpigerre (approved):

Jawshy (approved):

Jawshy:
Thanks for the pull request, @thegeekkid.

Issue #149, opened 2014-03-19
Add Evernote


Pichu0102:
https://evernote.com/evernote/guide/mac/#11

zach-taylor:
Hi @Pichu0102 Evernote is already there under the "Backup and Sync" section

jdavis:
Thanks for helping out though! ^_^

PR #2064, opened 2016-09-26
Add Planning Center Online

thegeekkid:
PCO (focused around church planning and record keeping) contains personal information such as volunteer contact information, information about children being checked into the nursery, etc.  No 2fa is supported as of yet.


stephengroat (approved):

stephengroat:
Ranks #149,199 on Alexa (https://www.domaintally.com/www/planning.center)

@Carlgo11, what do you think?

thegeekkid:
Just some background; while it probably has relatively low traffic (at least compared to other websites), PCO is pretty much the "standard" of church planning software, and stores some pretty sensitive data and is a payment processor.  It's mainly the type of data that presents itself as a concern and why I submitted it.  :)

stephengroat:
@thegeekkid I'm not against adding it, i'm just adding some extra info. I'd like a general guideline setup for the minimum Alexa rank (or some other metric) for sites to be added. Just some background info, that's all. trying to start a discussion

thegeekkid:
@stephengroat , no worries... I understood that.  Just figured I should add some more information about it since not everyone would be familiar with it.  :)

Not a bad idea about a general guideline for a minimum rank.

stephengroat:
@2factorauth/collaborators

fpigerre:
Given that the service seems quite polished and professional, with potential for an increase in its Alexa ranking, I think it would be a good idea if twofactorauth.org listed this website. :+1:

PR #2065, opened 2016-09-27
Add government category

Carlgo11:
This is a PR to add a government category for services like the IRS, the Canadian IRS, the Swedish IRS, did I mention the IRS? Maybe this category should just be named the IRS? :wink: 

More sites to follow but these are the ones that already were listed here.


stephengroat:
maybe add healthcare.gov?

Carlgo11:
@stephengroat the PR is open for "edits" by maintainers so I think you should be able to push commits to the branch.
If not, do they support TFA?

Jawshy:
Will this new category be purely for government websites or will it also include [government-owned corporations](https://en.wikipedia.org/wiki/State-owned_enterprise)?

Carlgo11:
@Jawshy so I thought about this and think the best approach so to list 2nd level government organisations.
_I tried to make a chart for this but apparently I don't have any simple image editing program on this distro so I'll try and explain it as best as I can using this wonderful thing called <b>words</b>._
- 1st level government org - Actual government, senate, congress etc.
- 2nd level government org- Ministries/Offices.
- 3rd level government org - Local offices such as police, schools, etc.

So for example, we could list the education ministry but not a local school.
I know that this doesn't apply to all countries' but at least the majority of western countries are covered.

A more in depth explanation is soon to come. Whenever I actually have an image editor that's not GIMP.

Carlgo11:
## Usage of the Government category

### Levels of government

Here's a more in depth explanation of what **I** think should go in the _government_ category.
As I mentioned before I'll categorize the different layers of a typical western government in a reverse tree structure where the first layer is the layer with the most control.
Here's an example I made:  

example government category
_Disclaimer: I'm not an artist. This example is not made to look good but to help visualize the basics of what I'm talking about._
_I refer to offices/departments as ministries in this example as I have more knowledge of those than their US counterpart._

My view on this is that most people will encounter the 2nd level sometime in their lives. Whether it's the department of education, homeland security or healthcare you as a citizen will probably have to deal with them at least once in your lifetime.
That isn't true for the 3rd level as it's often more localized. For an example, in many countries the universities are 3rd level governments where the principal is directly assigned by the head of state.
_Sounds odd but that's the case. Probably because of a long history of wars between the educated and the rulers. Who knows..._

Like sites in all other categories are checked for importance I think the same should be done here. 3rd level governments are often important to so few that I think we simply should leave all of them out.

### Government owned companies

Government owned companies such as Amtrak and USPS [[1]](https://en.wikipedia.org/wiki/List_of_government-owned_companies#United_States) in the US should by my views **NOT** be listed here as they still are companies and often fit other categories.

Carlgo11:
bump @2factorauth/collaborators

stephengroat:
I think it's a good idea and, I know this is sounds stupid, but won't having only one more categories throw the alignment of the homepage off a bit? maybe add 5 ad the same time to balance it out?

Carlgo11:
No I don't think that's necessary.

fpigerre (approved):

Carlgo11:
Fixed merge conflicts. Can we merge this now?

PR #2072, opened 2016-09-30
Add ieee

stephengroat:


mxxcon:
How would this fit into our criteria?

stephengroat:
IEEE stores payment information, IEEE Collabratec is their new social professional network feature. There's more I can list if needed

fpigerre (approved):

Jawshy:
They have several contact and support email addresses listed [here](https://www.ieee.org/about/contact_center/index.html?WT.mc_id=hpf_contact) and I am not sure which one - if any - would be most appropriate to include.

fpigerre:
For an organisation as large and with as broad-ranging responsibilities as the IEEE, I think it might be best to avoid listing an email address, as social media would be better suited. Otherwise, however, the following email address could be used: `contactcenter@ieee.org`.

stephengroat:
decided not to list email

PR #2079, opened 2016-10-03
Update developer with facebook

stephengroat:


_data/developer.yml
```diff
@@ -224,6 +230,7 @@ websites:
     - name: Papertrail
       url: https://papertrailapp.com/
       twitter: papertrailapp
+      facebook: papertrail.io
```
fpigerre: It looks like papertrailapp.com and papertrail.io might be separate companies. Their websites seem quite different. They also have different logos.

Jawshy: I concur. The Facebook page for [papertrailapp.com](https://papertrailapp.com/) appears to be @[papertrailapp](https://www.facebook.com/papertrailapp/), but it is not verified and has virtually no activity.

_data/developer.yml
```diff
@@ -352,5 +365,6 @@ websites:
     - name: Zapier
       url: https://www.zapier.com
```
Jawshy: Their current URL is now `https://zapier.com/`. Could you please update the `url` to reflect this?

_data/developer.yml
```diff
@@ -310,12 +320,14 @@ websites:
...
       url: https://www.statuspage.io
       twitter: statuspageio
+      facebook: statuspageio
       img: statuspageio.png
       tfa: No
 
     - name: Taiga.io
       url: https://taiga.io
       twitter: taigaio
+      facebook: taigaio
```
Jawshy, line 330: Their last post appears to be from November 2015 and they do not post often. Should this still be added?

_data/developer.yml
```diff
@@ -291,6 +300,7 @@ websites:
     - name: SourceForge
       url: http://sourceforge.net/
```
Jawshy: They are now using HTTPS. Could you please update the `url`?

_data/developer.yml
```diff
@@ -114,6 +118,7 @@ websites:
     - name: Docker
       url: https://docker.com/
```
Jawshy: Their current URL is now `https://www.docker.com/`. Could you please update the `url` to reflect this?

_data/developer.yml
```diff
@@ -94,6 +97,7 @@ websites:
     - name: Codeship
       url: https://codeship.io/
```
Jawshy: I am getting `NET::ERR_CERT_COMMON_NAME_INVALID` for this URL. `https://codeship.com/` works fine. Perhaps use the latter instead?

stephengroat: :heavy_check_mark:

_data/developer.yml
stephengroat: :heavy_check_mark:

_data/developer.yml
stephengroat: :heavy_check_mark:

_data/developer.yml
stephengroat: why not? same possibility with emails, at least facebook is public shaming

_data/developer.yml
stephengroat: :heavy_check_mark:

_data/developer.yml
Jawshy: Indeed. 

_adds label_

PR #2086, opened 2016-10-07
Added Cloudflare SMS support and correct caps (Cloudflare)

i-rme:
Added Cloudflare SMS support and correct caps (Cloudflare)


stephengroat (changes requested):
Can you show docs that support `sms: Yes`? I use Clouldflare and don't have an SMS option

i-rme:
Close the request, I thought SMS could be used as failback because the setup process ask for you phone number.

image

Issue #2093, opened 2016-10-10
add raizers


stephengroat:
```
name: 
url: https://www.raizers.com
img: raizers.png
tfa: No
twitter: Raizers
```

![raizers](https://cloud.githubusercontent.com/assets/1159138/19241525/e5a7ac40-8ec3-11e6-9cb5-f640183ad00d.png)

kartik95:
Added Raizers to investing.yml
PR #2106

PR #2106, opened 2016-10-13
Added Raizers to investing.yml

kartik95:
Referencing #2093 

Added Raizers to investing.yml


stephengroat (changes requested):
Please correct the `img` to 
image

mxxcon (changes requested):
@kartik95 our logos must always be 32x32
Also I don't think `investing` is the correct category for this. This is somewhat similar to kickstarter, so it should be moved to `other`

stephengroat:
@mxxcon Kickstarter has much more of a product focus, while raizers is focused on giving investors equity in x company. I think that's more `investing`, right?

kartik95:
Should I move Raizers to investing or other? Even imo, Raizers is best fit under investing.

mxxcon (approved):

PR #2126, opened 2016-10-16
Make it look nicer on mobile

dsoegijono:
Hi all! Just made a few changes for mobile so it looks nicer. Let me know if you guys (dis)like the changes, or if anyone has any suggestions to make the site more usable on mobile!
- Use flexbox to align items (mostly so icon and name are center-aligned, and when name wraps it doesn't go below icon)
- Remove button text
- Move buttons from the right side to the bottom, and it shrinks if there are more buttons
- Change "Supported" text so it only takes 1 line

Before / After
rsz_before rsz_after

Example of text wrap:
text wrap


dsoegijono:
I think I screwed up the circle/border around icons, but now that I see it without it, I think it's clearer like this? Maybe we can use a square border instead since the icons are square.

Carlgo11:
We can revisit that but let's stick with the circles for now.

dsoegijono:
@Carlgo11 👍 
- [x] I'll fix that later today. I also need to add prefixes for flex [DONE]

Carlgo11:
@psgs you'll have to review this one. If you think it looks good then you can merge directly.
Pass GO and collect $20 ➡️

fpigerre (approved):

PR #2131, opened 2016-10-17
Add Hexonet [Domains]

vgarmash:


RichJeanes:
Please add the `img` tag to the entry.

If you need more info, please check out the [Wiki](https://github.com/2factorauth/twofactorauth/wiki) or the [Contributing doc](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md).

_data/domains.yml
```diff
@@ -201,6 +201,12 @@ websites:
       tfa: No
       twitter: HeartInternet
 
+    - name: Hexonet
+      url: http://www.hexonet.net
```
Jawshy: That URL directs to `https://www.hexonet.net/` (HTTPS). Please use that instead.

_data/domains.yml
```diff
@@ -201,6 +201,12 @@ websites:
...
+      tfa: Yes
+      software: Yes
+      doc: https://cp.hexonet.net/cp2/downloads/2FA.pdf
```
Jawshy, line 209: Do you know if they have the documentation available on a normal webpage - not a PDF file? If so then please use that instead.

Jawshy (changes requested):
HTTPS and doc.

_data/domains.yml
vgarmash: Fixed

vgarmash: This is the only documentation link about it. The link itself is located under the authentication form.

_data/domains.yml
vgarmash: This is the only documentation I have found. The link to 2FA instructions is located under the authentication form. 


There is nothing about it at https://wiki.hexonet.net/wiki/Main_Page.

Jawshy (approved):

Jawshy:
@vgarmash, would you like to use the following image instead? It just has some minor tweaks. 

hexonet

vgarmash:
@Jawshy Your picture is 715 bytes while mine is 315 bytes. The actual X symbol is smaller on yours which should look better after the auto-rounding. Although you have transparent background instead of white. If your website engine will add the white background automatically then it is probably OK to use yours. Otherwise it will look weird as all other icons have white background.

Jawshy:
For comparison, [10Dollar](https://github.com/2factorauth/twofactorauth/blob/361f47a081a44820c93da8399f5778d2ee61332c/img/domains/10dollar.png) has a white background whilst [123 Reg](https://github.com/2factorauth/twofactorauth/blob/992db3ee802d6a605f7fe7dbe4420bb75c37559b/img/domains/123reg.png) does not. They both appear with a white background on the [website](https://twofactorauth.org/).

vgarmash:
Ok then. I will put your picture instead of mine.

vgarmash:
@Jawshy Please check why the build fails now. By the way - the last warning is about file size. Could it be that your image is too big?

stephengroat:
#2173 fixes the issue

vgarmash:
Thanks, @stephengroat. In this case I will just wait and watch.

Jawshy (approved):
`looks-good`

fpigerre:
Thanks for the pull request @vgarmash! :+1:

PR #2132, opened 2016-10-18
Add gravatar

stephengroat:


mxxcon:
Gravatar uses wordpress.com login, doesn't it?

Also is it worth listing it? It's just your avatar hosting..

Carlgo11:
Yup, it uses wordpress.com and all you can do with it is change your profile picture. Not worth adding IMO

stephengroat:
Changed to wordpress.

I could see a change to an offensive/problematic picture being an issue.

jamcat22:
I'd say we should add it, since they do support 2FA. If they didn't support 2FA, I wouldn't suggest adding it, but considering how we list eight entries for Google products and five for Microsoft, I don't think adding another WordPress site would hurt.

Also, should we use https://secure.gravatar.com/ for the link if we add it?

mxxcon:
@jamcat22 But none of the Google and Microsoft services we list are as single-purpose as avatar hosting and we already have WordPress.com listed

stephengroat:
Gravatar's reach is far beyond simply wordpress though. it's used on
github, stackoverflow, etc.

mxxcon:
Yes it is. But its access is covered by Wordpress.com entry

RichJeanes:
I agree with mxx on this one. Wordpress is the only Login option for Gravatar and it is very obviously branded that way.

PR #2133, opened 2016-10-18
Update verizon logo

stephengroat:
new logo, looks a lot cleaner than the old one


RichJeanes:
Does Verizon use the same logo for both Verizon and Verizon Wireless? (They are two distinct entities)

stephengroat:
From what I can see on their websites, yes.

Makes sense, since Verizon now completely owns Verizon Wireless and it's no
longer shared with Vodafone

Jawshy (approved):

PR #2141, opened 2016-10-20
Revision of Natwest

djrarky:
Natwest allow 2fa in their mobile app. 
They even allow you to use TouchID on iOS devices.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -667,3 +667,4 @@ websites:
       software: Yes
       hardware: Yes
       doc: https://www.zkb.ch/de/pr/pk/efinance/ebanking/login-verfahren.html
+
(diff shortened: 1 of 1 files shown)


_data/banking.yml
```diff
@@ -666,4 +667,3 @@ websites:
       software: Yes
       hardware: Yes
       doc: https://www.zkb.ch/de/pr/pk/efinance/ebanking/login-verfahren.html
-
```
stephengroat, line 669: we use newlines at the end of files, please remove change

stephengroat (changes requested):
newline comment, also need feedback from others

djrarky:
No problem! There's documentation here: http://communities.natwest.com/t5/Online-Mobile-Banking/You-asked-we-listened-Apple-Touch-ID/td-p/34759

PR #2142, opened 2016-10-20
Revision of WhatsApp

djrarky:
WhatsApp always verify your phone via SMS before signing in.
https://www.whatsapp.com/faq/en/iphone/20902747


jamcat22:
This PR is a duplicate of #2143.

PR #2143, opened 2016-10-20
Revision of WhatsApp

djrarky:
WhatsApp always verify your phone via SMS before signing in.
https://www.whatsapp.com/faq/en/iphone/20902747

### DIFF
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -192,10 +192,10 @@ websites:
 
     - name: WhatsApp
       url: https://www.whatsapp.com/
-      twitter: whatsapp
-      facebook: WhatsApp
       img: whatsapp.png
-      tfa: No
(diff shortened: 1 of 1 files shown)


jamcat22:
Hello @StevenChorkley!

Two factor authentication involves using two different factors to sign in. These factors can be something you know (like a password), something you have (like a phone or hardware security token), and/or something you are (like a fingerprint).

Because WhatsApp only uses phone-based SMS authentication (something you have), and doesn't utilize another factor (like a password), it doesn't have 2FA.

If you'd like to see an example of a messaging app which uses SMS authentication, but also allows 2FA in the form of an optional additional password, you may want to look at [Telegram's](https://telegram.org/blog/sessions-and-2-step-verification) implementation.

Thanks!

PR #2147, opened 2016-10-20
update entry for PythonAnywhere, now has software 2fa

hjwp:


stephengroat (approved):
thanks for the pr @hjwp!

Jawshy (changes requested):
@hjwp, is there any documentation about their support for software 2FA?

hjwp:
there is not, no.  it's a soft launch.  i'll update the entry once docs are available though.

Jawshy (approved):

PR #2149, opened 2016-10-20
Update Aol Mail exception

joelschnee:
AOL Mail 2FA does support international users.  Updated to reflect this.

### DIFF
--- a/_data/email.yml
+++ b/_data/email.yml
@@ -5,8 +5,6 @@ websites:
       tfa: Yes
       sms: Yes
       phone: Yes
-      exceptions:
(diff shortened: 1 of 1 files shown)


Jawshy (approved):
@joelschnee, thanks for the PR.

stephengroat (changes requested):
according to the `doc`, this isn't the case. 
Does AOL’s 2-Factor Authentication work anywhere other than in the US?
Yes, our 2-Factor Authentication supports non-US accounts and phone numbers.

RichJeanes:
Sounds to me like either way the exception should be removed.

stephengroat (approved):

Issue #2150, opened 2016-10-21
Add University of Michigan (with support for Duo Security)


vielmetti:
News release is here

http://its.umich.edu/accounts-access/uniqnames-passwords/two-factor-authentication

PR #2152, opened 2016-10-21
Add umich

stephengroat:
#2150


Jawshy (approved):

jamcat22:
We're even mentioned on their [help site](http://its.umich.edu/accounts-access/uniqnames-passwords/two-factor-authentication/turn-two-factor-weblogin)!

jamcat22 (approved):

PR #2153, opened 2016-10-21
CIBC bank now supports 2FA

andmalc:
Added 2FA fields to CIBC entry in banking.yml

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -146,7 +146,13 @@ websites:
       url: http://www.cibc.ca
       twitter: cibc
       img: cibc.png
-      tfa: No
+      tfa: Yes
+      sms: Yes
+      software: No
+      phone: Yes
+      email: Yes
+      hardware: No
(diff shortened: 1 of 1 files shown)


stephengroat:
note: according to `doc`, used for "sensitive transactions only"

andmalc:
Yeah saw that.  I suppose they don't want to bug their customers for routine activities.  Does this mean 2FA is not supported according to the site's definition?

Jawshy:
It is not because according to the [documentation](https://www.cibc.com/ca/legal/two-step-verification.html) the second step is for authorisation (of sensitive actions, as pointed out by @stephengroat) and not for account authentication (login) which does not meet [Two Factor Auth's definition](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions). 

Thanks for taking the time to submit a pull request, @andmalc, but unfortunately I believe it will have to be closed. Could another @2factorauth/collaborators confirm (the video mentions other reasons for needing an OTP)?

PR #2163, opened 2016-10-25
Correct pnc entry

thegeekkid:
While I couldn't find any decent documentation on this (only a few security press releases that vaguely mentioned 2fa), I can confirm that PNC does support SMS based 2fa now.  The method they used to validate returning (previously authenticated) users without requiring a text is easily bypassed; but I believe that by the definition of 2fa, they do support it now.


Carlgo11:
So have they fixed the implementation?
If it still can be easily bypassed then I doubt we'll add it.

thegeekkid:
No, they didn't really fix the implementation - it can be easily bypassed on the app and I've seen it not prompt on occasion on the web as well.  So ya... if you would prefer not to change it until that gets fixed, that's totally fine.  :)

Carlgo11:
Yeah I think we'll stick with labeling them as not supporting 2FA for the time being.
Wouldn't want to give users a false sense of security. 😉 

Thanks for your contribution tho! Sorry this one couldn't be merged.

PR #2169, opened 2016-10-28
update gaming with facebooks

stephengroat:


Jawshy (approved):
`looks-good`

_data/gaming.yml
```diff
@@ -134,7 +134,7 @@ websites:
       doc: http://itch.io/developers/two-factor-authentication
 
     - name: League of Legends
-      url: http://leagueoflegends.com
+      url: http://oce.leagueoflegends.com/
```
jamcat22: The old URL automatically redirects users to the correct server based on their location, for example, http://na,leagueoflegends.com for users in North America. This is by design, and no change needs to be made to the previous URL.

_data/gaming.yml
```diff
@@ -248,7 +248,7 @@ websites:
       doc: http://forums.ubi.com/showthread.php/1463703-Secure-your-account-with-2-Step-Verification
 
     - name: Wargaming
-      url: http://wargaming.net/
+      url: http://asia.wargaming.net/
```
jamcat22: The old URL automatically redirects users to the correct server based on their location, for example, http://na,wargaming.net for users in North America. This is by design, and no change needs to be made to the previous URL.

jamcat22 (dismissed):
Changes for the URLs should be reverted, because the old, non-region-specific URLs automatically redirect to the correct server based on the user's location.

Jawshy (approved):
Reverted 1686957212cd2eea7895458b0f2fdc0f45ac5a2b

Jawshy:
Could another @2factorauth/collaborators please take a look at this when possible?

mxxcon (approved):

stephengroat:
merging (even though it's mine) since there's 2 reviews on it

PR #2172, opened 2016-10-30
Add receivabl.es to Payments

tehpeh:


mxxcon (changes requested):
@tehpeh, according to your exception text that makes this two-factor _authorization_, not authentication, which means it should be `tfa: no` and other relevant fields updated as well

tehpeh:
Ok, let me rephrase: payment senders are required to enter their email address and a one-time SMS code sent to their phone to access their saved personal data, conversely payment receivers are authenticated via OAuth with Xero (which supports TFA). Does that count?

mxxcon:
@tehpeh See https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions if that helps you clarify things.

tehpeh:
Ok I'm not sure now, because if your personal information is stored on the site, you are prompted to enter a token during login. But this only applies to half the user base (ie. the payment senders).

mxxcon:
How does it know that today I want to send vs receive a payment? Are those completely different accounts?

tehpeh:
Yes, completely different accounts. I can email you a link to a demo if you'd like?

mxxcon:
Is there a publicly reachable docs link you provide?
At the very least the exception text would need to rephrased clarifying that only payment accounts have tfa, not recipients.

tehpeh:
No documentation yet, but I'll send an update as soon as it's available.

Jawshy:
Hello, @tehpeh. Would you be able to provide an update on the site's documentation of how it has implemented two-factor auth?

tehpeh:
Still no documentation, but this PR can be closed now thanks.

PR #2171, opened 2016-10-30
Add Booking.com

Carlgo11:
This is an update to #2012 as the creator of that PR seems to be unwilling to update their PR.

PR #2173, opened 2016-11-01
Fix rubocop

stephengroat:
improves efficiency, fixes #2171


Carlgo11 (approved):
RuboCop seems to check a little too much but I'll approve for now so that we can continue to work like usual.
In the future I'd like to discuss if we really need RuboCop as it just seems to cause trouble for us.

PR #2181, opened 2016-11-17
Add Bitbond

jareknowotka:
Bitbond requires users to use 2FA when signing in from a new device. If you login from the same device twice you won’t be asked for 2F.


Jawshy (changes requested):
Thanks for the pull request, @jareknowotka. Does Bitbond have any documentation regarding their implementation of two-factor authentication?

jareknowotka:
Thanks for asking, unfortunately not, we don't provide documentation regarding 2FA.

Carlgo11:
@jareknowotka if that's the case could you send some screenshots and describe how you've implemented 2FA?
What 2FA options you support and so on.

Jawshy:
@jareknowotka https://github.com/2factorauth/twofactorauth/pull/2181#issuecomment-261291459

stephengroat:
screenshot 2016-12-11 at 15 00 58

jareknowotka:
Thanks for registering! The authorisation flow for a user singing in from a new device looks like that: 

screen shot 2016-11-18 at 10 30 13
screen shot 2016-11-18 at 10 30 39
screen shot 2016-11-18 at 10 30 53

The last screen requires a token from a method user has selected as his 2FA.

Jawshy (approved):

PR #2183, opened 2016-11-20
Update Paysafecard

Carlgo11:
@olback informed me that Paysafecard offers 2FA in the form of TOTP tokens.
SMS and Email is asked for upon registration but not used on every login.

Couldn't find a doc page but I do have a screenshot of the settings page to confirm.
_Please note that the settings page is in Swedish as they apparently change language per geo-ip location._ ☹️
screenshot


Jawshy (approved):
Added `doc` - `looks-good`

PR #2184, opened 2016-11-20
Add INWX in Domains

ambux:
[INWX](https://www.inwx.com/en/) is a domain registry service and offers 2FA via software tokens. 

Here is the FAQ on it: https://www.inwx.com/en/offer/mobiletan


Jawshy:
@ambux, there are a few issues with your pull request. You are merging from `master` into `gh-pages` when it should be from `%your-branch%` into `master`. Also the site you wish to add is not in any of the commits.

PR #2187, opened 2016-11-21
Add InterNetX

InterNetX:
Add InterNetX Icon and Data


stephengroat (changes requested):
@InterNetX we tend to only list companies under a single category. what do you think the primary category for InterNetX is, maybe `domains`?

_data/domains.yml
```diff
@@ -262,6 +262,14 @@ websites:
       software: Yes
       doc: https://www.inwx.de/en/offer/mobiletan
 
+    - name: InterNetX
+      url: https://www.internetx.com
+      img: internetx.png
+      tfa: Yes
+      sms: Yes
+      software: Yes
+      doc: https://www.internetx.com/en/domains/autodns/
```
Jawshy, line 271: I cannot see any documentation about SMS or software 2FA on that webpage. Are you sure that is the correct page?

Jawshy (changes requested):
`doc`

_data/domains.yml
InterNetX: There it is https://www.internetx.com/en/domains/autodns/domain-safe/ or do you need full documentation?

Jawshy: Full documentation would be ideal.

stephengroat (approved):

Carlgo11:
How should we continue with this PR?

Jawshy:
Full documentation or close.

Carlgo11:
And since the last response from the author is on the 21st of November I guess it's the later option.

Issue #2193, opened 2016-11-26
Mailfence


Hillside502:
Email privacy | Mailfence protects your mail privacy
https://mailfence.com/

Two-factor authentication
https://mailfence.com/c/mailfence/more/totp.jsp

stephengroat:
thanks for the suggestion @Hillside502!

MSalmanN:
The issue has been closed, and site has been reviewed #2194. But it's not included yet in the current repository! @stephengroat

Jawshy:
That PR just needs to be merged then `gh-pages` synced with `master`, @MSalmanN.

jamcat22:
@MSalmanN You can learn more about our merging process on the [wiki](https://github.com/2factorauth/twofactorauth/wiki/FAQ-Merge). 😄

MSalmanN:
Thanks @jamcat22, I was not aware of this.

PR #2194, opened 2016-11-26
add mailfence

stephengroat:
#2193


Jawshy (approved):
`looks-good`, although not particularly popular ([Domaintally](https://www.domaintally.com/www/mailfence.com) and [Alexa](http://www.alexa.com/siteinfo/mailfence.com)).

PR #2202, opened 2016-11-28
Add LCN.com

rubenarakelyan:


stephengroat (changes requested):
could you fix the image size? it should be 32x32

stephengroat (changes requested):
try using tinypng.com to reduce the image size

stephengroat (approved):

fpigerre:
Thanks for the pull request @rubenarakelyan! 👍

PR #2204, opened 2016-11-28
Add Thriva

rubenarakelyan:


stephengroat (changes requested):
Could you fix the image size? it should be 32x32

stephengroat (changes requested):
Could you try running the image through tinypng.com? gave me compression of about 50%

stephengroat (approved):

PR #2205, opened 2016-11-28
Add Vitality

rubenarakelyan:


stephengroat (approved):
thanks for the pr @rubenarakelyan!

mxxcon:
Hmm..Should insurance company be under _finance_ category?

stephengroat:
the adirondack insurance exchange is under finance...

mxxcon:
doesn't mean it's in the right spot either.. :)

rubenarakelyan:
That was my reference point as well...

rubenarakelyan:
Do you see insurance as needing its own category?

mxxcon:
It's a sensible category, but does it make sense for just 2 items?

rubenarakelyan:
Personally, I think it makes sense to split out insurance once it gets a few more links so that they don't crowd out the rest of the financial sites.

Carlgo11:
Let's stick with finance for now and if we get some more insurance entries we can make another category for them. Just like we have with government and transportation.

PR #2230, opened 2016-12-09
Add exceptions text to Telegram entry

ghost:
Add exceptions text to Telegram entry: "SMS codes are the default method of logging in. Two-factor authentication is available only after you have enabled it by setting a password through the Privacy and Security settings page."


ghost:
I'm closing this, because it might be better to start a separate issue thread to discuss the issue of twofactorauth.org not showing what each service uses for single-factor authentication.

PR #2231, opened 2016-12-09
Update WhatsApp entry

ghost:
WhatsApp has added 2FA: https://www.whatsapp.com/faq/en/general/26000021

### DIFF
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -208,10 +208,12 @@ websites:
 
     - name: WhatsApp
       url: https://www.whatsapp.com/
-      twitter: whatsapp
-      facebook: WhatsApp
       img: whatsapp.png
-      tfa: No
+      tfa: Yes
+      sms: Yes
+      doc: https://www.whatsapp.com/faq/en/general/26000021
+      exceptions:
+            text: "SMS codes are the default method of verification. Two-factor authentication is available only after you have enabled it by setting a six-digit passcode through the Account settings page."
 
     - name: Yammer
       url: https://www.yammer.com
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -212,8 +212,6 @@ websites:
       tfa: Yes
       sms: Yes
       doc: https://www.whatsapp.com/faq/en/general/26000021
-      exceptions:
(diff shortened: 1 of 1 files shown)


Carlgo11:
I don't really see why it's necessary with an exception text for this.

jamcat22:
I agree with Carl here. The doc page explains how to setup 2FA, so I don't think the exception is necessary.

ghost:
When I first visited twofactorauth.org, I interpreted the tables so that the columns that are on the right side of the docs column are what each service uses as the secondary form of authentication. Some services like WhatsApp use SMS codes for single-factor authentication, and optional user-defined passcodes/passwords are what ultimately complete the picture so that the users have 2FA in place.

I would like it if twofactorauth.org added two new columns: One for what each service uses for single-factor authentication and one named Passcode/Password. Then the WhatsApp entry could say SMS in the SFA column and have a check mark in the Passcode/Password column.

As long as there isn’t a separate column for what each service uses for single-factor authentication, I think it would be reasonable to add exception notes to the exceptions that use something other than user-defined passwords for single-factor authentication.

ghost:
I think it's important to distinguish what each service uses for single-factor authentication, because there <s>are cases</s> _is one case_ where – due to the service using SMS for SFA – an attacker could get access to most if not all of the user's messaging history by intercepting a single SMS code:

https://www.fredericjacobs.com/blog/2016/01/14/sms-login/
https://www.fredericjacobs.com/blog/2016/04/29/more-on-sms-logins/

Edit: On the other hand, I can see why making this distinction wouldn't be that important in the case of WhatsApp, because they don't store their users' messages on servers like Telegram does.

ghost:
Judging by the silence, I don't think the exception text I added is going to receive any more approval at this point, so I have now removed it.

The issue of twofactorauth.org not showing what each service uses for single-factor authentication can probably be discussed in a separate issue thread.

mxxcon:
@echo743 silence sometimes simply means that people are busy, not always that they disagree. 😉 
However, this site is about 2FA support. _I_ think cataloging the primary login system is somewhat beyond our scope. For example, I think there was a fork of this site that was documenting password length/strength of various sites. Perhaps the info you want to include is better suited there?

stephengroat (approved):

PR #2235, opened 2016-12-10
Fix quoting for Watchman Monitoring

unlobito:
Double quotes used inside the string cause the layout to break as the generated HTML is invalid.

```html
<span class="popup exception"
    data-position="bottom center"
    data-html="<div class='header'>    Exceptions &amp; Restrictions</div><div class='content'>            Two factor authentication unavailable for "End User" access, which is read-only.    </div>">
    <i class="warning sign red link icon"></i>
</span>
```

### DIFF
--- a/_data/other.yml
+++ b/_data/other.yml
@@ -408,7 +408,7 @@ websites:
       software: Yes
       doc: https://www.watchmanmonitoring.com/two-factor-authentication
       exceptions:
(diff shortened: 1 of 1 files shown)


Carlgo11:
Urgh this new "review request" feature doesn't work for me. I'll just assign you to this PR instead @psgs.

mxxcon (approved):

fpigerre (approved):

PR #2238, opened 2016-12-12
Created basiscode.png using notepad

baileybasiscode:
Hello, I'm creating the new basiscode.png file. I couldn't upload and figured, I may be able to copy the notepad version of the png file and submit it. If this is in error, feel free to let me know.

The online version can be found here: http://basiscode.com/wp-content/uploads/2016/12/compLogo32x32.png


RichJeanes:
Can you add the image as as another commit on the same branch that you submitted a PR for already? If you push to the same branch, it will shoe up in the PR automatically.

What do you mean by "the notepad version" of the icon? May I recommend using a logo more like your favicon? Text does not turn out to well at resolutions as low as 32x32.

The reason you can't upload images to this repo is that you do not have permission to modify 2factorauth/twofactorauth, however you should be able to modify your own fork, baileybasiscode/twofactorauth

PR #2234, opened 2016-12-09
As of Dec 1, 2016, ProtonMail supports 2FA

seanmcelroy:

PR #2207, opened 2016-12-01
ProtonMail (v3.6) now supports 2fa with a software token.

vab:
Support status updated. Documentation link added.

Ref: https://protonmail.com/blog/protonmail-v3-6-release-notes/

PR #2241, opened 2016-12-13
Update ProtonMail as it now has 2FA

jdavis:
As per an email I received


mxxcon:
dupe of #2234 and #2207

Carlgo11:
Solved by d22e5f7.

PR #2248, opened 2016-12-14
1Password supports 2FA for Teams accounts using Duo Security

gingerbeardman:
See https://twitter.com/1Password/status/809091553381777409

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -23,6 +23,16 @@ websites:
           text: "SMS or phone call required to enable 2FA."
       doc: https://www.amazon.com/gp/help/customer/display.html?nodeId=201596330
 
+    - name: Amazon UK
+      url: https://www.amazon.co.uk
+      img: amazon.png
+      tfa: Yes
+      software: Yes
+      sms: Yes
+      exceptions:
+          text: "Requires a second account on Amazon.com where 2FA is activated."
+      doc: https://www.amazon.com/gp/help/customer/display.html?nodeId=201596330
+
     - name: Apple
       url: https://appleid.apple.com
       img: apple.png
@@ -304,4 +314,3 @@ websites:
       tfa: No
       twitter: zazzle
       facebook: zazzle
-
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -23,16 +23,6 @@ websites:
           text: "SMS or phone call required to enable 2FA."
       doc: https://www.amazon.com/gp/help/customer/display.html?nodeId=201596330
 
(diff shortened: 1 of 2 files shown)


mxxcon (changes requested):
please don't include amazon change with this PR. please make it a separate PR.

gingerbeardman:
Apologies, GitHub Desktop didn't warn me.

mxxcon (approved):

mxxcon:
That's a pretty crappy setup of them. 😕 Use only 2fa service which requires payment to that service.

_data/retail.yml
```diff
@@ -304,4 +304,3 @@ websites:
       tfa: No
       twitter: zazzle
       facebook: zazzle
-
```
RichJeanes, line 307: Please do not remove new lines at the end of files. They should be there in all of our `.yml` files. See our `.editorconfig` for our formatting rules. If you use an editor that supports plugins, you can look for and install the Editor Config plug-in.

RichJeanes (changes requested):

gingerbeardman:
ok, thanks for the info.

changed

Carlgo11:
I think that maybe we should wait for them to roll it out everywhere. From past experience they could very well disable 2FA again or never activate it for individuals.

mxxcon:
@Carlgo11 but that's why there's exception text?

Carlgo11:
@mxxcon yeah but then `tfa` is set to true.

RichJeanes (approved):

Carlgo11 (changes requested):
Only partially added. Don't think `tfa` should be set to `Yes` then.

jamcat22 (approved):
I say we should add it with the exception. We've taken this approach on other sites we've added, including Square, Dyn, Coinify, and arguably PayPal and Apple, whose 2FA systems don't protect their entire userbase and various sign in systems.

gingerbeardman:
@jamcat22 @Carlgo11 FWIW I searched for other entries that use Duo Security and copied the way their entry was set up. Specifically, University of Michigan.

Carlgo11:
@jamcat22 what if we put it as a `status` link then? To still put pressure on them.

fpigerre:
A `status` link sounds like a good compromise, given that payment is required in order to use 2FA for 1Password.

gingerbeardman:
redone as status link @Carlgo11 @psgs 

still think exception is better as does @jamcat22

Carlgo11 (approved):

PR #2249, opened 2016-12-14
Amazon UK, DE, etc. support 2FA via Amazon.com

gingerbeardman:


_data/retail.yml
```diff
@@ -304,4 +314,3 @@ websites:
       tfa: No
       twitter: zazzle
       facebook: zazzle
-
```
RichJeanes, line 307: Please do not remove new lines at the end of files. They should be there in all of our `.yml` files. See our `.editorconfig` for our formatting rules. If you use an editor that supports plugins, you can look for and install the Editor Config plug-in.

RichJeanes (changes requested):

_data/retail.yml
mxxcon: [.editorconfig](https://github.com/2factorauth/twofactorauth/blob/master/.editorconfig) is like a style guide for files. how many spaces to use, what kind of linebreaks, etc etc. see http://editorconfig.org/

gingerbeardman:
understood, thanks for the info.

changed

mxxcon:
I'm a little bit worried by this. If this is a glitch/bug in Amazon UK's login system, once Amazon fixes this might bite people in the butt and lock them out of their accounts.
It would be nice to contact Amazon UK support and confirm with them that this is safe to use. 🤔

RichJeanes (approved):

Carlgo11:
Does it make sense to have multiple Amazon sites listed here? As far as I can tell 2FA works on all of them the same way.

gingerbeardman:
@Carlogo11 I think you're right, maybe just stick with .com URL, make sure Amazon name doesn't include any domain suffix, and make a note in the exception text?

Carlgo11 (approved):
Looks good to me! 👍

PR #2276, opened 2016-12-23
Vanguard u2f/alternate method update

kaytwo:
While only available to logged-in users, Vanguard's u2f documentation is available here: https://personal.vanguard.com/us/U2FKeyEnrollment.

I assume it has the same restriction to personal investing accounts only, so the exception still stands.

I also added a link to the blog post which outlines that both SMS and voice call 2fa are available.

### DIFF
--- a/_data/investing.yml
+++ b/_data/investing.yml
@@ -177,7 +177,9 @@ websites:
       img: vanguard.png
       tfa: Yes
       sms: Yes
-      doc: https://personal.vanguard.com/us/insights/article/Account-security-092015
+      phone: Yes
+      hardware: Yes
+      doc: https://personal.vanguard.com/us/insights/article/security-codes-112015
       exceptions:
           text: "TFA is only available for Personal Investor accounts"
 


_data/investing.yml
```diff
@@ -177,7 +177,9 @@ websites:
       img: vanguard.png
       tfa: Yes
       sms: Yes
-      doc: https://personal.vanguard.com/us/insights/article/Account-security-092015
+      phone: Yes
+      hardware: Yes
```
Carlgo11, line 181: I don't see anything about hardware on the [doc](https://personal.vanguard.com/us/insights/article/security-codes-112015) page. Do they support it?

Carlgo11 (dismissed):

_data/investing.yml
kaytwo: As mentioned in the PR, the information is only available while logged in, here's a screenshot:

screen shot 2017-01-04 at 8 59 06 pm

Carlgo11 (approved):
👍

Jawshy (approved):

PR #2277, opened 2016-12-26
Remove hardware support for Name.com

Parent5446:
I am not sure if support for hardware tokens existed previously, but currently, as far as Name.com docs and account settings show, there is no support for hardware tokens. Only TOTP using Google Authenticator is supported.


mxxcon (approved):
It used to use different system that supposed hardware tokens.

stephengroat (approved):

PR #2281, opened 2016-12-30
Add UpCloud

cez81:


Jawshy (approved):
Thanks for submitting another pull request, @cez81.

cez81:
As I understand it only subaccounts can use sms and phone. Main account is TOTP only. Maybe worth a note if they're added?

stephengroat:
since sms/phone call is used for account recovery and overrides the TOTP token, I think it makes sense to include that

cez81:
Ah yes, that's true. I agree.

Carlgo11 (changes requested):
[@stephengroat](https://github.com/2factorauth/twofactorauth/pull/2281#issuecomment-269793163) I disagree. It clearly states that it's only used for recovery.
Yes one could use it instead of TOTP but TOTP is what they advertise as the only 2FA method.

stephengroat:
alright, what about an exception with `sms is required for password
recovery`?

Carlgo11:
Go for it

Carlgo11 (approved):

PR #2284, opened 2017-01-02
Add Orbit bhyve

stephengroat:
An IoT sprinkler


Jawshy (commented):
How does this site meet the [criteria](https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria)? I am guessing 4 (Control), but is it significant enough to warrant listing?

stephengroat:
Yes and i think that any control item should be listed. Even if it only scales to the 10s of thousands, that's a lot of sprinklers firing off

Carlgo11 (changes requested):
I don't think meets the criteria.

stephengroat:
how not? it directly controls relays that control sprinkler systems for
homes

Carlgo11:
> This criteria is more general, in that it includes sites that give access to things that may infringe upon criteria 1, 2, or 3. An example of this is a website that allows remote access to a device.

While it may cause some inconvenience I wouldn't say it could cause any long term effects if someone got remote access to your outdoor sprinkler system.

stephengroat:
you can flood yards, crack foundations, etc. it's actually pretty powerful

Carlgo11:
Maybe we can come back to this in the future. As of right now I'd say it's too insignificant to be on our list.
Control to this app also doesn't affect criteria 1 and 2.

PR #2288, opened 2017-01-05
Update Outlook

Ndragonawa:
Updated outlook.com; yes on email and phone 2fa

### DIFF
--- a/_data/email.yml
+++ b/_data/email.yml
@@ -92,7 +92,7 @@ websites:
       phone: Yes
       email: Yes
       software: Yes
-      doc: http://windows.microsoft.com/en-us/windows/two-step-verification-faq
+      doc: https://support.microsoft.com/en-au/help/12408/microsoft-account-about-two-step-verification
 
     - name: Pobox
       url: https://www.pobox.com/
--- a/_data/email.yml
+++ b/_data/email.yml
@@ -89,6 +89,8 @@ websites:
       img: outlook.png
       tfa: Yes
       sms: Yes
+      phone: Yes
+      email: Yes
       software: Yes
       doc: http://windows.microsoft.com/en-us/windows/two-step-verification-faq
 


Jawshy (changes requested):
Hi, @Ndragonawa. Thanks for submitting a pull request. 

`phone: Yes` means phone call based authentication (see [FAQ Categories](https://github.com/2factorauth/twofactorauth/wiki/FAQ-Categories)) and I cannot see that explictly specified in their documentation (then again neither is text/SMS). 

A text message (SMS) can be sent to my phone as the second factor in 2FA, but there is no phone call option available even though the phone number can be initially verified via text or call. 

Here is a screenshot that shows the possible 2FA options that can be used: 
outlook-2fa-options

jamcat22 (dismissed):

jamcat22:
@Jawshy, to use phone call 2FA with a Microsoft Account, you must select "Call" when adding your phone number to your security info. The screenshots below demonstrate this:
Adding a phone to a Microsoft Account, and selecting "Call" as the method of verification
Logging in with phone call 2FA

Jawshy:
@jamcat22, I did try that, but I still did not receive _call_ as a verification option when signing in.

Carlgo11:
I don't have a Microsoft account so I can't check this but we somehow need to confirm that this works.
Maybe calls are only supported in the US or NA?

jamcat22:
@Jawshy, did you add a new phone, or completely remove the old version before adding the new version? I had to add a new phone number (through Google Voice in my case) or remove the old one, wait a bit, then readd it as a calling phone number before it would work.

Jawshy:
@jamcat22, I completely removed my existing phone and then re-added it. I shall try again to confirm. 

*A few moments later...*

I still do not have a *call* option for receiving the verification code.

jamcat22:
@Jawshy, but you were able to verify the phone via a call? You just couldn't login using a call after?

Jawshy:
Correct. I verified my mobile phone via call without issue, but I could not authenticate via call.

Carlgo11:
Looks like we're in a stalemate. How should we go forward?

Hopefully we won't need an official Aperture Science Stalemate Associate.
https://youtu.be/SIzfnY-umzA

mxxcon:
Maybe this is related to region-specific outlook accounts?
I do get call/text verification choices when I *ADD* a phone number to my account,
But I get only text option when when it comes 2fa prompt.

jamcat22:
@Carlgo11, you need to be a trained stalemate associate to press that button. You're unqualified.

If we determine that my account is just weird for some reason, then I'd say we just remove the `phone: Yes` line and keep the `email: Yes` line.

Otherwise, I'd say we add an exception explaining how phone call based 2FA is only available within certain regions, or only available within the US (if we determine that). I'm testing Canadian, German, French, and UK phone numbers right now, and I'll edit this comment with the result.

**Edit:** I just tested this with the above-mentioned numbers on a US Microsoft Account. All the phone numbers allowed me to use the "Call" option.
Result of Phone Number Test

Carlgo11:
@jamcat22 Looks like this only works in certain regions.
I tried to set up 2FA thru phone calls but didn't get the option when logging in.
(Only tried with Swedish phone number)
firefox_2017-03-15_22-05-04
firefox_2017-03-15_22-08-28

RichJeanes:
So do we care to update this with an exception saying that Phone is only available in some regions?

Jawshy:
@RichJeanes, yes, let's add an exception. 

Regions / phone numbers that **do** have phone call available for 2FA. 

* United States (+1)
* United Kingdom (+44)
* Canada (+1)
* Germany (+49)
* France (+33)

Are there any others? 

Regions / phone numbers that **do not** have phone call available for 2FA. 

* Australia (+61)
* Sweden (+46)

Are there any others?

Carlgo11:
I think all countries except the ones listed as supporting phone 2FA, have it disabled so there should be no need to list those that don’t have it.

jamcat22:
Could someone with an account in a **non-North American** region try adding a US phone number to their account using the “Call” option and see if they’re able to use phone 2FA to login? This would let us know if it’s based on _account region_ or _phone number region_.

kmpoppe:
I've tried a few language-codes on https://support.microsoft.com/xx-yy/help/12428 - one of the questions is "If I can't receive texts, can I get a phone call" and all languages say yes. I've tried outlook.com with a German Microsoft account and was able to add a new authentication method via Phone Call for a +1 number.
So _I_ think, it's safe to assume, that `sms`, `phone`, `email` and `software` are valid 2FA options for every Microsoft-Account-service.

Carlgo11:
Moving this to a separate PR updating all MS services

PR #2289, opened 2017-01-05
Added Dutch Bangla Bank Limited (DBBL)

aniskhan001:


stephengroat:
thanks for the pr @aniskhan001. the image is a little too large, could you try using tinypng.com?

_data/banking.yml
```diff
@@ -260,6 +260,14 @@ websites:
       software: Yes
       doc: https://www.dnb.no/privat/kundeservice/innlogging-nettbank.html
 
+    - name: Dutch Bangla Bank Limited (DBBL)
```
Jawshy: Looking at their website they appear to hyphenate Dutch-Bangla. So in full it is `Dutch-Bangla Bank Limited`.

Jawshy (changes requested):
Thanks for submitting a pull request, @aniskhan001. 

It appears that their implementation of two-factor is for authorization of transactions and not authentication of user login. See also [A Note on Definitions](https://github.com/jdavis/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions). 

[Guidelines for using DBBL Two Factor Authentication (2FA) Security Solution](http://www.dutchbanglabank.com/internet-banking/2fa-guidelines.html): 
> [...] Combine this number with your existing user name and password **to make a transaction using our web services**. This new feature increases the security of your accounts. **There are 2 types of token we are offering to make a secure transaction.** These are Hardware and Software token.
> [...] 
> 4. Now pin has been set. **Please use this pin and token code (displayed in hardware token) as passcode for transactions.**
> [...] 
> 9. Hold on! Before you proceed to generate passcode **(which will be required for doing transactions)** using this token generator, you need to change the PIN for the first time. [...] 
> [...]
> **Passcode is required for doing transactions through internet banking.** Whenever you initiate a transaction process, you need to generate the Passcode from RSA Software and use the same Passcode within the life-time of the passcode in the required field. [...] 

And so on. 

It is also interesting that they have two-factor but a landing page that uses HTTP. :open_mouth:

_data/banking.yml
aniskhan001: Should I remove DBBL also? This short name is popular here.

Jawshy: If the acronym is commonly used then I would say keep it. It would make it easier to find the site via searching.

Jawshy (approved):

Carlgo11 (changes requested):
I recommend using their email address as well. `cards@dbbl.com.bd`.
Plus a lang tag.

aniskhan001:
@Carlgo11 The contact us section says: hrd_head@dbbl.com.bd is the email to contact them.
Also, the site language is English, although the local lang is bd (Bangladesh). Which should I provide?

Carlgo11:
@aniskhan001 go with hrd_head@dbbl.com.bd and use `lang: bd`.

Carlgo11 (approved):

Jawshy:
[The `lang` fields use ISO 639-1 codes.](https://github.com/2factorauth/twofactorauth/blob/13b772d54ae1ee0c67a84708595a8ec791a44740/_data/languages.yml#L1) The [ISO 639-1 code](https://en.wikipedia.org/w/index.php?title=List_of_ISO_639-1_codes&oldid=756635943) for [Bengali](https://en.wikipedia.org/w/index.php?title=Bengali_language&oldid=757540045), the official language of [Bangladesh](https://en.wikipedia.org/w/index.php?title=Bangladesh&oldid=758635578), is `bn`; `BD` is the country code, @Carlgo11 and @aniskhan001. I will change it now.

aniskhan001:
@Jawshy my bad. Cheeky mistake. Thanks for clarifying.

Jawshy (approved):
As good as it is going to get. :+1: 

*(They appear to have a Twitter handle at [dutchbanglabank](https://twitter.com/dutchbanglabank), but there are no tweets and it is not verified.)*

aniskhan001:
Yeah, but the link is not attached to their website. So, didn't add here.

Jawshy:
Indeed. No reason to add it.

PR #2292, opened 2017-01-06
Skrill info updated

aniskhan001:
- they added software token
- updated support link


Jawshy (approved):
Thanks again for another pull request, @aniskhan001.

Carlgo11 (approved):

PR #2293, opened 2017-01-07
Fixed broken Guild Wars 2 documentation link

Kezxo:
Small fix.

### DIFF
--- a/_data/gaming.yml
+++ b/_data/gaming.yml
@@ -114,7 +114,7 @@ websites:
       email: Yes
       phone: Yes
       software: Yes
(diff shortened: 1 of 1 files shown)


Jawshy (approved):
Thanks for submitting a pull request, @Kezxo.

Carlgo11:
Do they still support email authentication?

Kezxo:
Yes, I think it's required but only when you connect from a new network. They email you a link that you have to click when you log in somewhere new. You can mark a PC as trusted or not when you log in, then it won't ask for codes and maybe email authentication again.

https://help.guildwars2.com/hc/en-us/articles/201862918-Trouble-Logging-In

Carlgo11 (approved):

Issue #2188, opened 2016-11-22
ebay: Hardware only


oilervoss:
ebay demands a hardware key for 2FA. There is no way to register a software key.

Jawshy:
Hi, @euleralves. Thanks for reporting that. Would you like to change eBay's [2FA category](https://github.com/2factorauth/twofactorauth/wiki/FAQ_Categories) in a pull request?

stephengroat:
@euleralves, have you tried using the software Symantec key with eBay's
2FA? Been working for me for a time

stephengroat:
closing due to inactivity and belief that [this](https://nakedsecurity.sophos.com/2016/07/15/setting-up-two-factor-authentication-on-ebay-harder-than-it-should-be/) method still works

oilervoss:
The hardest part to activate eBay 2FA is found the right URL, as the article says. 
Here you are: https://scgi.ebay.com/ws/eBayISAPI.dll?ActivateSecurityToken

I didn't know that I could activate eBay 2FA using the Symantec VIP Access softkey (https://m.vip.symantec.com) because the eBay page states the hard token as only option. You need to use the VIP Access Credential ID as the hard key serial number and wait for two timed Security Codes to use as OTP.

But I found a even better alternative: https://github.com/cyrozap/python-vipaccess is a python FOSS "implementation of Symantec's VIP Access client. It is able to generate OATH URIs and their corresponding QR codes so any TOTP-generating application can be used as a VIP OTP token."

You don't need two apps. You can generate a QR code and register it in an app like [Authy](https://www.authy.com/) or [Google Authenticator](https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2&hl=en)

PR #2294, opened 2017-01-08
Change eBay Software status to No

swrobel:
I cannot find any evidence that eBay supports Software 2FA. The linked URL references a hardware token that is available. There is no information about using software tokens.

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -81,7 +81,7 @@ websites:
       img: ebay.png
       tfa: Yes
       hardware: Yes
-      software: Yes
+      software: No
       exceptions:
           text: "eBay only supports 2FA in some countries."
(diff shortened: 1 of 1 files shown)


stephengroat:
please see #2188, [this](https://nakedsecurity.sophos.com/2016/07/15/setting-up-two-factor-authentication-on-ebay-harder-than-it-should-be/) method still works

swrobel:
If you don't figure out some way to link to that, people will keep opening issues...

As the post states, it's much harder than it should be

Carlgo11:
Or you could try to push eBay in to making it easier to set up software authentication.

swrobel:
Like this? https://twitter.com/swrobel/status/817889139718688768

Carlgo11:
Just like that. 😉

PR #2308, opened 2017-01-15
Add SRI Hash for jQuery from CDN

mik-laj:
Hello,

It's a security improvement. You could read more about it on: https://www.srihash.org/

Thanks in advance

### DIFF
--- a/_layouts/secondary.html
+++ b/_layouts/secondary.html
@@ -60,7 +60,7 @@
 </div>
 
 <!-- 3rd Party Libraries -->
-<script src="//ajax.googleapis.com/ajax/libs/jquery/2.1.4/jquery.min.js"></script>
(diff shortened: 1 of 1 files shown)


stephengroat:
thanks for the info @mik-laj 

i think the idea of the hash is great, but i'm not 100% on the `https:` added to the `src`.
seems like `https` is more of a transport layer configuration issue that should be configured at that layer, not in a application/presentation layer function

mxxcon (approved):
@stephengroat But we are using HSTS, so pretty much all requests to our site are over HTTPS. Might as well hardcode all external resources to use HTTPS too. There's no drawback if somebody still manages to get to the site using HTTP.

jamcat22 (approved):
I agree with @mxxcon. Additionally, as of [v2.1](https://github.com/2factorauth/twofactorauth/releases/tag/v2.1), HTTPS is hardcoded into the configuration file, and any HTTP version of a page is now forced to redirect to the HTTPS version, even in browsers without HSTS support.

Carlgo11 (approved):

PR #2312, opened 2017-01-18
BitGo has removed SMS 2FA support and add hardware tag

jlopp:
We have removed SMS 2FA due to the rise in phone porting attacks https://blog.bitgo.com/bitgo-phasing-out-sms-based-2-factor-authentication-96db16fdd63a

### DIFF
--- a/_data/cryptocurrencies.yml
+++ b/_data/cryptocurrencies.yml
@@ -33,8 +33,8 @@ websites:
       url: https://www.bitgo.com
       img: bitgo.png
       tfa: Yes
-      sms: Yes
       software: Yes
+      hardware: Yes
       doc: https://bitgo.zendesk.com/hc/en-us/sections/201009049-Two-Factor-Authentication-Help-Page
 
     - name: Bitlish


Carlgo11 (changes requested):
```<------------ ERROR ------------>

1. BitGo: The YAML tag 'sms' should either have a value set to 'Yes' or not be used at all. (Current value: 'false')
```

jlopp:
Removed sms tag.

Carlgo11:
The doc link still refers to SMS authentication.

https://bitgo.zendesk.com/hc/en-us/articles/205139615-Two-Step-Authentication-via-SMS

mxxcon:
from their blog:

> Additionally, effective immediately, BitGo has withdrawn SMS as a 2FA method for all new user signups. New users must use client-side 2FA systems such as TOTP with Google Authenticator.
If you’re an existing BitGo user using SMS 2FA, you can continue to use it. However, we strongly recommend that you migrate to alternate methods as soon as possible.

jlopp:
Yes, we need to update the Zendesk article - I've tasked our Zendesk person with doing that.

jlopp:
SMS zendesk post has been removed; should be good to go!

Jawshy (changes requested):
[How to setup Two Factor Authentication?](https://bitgo.zendesk.com/hc/en-us/articles/211880046-How-to-setup-Two-Factor-Authentication-): 
> 4. You can setup Two Factor Authentication using Google authenticator or Yubikey

So there is [hardware](https://github.com/2factorauth/twofactorauth/wiki/FAQ_Categories) support too. @jlopp, would you like to add the following on a new line below `software`? 

```ruby
       hardware: Yes
```

jlopp:
Good catch; thanks. Added hardware tag.

Jawshy (approved):

jamcat22 (approved):

Carlgo11 (approved):

PR #2119, opened 2016-10-14
Mint doesn't support software token for 2FA

vgarmash:
Read more at https://mint.lc.intuit.com/announcements/1286158

PR #1791, opened 2016-06-03
Removed SW Token for Mint

ajixnet:
Contacted Mint via Twitter and they indicated that they only support SMS and email for 2FA. Response linked me to the following page for confirmation of just sms and email. https://mint.lc.intuit.com/announcements/1286158

PR #1517, opened 2016-01-13
Updated Mint

rworsnop:
They now have 2fa by email

PR #2314, opened 2017-01-18
Mint does not support 2FA

jlopp:
Mint has no support for hardware or software 2FA. If you're signing in from a new device they will email an OTP, but I wouldn't consider that 2FA since it doesn't occur for every login nor is it configurable.

### DIFF
--- a/_data/finance.yml
+++ b/_data/finance.yml
@@ -123,10 +123,10 @@ websites:
     - name: Mint
       url: https://www.mint.com
       img: mint.png
-      tfa: Yes
-      sms: Yes
-      email: Yes
-      software: Yes
+      tfa: No
+      sms: No
+      email: No
+      software: No
 
     - name: Mint Bills (formerly Check)
       url: https://bills.mint.com


Carlgo11:
Output from Travis:
```

1. Mint: The YAML tag 'email' should NOT be present when TFA is disabled.

2. Mint: The YAML tag 'email' should either have a value set to 'Yes' or not be used at all. (Current value: 'false')

3. Mint: The YAML tag 'software' should NOT be present when TFA is disabled.

4. Mint: The YAML tag 'software' should either have a value set to 'Yes' or not be used at all. (Current value: 'false')

5. Mint: The YAML tag 'sms' should NOT be present when TFA is disabled.

6. Mint: The YAML tag 'sms' should either have a value set to 'Yes' or not be used at all. (Current value: 'false')
```

RichJeanes:
Haven't we been through this before? A few times?
#2119 
#1791 
#1517

jlopp:
Wow, that's news to me. I opened a support ticket with Mint yesterday and the response was that they don't support 2FA.

PR #2310, opened 2017-01-18
Update a whole lot of dependencies

fpigerre:
Hello!

This pull request updates a significant number of dependencies for twofactorauth.org, notably the [jQuery](https://code.jquery.com/) and [Jets.js](https://github.com/NeXTs/Jets.js) dependencies.

* jQuery has been updated from version `2.1.4` (29th of April, 2015) to `3.1.1` (23rd of September, 2016)
* Jets.js has been updated from version `0.6.0` (15th of December, 2015) to `0.13.0` (31st of October, 2016)
* Semantic UI has been updated to version `2.2.7` (December 21st, 2016) from version `2.0.7` (JavaScript) and version `2.1.4` (CSS)
* The `github-pages` Gem has been updated from version `112` to version `116`
* The version of Ruby used by TravisCI has been updated from version `2.3.0` (25th of December, 2015) to version `2.4.0` (25th of December, 2016)

Thanks,
psgs 🌴

PR #2315, opened 2017-01-18
Fix default layout 

stephengroat:
to match #2308


mxxcon (approved):
I think #2310 will make this PR obsolete since it updates to newer version of jquery.

stephengroat:
gotcha, missed that.

PR #2316, opened 2017-01-18
Fix README to instruct to use bundle installed gems

stephengroat:
Otherwise, dependencies in defaultly installed gems may cause issues


Carlgo11 (approved):

fpigerre:
Thanks for the pull request @stephengroat! 👍

PR #2324, opened 2017-01-21
Add AAI@EduHr

anotherus3r:
Added AAI@EduHr (Authentication and Authorisation Infrastructure of science and higher education in Croatia)

PR #2326, opened 2017-01-22
Add Moj Tele2

anotherus3r:


Carlgo11 (changes requested):
Same as with #2324. Add the language tag.

Carlgo11 (approved):

PR #2331, opened 2017-01-27
Add JigoCloud

jameswoodmancy:
Hi,

Like this idea :-).  I've been using some SaaS based software that doesn't offer 2 factor auth yet..I've added it to the list.

Cheers
James


Carlgo11:
Hello @jameswoodmancy,
Thank you for your interest in twofactorauth.org!

While we'd love to add all sites it's just not feasible. At the moment we try not to add sites with a too low popularity ranking just to keep the list managable.
By looking at [alexa](http://www.alexa.com/siteinfo/jigocloud.com) it looks like JigoCloud isn't used by a lot of people and therefore I'm afraid I can't accept this pull request. I'll await another collaborator of  @2factorauth/collaborators to give their view on it but most likely they will agree with me.

Carlgo11 (changes requested):
https://github.com/2factorauth/twofactorauth/pull/2331#issuecomment-275659203

stephengroat:
sorry @jameswoodmancy, but since twofactorauth.org is a static site, we
don't want to force people to load too much data to view the data.

maybe we could start thinking about a way to use jquery to allow smaller
sites to only show up in searches and not load by default?

PR #2335, opened 2017-01-30
Add Indeed

Stevoisiak:
* Add Indeed.com to Other

Note: I wasn't sure whether to add this to Other or Social


stephengroat (approved):

jamcat22 (approved):
I think we should go with `Other` for now.

PR #2345, opened 2017-02-01
Autofocus search box

hickford:
I assume most visitors to https://twofactorauth.org/ immediately use the search box, so automatically focus the search box.

### DIFF
--- a/index.html
+++ b/index.html
@@ -58,7 +58,7 @@ hash: SupportTwoFactorAuth
       <i class="search icon"></i>
     </label>
     <input type="search" id="jets-search" placeholder="Search websites" autocomplete="off" spellcheck="false"
-           tabindex="0">
+           tabindex="0" autofocus>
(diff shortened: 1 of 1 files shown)


mxxcon:
I'm not sure on mobile it's a good behavior. I personally find it pretty annoying when the moment I get to some site my keyboard pops up before I even had a chance to browse the site.

hickford:
@mxxcon what browser's that? That sure would be annoying.

Chrome for Android? http://caniuse.com/#feat=autofocus

mxxcon:
Chrome and Samsung's browser.

fpigerre:
It would also stand to reason that most new visitors to [twofactorauth.org](https://twofactorauth.org) would browse through the categories for a moment before searching for sites that they use. Autofocusing the search box immediately would likely make it harder to scroll through for those people.

PR #2350, opened 2017-02-02
check section alphabetization

stephengroat:


fpigerre:
@stephengroat, Weren't these changes already incorporated in commit 9e8b7e333f53c8b548b2b9daa8c356f0557c73f8?

screen shot 2017-02-06 at 15 41 58

From the the blame for `verify.rb`, it looks like everything's already been added.

stephengroat:
that only checks the alphabetization of the `websites` yaml files, not the `section` yaml. i haven't made a `section` yaml schema yet, but it's on the list 🎉

verify.rb
```diff
@@ -74,18 +74,18 @@ def test_img(img, name, imgs)
 # as well as if an image is missing
 begin
   sections = YAML.load_file('_data/sections.yml')
+  # Check section alphabetization
```
jamcat22: This comment is the same as line 91. Perhaps it should be changed?

stephengroat: i can change it to "Check sections alphabetization"

jamcat22: Maybe we should say "Check sections.yml alphabetization"?
Just trying to think of the least ambiguous way of saying that.

stephengroat: done

jamcat22: Great!

jamcat22 (approved):
:+1:

dsoegijono (approved):
👍

PR #2351, opened 2017-02-03
Add onename

stephengroat:


Carlgo11 (commented):
Is the site popular enough?

stephengroat:
it's the identity service of https://blockstack.org/, which is gaining popularity

Issue #1811, opened 2016-06-13
How We Classify Banks


jamcat22:
Currently, we classify a number of banks as having 2FA, when they only require it for certain transactions, not for logging in to their website. My question is whether or not we should change any of the banks listed below to `tfa: No`, add an exception to each of them, or find a way to include a `transactional` flag on the website.

Below is a list of banks that do not currently fall under our definition of 2FA, despite having the `tfa` flag set to `Yes`. I'm curious as to how everyone feels we should handle these types of situations with these banks and any additional ones going forward.
- Barclays UK's PINSentry doesn't qualify for our definition of 2FA, as their [website](http://www.barclays.co.uk/Helpsupport/UpgradetoPINsentry/P1242559314766) shows that it is only required for certain transactions.
- Citibank Australia's OTP doesn't qualify for our definition of 2FA, as it seems that they only require it for certain transactions. Their [website](http://www.citibank.com.au/otp/) says:
  ![You'll need to enter this anytime you want to perform select online transactions or query through Citibank Online.](https://cloud.githubusercontent.com/assets/1897053/15996479/f922f71c-30e4-11e6-8461-2d5fe503aff5.JPG)
- Commonwealth Bank of Australia's NetCode doesn't qualify for our definition of 2FA, as their [website](https://www.commbank.com.au/security-privacy/netcode.html) states that it's only required for certain transactions.
- first direct's Secure Key doesn't qualify for our definition of 2FA, as it can be bypassed using a password to login. Their [website](http://www1.firstdirect.com/1/2/securekey/options) says:
  ![...this doesn't mean you can't access your accounts if you don't have it with you. You can still log on without it and have limited access to Internet Banking by selecting the link at the top of the log on page...](https://cloud.githubusercontent.com/assets/1897053/15994744/e4b6055a-30c9-11e6-8b56-9d38756f6184.JPG)
- HSBC's Secure Key doesn't qualify for our definition of 2FA, as their [website](https://www.hsbc.co.uk/1/2/customer-support/online-banking-security/secure-key#demos) shows how you can choose the "Without Secure Key" tab when logging in to bypass 2FA.
- Nationwide Building Society's card reader doesn't qualify for our definition of 2FA, as the it can be bypassed using a password to login. Their [website](http://www.nationwide.co.uk/support/security-centre/internet-banking-security/card-reader-and-security-questions#xtab:what-is-a-card-reader) says:

[...]

PR #2354, opened 2017-02-04
Update Barclays UK doc URL

Stevoisiak:
The old Barclays UK doc page redirects to a general security FAQ with no mention of 2FA.

I'm not entirely sure if Barclays UK still supports 2FA. Can someone with an account confirm?

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -102,7 +102,7 @@ websites:
       tfa: Yes
       hardware: Yes
       software: Yes
-      doc: https://ask.barclays.co.uk/help/security/protection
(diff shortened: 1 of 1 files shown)


jamcat22 (changes requested):
While whether or not Barclays UK actually provides two factor authentication, rather than two factor authorization is brought up in #1811, until that issue is resolved we should continue to list it as providing 2FA.

Please update the doc link to their new page: `https://www.barclays.co.uk/Helpsupport/UpgradetoPINsentry/P1242559314766`

Stevoisiak:
@jamcat22 Updated the URL as requested

jamcat22 (approved):

PR #2356, opened 2017-02-04
Update XXL Cloud

Stevoisiak:
The original 2015 post on 2FA is no longer available.


Carlgo11 (approved):
Doesn't look like they support 2FA anymore as there are no current pages mentioning it. At least not that I could find.

fpigerre (approved):

fpigerre:
Thanks for the pull request @Stevoisiak! 👍

PR #2357, opened 2017-02-04
Contributing: Add "lang" field to example

Stevoisiak:
Added the 'lang' tag to the example new site

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -120,9 +120,11 @@ websites:
     software: Yes
     hardware: Yes
     doc: <link to site TFA documentation>
+    lang: <language code>
 ```
 
 The fields `name:`, `url:`, `img:`, `tfa:` are required for all entries.
+The field `lang:` is only used for non-English websites.
 
 #### Adding a site that *supports* TFA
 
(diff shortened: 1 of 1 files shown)


fpigerre:
Hi @Stevoisiak,

At the moment, the `lang` tag currently isn't required for websites that only use English. This is so that the majority of websites that are listed don't need a `lang` tag. If a website uses a different language, however, such as german, the `lang: de` tag is appended to that website's entry.

Thanks,
psgs 🌴

Stevoisiak:
@psgs I updated my PR to explain the `lang` tag is only used for non-English websites.

Carlgo11:
Maybe add a note that [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) codes should be used?

Stevoisiak:
@Carlgo11 Added the ISO 639-1 specifier

RichJeanes:
Perhaps include the link that Carlgo11 provided in the document so people know what that means.

Stevoisiak:
@RichJeanes I can't use hyperlinks inside a code block.  I'll have to mention ISO 639-1 below the code block instead.

RichJeanes:
That's fine. I just think it would be helpful to include it somewhere.

Carlgo11 (approved):

PR #2358, opened 2017-02-04
Update ING (Luxembourg)

Stevoisiak:


Carlgo11 (changes requested):
I think it's better to keep url as short as possible. That way the site can redirect the user more easily.
Plus, sites often change their URL structure causing our links to become invalid.

Stevoisiak:
I made this PR because I wasn't able to figure out the policy for language-specific URLs. The general trend seemed to be use the English site when available.

Carlgo11:
That is true but URL cleanliness is also a rule of thumb.

Stevoisiak:
Makes sense. The policies on page language and URL cleanliness should probably be included somewhere in the [contributing guide](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md).

Carlgo11:
Closing as discussed in https://github.com/2factorauth/twofactorauth/pull/2358#issuecomment-277482157

PR #2364, opened 2017-02-05
Basic schema for websites yml

stephengroat:


mxxcon:
yaml schema! 🙌 😄

stephengroat:
i wouldn't say it's great, but it's a start. `kwalify` is a mess, but it's something

.travis.yml
```diff
@@ -3,6 +3,8 @@ language: ruby
 cache: bundler
 rvm:
     - 2.4.0
+before_install:
+  - gem update --system --no-doc
```
Carlgo11, line 7: why is this necessary?

Carlgo11 (approved):

.travis.yml
stephengroat: `gem` has a bug that's creating a frozen sting issue, fixed in v2.6.9

jamcat22 (approved):

PR #2371, opened 2017-02-12
Travis check for bad lang tags

stephengroat:
`lang` tags should only be used on `tfa: No`, have travis check for this and fix current bad instances

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -351,7 +351,6 @@ websites:
       hardware: Yes
       software: Yes
       doc: https://www.handelsbanken.se/shb/INeT/ICentSv.nsf/Default/qAE99695C4C1D033DC12579120036EB43?Opendocument
-      lang: sv
 
     - name: HSBC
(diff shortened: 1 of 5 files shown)


jamcat22 (commented):
I'm not sure I agree with not having `lang` tags for sites with 2FA, mostly because it might aid in any future plans for a more internationally friendly design. Curious to hear other's thoughts on this.

mxxcon:
@jamcat22 But then it's the same approach as removing twitter/facebook/etc for sites that do have tfa...why remove that information since it can be useful for the future as well. 😕

jamcat22:
@mxxcon: I guess we could always add them back later if needed. 😄

Carlgo11:
@stephengroat are you okay with f8c0697?

stephengroat:
@carlgo11 of course! Thanks! The more comments the better! Isn't it a dream to have someone else comment your code for you?

Carlgo11 (approved):

PR #2373, opened 2017-02-14
Add chime

stephengroat:
#2359


mxxcon:
Should we somehow distinguish it from https://chime.aws ?

stephengroat:
it's calling itself "Amazon Chime," I think that's good enough

Carlgo11 (changes requested):
I suggest using this logo instead.
It's better for lower resolutions and the one that Chime uses as their favicon.
chimebank

stephengroat:
image updated

Carlgo11 (approved):

jamcat22 (approved):

PR #2374, opened 2017-02-15
Add Hostiso

nitrohorse:
[Hostiso uses cPanel](https://hostiso.com/speed-security/) which [has TFA](https://documentation.cpanel.net/display/ALD/cPanel+Two-Factor+Authentication). However they're lacking their own documentation formally stating so.


_data/cloud.yml
```diff
@@ -83,6 +83,13 @@ websites:
       software: Yes
       doc: https://devcenter.heroku.com/articles/two-factor-authentication
 
+    - name: Hostiso
+      url: https://hostiso.com/
+      img: hostiso.png
+      tfa: Yes
+      software: Yes
+      doc: https://hostiso.com/speed-security/
```
stephengroat: @trance-with-me since the `doc` link doesn't mention 2fa and isn't required, could you remove it?

stephengroat (changes requested):

_data/cloud.yml
nitrohorse: Sure

stephengroat (approved):

stephengroat:
@trance-with-me thanks for the PR!

nitrohorse:
No prob :+1:

Carlgo11:
I'm a bit unsure if we should add it when it uses cPanel. Maybe that falls under the self-host category or third party login category?

@2factorauth/collaborators Any thoughts on this?

jamcat22:
@Carlgo11 I think it falls under the same category which other sites like Bluehost, sites hosted with a 2FA WordPress plugin, and sites using WHM. I think it belongs in either the `Hosting/VPS` or `Cloud` categories because it provides shared hosting with 2FA (similar to DreamHost or Bluehost), rather than providing instructions for a host (running cPanel) to enable 2FA. cPanel doesn't have a single-sign-on version of their accounts, and each account on a hosting website is completely separate (again, much like it would be on WordPress or Drupal).

@trance-with-me, does 2FA also protect the [client area](https://my.hostiso.com/clientarea.php), or only the shared hosting cPanel?

nitrohorse:
@jamcat22 yeah 2FA (software token) protects the client area and a separate 2FA software token protects the shared hosting cPanel.

Carlgo11 (approved):
Guess it's good to go 👍

jamcat22 (changes requested):
I think we should add an exception stating that 2FA for the client area and shared hosting panel are separate and each require setup. Maybe something like: `While both the main client area and the shared hosting panel support 2FA, each system is separate and requires its own setup.`

jamcat22 (approved):

jamcat22:
Thanks @trance-with-me for the PR!

PR #2377, opened 2017-02-16
Add Holvi

matiaskorhonen:
Holvi provides banking and other financial services to small businesses, non-profits, and individuals in the EU.

> Founded in 2011, Holvi is a banking service designed for the new era, all about connectivity and full flexibility. Holvi is the world’s first financial hub operating with its own Payment Institution Licence authorised for operations across Europe by the Financial Supervisory Authority of Finland (FIN-FSA).

<https://www.holvi.com>


stephengroat (approved):
thanks for the pr @matiaskorhonen

_data/banking.yml
```diff
@@ -351,6 +351,13 @@ websites:
       software: Yes
       doc: https://www.handelsbanken.se/shb/INeT/ICentSv.nsf/Default/qAE99695C4C1D033DC12579120036EB43?Opendocument
 
+    - name: Holvi
+      url: https://www.holvi.com/
```
Carlgo11, line 355: @2factorauth/collaborators should we keep the URL or use the one that the current one redirects to?

RichJeanes: I don't think the redirect is an issue, especially since this listing will probably always be live, even if they stop using the "about" subdomain in the future.

RichJeanes (approved):

Carlgo11 (approved):

_data/banking.yml
Carlgo11: Thanks

PR #2380, opened 2017-02-16
Update Discover

ameykelkar:
Added two factor authentication information for Discover

[Here is the information on their Two Factor Authentication](https://www.discover.com/credit-cards/member-benefits/security-center/protect-account/account-verification.html)

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -254,9 +254,11 @@ websites:
     - name: Discover
       url: https://www.discover.com/
       img: discover.png
-      tfa: No
-      twitter: Discover
-      facebook: discover
+      tfa: Yes
+      email: Yes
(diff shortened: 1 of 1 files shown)


Carlgo11:
Based on the doc link you provided it sounds like they offer two step verification rather than two factor authentication.

ameykelkar:
@Carlgo11 Oh yes, that is right. However, when I tweeted them [this was their response](https://twitter.com/Discover/status/832127301768855552).

RichJeanes:
Considering that a code is required for logins from unrecognized devices, it would seem that this does qualify as 2FA.

RichJeanes (approved):

jamcat22 (approved):

PR #2393, opened 2017-02-21
Add Flickr

pedc:
Flickr is part of yahoo but unlike yahoo mail it does not seem to have 2FA


pedc:
oops, screwed this up somewhere....

mxxcon:
flickr uses your standard yahoo login. if you have 2fa enabled there, you'll get 2fa on flickr.

PR #2405, opened 2017-02-27
Add GamersGate

morguldir:


morguldir:
delet this

Jawshy:
Pull requests and issues cannot be deleted; only closed and/or locked.

morguldir:
@Jawshy Right, i was even reading that it cant be deleted so you cant hide what happened during development, silly me.

PR #2406, opened 2017-02-27
Add GamersGate

morguldir:


Carlgo11:
Duplicate of #2405 however I think this should be the primary one as the other one is closed.
In the future @Criminalllz, try to use just one PR.

Carlgo11 (changes requested):
Try this image instead. It's smaller
gamersgate

morguldir:
@Carlgo11 I messed up and didnt reset my local repo and it included the previous pull request aswell. And i made a new pr to only include the add-gamersgate.

stephengroat (approved):
thanks for the pr @Criminalllz!

Carlgo11:
Maybe retail would fit better?

stephengroat:
@Carlgo11 i think it's better in `gaming`, otherwise anyone that sells anything should be `retail`, right?

fpigerre:
Given that GamersGate's front page is almost solely dedicated to selling items (with prices listed), I think it would be best if it was placed in the `retail` category. It does seem like the website does have some "Achievements" and "Friends" features, however, and not only sells games but also strategy guides.

screen shot 2017-03-01 at 18 18 38

mxxcon (approved):

mxxcon:
I think either category is good enough for the site.
Right now I'm leaning more towards `gaming` though.
If they were selling gaming-related apparel or hardware or some other stuff, then I'd be leaning more towards `retail`.

Carlgo11:
From looking on their description about themselves it seems like their main feature is selling games.

Carlgo11 (approved):
Moved GamersGate to `retail`.

fpigerre (approved):

PR #2420, opened 2017-03-03
Semantic and jets from cdnjs

stephengroat:
# **DO NOT MERGE YET**
Currently changes the way the page looks (on my box locally at least)
Need @psgs to look it over and see what I messed up


fpigerre:
Hi @stephengroat,
Awesome stuff! 👍 

Because Semantic-UI is no longer built manually, the [EM (relative) units](https://j.eremy.net/confused-about-rem-and-em/) that determine the size of fonts and various other relative elements are now set to the default value of `14px`. This means that the version of the site served solely through the CDN (using default values) appears slightly smaller than the primary version of the website, where the base size for EM values had been set to `18px`.

These variables are set in [`src/themes/default/globals/site.variables`](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/themes/default/globals/site.variables#L28) before compiling Semantic-UI.

I don't know whether it's possible to change these values when serving Semantic-UI from a CDN, but it might be possible to find a CDN that hosts different versions of the library, with different EM sizes.

Also, would it be possible to add a [Subresource Integrity Hash](https://www.srihash.org/) for each of the files served from the CDN? Because we're now serving them from a different domain that we have no control over, adding an SRI hash will help mitigate the any issues that may involve the CDN serving the wrong files, or an attacker using the CDN to inject code not intended to be used on twofactorauth.org.

fpigerre (changes requested):
* Default EM value should be `18px`
* Files loaded from separate domains should have SRI hashes

mxxcon:
Also let's hardcode urls to be https:// rather than protocol-relative. No reason to serve it over plain text http.

Carlgo11:
Closing for the time being. Reopen this PR when it's good to go

jamcat22:
This PR has been superseded by #2427.

PR #2426, opened 2017-03-05
Add myQNAPcloud

cez81:


Carlgo11:
@2factorauth/collaborators Does this fall under the self-hosting category?
It relies on your own NAS storage but it would appear that the authentication is centralized.

stephengroat:
i think centralized auth is what this site is focused on

stephengroat (approved):

jamcat22 (changes requested):
Currently, we list both QNAP and Synology NAS devices. Technically speaking, these should both fall under the umbrella of "self-hosted" and most likely should not be posted on twofactorauth.org. However, in the instance of Synology devices, the 2FA enabled directly on the device also affects accessing the device via a Synology account, even though the account itself isn't protected. In the case of QNAP, it appears there's both 2FA functionality for the user account on the device and their myQNAPcloud account.

This entry would not be considered "self-hosted", as it pertains to an online account system, rather than the local user account on the device itself. The same can be said for Synology's online account system.

I think we need to address this issue before we can merge this PR.

Questions to consider:
1. Should we remove the QNAP entry and instead post this entry for their online account service?
2. Should we remove the Synology entry? Does the fact that it protects the device when being accessed through an account count as 2FA?
3. If we keep both QNAP and Synology, should we post this entry as well?
4. If we remove the Synology entry, should we create an entry for their online account system?

Carlgo11:
Closing due to inactivity

PR #2427, opened 2017-03-06
Load third-party libraries using Cdnjs

stephengroat:
Include semantic ui from cdn.

still need a tiny bit of help, my font-size override trick works for everything except the `icon` in the `header`


mxxcon:
Let's implement SRI tags. You can use https://www.srihash.org/ to speed things up.

stephengroat:
OK. Let's get the sizing down first

fpigerre:
Awesome stuff @stephengroat! 👍 

At the moment, the Semantic-UI library sets the icon font-size to `2em` through the following classes: 
```.ui.icon.header .circular.icon, .ui.icon.header .square.icon {```
It's quite likely that somewhere, either in our own CSS or in Semantic-UI's CSS, there is a line of code that sets a base font-size lower than 18px specifically for a class which affects the logo icon. I'll have a dig around over the next day or two and see what I can find.
You can also see exactly which classes have what effects on particular elements by right-clicking on an element and selecting "Inspect". From there, specific classes provided for by the CSS appear in the Styles tab. The "Computed" tab is also really useful, as it shows the whole collection of styles applied to that specific element, regardless of how they were applied.

Would you be able to re-factor the `style="font-size:18px"` snippet into the `css/base.scss` file by any chance? This helps keep all the CSS in one place.

Thanks,
psgs 🌴

**Review by fpigerre — CHANGES_REQUESTED** [review]
* Refactor style tag into `base.scss`
* Fix logo icon sizing
* Implement SRI hashes

**Comment by stephengroat** [comment]
@psgs thanks for the help, i don't really know what i'm doing with css (learning curve is steep).

i moved the `font-size` to the css and it seems to fix the `icon` issue, could you look the size over and see if you can tell any major differences? i tried it on chrome and firefox and it looks the same (to me)

**Comment by fpigerre** [comment]
Awesome work @stephengroat! 👍 

The only difference with the sizing of elements that I can see is a slight difference in the width of table columns. It's probably caused by several elements gradually inheriting a larger font size. Visually, the difference is negligible, however, so I think we can go ahead and merge the pull request.

@Carlgo11, @mxxcon: Should we specify one particular type of hash to use for SRI hashes (eg. `sha-256`), or allow different hash types to be used for each resource?

**Review by fpigerre — APPROVED** [review]

**Comment by mxxcon** [comment]
>Should we specify one particular type of hash to use for SRI hashes (eg. sha-256), or allow different hash types to be used for each resource?

I think it doesn't really matter as long as they are secure hashes(not sha1/md5). We could stick to one type just for consistency sake.

**Comment by jamcat22** [comment]
I think if we decide to stick to one type, we should use sha-384, since that's what https://www.srihash.org uses when generating hashes.

**Comment by stephengroat** [comment]
The srihash.org generated hashes weren't working for the Google fonts, so I
switched to Chrome generated 256 that worked

**Comment by stephengroat** [comment]
@psgs i'm still seeing a problem for the `backup` entry `pogoplug`

do you see anything?

**Comment by fpigerre** [comment]
@stephengroat, I can't seem to see any problems with Pogoplug.
Do the screenshots below seem right to you?

screen shot 2017-03-08 at 22 17 27


screen shot 2017-03-08 at 22 17 09

**Comment by stephengroat** [comment]
@psgs you're 100% right. i was comparing against an unmerged live copy that still had the `in progress` badge on it and was getting VERY confused!

**Comment by stephengroat** [comment]
@2factorauth/collaborators can some more people look this over? i think this one needs more than 2 reviews

File: _layouts/default.html [hunk 1]
```diff
@@ -16,8 +16,10 @@
 
   <title>{{ site.title }}</title>
 
-  <link rel="stylesheet" href="//fonts.googleapis.com/css?family=Open+Sans:400,700">
-  <link rel="stylesheet" href="/lib/semantic/css/semantic.min.css">
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:400,700"
```
mxxcon: I think we should NOT use SRI for Google Fonts. Apparently CSS they serve is intentionally unversioned and it changes depending on which browser requests it.
See https://github.com/google/fonts/issues/473

mxxcon (approved):

jamcat22 (changes requested):
Is it possible to change the hashes of our jquery source to be sha384?

jamcat22 (approved):

fpigerre:
👍

Issue #2441, opened 2017-03-10
Please Add TaxAct (SMS Support)


CometDog:
TaxAct currently supports SMS (or I suppose email based) 2FA.

There is no direct information on their site about it past their general Security page

https://www.taxact.com/security/

stephengroat:
thanks for the suggestion @PiwwowPants!

PR #2442, opened 2017-03-10
add taxact

stephengroat:
#2441


mxxcon (approved):

jamcat22 (changes requested):
I think we should use this link, as it talks about both email and text codes: https://www.taxact.com/support/24138/2016/how-is-a-verification-code-used

jamcat22 (approved):

PR #2444, opened 2017-03-11
Update Instagram

anotherus3r:
Instagram supports 2FA using SMS, but I couldn't find any documentation for it on the web.
instagram-2fa-screenshot


stephengroat (approved):

jamcat22 (approved):
It looks like they've finally enabled it for everyone! Only took a little more than a year.

PR #2448, opened 2017-03-14
add Charles Schwab bank

michaelhwan:


stephengroat:
thanks for the pr @michaelhwan, but we already have Schwab listed under `investing`

If you have any other ideas, please let us know!

PR #2450, opened 2017-03-15
Update fidelity with real doc

stephengroat:


fpigerre (approved):

jamcat22 (changes requested):
Fidelity's doc does not mention the fact that they support hardware-based 2FA. I'm worried that by linking to their doc, we may confuse users who do not understand that hardware tokens are supported. See #1519.

stephengroat:
since fidelity no longer recognizes on their website, i've removed it

jamcat22:
Technically speaking though, they do support it. Much like how PayPal does not document support for software 2FA anymore, but they do still support it in their account settings.

PR #2451, opened 2017-03-15
Update eso doc

stephengroat:


jamcat22 (approved):
Even though their FAQ section has multiple docs pertaining to this, I'd say the one linked in this PR is the most through.

fpigerre (approved):

fpigerre:
Although it seems as if the Notes page is worded more clearly and oriented more specifically towards two factor authentication, in this case it is probably preferable to link to an external website rather than maintaining a documentation page. 👍

PR #2453, opened 2017-03-16
Move chase notes to exception

stephengroat:


jamcat22 (commented):
I'm not sure this makes sense as an exception. Usually, we only use exceptions to notate when a service has either limited 2FA coverage (ex: Square's 2FA only covers the dashboard), requires SMS to enable 2FA, or when 2FA is only available to certain customers (ex: only paid plans offer it, or SMS is only supported for the US).

I think a note makes more sense than an exception in this case, as we can provide more detailed information on how to enable it, in an easier to read format.

stephengroat:
centralizing the data in yaml files is also really important though. very few people are going to edit the raw html. also, we don't give detailed instructions for any other sites, so why do it for 6 cases? the site www.turnon2fa.com is give screenshot walkthrough the turn on 2fa, but i don't think this site was to try to maintain that detail

Carlgo11 (changes requested):
I agree with [@jamcat22](https://github.com/2factorauth/twofactorauth/pull/2453#pullrequestreview-27487578) here.

PR #2454, opened 2017-03-17
Added Länsförsäkringar to Banking section

bjorno:
Added the Swedish bank "Länsförsakringar" to the banking section.


stephengroat (approved):
thanks for the pr @bjorno! i just removed the `lang` tag (we only use that for `tfa: No`) and compressed the img using tinypng.

fpigerre (approved):

fpigerre:
Although the page listed for the documentation site seems to provide all the necessary answers, it might be more appropriate to list the [Säkerhetsdosan](http://www.lansforsakringar.se/privat/att-vara-kund/digitala-tjanster-telefonbank/internetbanken/sakerhetsdosa/) page instead.
@Carlgo11, would you like to weigh in on this issue?

Carlgo11 (approved):
@psgs no I think the existing link is better as it lists all 2FA models and not just hardware.

PR #2456, opened 2017-03-21
Add bitsewa

tuladhar:
Bitsewa is Nepal's most popular Bitcoin and Blockchain company. We have just recently added 2FA, it would be great if we can get listed under https://twofactorauth.org/#cryptocurrencies


stephengroat:
could you update the image with this? 60% smaller

bitsewa

tuladhar:
Thanks for the advise!

I have added the optimized image. Sorry for duplicate commits, as I had accidently forgot to rename the filename. I have fixed it. So everything should be good. 

Cheers!

stephengroat (approved):
thanks for the pr @tuladhar! don't worry about the multiple commits, whoever merges can squash all of the commits (even in the web interface now!)

fpigerre (approved):

PR #2460, opened 2017-03-23
Update Square

jtamplin:
Google Authenticator (or any other software/hardware that supports RFC6238 TOTP)
is now supported everywhere, while SMS 2FA is only supported in US/CA.  I didn't
list it as supporting hardware tokens despite some hardware supporting TOTP
since otherwise there would be no point of having a hardware entry separate from
software.

### DIFF
--- a/_data/payments.yml
+++ b/_data/payments.yml
@@ -125,8 +125,9 @@ websites:
       img: squareup.png
       tfa: Yes
       sms: Yes
+      software: Yes
       exceptions:
-        text: "Currently only US merchants, protects only Dashboard access"
+        text: "SMS only supported in US and Canada"
       doc: https://squareup.com/help/us/en/article/5593-2-step-verification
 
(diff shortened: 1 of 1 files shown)


_data/payments.yml
```diff
@@ -125,8 +125,9 @@ websites:
       img: squareup.png
       tfa: Yes
       sms: Yes
+      software: Yes
       exceptions:
-        text: "Currently only US merchants, protects only Dashboard access"
+        text: "Outside US/CA, only Google Authenticator (or any TOTP client) are supported"
```
mxxcon: Perhaps rephrase it that SMS is supported only in us/canada? Since that's the limitation, not totp.

jtamplin: Updated.

jamcat22 (approved):

mxxcon (approved):

PR #2465, opened 2017-03-25
Add Government Gateway

djrarky:
Government Gateway is the main UK government login for many sites such as the DVLA (car tax, licence sharing), HMRC (PAYE and self-assesments), Universal Credit, state pension.

It definitely needs to be included in the two-factor auth document as it isn't well documented.


djrarky:
HMRC doesn't have its own portal.
 HMRC is Government Gateway. 
DVLA is Government Gateway.
etc.

Carlgo11:
Unable to merge as the branch has been removed.

Issue #2455, opened 2017-03-20
ebay


ghctim:
ebay no longer supports 2fa.  All links to register are dead.

stephengroat:
eBay now support 2fa over SMS only, no longer uses the Symantec VIP soft token

not the best source, but a source (https://tamebay.com/2017/03/ebay-to-switch-to-mobile-2-factor-authentication-instead-of-paypal-key.html)

ghctim:
How do you enable it? All instruction links don't work

stephengroat:
they seem pretty worthless at documentation.

I've [tweeted at them](https://twitter.com/stephengroat/status/843855195717099520), hopefully they'll show us some documentation soon!

I'll try to update this thread with whatever they tell me!

ghctim:
@stephengroat that article references a link for the UK, doesn't work with ebay.com....

stephengroat:
try this

http://rover.ebay.com/rover/1/710-53481-19255-0/1?campid=5335837209&customid=&toolid=10001&mpre=http%3A%2F%2Fsigninalert.ebay.com/sgninalrt/twofa/onboard

ghctim:
signed up! That is sweet! Thank you. Recommend updating the site doc for this....

stephengroat:
from [AskeBay](https://twitter.com/AskeBay/status/844014526492303362?refsrc=email), new 2fa users currently can't turn on the feature.

do we want to update ebay with 2fa pending status @2factorauth/collaborators?

melvynm:
Since that link works, and it's being emailed around by eBay and 2FA enthusiasts, why "Pending"? 

eBay now supports 2FA by SMS.

stephengroat:
@melvynm Only works for existing 2fa users. Users who didn't have the feature before won't be able to turn it on

ghctim:
I didn't and I was able to turn on

ideologysec:
Worked for me as well, and I have never had 2FA on eBay enabled.

ohheyrj:
Same for me, never had 2FA on my account enabled and it is now working.

gingerbeardman:
I just turned this on for the first time and it's working just fine for me

ghost:
@stephengroat Your link works fine even with eBay Italy; just replace both instances of ebay.com with ebay.it in the URL. Thanks!

gingerbeardman:
[...]

PR #2468, opened 2017-03-25
Update eBay

djrarky:
As per #2455

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -80,11 +80,7 @@ websites:
       url: http://ebay.com
       img: ebay.png
       tfa: Yes
-      hardware: Yes
-      software: Yes
-      exceptions:
-          text: "eBay only supports 2FA in some countries."
-      doc: http://pages.ebay.com/securitykey/faq.html
+      sms: Yes
 
     - name: Etsy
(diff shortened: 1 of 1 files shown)


stephengroat (changes requested):
because of [the specific twitter advice from the official AskeBay account](https://twitter.com/AskeBay/status/844014526492303362), I don't think it's a good idea to update this yet with `sms: Yes`. if anything, update with `tfa: No` or some strange `exception`

djrarky:
I think an exception might be a good idea, I activated 2fa with the link in the issue thread just before I made the pull request to test. 

What do you think? I can make the change tonight.

Carlgo11:
I just signed enabled SMS 2FA so it seems to work okay as long as you know the link to enable it.
@stephengroat what's your reasoning for not adding `SMS: Yes`?

stephengroat:
according to [their tweet](https://github.com/2factorauth/twofactorauth/issues/2455#issuecomment-288090016), this isn't really a supported feature yet. i don't know if we should update until it's actually supported

Carlgo11 (approved):

stephengroat:
Closing since it's not officially supported yet #2455

PR #2476, opened 2017-03-28
Add Real Me

QuestionableChanges:


_data/government.yml
```diff
@@ -11,6 +11,15 @@ websites:
       img: healthcaregov.png
       tfa: No
 
+    - name: RealMe
+      url: https://www.realme.govt.nz/
+      img: realme.png
+      tfa: Yes
+      sms: Yes
+      hardware: Yes
```
mxxcon, line 26: What kind of hardware 2fa does it support? It seems like only sms and google authenticator.

QuestionableChanges: Looks like an RSA token that can be provided by another organisation.

On the documentation link provided (https://www.realme.govt.nz/help/#second-factor-authentication) the following texr can be found,
"If you need to use a code, you’ll be told how you’ll get it when you log in. This could be: 
*   by text to your registered mobile phone  
*  using Google Authenticator. 
*   using a RealMe token (you’ll be sent this by the organisation that you’re logging into)"

Looking at the glossary of terms for developers (https://developers.realme.govt.nz/about/glossary/) we see,\
"RealMe code by token	A one time code generated by the RSA token that is sent to the user...."

stephengroat (approved):

QuestionableChanges:
Any reason this hasn't been merged for 3 months?
Is there anything I need to fix?

mxxcon (approved):

PR #2480, opened 2017-03-30
Updated Patreon

vlanard:
We've added sms 2fa support.

### DIFF
--- a/_data/other.yml
+++ b/_data/other.yml
@@ -279,6 +279,7 @@ websites:
       url: https://www.patreon.com/
       img: patreon.png
       tfa: Yes
+      sms: Yes
       software: Yes
       doc: https://patreon.zendesk.com/hc/en-us/articles/206538086-How-Do-I-Set-Up-2-Factor-Authorization
 
--- a/_data/other.yml
+++ b/_data/other.yml
@@ -280,7 +280,6 @@ websites:
       img: patreon.png
       tfa: Yes
       sms: Yes
-      software: Yes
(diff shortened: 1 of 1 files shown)


fpigerre:
Thanks for the pull request @vlanard! 👍

fpigerre (approved):

jamcat22:
It looks like software 2FA support is no longer an option for those without 2FA currently setup. Should we keep the `software: Yes` tag on it?

What do you think @vlanard and @2factorauth/collaborators?

RichJeanes:
To quote the [Support page](https://patreon.zendesk.com/hc/en-us/articles/206538086-How-Do-I-Set-Up-Two-Factor-Authorization-):

> If you have enabled totp two factor (Google Authenticator) you do not need to switch to sms. If you decide to remove two factor or switch to sms you will lose access to totp two factor authentication on Patreon. Moving forward, SMS two factor authentication will be our main two factor security preference.

stephengroat:
i'd say remove `software`

RichJeanes:
Agreed.

vlanard:
Will do. It felt kind of ambiguous since existing users can use Software SMS.

RichJeanes:
While users that already have a software (TOTP) token enabled can continue to use it, users who do not have 2FA enabled are more likely to be looking at 2FA.org for what is available and will not be able to enable a software token.

vlanard:
done !

stephengroat (approved):

PR #2481, opened 2017-03-30
Corrected Loopia exception.

westlund:
Citizenship is not required - a permanent residence with a Swedish personal number from the Swedish tax authorities is enough.

### DIFF
--- a/_data/domains.yml
+++ b/_data/domains.yml
@@ -284,7 +284,7 @@ websites:
       software: Yes
       doc: https://blogg.loopia.se/2015/03/16/nu-kan-du-logga-in-med-tvafaktorsautentiering-hos-loopia-genom-bankid/
       exceptions:
-          text: "TFA only available for Swedish citizens"
+          text: "TFA only available for Swedish residents with BankID."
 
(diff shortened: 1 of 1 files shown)


_data/domains.yml
```diff
@@ -284,7 +284,7 @@ websites:
       software: Yes
       doc: https://blogg.loopia.se/2015/03/16/nu-kan-du-logga-in-med-tvafaktorsautentiering-hos-loopia-genom-bankid/
       exceptions:
-          text: "TFA only available for Swedish citizens"
+          text: "TFA only available for Swedish residents with BankID."
```
Carlgo11, line 287: So you don't have to be a Swedish citizen to use BankID? Because it asks for your 'personnummer' which only works with Swedish citizens

Carlgo11: From their website: 
```
BankID is a citizen identification solution that allows companies, banks and governments agencies to authenticate and conclude agreements with individuals over the Internet.
```
\- https://www.bankid.com/en/

Carlgo11 (approved):

PR #2496, opened 2017-04-12
Add Ulule

TJ619:
The EU crowdfunding website (like Kickstarter for US)


stephengroat (approved):
thanks for the pr @TJ619

_data/other.yml
```diff
@@ -343,6 +343,13 @@ websites:
       tfa: Yes
       software: Yes
 
+    - name: Ulule
+      url: https://ulule.com/
```
Carlgo11: Please change the URL to `https://www.ulule.com/`

Carlgo11 (dismissed):

_data/other.yml
```diff
@@ -343,6 +343,13 @@ websites:
       tfa: Yes
       software: Yes
 
+    - name: Ulule
+      url: https://www.ulule.com/
+      twitter: ulule
+      facebook: ulule
```
jamcat22: Their Facebook page redirects to `ululeEN` for me. Does anyone else get a different result after going to [facebook.com/ulule](https://www.facebook.com/ulule)?

jamcat22 (changes requested):

_data/other.yml
stephengroat: it's doing that based on globalization techniques available for facebook pages. it's the best option to link against

Carlgo11: Well `ulule` seems to be in French so why not use `ululeEn`? Also, it doesn't redirect for me.

TJ619: For me (in FR) the ulule facebook didn't redirect me to the FR one, seem to be better for all.

stephengroat: i believe that `ulule` is the company's [Facebook root or default page](https://www.facebook.com/business/help/331800410323820)

Carlgo11: I'm not saying it redirects to another facebook page. Just that `ulule` seems to be a French version.

## ulule
ulule-screenshot
## ululeEn
ululeEn-screenshot

stephengroat: for me, `ulule` redirects to `ulueEn`

Carlgo11: @stephengroat are you signed in to Facebook?

stephengroat:
i get redirects when signed in and when not

i am behind a pretty aggressive proxy though

_data/other.yml
TJ619: Oh yes i just see.
Yes the **ulule** page is the French Page (cause created in France before expand for EU)
So yes, better for here to put **ululeEn** page

TJ619:
Yes the ulule page is the French Page (cause created in France before expand for EU)
So yes, better for here to put ululeEn page for all.

jamcat22:
I think in the context of how we currently handle things (forcing English websites unless they're a regional service such as a bank) we should probably force their English Facebook page as well. If we later implement options for other languages then we can figure something out.

Cross ref #2510.

jamcat22 (approved):

PR #2505, opened 2017-04-15
Add ClearScore

ohheyrj:
ClearScore is a credit score tool provided for the UK. It is similar to Noddle which is already listed.


Carlgo11:
Maybe this logo is better? 
img

ohheyrj:
Yup, I agree. I tried to get a image with the company name under it. But I'll update the image to that one instead 😊

Carlgo11:
Text in the logo is often something to avoid as it will look distorted when shrunken down to 32x32 pixels.

ohheyrj:
That should be the image updated now

Carlgo11 (approved):

jamcat22 (approved):

jamcat22:
Thanks for the PR!

PR #2522, opened 2017-04-28
Add affinity federal credit union

stephengroat:


RichJeanes (approved):

RichJeanes (changes requested):
I see that the original submitter stated they support SMS, but that linked doc page makes no mention of it and just seems to be a superfluous marketing page. I would suggest removing it and (if one can be found) adding a more appropriate support page or something that talks about how to enable SMS verification. Or at least something that says it is enabled by default if that is the case.

PR #2523, opened 2017-04-29
Add Linevast

redsuperstructure:


stephengroat (approved):
thanks for the pr @redsuperstructure. i couldn't find any online documentation to add to the `doc` tag, could you confirm that the site is using TOTP-style authentication (Google Authenticator style)?

redsuperstructure:
yes, linevast is using whmcs which has an google authenticator integration.

Carlgo11:
Doesn't look very popular http://www.alexa.com/siteinfo/linevast.de.
Since we can't list all sites on the Internet we try to limit the list to only include fairly popular services.

PR #2526, opened 2017-05-04
Update TransferWise to have TFA for SMS and Phone

tyiu:


stephengroat (approved):
thanks for the pr @tyiu

Carlgo11:
The doc link doesn't mention how the authentication works. Are you required to use 2FA every time you log in? Is it only when performing certain actions?

tyiu:
@Carlgo11 They have a different doc explaining how it works, but not how to enable it: https://transferwise.com/help/topics/4/topics/19/questions/2597959/what-is-2-step-login

Should I link to that URL instead?

To answer your question, you are required to use 2FA every time you log in after you have enabled it. They also claim, in that URL, that "In order to keep your money safe, setting up 2-step is mandatory when using certain features of your account, including uploading money and setting up account numbers."

Carlgo11:
@tyiu No the current doc link is fine. I just wanted to know how they have implemented 2FA.
One last thing. They don't mention 2FA via phone calls. Do they actually support that?

tyiu:
@Carlgo11 Sounds good. Yes, they can call you with a code. Just tried it out!

Carlgo11 (approved):

PR #2535, opened 2017-05-11
Add ImmobilienScout24 as a service

ruempel:
ImmobilienScout24 supports 2FA via authenticator app, but does not provide a public documentation.


_data/other.yml
```diff
@@ -203,6 +203,13 @@ websites:
       software: Yes
       doc: https://ifttt.com/settings#two-step-verification
 
+    - name: ImmobilienScout24
+      url: https://www.immobilienscout24.de
+      img: immobilienscout24.png
+      tfa: Yes
+      software: Yes
+      lang: de
```
stephengroat: could you remove the `lang` tag? we don't use it when `tfa: true`

stephengroat (changes requested):
also, do you have any links to documentation that we could add to the `doc` tag?

ruempel:
There is no publicly available documentation on 2FA. However, there are hints available at https://sso.immobilienscout24.de/account/verwaltung when logged in, but this link is useless for public audience.

stephengroat (approved):

stephengroat:
@ruempel no problem. thanks for the addition and quick fixes!

fpigerre (approved):

PR #2540, opened 2017-05-16
Fix Dominion Energy

stephengroat:


_data/utilities.yml
```diff
@@ -73,9 +73,8 @@ websites:
       tfa: No
 
     - name: Dominion
```
jamcat22: It appears they've [rebranded](https://www.dominionenergy.com/NewBrand) as "Dominion Energy", so I think we should update their name to match that.

_data/utilities.yml
```diff
@@ -73,9 +73,8 @@ websites:
...
-      url: https://www.dom.com/
-      twitter: DomVAPower
-      facebook: dominionvirginiapower
+      url: https://www.dominionenergy.com/
+      twitter: dominionenergy
```
jamcat22: It appears that this account is solely used as a corporate news Twitter, and it doesn't look like they reply to tweets. Should we use a different account?

jamcat22 (changes requested):

_data/utilities.yml
stephengroat: there isn't any other official account linked on their website

Carlgo11 (approved):

_data/utilities.yml
Carlgo11: chrome_2017-06-22_20-30-11
There are actually a few replies from the account.

jamcat22 (approved):

PR #2542, opened 2017-05-16
Add 1984 hosting

Efreak:
1984 hosting doesn't have a favicon. Their only logo is a stylized 1984 with a 4:1 aspect ratio, so I'm not really sure what to do for an icon...

### DIFF
--- a/_data/domains.yml
+++ b/_data/domains.yml
@@ -28,6 +28,12 @@ websites:
       twitter: 123reg
       facebook: 123regfans
 
+    - name: 1984 Hosting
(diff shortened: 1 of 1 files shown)


stephengroat:
i looked too, nothing easy on their official site. i think taking that 4:1 icon and, keeping aspect ratio, putting it into a `32x32` img makes it illegible, right?

Carlgo11 (changes requested):
Looks like the company isn't very big and that the service isn't used by that many.
We try to only list sites that are fairly popular as listing every single website on the internet would be petabytes of data.

PR #2544, opened 2017-05-18
Add facebook to quora

stephengroat:


mxxcon (approved):

Carlgo11 (changes requested):
There's no way to send messages to them on Facebook

PR #2547, opened 2017-05-20
Change LICENSE from MIT to GPLv3

trevethantrump:
May 20 2017
Emily trevethan lozano


Carlgo11:
And you propose this change because?

PR #2548, opened 2017-05-21
Add Figma

bfarooqi:
Add [Figma](https://www.figma.com/).


stephengroat (approved):
thanks for the pr @bfarooqi

Carlgo11 (approved):
I have no idea how you compressed the image that good but still kept the lines so smooth. 👍

jamcat22:
It looks as if SMS is only offered as a backup option...

@bfarooqi can you confirm that this is the case?

bfarooqi:
@jamcat22 Figma supports both software and SMS tokens. This is our [help article](https://help.figma.com/#_two-factor-authentication) specifying both (you can also sign up for free to double check if you're interested).

Carlgo11:
@bfarooqi

You're saying it's not a backup method?

PR #2551, opened 2017-05-27
Update Runescapes Doc link

tim-schneider:
fix RuneScapes documentation link


Tim-Schneider:
I'd also like to update the icon but they only granting fair use. Is that an issue?

jamcat22:
This PR is a duplicate of #2552.

PR #2478, opened 2017-03-29
Add html-proofer, Fix Fidelity Notes, Update Various Sites

stephengroat:
fixes bugs in fidelity notes and updates sites (identified using html-proofer)

html proofer now has a cache (it can cache websites that it searches). i'm working on getting that working 100%, but it's not there yet

PR #2554, opened 2017-05-27
Fix 404 s

galeksandrp:


stephengroat:
thanks for the pr @galeksandrp, we have all of these covered in #2478

Issue #2562, opened 2017-06-09
lobste.rs supports TOTP 2FA


strugee:
https://lobste.rs supports two-factor authentication via TOTP. Unfortunately I don't have a lot of time or I'd just PR this myself, but I wanted to at least file an issue so I don't forget :)

Thanks for maintaining this project <3

stephengroat:
thanks for the info @strugee.

i submitted a request to get an account to confirm. once i do, i'll make the pr

strugee:
Whoa, that was fast! Thanks @stephengroat :)

PR #2563, opened 2017-06-09
Add lobsters

stephengroat:
#2562


mxxcon (approved):
Nothing wrong with the data, but does this site fit our criteria to be listed?

jamcat22:
@mxxcon Usually we allow all sites that _support_ 2FA, only applying the criteria to sites which _do not_ support 2FA. An exception I've seen for this is when a site which only contains a forum is submitted which supports 2FA and is not popular enough to justify adding.

Carlgo11 (dismissed):
Too insignificant

stephengroat:
@Carlgo11 curious why that is. it's an alexa rank ~170k. just ran a quick script, we currently have a lot of other sites about the 170k mark.

should we implement an alexa score minimum?

RichJeanes:
Site seems to be growing. Now in the top 110k.
http://www.alexa.com/siteinfo/lobste.rs
@Carlgo11 ?

jamcat22 (approved):
As I said earlier, I'd likely not add this site if they did not support 2FA. But since they do, I believe they should be added.

fpigerre:
Would it be worth adding [this link](https://lobste.rs/s/1cyltz/two_factor_authentication_now_available) as the `doc` link?

fpigerre (approved):

PR #2573, opened 2017-06-16
Add Readthedocs

stephengroat:
I understand that the `twitter` is not a support channel, but it does provide public shame


fpigerre:
Hi @stephengroat 😄 

How would this site fit into the current [website criteria](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria)?

stephengroat:
I think it has to deal with Data (2) and Control (4). Sites like to auto-push docs to readthedocs.io, don't want those changing, providing mis-information, etc.

_data/developer.yml
```diff
@@ -315,6 +315,13 @@ websites:
       software: Yes
       doc: http://support.beanbaginc.com/support/solutions/articles/3000006998-enabling-two-factor-authentication
 
+    - name: Read the Docs
+      url: https://readthedocs.org
+      img: readthedocs.png
+      tfa: No
+      twitter: readthedocs
```
Carlgo11: Doesn't seem like they're monitoring their twitter account. I suggest we only use the email address.

Carlgo11 (changes requested):

Carlgo11:
I think it can pass. Looks like a pretty big service and it shouldn't be very difficult for them to add 2FA.

Maybe a ticket on their Github could be a good first stage though? They're a open source project like us and so why not give them the benefit of not getting their email address spammed before they get a chance to reply.

Carlgo11:
@ericholscher and/or @agjohnson would any of you like to comment?

agjohnson:
2fa is on our list, and something we'd certainly take contributions on. Currently it isn't implemented however, and isn't on our roadmap just yet.

I'm not familiar with the purpose of the above changes, but as to the question around twitter or email: we don't consider either as support channels for our community platform, so both are highly likely to be ignored.

If the point is to encourage us to implement 2fa, we'd just tell anyone looking for this feature that patches are encouraged. Perhaps you need a "contribute a patch" call to action on the site? 🙂

mxxcon:
Perhaps `exceptions:` is in order then indicating that 2fa is not planned but could be contributed?

ericholscher:
Opening a ticket on https://github.com/rtfd/readthedocs.org is probably the best course of action. Definitely would love to have 2fa, but needs to not be a large operational overhead. We're Django, with django-allauth as our auth mechanism. I wouldn't be surprised if there are plugins for this.

jamcat22:
Referencing rtfd/readthedocs.org#3523 in case we decide to link to it.

Carlgo11 (approved):

mxxcon (approved):

PR #2576, opened 2017-06-19
Add Fastly to cloud section

tobiastornros:


stephengroat:
@tobiastornros thanks for the pr, but we already have [fastly in Hosting](https://twofactorauth.org/#hosting)

PR #2589, opened 2017-06-24
Add a notice about Launchpad being still in beta.

Flameeyes:
Even three years later it appears you need to join the beta testers group to be able to use 2FA on Launchpad.

### DIFF
--- a/_data/developer.yml
+++ b/_data/developer.yml
@@ -209,6 +209,8 @@ websites:
       hardware: Yes
       software: Yes
       doc: https://help.ubuntu.com/community/SSO/FAQs/2FA
+      exceptions:
+          text: "TFA available only after joining beta testers group"
 
     - name: Looker
       url: https://looker.com/


RichJeanes:
This is literally the first step listed in the linked instructions. What purpose does it serve to add this exception to the site?

Flameeyes:
I'll be honest: I don't usually look at instructions.

For most website when there is a tick on the options, you can just navigate to Account->Security and you find what you want. Launchpad is clearly an exception to this rule, as you actually *have* to read those instructions and do a number of steps before getting there.

RichJeanes:
That's fair, but the exception text that you have submitted doesn't seem to be directly helpful. This is why we link to the documentation in the first place; not every site makes setup obvious.

And you say you don't _usually_ look at instructions. When you can't figure something out on your own, do you then go back and look at the instructions? That would seem the logical course of action to me.

fpigerre (approved):

stephengroat:
i kinda agree with @RichJeanes on this one

also, it one more piece of info to maintain that will eventually go out of date. at least the doc seems to be better maintained

Flameeyes:
I'll withdraw this, then. Although I still find the Launchpad is egregiously complicated for a service that has been tested for this long :(

PR #2590, opened 2017-06-24
Add a note about Online.net SMS only available in France

Flameeyes:


mxxcon:
Is there any documentation mentioning that?

Flameeyes:
Not that I can find documented, but this is what I have on my console.

image

mxxcon (approved):

PR #2594, opened 2017-06-24
Add Vodafone Italy.

Flameeyes:
Same logo as Vodafone UK, no 2FA option.

### DIFF
--- a/_data/utilities.yml
+++ b/_data/utilities.yml
@@ -327,7 +327,7 @@ websites:
 
     - name: Vodafone (IT)
       url: http://www.vodafone.it/
-      twitter: vodafoneit
+      twitter: VodafoneIT
       facebook: vodafoneit
       img: vodafone.png
       tfa: No
--- a/_data/utilities.yml
+++ b/_data/utilities.yml
@@ -325,6 +325,13 @@ websites:
       img: virginmobile.png
       tfa: No
 
+    - name: Vodafone (IT)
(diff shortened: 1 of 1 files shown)


RichJeanes:
Does every region of Vodafone have an independent login system? As in, if one gets 2FA, would they all? I don't know how independently the branches of the company operate. It just seems excessive to list every country they operate in if it isn't completely necessary.

Flameeyes:
Unfortunately they all operate independently :( Effectively Vodafone does not share anything but marketing material between countries.

For example, Vodafone Italy even has a terrible website without HSTS and with a number of entrypoints for downgrades.

stephengroat (approved):
thanks for the pr and explanation @Flameeyes

Carlgo11 (approved):

Carlgo11 (changes requested):
Please change the Twitter handle to `Vodafoneit`

Flameeyes:
I've styled it the way it's styled on their twitter page (VodafoneIT). Previously I used the way it's styled on their homepage.

Carlgo11 (approved):

Carlgo11:
Wow, they have a valid HTTPS cert that they use to redirect users to their HTTP version... 😒

Flameeyes:
They don't even force-redirect on pages where they ask for credit cards :|
I have been trying to get in touch with someone with a clue about their IT for a few months, but mostly gave up.

PR #2603, opened 2017-06-29
Enforce normalization on urls

stephengroat:
Creates consistency

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -7,7 +7,7 @@ websites:
     doc: https://support.aerofs.com/hc/en-us/articles/202610424-How-Do-I-Set-Up-Two-Factor-Authentication-
 
   - name: Apple iCloud
-    url: https://www.icloud.com
+    url: https://www.icloud.com/
     img: icloud.png
     tfa: Yes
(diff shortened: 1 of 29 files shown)


fpigerre:
I'm a little concerned that addition of URL normalisation might make it harder for new contributors to add new sites to the repository. Contributors would have to create their pull request before seeing the TravisCI warning, then figure out why the build was failing, before finally creating a second commit that would normalise the URL.

stephengroat:
@psgs i thought about that too.

it seems like contributors who understand why the build is failing try to address it, while others don't 100% understand and don't seem to pull back at all

mxxcon:
Isn't this is a bit pedantic? 😉 
What does trailing / gives us? What kind of errors/mistakes would enforcing this rule address?

stephengroat:
to me it's similar to correcting the capitalization of twitter usernames. creates verifiable consistency

mxxcon:
If twitter names are case insensitive then to me we are being overly pedantic with that too :)

Carlgo11:
Nah I think this is an overkill, at least for now.

PR #2627, opened 2017-07-11
Add GoDaddy Workspace E-mail

ghost:
GoDaddy Workspace E-mail products are NOT protected by 2FA (different from the GoDaddy account administration which is protected by 2FA - currently listed under the Domain category). Please add this as well so it's easy to tell that the end user's Webmail is NOT protected.


mxxcon (changes requested):
Please correct your pull request since it includes unrelated changes.

PR #2628, opened 2017-07-12
Create Legal section & populate w/ 40 companies

danrubins:
Legal information is extremely sensitive and all of the big vendors, plus many of the smaller companies in this industry just have ignored 2FA. I was going to just add Legal Robot to the list since we just added U2F on top of TOTP, but I discovered there wasn't even a Legal section yet!

There were a few Legal companies scattered about in the Other section and in the Identity section, so I moved those into Legal as well.


mxxcon:
@dhrubins you obviously put a lot of time and effort in aggregating that list. It would've been best if each site was submitted as a separate PR. That way issues with any one won't prevent others from getting merged. 😅

_data/legal.yml
```diff
@@ -0,0 +1,278 @@
...
+      facebook: BarracudaSignNow
+
+    - name: SignRequest
+      url: https://signrequest.com
+      img: signrequest.png
+      tfa: No
+      twitter: SignRequest
+      facebook: SignRequest
+
+    - name: Spring CM
```
stephengroat, line 247: Spring CM seems to offer solutions other than simply legal.

anyone else think may not be the best category?

danrubins:
@mxxcon Oh, no worries. I'll resubmit as separate PRs.

_data/legal.yml
danrubins: ... probably a good reason why I should have submitted each site separately. I'll reference this comment on the PR for moving Spring CM over to Legal so it doesn't get lost.

PR #2630, opened 2017-07-12
Remove version hold on ghpages gem

stephengroat:
It's already bumped another version from #2616, I think we can trust that github will be using the latest version


fpigerre:
Because we don't have any other Gems that are used in production, we can now let GitHub sort out breaking changes, meaning we don't have to freeze versions ourselves 👍

fpigerre (approved):

PR #2633, opened 2017-07-13
Add legal section

danrubins:
This is an empty section, for now. Will add new companies and move some from other sections via other PR's.


stephengroat:
favoring #2634 since it creates the legal section as well

jamcat22:
This PR supersedes #2628.

PR #2636, opened 2017-07-14
update backblaze

stephengroat:
move to support article that should update with totp support

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -21,7 +21,8 @@ websites:
     img: backblaze.png
     tfa: Yes
     sms: Yes
-    doc: https://www.backblaze.com/blog/two-factor-verification-for-backblaze/
+    software: Yes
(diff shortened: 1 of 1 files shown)


_data/backup.yml
```diff
@@ -21,7 +21,8 @@ websites:
     img: backblaze.png
     tfa: Yes
     sms: Yes
-    doc: https://www.backblaze.com/blog/two-factor-verification-for-backblaze/
+    software: Yes
```
mxxcon, line 24: Why `software:yes`? It looks like it's still only sms.

stephengroat: sorry, forgot to copy and paste this link in the description

https://www.backblaze.com/blog/two-factor-verification-via-totp/

mxxcon (approved):

jamcat22 (approved):

PR #2666, opened 2017-07-27
Add ING-DiBa Austria

tgamauf:


stephengroat (approved):
thanks for doing the split and the pr @eega!

Carlgo11 (changes requested):
Here's a more compressed logo. Please use that one instead.
ingdibaaustria

tgamauf:
Sure, replaced the icon with the compressed version you provided.

Carlgo11 (approved):

PR #2672, opened 2017-07-28
Add Mailgun

alexcarous:
Mailgun now supports software 2FA

### DIFF
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -108,12 +108,9 @@ websites:
 
     - name: Mailgun
       url: https://www.mailgun.com
-      twitter: Mail_Gun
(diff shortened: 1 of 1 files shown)


stephengroat (approved):
thanks for the pr @alexcarous 

i just cleaned up the social media links (they are removed when `tfa: Yes`) and removed the doc link that was behind an account-wall (public links are preferred)

Carlgo11:
I'd prefer some sort of doc link or at least some proof that they support 2FA.

stephengroat:
i've confirmed that they now support `software` tfa

Carlgo11 (approved):

PR #2682, opened 2017-08-17
Add StartMail

jap:


Carlgo11:
Hello Jasper,
the logo you chose won't go well on our site. Letters and numbers tend to look distorted when shrunken down to 32x32 pixels. Maybe the favicon of StartMail would be better?
As a rule of thumb, the fewer characters in the image the better it looks on our site.

//Carl

jap:
Hi Carl, 
Thanks for the review. I'll check if @StartMail 's marketing team can come up with a better logo - not sure if there is an icon-like one, but maybe the font-size can be increased by repositioning the letters. Will keep you posted!

Jasper

Carlgo11 (approved):
Looks like they don't have any other logo. Guess we'll have to go with this one

mxxcon (approved):

PR #2697, opened 2017-09-03
KeyBank

dphiffer:
KeyBank does have a phone/SMS-based one time password, but an attacker can answer a security question instead. If they add an option to require a second factor I will update the status here. I will update and praise them profusely if they implement hardware token 2FA.

See also: https://twitter.com/KeyBank_Help/status/904367105071566848


stephengroat (approved):
thanks for the pr and the details @dphiffer

Carlgo11 (approved):

PR #2699, opened 2017-09-06
Update entry for Sparkasse.de

wnklmnn:
I'm not sure if this qualifies for an updated entry since the TFA only affects the final confirmation for  transactions (sending money, etc).
For the login there is no TFA available.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -728,8 +728,14 @@ websites:
       facebook: sparkasse
       email_address: sparkasse@sparkassen-finanzportal.de
       img: sparkasse.png
-      tfa: No
+      tfa: Yes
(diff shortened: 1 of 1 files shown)


Jawshy:
Hi, @wnklmnn. Thanks for your interest in @2factorauth. 

What you described is what we would consider to be two factor authorisation because a token (second factor) is needed to _authorise_ a sensitive action (e.g. a transaction). 

A token required on login would be considered two factor authentication. Therefore Sparkasse should not be listed as supporting 2FA. Further reading: [A Note on Definitions](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions).

PR #2711, opened 2017-09-10
Adding Nicor gas

wisco24:


mxxcon:
A mistake you made here is that your `ComEd` addition was made into your _master_ branch. So now every new branch which is based on master will have ComEd.
You have 2 options here, either make branches from master commit _prior_ to your addition of ComEd, or a better approach (but a bit more work) is to start from scratch and create separate branches for each site you'd like to add.

jamcat22:
Duplicate of #2712

PR #2714, opened 2017-09-10
Adding Cellcom

wisco24:


mxxcon (commented):
Is this site significant enough? 🤔alexa lists it at almost 400k and if I remember correctly we might try to put a cut-off at 200k+

stephengroat:
i'd say no

wisco24:
Where is that documented?  What are you using as a source to make that determination?

stephengroat:
@wisco24 it's not documented (something that we're working on), but we use www.domaintally.com (and the alexa ranking) to make the determination)

wisco24:
Company to small for now to add.

PR #2715, opened 2017-09-10
Removed e-mail that was causing Travis-CI to fail.

wisco24:


mxxcon:
I think this might be a wrong way to fix this? 🤔
I think we might have wrong logic in yaml validation...

wisco24:
I think he intended to provide the e-mail to allow people to e-mail the company to request TFA be added like they can with Twitter/Facebook, but the current layout, uses it if a company does have e-mail TFA.

wisco24:
found the issue.  Will resubmit the PR in a second.

stephengroat (approved):
thanks for the help

PR #2716, opened 2017-09-11
Add Telzio

petefox:


mxxcon (commented):
Perhaps `communication` is a better category for this?
Also if tfa:yes, you don't need twitter and facebook attributes.

stephengroat:
@PeteFox please also remember to alphabetize the `.yml` file that you add to by the `name` tag

petefox:
@mxxcon I did look at Communications, but it seems like all other VoIP and telecom providers are placed in Utilities.

stephengroat:
i agree with @PeteFox. seems like that's where we've listed `VoIP.ms` and some other VoIP providers.

maybe we should revisit in the future? for now, I'm approving

stephengroat (approved):

mxxcon (approved):

PR #2727, opened 2017-09-13
Losslessly recompressed all png files.

mxxcon:
1011876 bytes compressed down to 930454 bytes


Jawshy (approved):
Looks good to me.

jamcat22 (approved):
:+1:

PR #2731, opened 2017-09-15
Adds contabo.de

mxxcon:


mxxcon:
@2factorauth/collaborators bump?

Jawshy (approved):
Their Twitter account isn't terribly active, but other than that this looks good (as far as a non-German speaker can tell.)

jamcat22 (approved):

PR #2739, opened 2017-09-20
Reddit is beta-testing 2FA

haykam821:
Reddit has started enabling a software-generated token. Because of this, I made this pull request but made the exceptions state it's only for beta testers.

image

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -183,7 +183,6 @@ websites:
 
     - name: Reddit
       url: https://www.reddit.com/
-      twitter: reddit
       img: reddit.png
       tfa: Yes
       software: Yes
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -185,7 +185,9 @@ websites:
       url: https://www.reddit.com/
       twitter: reddit
       img: reddit.png
-      tfa: No
+      tfa: Yes
+      exceptions:
+          text: "Only for opt-in beta testers."
 
     - name: Snapchat
       url: https://www.snapchat.com/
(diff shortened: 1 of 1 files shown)


_data/social.yml
```diff
@@ -185,7 +185,10 @@ websites:
       url: https://www.reddit.com/
       twitter: reddit
```
mxxcon: Remote `twitter` if `tfa:yes`

mxxcon (dismissed):

jamcat22:
Additional info seems to be available here: https://www.reddit.com/r/modnews/comments/6x1uh7/twofactor_authentication_beta_for_moderators/

stephengroat:
since this is for moderators only and the such a small subset of their users, i'm not sure this one would be ready to post.

mxxcon:
If you join beta thing on reddit and go to their beta subreddit, you'll see some posts about tfa too..

jamcat22:
Maybe we should use a status label?

RichJeanes:
I think it's fair to add this since it is a pretty big deal for moderators, more so than your every day user. A user gets their account hacked, they might lose some fake internet points. A mod gets their account hacked and a whole community (potentially) gets nuked. But, yes, put an exception text on it that it is beta and mod-only at this time.

Carlgo11 (changes requested):
As per [331440306](https://github.com/2factorauth/twofactorauth/pull/2739#issuecomment-331440306) and [331475143](https://github.com/2factorauth/twofactorauth/pull/2739#issuecomment-331475143), add a note that it's only applied to moderators or set tfa to no.

Carlgo11:
Closing due to inactivity.

PR #2745, opened 2017-09-22
Add PosteItaliane

Nicos18:


mxxcon (changes requested):
Please add only 1 site per pull request.
And please add 32x32 logo file

stephengroat (dismissed):
thanks for the pr @Nicos18

mxxcon (dismissed):
Semantically data is correct, however since I don't read Italian I can't confirm if content is accurate.

mxxcon (commented):
Looking at their twitter, this looks like a postal site? If so then I don't think `banking` is the correct category.  We have _Australia Post_ in `other` category; perhaps this entry should be moved there as well?

Nicos18:
They also offers financial services under the name "BancoPosta" (but the company is the same) as stated in the official page of the company "http://www.posteitaliane.it/en/company/business-areas/index.shtml" (in English)

If this can bring a confusion, I could change the file leaving the "PosteItaliane" referring, and adding "BancoPosta". Or I can change this entry to "other", and create a new line in banking with the "BancoPosta" name. Please let me know @mxxcon

mxxcon:
Does BancoPosta uses the same login system as PosteItaliane?
If they are different login systems, then I think BancoPosta should be listed. If they are the same and BancoPosta is just a service by PostelItaliane, then I think it should be moved to `other` since banking is not the primary purpose of that business.

Nicos18:
@mxxcon yes, the login is the same. There is a private area where, after you sign in, you can choose which service do you need; but, also, there are some specific websites to access a specific service, but you have to use the same user and password of the main site

Carlgo11 (approved):

Carlgo11:
Sorry for the delay @Nicos18.
I changed the Twitter handle so that it matches their capitalization.

Btw, since you know Italian, could you possibly add an Italian translation to our default Twitter messages on https://github.com/2factorauth/twofactorauth/blob/master/_data/languages.yml?

Nicos18:
@Carlgo11 I added an Italian entry in the languages file.

Carlgo11:
@Nicos18 Nice! Could you put it in a pull request please? (a separate pull request from this one that is)

Nicos18:
@Carlgo11 done!

brycx (approved):

PR #2753, opened 2017-09-25
Add Askfm

Nicos18:


stephengroat:
@Nicos18 we tend to only count `tfa: yes` when the site protects it's login with it, not when oauth/third party provided 2fa is the only protection

Nicos18:
@stephengroat I didn't know. I will update the file according to this

_data/social.yml
```diff
@@ -5,6 +5,13 @@ websites:
       img: aboutme.png
       tfa: No
 
+    - name: Askfm
```
Carlgo11: Change the name to `ASKfm`

Carlgo11 (changes requested):

Carlgo11 (changes requested):
Please use this logo instead.
askfm

Carlgo11 (approved):

mxxcon:
Why do you think this site qualifies to be listed?

Nicos18:
I think that it should be listed because ASKfm is a site used especially by bullies (there are many cases about this site), and requesting a 2FA method, many bullies could be discouraged to do this.

mxxcon:
If this is that kind of site, I'm not sure it is even worth getting publicity by being listed here. 🤔

Nicos18:
This is the Wikipedia's page about ASKfm: https://en.wikipedia.org/wiki/ASKfm

If you think that it shouldn't be listed on the site, close this pull request. No problem!

Carlgo11:
@mxxcon ASKfm was popular a couple of years ago with pre-teens and teens (say around 10-13yo).

I’ve never used it myself but I’ve seen some people use it and it’s mostly just anonymous questions to others. ‘Do you think x is cute?’ ‘Did you see what y did today in class?’ and things like that.

By the looks of it most of their accounts seem to be inactive but it was a pretty big social network at that time. That means they still have a lot of data and so it would probably be worth protecting.

As I’ve said earlier, I don’t have any first hand experience with the site but I would assume there would be some bullying happening on such sites. If the target audience is of that age it’s almost impossible not to have that. Especially when people can post anonymously.

As with all social media the negative posts are in a minority but of course that’s the posts that get attention.

Nicos18:
@Carlgo11 I saw myself that, especially for profile with more than 10K answers, in many cases there are very sensitive data, such as the school that the user attends, and so on

mxxcon (approved):

PR #2756, opened 2017-09-25
Add U2F (hardware) to gandi.net

kalou:
As of today Gandi's new website supports U2F authentication

### DIFF
--- a/_data/domains.yml
+++ b/_data/domains.yml
@@ -193,6 +193,7 @@ websites:
       img: gandi.png
       tfa: Yes
       software: Yes
+      hardware: Yes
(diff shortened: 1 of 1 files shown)


mxxcon (commented):
Is there any documentation on this? the doc page still only lists totp method.

kalou:
Hi, thanks for looking into it

No update to the doc yet -
The feature is available on https://account.gandi.net in the security tab

Also we don't plan on making it work for the old website, I don't know if that's a thing worth mentioning ?

Carlgo11 (approved):

stephengroat (approved):

PR #2759, opened 2017-09-25
Add Stack Overflow

Nicos18:
I chose other because I think that it doesn't fit in any other category


stephengroat:
covered by Stack Exchange entry

PR #2765, opened 2017-09-26
Add Zooplus

Nicos18:
I added the twitter and Facebook of German Zooplus, because they don't have a unique account, so I inserted the DE accounts, because the central offices are in Germany


stephengroat:
replacement image

favicon

stephengroat (approved):

mxxcon (approved):

PR #2768, opened 2017-09-26
Add ePRICE

Nicos18:


stephengroat (approved):
i agree with you on the logo too, they even make their favicon illegible with text

mxxcon (approved):

PR #2777, opened 2017-09-26
Add PAYEER

Nicos18:


stephengroat:
i cannot check without moving my ip outside of the us

mxxcon (changes requested):
This logo doesn't have that white line under P.
payeer

mxxcon (commented):
How did you find that sms cost $0.05? I tried browsing their site, but it's very sparse.
Also what do you mean by the word "personalizable"?

Nicos18:
Unfortunately, these info are only reported after logging in. In the screenshots, there is the $0.05 cost for SMS, and for personalizable I intend the fact that you choose if you receive a code or not, as shown in the screenshots

payeer_2fa1
payeer_2fa2
payeer_2fa3

mxxcon:
If it uses `Telegram` to deliver 2fa codes then I think it should also be `software: yes`

mxxcon (approved):

jamcat22 (approved):

PR #2781, opened 2017-09-27
Add Zalando

Nicos18:


Carlgo11 (dismissed):
:+1:

mxxcon:
Do all these Zalando sites have separate login systems or are they all connected?
If they are all connected then I think it's better to use a single global entry such as https://zalando.com/

Nicos18:
I am registered on the Italian site, and I tried to login the UK site. The login system is the same, but except this, on the UK site, for example, I didn't find my Italian address and some other data that exists on the IT site.

mxxcon:
So if the login system is shared, I think it's better to have a single entry rather than duplicating all that info.
Their `zalando` twitter and facebook accounts are in German..
@2factorauth/core feedback?

jamcat22 (dismissed):
I agree with @mxxcon here. It's similar to an Amazon.com situation.

Carlgo11:
@jamcat22 @mxxcon I merged it into one listing and made the url zalando.com. That url redirects the user to their own country's version of Zalando.
I also changed the Twitter handle to Their Technology dep. They're not very active there but have responded to a couple of user tweets. Hopefully by sending directly to them, it's easier for the change to be made.

brycx (approved):
This looks like the best solution to me. Even though the Facebook account and email address is on .co.uk domain, the facebook.com/zalando is also in german. I'd say keep it like this.

mxxcon (approved):

PR #2787, opened 2017-09-28
fix line facebook

stephengroat:


Jawshy (approved):
Should have been from your fork, but 👍.

mxxcon (approved):

PR #2794, opened 2017-09-30
Add TIMvision

Nicos18:


mxxcon:
Does this also use yet another different login system from tim and timmail?

Nicos18:
Partially. For who is a TIM customer that has a landline, it uses the "timmail" login; otherwise, it uses a different and separate login system

brycx (approved):

PR #2799, opened 2017-10-08
Add Indiegogo

Jawshy:
As far as I can tell the crowd funding site [Indiegogo](https://www.indiegogo.com/) does not offer two-factor authentication.


mxxcon (approved):

_data/other.yml
```diff
@@ -252,6 +252,13 @@ websites:
       twitter: indeed
       facebook: Indeed
 
+    - name: Indiegogo
+      url: https://www.indiegogo.com/
+      twitter: Indiegogo
```
Carlgo11: Doesn't look like this account is monitored. What do you think about using @IndiegogoUK instead?

Carlgo11 (approved):

PR #2800, opened 2017-10-09
Add GoFundMe

Jawshy:
[GoFundMe](https://www.gofundme.com/) is a social fundraising platform. 

GoFundMe appears to offer two-factor authentication via SMS/text message and phone/voice call, [but only to US and Canadian phone numbers](https://support.gofundme.com/hc/en-us/articles/203778544-Do-I-have-to-be-in-the-US-to-verify-my-account-by-phone-). 

[This help article](https://support.gofundme.com/hc/en-us/articles/204138130-Why-do-I-have-to-verify-my-phone-number-before-I-withdraw-) says that verification is required before withdrawing, which is distinct from [twofactorauth's definition](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions).


mxxcon (approved):

mxxcon:
Should it go into `finance` or `payments` categories perhaps?

jellemdekker:
If the category is changed, I feel Kickstarter should be moved too.

Jawshy:
I don't understand how GoFundMe would fit under finance or payments.

jamcat22 (approved):
Because IndieGoGo was merged into the Other category, along with Kickstarter being in Other, I’m going to merge this PR in its current state. A new PR moving the three services into another category can be made later if applicable.

PR #2804, opened 2017-10-09
Reddit 2FA is in beta

kailan:


RichJeanes:
Duplicate of #2739, but that user has not updated their PR to meet our requests. If you would be able to do so, that would be fantastic.

- Set `tfa: Yes`
- Add `software: Yes`
- Change your `status` link to a `doc` link
- Add an exception stating that it is currently in beta and only for Moderators.

jamcat22:
Duplicate of #2866

PR #2807, opened 2017-10-10
Add RimuHosting

stephenreay:


mxxcon:
@stephenreay Thanks for this PR. But looking at https://www.alexa.com/siteinfo/rimuhosting.com it doesn't seem like this is a popular enough site. We can't list every single out there, so we tend to cut off at 200k+ popularity.
However on a plus side, this site already supports 2fa, so there's nothing else for them to do and their users are already protected. 👍

stephenreay:
Honestly, I don't see why its a problem to "have every single one" - particularly when you're getting the data via PR's. This list is a valuable resource, I'd have thought more information is always better.

If you're adamant about only having "popular" sites on the list, maybe you need to codify that: the text of the site *encourages* people to "add your favourite site via pull request", and checking a couple of others from the hosting section, one was at 600K, one was at 1.1M on Alexa.

brycx (commented):
Thank you for the PR @stephenreay. Due to a recent limitation where a site needs to be within the Alexa top 200k, I have to close this PR since RimuHosting [is not](https://www.alexa.com/siteinfo/rimuhosting.com).

stephenreay:
You’re faced with two choices:

- click a “merge pr” button, and accept the changes you *specifically ask users to submit*.

- invent an arbitrary rule about a sites *estimated* traffic ranking, and then check each and every site against that rule to determine if the site should be included, because everyone knows only the most popular sites need two factor auth.


Seriously, you can all go eat a bag of rotten dicks. Enjoy your fiefdom of bullshit.

PR #2808, opened 2017-10-12
Add DigiD.nl

jellemdekker:
Added [DigiD.nl](https://www.digid.nl/over-digid/), the [universal sign-in platform for government websites in the Netherlands](https://www.rijksoverheid.nl/onderwerpen/digitale-overheid/vraag-en-antwoord/wat-is-digid).

Using two-factor authentication is optional. However, government websites can now decide individually whether or not the use of [2FA should be required during log-in](https://www.digid.nl/nl/vraag-en-antwoord/waarom-moet-ik-mijn-digid-uitbreiden-met-sms-controle/). Therefore, most DigiD users should enable it or otherwise they will lose access to those government websites.

2FA is supported through SMS, [receiving spoken messages through calls](https://www.digid.nl/nl/vraag-en-antwoord/ik-heb-geen-mobiele-telefoon-kan-ik-toch-van-de-sms-functie-gebruik-maken/), and by using the [DigiD app](https://www.digid.nl/over-digid/app) which scans a login request QR-code and lets the user approve or deny it on their mobile phone.

The websites that are linked are in Dutch, please use Google Translate to view them in English.


Jawshy (commented):
Thanks for submitting a pull request, @jellemdekker. 

There was one error which I fixed for you.

jellemdekker:
Thank you, @Jawshy.

mxxcon (approved):

jellemdekker:
@mxxcon Can this be merged?

mxxcon:
Need 2nd review from @2factorauth/collaborators

_data/government.yml
```diff
@@ -13,6 +13,17 @@ websites:
...
+    - name: DigiD
+      url: https://www.digid.nl/
+      img: digid.png
+      tfa: Yes
+      sms: Yes
+      software: Yes
+      phone: Yes
+      doc: https://www.digid.nl/nl/vraag-en-antwoord/hoe-kan-ik-de-sms-controle-toevoegen/
+      exceptions:
+        text: "Software implementation requires use of the DigiD app. It does not support multiple accounts on the same device. To enable the app, the user must already have the SMS option configured."
```
jamcat22: Does the app require SMS, or just a linked phone number? For example, could I register a landline and use phone call verification for the app? If it does require SMS, I think we could simplify this to: `SMS required for software 2FA. 1 account per device.` If it only requires a valid phone number, we could simplify to: `SMS or phone call required for 2FA. 1 account per device.`

jellemdekker: Activating the app (2FA through software, that is) requires you to enter a code once that you receive through SMS or a phone call. So is rephrasing it to the following better or is it still too long: `Software 2FA requires DigiD app. 1 account per device. SMS or phone call required once for activating app.`?

jamcat22: Does the phone number used to activate the app need to remain on the account after the app is activated?

jellemdekker: Yes, that's required. After activating the app you can then choose to log in using the app (by scanning a QR code that's displayed and approving the login request) or by logging in using username, password and a code (from SMS or phone call).

jamcat22: In that case, use `Software 2FA requires DigiD app and SMS or phone call. 1 account per device.` since the phone number must remain on the account after activating the app.

jellemdekker: I have changed the exception text.

jamcat22 (approved):
Thanks @jellemdekker!

PR #2816, opened 2017-10-21
Add Paytm

jobinlawrance:


Jawshy (approved):
Thanks for the pull request, @jobinlawrance. I changed the image because I think that this new one will look better on the site.

mxxcon (approved):

PR #2817, opened 2017-10-21
Add RedBus

jobinlawrance:


Jawshy:
Hi, @jobinlawrance. Are you sure that redBus does not support [two-factor authentication](https://github.com/2factorauth/twofactorauth/blob/dd6c07185e3b834ba5c896f772781ccd4380e005/CONTRIBUTING.md#a-note-on-definitions)? 

I ask because I see [OTP](https://en.wikipedia.org/wiki/One-time_password) being mentioned when I search `two-step verification site:www.redbus.in`. Here's one example: [Offer Terms](https://m.redbus.in/offers/581).

jobinlawrance:
Yeah they use OTP only during the first time registration for verification just like how most websites use email to verify, not as an extra authentication during login or while booking buses.

Jawshy (approved):
Thanks for another pull request, @jobinlawrance. I made some changes as you can see.

mxxcon (commented):
Does the site store your payment information?
If not, doesn't sound like too much of a sensitive site..

Carlgo11 (approved):

jobinlawrance:
Yes they do.  @mxxcon 
> If you transact with us, we collect some additional information, such as a billing address, a credit / debit card number and a credit / debit card expiration date and/ or other payment instrument details.  

> With your explicit approval and to ease the payment process we may from time to time prompt you to allow us to save your credit/debit card information. With the help of Express check out, you can pay by credit/debit card using just one click. This will expedite the payment process and will only be done with your explicit approval. The details are stored on encrypted servers and you have our word that they will be safeguarded to the best of our ability.  

[redBus Privacy Policy](https://www.redbus.in/mob/PrivacyPolicy.aspx)

PR #2798, opened 2017-10-06
Add United Domains USA

seitsu:
Add United Domains USA. Please keep the USA in the title, because there is also a German website (https://www.united-domains.de).

PR #2546, opened 2017-05-19
Added united-domians Reselling

seitsu:
* Added udreselling.png to img/domains/
* Added united-domains Reselling to _data/domains.yml

PR #2824, opened 2017-10-25
Add united-domains

seitsu:
This is different to the US and the reselling website, which are both already included.


mxxcon:
Do they use different login systems?
Do they have different logos? Because right now you are using the same logo for both entries.

seitsu:
@mxxcon Yes, it is different to #2546 and #2798. All of them are using different login systems as they are different companies. The logo of #2546 (udreselling.png) seems similar to #2824 (united-domains.png), but it is different.

seitsu:
@mxxcon Any idea how to proceed with this request?

mxxcon:
I guess perhaps rename it to be `united-domains DE` or maybe even write it out as a domain `united-domains.de`? to clearer differentiate it from the other entries.

seitsu:
@mxxcon Ok, I renamed it to united-domains.de.

seitsu:
@mxxcon / @Jawshy  Can we proceed with that?

mxxcon (approved):

Jawshy (approved):
The united-domains.de FAQ article said that a phone number is required so I added that as an exception.

Jawshy:
Thanks for the pull request, @seitsu.

PR #2825, opened 2017-10-25
Add Breeze

t-richards:
The project management tool. https://www.breeze.pm 

From an email conversation with their support earlier this year,

>We don't have two-factor authentication support at the moment but I added this to our todo list.


Carlgo11 (approved):
While we usually don't add small sites, they seem to respond regularly on Twitter so maybe adding this would make them prioritize security. 🤔

jamcat22 (approved):

PR #2826, opened 2017-10-27
Add SmartSimple

loracsilva:


seitsu (changes requested):
I think, you need to keep the alphabetical order of the entries, the image should be uploaded as well, the indent for 'name' is to much, and the key 'tfa' is missing. Is there also a documentation for 2FA at smartsimple available?

stephengroat:
since this seems stale, closing in favor of #2908

PR #2837, opened 2017-10-30
Update Wealthsimple 2FA details

LeeBrotherston:
Wealthsimple will shortly be launching 2FA, supporting both TOTP software and SMS as authentication methods.

I wanted to submit a PR now so that we can keep you up to date with the change.

### DIFF
--- a/_data/investing.yml
+++ b/_data/investing.yml
@@ -202,7 +202,9 @@ websites:
     - name: Wealthsimple
       url: https://www.wealthsimple.com
       img: wealthsimple.png
-      tfa: No
+      tfa: Yes
+      sms: Yes
+      software: Yes
       twitter: Wealthsimple
       facebook: wealthsimple
 
--- a/_data/investing.yml
+++ b/_data/investing.yml
@@ -205,8 +205,6 @@ websites:
       tfa: Yes
       sms: Yes
       software: Yes
-      twitter: Wealthsimple
-      facebook: wealthsimple
 
     - name: WiseBanyan
       url: https://www.wisebanyan.com/


Jawshy:
Thanks for the pull request, @LeeBrotherston. Does/will Wealthsimple have any documentation on how to enable 2FA?

mxxcon:
Also perhaps it's better to submit/merge this once tfa actually goes live so that people don't go looking for this feature while it's still not available..?

LeeBrotherston:
Hey all, to cover both questions:

@jawshy - There will be - but that's lagging slightly behind the launch.  My plan was to make another PR for that.

@mxxcon - I agree with that too.  To be honest I wasn't sure how long it would take for someone to see the PR and we're launching this week, so I was anticipating that that was what I was actually doing ;)

LeeBrotherston:
Hi - we have now gone into production on both web and mobile platforms, so I think we can officially say that we're live.  So I think it's safe to merge this PR if you feel comfortable with it.

CC: @Jawshy @mxxcon

mxxcon (approved):

Jawshy:
That's good to hear, @LeeBrotherston. Do you know when the documentation for 2FA will go live? If it's soon then I was thinking of having the PR updated before merging, else we could merge now and add the `doc` later as you said.

LeeBrotherston:
@Jawshy I must admit that the (tech) docs may be delayed a little.  We are doing a more marketing/business focussed blog post on Mon 6th Nov.  We haven't scheduled up the tech documentation yet though.

Jawshy (approved):

PR #2842, opened 2017-11-10
Add BreachInsider.com 2FA to Security

graystevens:
No documentation yet, but will arrange a blog post to show the capability. Already implimented.


Carlgo11:
Hello @graystevens,
while we want every site to support two factor authentication we have a finite amount of space to list sites.
Therefore we have to choose which sites to list and which sites not to list.
We choose this based on site ranking on sites like alexa.com and on social media following.

As of right now it looks like Breach Insider is too small to be listed on our site.
I apologize for that we couldn't add your PR this time.

//Carl

PR #2843, opened 2017-11-11
Update EA/Origin name

DandelionSprout:
EA has had 2FA for a while by now, so I figured out that I could just as well make this pull request to add it to the site.


mxxcon:
Perhaps use this logo? It's smaller in size and less jaggy edges
ea

DandelionSprout:
Ah, thanks. I'll get around to adding it.

mxxcon (approved):

Carlgo11:
How is this different to Origin which already is on our list? From what I can tell they both use the same authentication system.

mxxcon:
Oh ya, we already have Origin. No wonder this seemed familiar.

DandelionSprout:
...Somehow I failed to see that Origin was on the list before now. :{

I am not 100% certain about the situation, but it appears that Origin accounts became EA accounts some time in 2015. Whether it was a mere renaming, or if they added distinctly non-Origin features to the accounts, I don't personally know of. If anyone here are playing a lot more EA games online than what I do (who have just played Sims 3 and tried out EA Access), they can give more info on this than what I can.

If it turns out to have been a renaming, I suppose we can replace the Origin entry with the EA entry, or?

mxxcon:
EA account and Origin account are the same thing.
screenshot-2017-11-10 log in

DandelionSprout:
Googling a bit better, I too can see that it's the same thing.

So, I'll leave it up to you what to do next with this. Personally I'd advocate a renaming of the Origin entry and a logo change, but I'm not particularly picky with this now.

mxxcon:
I think perhaps we'll update existing Origin entry to something like "EA Origin".
But I don't think we should have 2 separate entries. Makes updating/maintaining things harder.

DandelionSprout:
I too agree that it'd be dumb to have two entries.

DandelionSprout:
So... shall you do the edits of your choice, in which case this pull can just as well be closed? Or shall I do the edits (with or without the black EA logo)?

jellemdekker:
@DandelionSprout I would use `Electronic Arts (Origin)` as the name and remove origin.png from the repo, replacing it with the black EA logo (perhaps named `electronic_arts.png`, because there may be multiple services using the acronym EA). Would you take the initiative to make this happen?

DandelionSprout:
Sounds like a good idea to me. So I've attempted to do that now.

_data/gaming.yml
```diff
@@ -36,6 +36,15 @@ websites:
       tfa: No
       twitter: curseapp
       facebook: curseapp
+      
+    - name: Electronic Arts (Origin)
+      url: https://www.ea.com/
+      img: ea.png
+      tfa: yes
```
jellemdekker: Line 43 to 46: Please change `yes` to `Yes` (capital Y).

jellemdekker:
@DandelionSprout Thanks for your edits, there's one more thing. The Travis build is failing, because the order of entries should be alphabetical. `Electronic Arts (Origin)` should come after `Elder Scrolls Online`, so please move the entire entry so that the order of the section is `Elder Scrolls Online`, `Electronic Arts (Origin)` and then `Enjin`.

jamcat22 (approved):

PR #2847, opened 2017-11-13
Add Bitcoin Vietnam

spazvt:
* add 'Bitcoin Vietnam' crypto currency exchange from fork
  https://github.com/acceptbitcoincash/acceptbitcoincash

On their main website, under Security they claim "Your connection is
SSL-encrypted and your account secured with Two Factor Authentication."
but I don't see any other publicly available information about their
2FA methods. In the API I don't see 2FA-related interfaces.


kenman345:
bitcoin.vn shut down....may want to close this PR

PR #2852, opened 2017-11-14
Add WEX

spazvt:
btc-e.com was shutdown but 2/3 of bitcoins and the userdatabase was restored and they've reorganized as WEX.nz legally based in Singapore. Therefore re-adding them.


damner (approved):

spazvt:
@Carlgo11 or @mxxcon please review.

 Alexa traffic stats rank wex.nz 17k globally but 2009 in Russia (https://www.alexa.com/siteinfo/wex.nz)

mxxcon (approved):

_data/cryptocurrencies.yml
```diff
@@ -229,6 +229,13 @@ websites:
       software: Yes
       doc: https://support.uphold.com/hc/en-us/articles/206119683-How-does-phone-number-verification-work-
 
+    - name: WEX (BTC-E)
```
Carlgo11, line 232: Does it make sense listing BTC-E in their name?
They are two different companies are they not?

spazvt: I guess it depends on how you look at it ... legally it's a new company that bought the assets of the old company. To users, the UI is exactly the same and their balances transferred over. I've listed the old name in parens so that people mentally link them.

Carlgo11: Alright that make sense.

Carlgo11 (approved):

PR #2862, opened 2017-11-15
Add puroxy.org to list of sites that support BCC/BCH

ex8:
name: Puroxy
url: https://puroxy.org/
img: puroxy.png
bcc: Yes
btc: Yes
othercrypto: Yes


mxxcon:
Looks like a commit to wrong repo.

PR #2875, opened 2017-11-30
Okta now supports Email as a factor

jpf:


Carlgo11 (approved):
Thanks for the PR!
Not sure what to think about adding email as a factor in 2017 but at least you support a bunch of other factors as well 👍

mxxcon (approved):

PR #2878, opened 2017-11-30
Paycor supports software 2fa now

KeenRivals:
They don't mention it on their doc page, but in the user settings there's an option to use Google Authenticator. This displays a regular QR code that works in any TOTP app.

### DIFF
--- a/_data/finance.yml
+++ b/_data/finance.yml
@@ -189,6 +189,7 @@ websites:
       sms: Yes
       email: Yes
       phone: Yes
+      software: Yes
       doc: https://www.paycor.com/security/how-to-keep-your-account-safe
 
     - name: Paytrust


mxxcon (approved):

jamcat22:
Since they don't mention it in their doc page, could you upload a screenshot of the process, making sure to censor any identifying information (QR codes, secrets, emails, etc.)?

KeenRivals:
# MFA Settings Page
MFA is enabled by going to "My Settings" > Security, then ticking the relevant box(es). I'll also note that what prompted me to notice MFA was available is they forced email MFA on me and suggested I set up SMS verification. When I noticed they had Google Authenticator support I opted for that instead.



# Screen when logging in with MFA enabled

jamcat22 (approved):
Thanks @KeenRivals!

Issue #2581, opened 2017-06-21
Fine-grained list needed


ghost:
The list of services supporting is pretty coarse grained between Hardware and Software tokens.

I think it should be more fine-grained and list specific protocols, for hardware this will be FIDO U2F and FIDO UAF and software it will be HOTP and TOTP.

mxxcon:
We specifically switched so such broad and generic terms to simplify and speed up addition of sites.
If we start listing all the specific implementations, the list would grow infinitely long.

jamcat22:
This issue has been addressed in multiple other issues, including #1369, #965, #852, #842, and #493. Ultimately, #333 was the PR which switched the site from focusing on specific methods, like Google Authenticator, Authy, and Verisign VIP, to more broad usage types, like software and hardware. I would suggest looking at the prior issues and PRs relating to this issue to understand the mentality behind the decision to not include specific providers. If you have a suggestion on how to tackle this problem that hasn't already been addressed, you're welcome to talk about it here.

Issue #2067, opened 2016-09-29
Creating an optional popup for software checkmark


stephengroat:
Looking at #2061, #2060, and #2057, it seems like if a specific software app is required, making the software checkmark popup a link to it would be a good idea (similar to the current exception system). It should probably be optional and need to have a list of the links to all of the different OS implementation (Android, iOS, etc.).

I'm envisioning the `_data` file looking like this:

```
  - name: Apple iCloud
    url: https://www.icloud.com
    img: icloud.png
    tfa: Yes
    sms: Yes
    software: Yes
      text: "iOS: www.apple.com"
    exceptions:
        text: "See http://support.apple.com/kb/HT5593 for a list of supported SMS carriers."
    doc: http://support.apple.com/kb/ht5570
```

and the website looking like it current

@psgs, any way you could help with this? I tried, but turned everything into spaghetti. I'm particularly having a hard time understanding what the `data.json` file does

fpigerre:
Sure @stephengroat :smile: 
The `data.json` file allows machines to access data from the twofactorauth.org website using JSON. For example, a program could be written that would pull all the entries from https://twofactorauth.org/data.json and make a decision that would show people which website in a particular category provides the most comprehensive TFA solution.

Carlgo11:
Let's wait with this for some time. First we need to decide if we're going to list things like this at all.

kiwiant:
This would be great especially when someone like Apple forces users to use an iPhone for software tokens rather than standard authenticator software like the rest of the industry.

peterbraden:
Indeed, there should be some way to differentiate between open OTP app usage, and specific app requirements. (Such as authy etc.)

conorgil:
@stephengroat @psgs @Carlgo11 Have you given this any further thought? I think it would be incredibly useful to know which services require some specific app or proprietary solution, etc.

Carlgo11:
I’d be open to the idea. Lots of people have requested a way to set what software services a site allows. But before we create some design for it I think the first step would be to try and create some sort of standard, list it in countributing.md & the wiki, create checks for it and then add it to data.json to start with.

twilsonco:
+1 for specifying between generic software token (e.g. 2FA), and proprietary, e.g. Authy)

Carlgo11:
Fixed with separate TOTP/Proprietary tags in v3.0

PR #2880, opened 2017-12-01
Update Humble Bundle

schlagi123:
Add exception to Humble Bundle implementation

### DIFF
--- a/_data/gaming.yml
+++ b/_data/gaming.yml
@@ -141,6 +141,8 @@ websites:
       sms: Yes
       software: Yes
       doc: https://support.humblebundle.com/hc/en-us/articles/202421374
+      exceptions:
+        text: "Software implementation requires use of the Authy Mobile app."
 
     - name: IndieGala
       url: https://www.indiegala.com/


mxxcon:
Is exception for this really necessary? 🤔

jamcat22:
Please see #2581 for more discussion on the topic of classifying software implementations. As of right now, we don’t mark specific implementations unless there’s a specific limitation associated with it.
Here are a couple examples:
- A banking website offers push notifications for 2FA, but they’re only implemented on their iOS app, not their Android app.
- A social networking website offers Google Authenticator-based 2FA, but they require that you register an SMS-capable phone number first.
- A gaming website offers 2FA through Authy, but they restrict the feature to accounts on a paid plan.

Feel free to comment if there’s a specific limitation to Humble Bundle you’d like to note.

Thanks!

schlagi123:
When I use 2FA the best case is, that I need only one app for token generation and that I am independent of the implementation (I can change the App). Examles for such services are Google, Nextcloud or Github, because I can choose the authenticator app (e. g. Authy, freeOTP, ...). For Humble Bundle I must use Authy. When my preferd authenticator app is another app, then it is the same restriction as for Steam (I need an extra app).

stephengroat:
While it would be nice if everyone used an OATH backed standard like TOTP of HOTP, Authy is a two factor authentication service provider. Their app covers multiple entities, so they get a little exception for that.
Steam is a little different because they use their own Steam app, making it implementation specific

jamcat22:
@schlagi123 thanks for comparing to Steam, as I had not noticed this earlier. Steam’s exception should actually be reworded to not mention the fact that it requires the Steam app, while still explaining that it’s 1 account per phone number. Others, including Blizzard, Garena, Square Enix, and Trion Worlds to name a few, require their own apps to use 2FA but do not specify this in an exception, due to reasons outlined in the aforementioned #2581 issue.

#2067 includes a proposal to include an optional dropdown with the software implementation checkmark to allow for explaining the specific software implementation used without crowding the columns or including an exception message on every service that supports something other than Google Authenticator.

PR #2900, opened 2017-12-18
Add RealVNC

### LINKED ISSUE #2885: RealVNC supports 2Factor authentication
https://www.realvnc.com/en/ has 2FA. :)

stephengroat:
closes #2885


jamcat22 (approved):

_data/remote.yml
```diff
@@ -52,6 +52,13 @@ websites:
       software: Yes
       doc: https://www.syspectr.com/en/faq/how-do-i-activate-deactivate-two-factor-authentication-for-my-account
 
+    - name: RealVNC
+      url: https://www.realvnc.com
+      img: realvnc.png
+      tfa: Yes
+      software: Yes
+      doc: https://www.realvnc.com/en/connect/docs/multi-factor-auth.html#protecting-your-realvnc-account
+
```
Carlgo11, line 61: Maybe a note that it's only available on enterprise and professional subscription?

stephengroat: it also works on the free home subscription for me.

are you seeing different results?

jamcat22: @Carlgo11 I think you may be reading the instructions for the VNC Server application, rather than the RealVNC account system. VNC Server would be considered self-hosted. Similar to TeamViewer accounts having 2FA, but directly connecting to the computer doesn't.

Jawshy (approved):
👍

PR #2906, opened 2017-12-19
Update Voyages to Oui (rebrand)

stephengroat:


mxxcon:
should the image be renamed? it sounds like `sncf` is a tld..?

stephengroat:
sncf is the overall company http://www.sncf.com/ https://en.wikipedia.org/wiki/SNCF, they just bought their own tld

mxxcon (approved):

jamcat22 (approved):

PR #2907, opened 2017-12-19
Update Discover

### LINKED ISSUE #2661: Discover correction (card vs bank)
Discover is listed as currently supporting 2FA, however this is only partially true.  It is supported for Discover _credit card_ customers, but not Discover _bank_ customers.  These two services should probably be separated so the bank's lack of 2FA gets highlighted.

stephengroat:
closes #2661


mxxcon:
Or should this simply be an _exception_ message rather than separate entries?

stephengroat:
since it's different auth features available, it seems like different authentication systems, which usually warrants multiple entries.

changes can be made either way, go at it if you're gung-ho 👨‍💻

mxxcon:
Different auth features, but not different accounts.. 🤔

mxxcon (approved):

Carlgo11 (approved):

PR #2920, opened 2017-12-27
Update natwest

stephengroat:


mxxcon:
Hmm, looking at https://supportcentre.natwest.com/olb/security/cs20/913222972/What-is-a-card-reader-and-how-do-I-use-one.htm makes it sound like it's _authorization_, not authentication device.. 🤔

Flameeyes:
FWIW all the RBS group banks (of which NatWest is one) do use the card reader (implementing the Chip Authentication Program) as an authorization rather than authentication device.

jamcat22:
See #1811.

PR #2928, opened 2018-01-02
Add Tsohost

IanOliver:
Daily do not support 2FA, so I'm adding them as such.


brycx (commented):
It seems Daily has changed to Tsohost. The current url redirects to [Tsohost's site](https://www.tsohost.com/welcome/daily). Daily's Facebook seems to be down but their Twitter still up. Here's Tsohost's [Facebook](https://twitter.com/tsohost)/[Twitter](https://www.facebook.com/tsohost/).

Would you like to add Tsohost instead of Daily?

IanOliver:
That is correct, they have recently been absorbed into Tsohost. Tsohost also do not support 2FA so I would like to add them instead - shall I create them afresh?

brycx:
You can just modify this PR, I'll just change it's name.

IanOliver:
Thank you. The details have now been changed from Daily to Tsohost. Looks like I need to submit a new PR for the image?

brycx:
Can't you just delete daily.png and add another picture with a commit?

IanOliver:
Must admit, I can't see a way to do that from here I'm afraid, sorry.

mxxcon:
Simply commit a new logo with name of `tsohost.png` and delete the old one.

IanOliver:
Believe I've submitted the addition of the Tsohost icon and removal of the Daily icon correctly. Let me know if there's anything else needed, thanks :)

brycx:
@IanOliver For future PR's, you can check what is wrong, by looking at the Travis build check under "Details". If you take a look, it says 
```
1. _data/domains.yml is not alphabetized by name
rake aborted!
```
So the entries must be in alphabetic ordering :)

brycx:
@IanOliver Is there a specific reason you put this into `domains`? As far as I can tell Tsohost also provides hosting. Maybe the `hosting` category would be a better fit?

IanOliver:
@brycx Ah, I see - thanks :)

The original listing for Daily was under domains, since they were more registrars than hosts I believe. But you're correct - Tsohost is probably more correct under hosting.

brycx:
@IanOliver Could you move it to that?

mxxcon (approved):

Carlgo11 (changes requested):
Move to hosting as requested in https://github.com/2factorauth/twofactorauth/pull/2928#issuecomment-403844812

Carlgo11:
Closing due to inactivity.

PR #2930, opened 2018-01-06
Update ebay exception text

Sunit-Kulkarni:
Ebay currently only supports SMS 2FA. There isn't exactly a documentation page for this on ebay so I put the instructions in the exception portion.

Need to go to Home > My eBay > My Account > Personal Information and scroll to the bottom to enable 2FA on eBay.

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -102,10 +102,11 @@ websites:
       url: http://ebay.com
       img: ebay.png
       tfa: Yes
+      sms: Yes
       hardware: Yes
       software: Yes
       exceptions:
-          text: "eBay only supports 2FA in some countries."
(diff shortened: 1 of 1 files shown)


Carlgo11 (approved):
Looks good to me. Nice work on keeping the text short yet informative! 👍

jamcat22 (changes requested):
I think we should remove `hardware` and `software` tags since they're not supported for new users. See #2480 for a similar situation we handled with Patreon.

Jawshy (changes requested):
I agree with @jamcat22 plus the url can be updated to use HTTPS: `https://www.ebay.com/`.

jamcat22:
Duplicate of #2987

PR #2934, opened 2018-01-10
update Commonwealth Bank of Australia and ANZ Australia

hmoffatt:
Updated data for two Australian banks.


_data/banking.yml
```diff
@@ -66,12 +66,12 @@ websites:
...
       tfa: No
 
-    - name: Australia and New Zealand Banking Group
+    - name: Australia and New Zealand Banking Group (ANZ)
       url: https://www.anz.com.au/
-      twitter: ANZ_AU
-      facebook: ANZAustralia
       img: anz.png
-      tfa: No
+      tfa: Yes
```
mxxcon, line 80: I don't think this is correct. on the `doc` URL it states

> Once you've registered for ANZ Shield, you'll be prompted to generate a Shield Code when making internet banking transactions or any changes to your account.

So this is two-factor _authorization_, not authentication..

hmoffatt:
True. The entries for the Commonwealth Bank of Australia and ING Australia are also incorrect then because those banks too use two factor codes to authorize transactions but don't require them to login.

Is this something that can be noted in the listings without removing their two factor status completely?

Sent from my android device.

mxxcon:
As we have it documented in https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions this project is specifically for two-factor _authentication_. So from our point of view such sites with two-factor _authorization_ should be listed as `tfa:no`

hmoffatt:
Fair enough. I think several of the Australian banking entries are wrong 
then. I will submit another PR soon.

Has there been any consideration given to listing two factor 
authorization separately? It would seem to be quite useful to have this 
information for banks that don't have two factor authentication. In the 
case of ANZ Australia the two factor auth is opt-in so listing it might 
increase awareness.

Carlgo11:
@hmoffatt we'd be happy to take a look at any other banks then! 😄 
We don't have any plans on listing sites with two factor authorization on our list. Authorization is a whole other topic and not related to what we're interested in, which is securing peoples' logins.

Echelon9:
@hmoffatt @mxxcon This isn't quite right (or at least has changed in the intervening time).

Commonwealth Bank of Australia ("CBA") does provide 2FA per the definition `2factorauth` adopts [0].
As @hmoffatt noted, the *SMS-based* approach CBA provides is two factor authorization, as it is requested after login, for certain larger transactions.

However, the *token-based* approach [0] provides an additional layer of security at login time that in my view does meet `2factorauth`'s standard for two factor authentication.

>"NetCode Token is a free service we offer to our customers who may not be suited to NetCode SMS or CommBank app NetCode Notifications. A NetCode Token is great for customers who live in an area with poor mobile phone reception or don’t use a mobile phone.
>
> The Token is a small electronic device about the size of a key ring. **You'll need to use it when you log on to NetBank** and ..."

Now, this hardware token is optional for customers of CBA. However, I didn't see on `2factorauth`'s documentation whether optional 2FA meant your website should report the service provider as providing it, but that would seem to be a sensible approach.

[0] https://www.commbank.com.au/support/faqs/72.html

mxxcon:
@Echelon9 the question was about ANZ bank, not CBA... Also it seems like they are related?

edit: But actually regarding CBA, https://www.commbank.com.au/security-privacy/netcode.html states that it's for auth_orization_ and not auth_entication_. Is it a matter of outdated docs? But if so, which one is more accurate?

@jimsug you made the last PR for CBA bank, do you have experience to clarify this?

mxxcon:
Oops. @hmoffatt I'm sorry, I tried to resolve this merge conflict in the other direction. github ui is not clear 😓. I think if you revert my last commit in your repo it should get things back.

Echelon9:
@mxxcon My earlier comment sets this out, but there are two levels offered by CBA:

* *SMS-based approach*: two factor authorization
* *token-based approach* (optional): two factor authentication

You identified a link to the first one; my earlier link is the support for the second one.

ANZ and CBA are not related to each other, they are separate competitors.

Echelon9:
The CBA parts of this pull request have been superseded by what is in master.

Carlgo11:
Merge conflicts and partially merged.

PR #2937, opened 2018-01-10
Update Facebook exception

### LINKED ISSUE #2935: Facebook does not require SMS for 2FA
Hey!

I recently enabled 2FA on a Facebook account without the use of SMS. Required instead was: a 2FA app for the phone and a U2F security key. The warning on the [site](https://twofactorauth.org/#social) says SMS is needed for 2FA on Facebook, maybe instead it should say "Depends on two independent 2FA methods"?

brycx:
Fixes #2935


Carlgo11:
maybe "security key" is a bit vague?  How about "hardware key" or "hardware token"?

brycx:
I wrote "security key" to be consistent with the related [docs](https://www.facebook.com/help/148233965247823) that say _"Keep in mind: You can use as many authentication methods as you'd like, but you need to have at least text message (SMS) codes turned on, or at least both a security key and Code Generator turned on."_ I would change it to "U2F security key" instead then. That could also suggest a YubiKey 4 with U2F configured slot then.

But if "hardware token" or "hardware key" fits better with the general site overall, then I'll change it promptly. I just need to know which one of them.

EDIT: I haven't tested it with anything but U2F.

conorgil:
I think "security key" is a fine term to use, since that is the same terminology used in the Facebook documentation and also by leading vendors (e.g. Yubico, etc). You could use "U2F security key" to be more specific, but those terms are basically interchangeable.

It would be nice to use consistent terminology throughout the entire site, but I realize that is out of scope for discussion in this issue.

Carlgo11 (approved):

brycx:
Fixed in #3217.

PR #2941, opened 2018-01-12
Add Planio

yeah:
This PR adds FOSS project management tool Redmine (where 2FA support is [in progress and scheduled for 4.1.0](https://www.redmine.org/issues/1237)). The 2FA patch was contributed by Redmine Hoster Planio where 2FA is already rolled out, so I'm adding Planio as well in the same commit.

two factor auth list 2018-01-12 17-54-50

Logos were optimized using `optipng`.


stephengroat (changes requested):
thanks for the pr @yeah. could you remove the `redmine` entry? we tend not to list self-hosted projects as a rule of thumb

yeah:
Sure, it's only Planio now. Thanks!

stephengroat (approved):
thanks for the work @yeah look good to me

Carlgo11 (dismissed):
@yeah here's a compressed version of the logo. Please use that instead:
planio

yeah:
I've added the compressed image, thanks. (I've also squashed the commits and rebased on current master.)

yeah:
As said before, I've uploaded the new image, ready to merge!

mxxcon (approved):

PR #2943, opened 2018-01-12
Add Insurance section & add Ottonova

derFunk:
Since ottonova is the third insurance company in the List - I also added an Insurance section and took over Adirondack and Vitality.


_data/sections.yml
```diff
@@ -66,6 +66,10 @@
   title: Identity Management
   icon: user
 
+- id: insurance
+  title: Insurance
+  icon: protect
```
mxxcon, line 71: Is this a lock icon, the same as for Security category?
Is there something else we can use more representative for "insurance? 🤔

derFunk: It's a "shield" icon which I found suitable.
https://semantic-ui.com/elements/icon.html

mxxcon (approved):

_data/sections.yml
Carlgo11: I agree with what @mxxcon said. It could be confused with security.
I also feel that a shield doesn't signal insurance as much as it does protection (such as law enforcement).

What about the `life ring` icon?
To me that signals insurance however that might be a cultural thing.
Appreciate feedback! 😄

_data/insurance.yml
```diff
@@ -0,0 +1,19 @@
+websites:
+    - name: Adirondack Insurance Exchange
+      url: http://www.aie-ny.com/
+      img: adirondackinsurnaceexchange.png
+      tfa: No
+
+    - name: ottonova
```
Carlgo11, line 7: Is there any documentation supporting this?

PR #2945, opened 2018-01-13
Add NASA Federal Credit Union

brycx:
Attempt at fixing: #2684


mxxcon:
Where did you find info that it's email and sms?

brycx:
On the site linked in the [docs](https://www.nasafcu.com/img/newebranch/ebranch_new.html), under "Advanced security features" it says "Options include mobile verification and more." That and the referenced issue. I know it's shaky but I thought since some sites don't have any doc links, I'd give it a try. I totally understand if it's not accepted.

mxxcon:
Just making sure that we list complete and correct info. I googled for a bit but haven't found anything conclusive. 😞
I asked the original issue creator for some help on that info, hopefully they'll reply soon.

brycx:
So I had a little chat with their support. It turns out they offer both SMS, phone and email as 2FA option. Unfortunately only required sometimes and not every time you log in.

Picture of chat

Carlgo11:
I think we need some clarification on what "at random" means. If it's randomly presented at logins or just randomly when using their site. 🤔

brycx:
@Carlgo11 I think by "No, it may not be required with each log in." it's meant that it's only presented at random during logins.

Carlgo11 (dismissed):
Just contacted them and I can confirm that it's for authentication and not authorization. Also, seems to be cookie based.

Carlgo11 (approved):

mxxcon (approved):

PR #2946, opened 2018-01-13
Add Bitwarden exception

brycx:
Bitwarden requires you to have a [premium](https://help.bitwarden.com/article/setup-two-step-login/) account to use hardware 2FA. I think it would be good to have an exception for this.

### DIFF
--- a/_data/identity.yml
+++ b/_data/identity.yml
@@ -25,6 +25,8 @@ websites:
       phone: Yes
       software: Yes
       hardware: Yes
+      exceptions:
+        text: "Premium account required for hardware 2FA."
       doc: https://help.bitwarden.com/article/setup-two-step-login/
 
     - name: Centrify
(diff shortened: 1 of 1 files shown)


mxxcon (approved):

Carlgo11 (changes requested):
To me it looks like it's not just only for hardware 2FA.
I suggest you rephrase it to `"Premium account required for SMS, Phone call and hardware 2FA"` or something similar.
I'm not a native English speaker so the structuring might need some improvements ¯\_(ツ)_/¯

mxxcon:
Carl, sms and phone call are properties of Duo's offering.
From what I can see you can't have them w/o Duo.
Perhaps change the wording further that Duo, Yubikey and FIDO U2F require paid plan?

brycx:
Is this better?

Jawshy (approved):
👍

Carlgo11 (approved):

PR #2949, opened 2018-01-16
Set Capital One tfa: to No

ted-dokos-zz:
As best I can tell, Capital One has backslid on their two-factor. I attempted to follow the instructions on the linked website (describing the process for "SwiftID"), but found that the settings they mention no longer appear in their mobile app (https://play.google.com/store/apps/details?id=com.konylabs.capitalone).

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -166,11 +166,7 @@ websites:
     - name: Capital One
       url: https://www.capitalone.com/
       img: capitalone.png
-      tfa: Yes
-      sms: Yes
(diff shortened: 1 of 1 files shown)


ted-dokos-zz:
I did some due diligence on this by talking to their Twitter team: https://twitter.com/The_Odor_E/status/953352948075114497

It turns out that even though the website documentation is stale (i.e. no SwiftID option), they still activate 2FA if you try to log in from a new device. Given that, I'm going to close this request.

PR #2950, opened 2018-01-17
Update TransferWise documentation, add software token.

Flameeyes:
The mobile app can act as a second factor with push notifications.

### DIFF
--- a/_data/finance.yml
+++ b/_data/finance.yml
@@ -275,7 +275,8 @@ websites:
       tfa: Yes
       sms: Yes
       phone: Yes
-      doc: https://transferwise.com/help/topics/4/topics/19/questions/2597993/how-can-i-setup-2-step-login
+      software: Yes
+      doc: https://transferwise.com/help/article/2811597/borderless-account/how-can-i-setup-2-step-login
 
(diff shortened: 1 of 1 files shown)


dsoegijono (approved):

Carlgo11 (approved):

joeldrapper:
TransferWise doesn’t actually support TOTP, so this PR breaks @1password Watchtower.

mxxcon:
@joeldrapper, `software: yes` doesn't mean exclusively TOTP. It is any kind of software 2fa implementation, such as how Steam mobile client does it or twitter or in this case TransferWise

joeldrapper:
Right, that makes sense. I guess this isn’t the right resource for that data then. 🤔

LloydThinks:
For the sake of services such as 1Password that are using https://twofactorauth.org/ as their source of information, shouldn't there be a separate flag for TOTP?

I came to this issue repo looking for why TransferWise was included on https://twofactorauth.org/. Now I know, but delivering information in a round-about way.

Carlgo11:
@LloydThinks see #3222

PR #2957, opened 2018-01-19
add uberspace.de

paradonym:
adding uberspace.de to category "hosting".
Uberspace is a popular "pay what you want" shared-hosting provider in Germany.

Google Authenticator isn't available for their Website login, just their SSH Login can be secured using OTP-Codes additionally to the password.
Unfortunately their english manual https://manual.uberspace.de/en/ doesn't have a section about 2FA, so I linked the german page under doc: and added the exception text.


Jawshy (changes requested):
There are several errors in the [Travis build log](https://travis-ci.org/2factorauth/twofactorauth/jobs/330703726): 
```
1. Uberspace: 'false': not matched to pattern /true/.
2. Uberspace: 'false': not matched to pattern /true/.
3. 'lang' should NOT be present when tfa: true.
4. img/hosting/uberspace.png is not 32x32 pixels.
5. img/hosting/uberspace.png should not be larger than 2500 bytes. It is currently 253276 bytes.
```

```
1. Uberspace: 'false': not matched to pattern /true/.
2. Uberspace: 'false': not matched to pattern /true/.
```
The above means that `sms: no` and `hardware: no` should only be `Yes` otherwise don't have them at all. 

Let us know if you need help correcting the errors.

paradonym:
commit 66b1bc8  
*  internal image /img/hosting/uberspace.png does not exist (line 44540)
*  internal image /img/hosting/uberspace.png does not exist (line 49114)

I think occurs because of the later commit of the new image

the travis log of commit b99ce29 and commit 82866cd actually looks okay to me

Carlgo11:
Unfortunately uberspace.de is ranked >200K on Alexa (currently above 600K) and as such it doesn't meet our requirements for new site additions (See our [guidelines](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#guidelines)).
Sorry for the inconvenience!

PR #2963, opened 2018-01-23
Move sl.se to https

stephengroat:


Jawshy (approved):
👍

Carlgo11 (approved):
I like how their main page uses an EV cert but their login page only uses a DV cert 😆

PR #2967, opened 2018-01-25
Update Skype

Carlgo11:
Looks like it's not possible to use ones Skype account nowadays without merging with a Microsoft Account (or creating a new one) and thus it makes it unnecessary to keep the (via MS Account) text.

:tay:


jamcat22 (dismissed):

mxxcon:
I think my mom still has skype-only account. I want to try to see how it behaves now. I'll try to get to it later today or tomorrow

jamcat22:
@mxxcon sounds good. I've been through the process. Logging in to https://go.skype.com/myaccount should start the migration. If she already has a Microsoft Account, be sure to enter that email when merging the account so as to not create multiple Microsoft Accounts.

mxxcon:
She doesn't have MS account. That's why i want to see what it'll look like.

fpigerre (dismissed):

fpigerre (approved):

PR #2968, opened 2018-01-26
Nordnet Norge

Avec112:
Method 1) They have support for software token (government standard called bank id)
Method 2) Username and password can be used on its own. There should be at least a software token added to this method

No hardware token


mxxcon:
@Avec112 did you mean to create a pull request or just an issue?

Avec112:
Oh. It was ment as an issue.

-Avec

PR #2987, opened 2018-02-05
Redo ebay

Sunit-Kulkarni:
Ebay only supports SMS for two factor authentication. Go to Personal Information part of ebay account and scroll to the bottom and there will be SMS option.


mxxcon:
Is it now supported globally? Do you links to documenting this?

Sunit-Kulkarni:
There isn't any documentation I can find on ebay per se, but yes it looks like it is global now. When you first enable SMS 2FA, It will ask for country code and seems like all are covered!

Sunit-Kulkarni:
It would be here once you log in: https://signinalert.ebay.com/sgninalrt/twofa/onboard?flow=myebay

_data/retail.yml
```diff
@@ -99,13 +99,13 @@ websites:
...
-      url: http://ebay.com
+      url: https://www.ebay.com
       img: ebay.png
       tfa: Yes
-      hardware: Yes
-      software: Yes
+      sms: Yes
       exceptions:
-          text: "eBay only supports 2FA in some countries."
+          text: "eBay now only supports SMS for 2FA."
```
jamcat22: I'm not sure this exception text is necessary since SMS is the only listed option that will be listed on the site, even if `software` and `hardware` used to be supported.

_data/retail.yml
```diff
@@ -99,13 +99,13 @@ websites:
...
+      doc: https://signinalert.ebay.com/sgninalrt/twofa/onboard?flow=myebay
```
jamcat22: This doc link requires logging in to eBay to view, which we typically don't allow. I'd recommend we remove it unless eBay can give us a public-facing doc link.

jamcat22 (changes requested):

_data/retail.yml
Sunit-Kulkarni: Alrighty. I can just delete the exception all together.

_data/retail.yml
Sunit-Kulkarni: All I can find that is related to this is the ebay Security Center: https://pages.ebay.com/securitycenter/protect_your_information.html#protect_account_info.

But it doesn't suggest 2FA on it even though it is now an option.

Sunit-Kulkarni:
I think it should be good to go now. Anything else needed?

stephengroat:
@Sunit-Kulkarni see https://github.com/2factorauth/twofactorauth/pull/2987#discussion_r166010727 probably should still remove the doc

Sunit-Kulkarni:
Alrighty, I just removed it. How about now?

stephengroat (approved):
Thanks for the pr @Sunit-Kulkarni

jamcat22 (approved):

Sunit-Kulkarni:
How long does it take for the changes to be reflected on the actual website? Currently ebay hasn't been changed.

Carlgo11:
Even though your PR has been merged, it might take some time before it shows up on the website. For more information, check out our [wiki](https://github.com/2factorauth/twofactorauth/wiki/FAQ-Merge).

PR #2989, opened 2018-02-07
Fix Smartbox

stephengroat:


Carlgo11:
Are you sure it's only available for enterprise customers?
I just asked their support and they said otherwise. (see screenshot below)

smartbox-screenshot

stephengroat:
i saw this in their documentation:

screen shot 2018-02-07 at 9 08 09 am

then i signed up and only got this in the user settings/account profile:

screen shot 2018-02-07 at 9 12 12 am

have you tried signing up and gotten different results?

Carlgo11 (approved):
Alright, I emailed their support again and they said the same thing as you.

jamcat22 (approved):

PR #2992, opened 2018-02-07
Charles schwab: update to new 2FA flow

nhfruchter:
Charles Schwab is changing its 2FA flow as of Februrary 2018. This PR reflects this.

Their help documentation doesn't reflect this change yet, [but here are some screenshots showing the new login flow live](https://imgur.com/a/1SaWP).

### DIFF
--- a/_data/investing.yml
+++ b/_data/investing.yml
@@ -26,6 +26,9 @@ websites:
       tfa: Yes
       software: Yes
       hardware: Yes
+      phone: Yes
+      email: Yes
+      sms: Yes
       doc: /notes/schwab/
 
     - name: COL Financial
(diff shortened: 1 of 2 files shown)


mxxcon:
Perhaps contact them to find if they'll be keeping software and hardware methods?

Sunit-Kulkarni:
I just contacted them. They are keeping the Symantec VIP app and the hardware tokens free of charge. They also told me they are rolling out text, email, and phone call as options and could activate those as well. Just need to call them and they will take care of it.

Carlgo11:
Thanks for that @Sunit-Kulkarni 👍

PR #2998, opened 2018-02-11
Add Wasabi Hot Storage

Sunit-Kulkarni:
Added Wasabi Hot Storage to backup section. Its a cheaper Amazon S3 alternative. Highly recommend using this in conjunction with Arq software.


mxxcon (changes requested):
Please update logo to be 32x32 and under 2500 bytes in size. (just the green circle is a good option)

Carlgo11:
Looks like it's not very popular. https://www.alexa.com/siteinfo/wasabi.com

To keep our list to a somewhat manageable size we try to limit site listings to rankings 200k or less.

Sunit-Kulkarni:
I'm a noob at the picture part. What do you suggest to downsample it to 32 x 32? I tried and it came out super blurry or does that not matter?

Hopefully with this site as awareness, Wasabi will become more popular haha! Its a fantastic competitor to Amazon S3.

mxxcon:
Image quality is not the primary purpose for this site to exist, but since good quality images exist they should be used. You could've used https://wasabi.com/wp-content/uploads/2017/04/fav.png and resized it down to 32x32.

This site is not intended to be abused for marketing purposes. We used to have 2fa providers section but deleted because it was nothing but spam. Clueless "directors of marketing" were coming here spewing their BS.

But I agree with @Carlgo11 regarding site being not high-traffic enough.

PR #3005, opened 2018-02-12
Add Planète OUI

nagromc:


Carlgo11 (changes requested):
Looking at [alexa.com](https://www.alexa.com/siteinfo/planete-oui.fr) it seems like the site is rather insignificant.
Our limit is around 200k.

nagromc:
> Carlgo11 **requested changes**

I'm not sure what you are expecting from me.

mxxcon:
@nagromc What Carl means that this site is not high traffic enough to list. Since there are essentially infinite number of sites we try to apply some sort of filtering criteria. As such we generally list sites which are <200k on alexa.

nagromc:
I do understand the fact the site is not popular enough to appear on twofactorauth.org

But, I don't understand what I am supposed to do with the **changes requested**.

PR #3008, opened 2018-02-13
Update CM Telecom doc

stephengroat:


Carlgo11 (dismissed):

fpigerre:
I can't seem to find any documentation regarding CM Telecom's software based two-factor authentication. Apart from that, this pull request seems good to merge.

mxxcon (commented):
Yes, both the new doc link and https://help.cmtelecom.com/en/account-wallet/account/how-can-i-login-to-my-cm-dashboard talk only about sms and voice call, but not software tfa. Perhaps it should be removed.

Carlgo11:
I emailed them, will keep y'all posted on any updates from them.

Carlgo11:
Response from CM:

```
Hi Carl,

Thank you for your interest in our 2FA products. Great that you've been looking at the help centre articles but there's also a lot of information on the products page.

If you dont have an app you can send sms, voice, hybrid: https://www.cm.com/products/access/one-time-password/
If you have an app we can do this (push or hybrid): https://www.cm.com/products/access/one-time-password/
If you dont have an app you can also use ours: https://www.cm.com/products/access/authenticator/


Kind Regards,
Matthijs Pepping

CM.com
```

conorgil:
If I am understanding this service correctly, they offer 2FA APIs for customers to use. If that is the case, then linking to the product page doesn't really make sense since we're interested in whether authenticating to CM.com supports 2FA. Since the support docs do not mention software tfa, I agree with @mxxcon that it makes sense to remove it

PR #3010, opened 2018-02-13
Fix xapo support doc

stephengroat:


Carlgo11 (approved):

Jawshy (approved):
👍

Jawshy (commented):
[This doc page](https://support.xapo.com/how-do-i-find-the-6-digit-code-for-second-factor-authentication-2fa/view=faq) says that the option to receive the login code via SMS was removed. Perhaps `sms: Yes` should be removed, but their [other doc page](https://support.xapo.com/how-do-i-enable-disable-the-second-factor-authentication/view=faq) still mentions verifying a mobile number.

jamcat22 (commented):
Maybe an exception that a linked phone number is required for 2FA, then remove the SMS and phone call tags?

fpigerre (approved):

PR #3012, opened 2018-02-14
Update Vanguard Doc

Sunit-Kulkarni:
Previous url link does not point to the correct information anymore. Found new link and replaced it.


stephengroat (approved):
thanks for the pr @Sunit-Kulkarni

Carlgo11 (approved):

PR #3013, opened 2018-02-15
Update Adobe ID - software

woodedjurassicpark:


Jawshy (changes requested):
Thanks for the pull request, @hellyarbm. 

Could you please add the following under `doc`: 
```ruby
      exceptions:
          text: "To activate 2-Step Verification, you must provide a mobile phone number."
```

woodedjurassicpark:
Nice catch Jawshy, as I definitely missed that

Jawshy (approved):

jamcat22 (approved):

Issue #2995, opened 2018-02-10
Add CSP and other headers to github pages


nitrocode:
Noticed the [securityheaders.io scan results](https://securityheaders.io/?q=https%3A%2F%2Ftwofactorauth.org&followRedirects=on) gave this site a D due to these missing headers

* Content-Security-Policy - [enabled using meta tag](http://qszhuan.github.io/technology/2015/08/12/add_csp_to_github_blog)
* X-Frame-Options - unsure if this is possible
* X-XSS-Protection - unsure if this is possible
* Referer-Policy - unsure if this is possible

Original issue closed when even fewer headers could be enabled on github pages: https://github.com/2factorauth/twofactorauth/issues/1527

ghost:
None of these are possible with GH. You could use Cloudflare workers to add the headers at the edge but they cost money.

PR #2893, opened 2017-12-15
Update js deps and move to unveil on cdn

stephengroat:
move to `unveil2` on `cdnjs`, update all other deps

functionality verified, should be reviewed again

PR #3015, opened 2018-02-16
Add CSP to _layouts files

brycx:
I took a swing at issue #2995 and included a CSP in the html files, in the _layouts directory.


Carlgo11:
How about this one?
It's set to report to report-uri so that we can monitor the results before enforcing the rule.

```
script-src https://ajax.cloudflare.com/cdn-cgi/scripts/b7ef205d/cloudflare-static/rocket.min.js;
style-src https://fonts.googleapis.com/css?family=Open+Sans:400,700 https://cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.2.7/semantic.min.css https://fonts.googleapis.com/css?family=Lato:400,700,400italic,700italic&subset=latin; 
font-src https://fonts.gstatic.com/s/opensans/v15/mem8YaGs126MiZpBA-UFW50bbck.woff2;
report-uri https://twofactorauth.report-uri.com/r/d/csp/reportOnly; 
```

brycx:
I read on MDN that `report-uri` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri). I'm not entirely sure when support for it will be dropped completely, but would it be good to include it anyway?

stephengroat:
@brycx mozilla may have deprecated it, but the spec didn't (https://www.w3.org/TR/CSP2/#directive-report-uri)

anyway, `report-uri` get ignored if it's in a `meta` tag, so it won't work. great thought though @Carlgo11 !

brycx:
@stephengroat That's my bad.

stephengroat:
np. i think it's a great idea, just always a pain to implement it at first :-) great start!

image

since we use SRI, i'm not sure that locking down to specific libraries is really necessary, what do you think? i'm not the biggest fan of having to update something in 2 different places, the lock to just the domain (`cdnjs.cloudflare.com`) seems good enough to me

we'll also have to include some fonts, what about something more like this? don't think this is complete, but i think it's a better start

```
+  <meta http-equiv='Content-Security-Policy' content="
+    script-src 'self' code.jquery.com cdnjs.cloudflare.com;
+    style-src 'self' fonts.googleapis.com cdnjs.cloudflare.com;
+    font-src 'self' fonts.gstatic.com cdnjs.cloudflare.com;
+    img-src 'self';
+  ">
```

brycx:
@stephengroat I think that makes sense, I can't see why we should specify the libraries when SRI is used. I added your CSP in the latest commit but also prefixed the domains with `https://` to force connections over HTTPS. I'm not sure what else could be added though?

stephengroat:
the site is delivered over HSTS so the `https` is superfluous

the current ruleset you have still currently breaks some stuff, have you look in the browser console to see what's being blocked?

i have to add `unsafe-inline` to the `style-src` directive and `data:` to the `font-src` directive to allow inlined js and locally hosted fonts, respectively. do you see those errors on your branch?

brycx:
I thought I had checked it but most likely I forgot to switch branches when testing it. Anyway, I see the breakage and need the `'unsafe-inline'` directive, so it has been committed. But I don't need the `'data:'` directive and get no error reports in the console when leaving it out.

stephengroat:
great work, i know this is super annoying to do but i think it's a great step forward.

i think the `data:` comes from the fact that it allows a site to user a browser's locally cached fonts instead of having to redownload them each time.

are you using chrome or firefox? maybe since they're google fonts, chrome stores them locally and that's why i'm seeing the error? errors that i'm seeing on the tip of your branch:

screen shot 2018-02-24 at 2 51 19 pm

brycx:
I actually don't think this is annoying, I've never done it before so I'm getting a lot out of it.

Your theory seems to be true. I used Firefox before but after running it through Chromium I got the error on the `data:` directive too. It's been committed. Thanks for the screendump btw.

stephengroat (dismissed):
this pr should not be merged until #2893, which has some superseding updates that are important

fpigerre (approved):

fpigerre (changes requested):
It looks like we can add custom HTTP headers using the [Jekyll configuration](https://jekyllrb.com/docs/configuration/#custom-webrick-headers). I'll make changes to this pull request shortly to support that.

stephengroat:
https://github.com/2factorauth/twofactorauth/pull/3015#pullrequestreview-109574001 @psgs since webrick isn't acutally serving the content, i don't think those changes have any affect on CSP for Github Pages/Cloudflare

fpigerre (dismissed):

fpigerre:
In that case, it looks like this pull request is ready to be merged. 👍

fpigerre (approved):

PR #3016, opened 2018-02-16
Correct Twitter exception.

### LINKED ISSUE #3014: Twitter no longer requires SMS for 2FA
Title.

This is listed as an exception, but this (thankfully!) isn't the case anymore.

brycx:
Fixes #3014.

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -223,7 +223,7 @@ websites:
       tfa: Yes
       sms: Yes
       exceptions:
-          text: "SMS only available on select providers. SMS required for 2FA. 10 accounts per phone number, 1 phone number per account."
+          text: "SMS only available on select providers. Phone number associated with account required for 2FA. 10 accounts per phone number, 1 phone number per account."
       software: Yes
       doc: https://support.twitter.com/articles/20170388
 


Jawshy (approved):
Looks good. Thanks for the pull request.

Carlgo11:
I don't know. It has to be a SMS capable phone and so you'd have to do the same steps as setting up SMS 2FA just without choosing SMS at the end.

brycx:
> I don't know. It has to be a SMS capable phone and so you'd have to do the same steps as setting up SMS 2FA just without choosing SMS at the end.

I see what you mean. To me, saying that SMS is required for 2FA, suggest that you need both enabled and that SMS then will be a registered option to use as 2FA. I guess it's just a matter of wording.

Carlgo11:
Geez Rick, I don't know if this is a good idea. Don't you think it would be better with something like `"SMS only available on select providers. An SMS capable phone is required to set up any 2FA. 10 accounts per phone number, 1 phone number per account."`?

(I'm not a native English speaker so maybe it would be good to have someone else look at this before committing (@2factorauth/collaborators))

brycx:
I'm no native English speaker either, so let's do that.

fpigerre (approved):

PR #3017, opened 2018-02-16
Add HTTPS rule to contribution guidelines

### LINKED ISSUE #2955: Contribution guidelines - Requiring HTTPS links if the site supports it
I think it would make sense to have a rule, or at least mention, that sites which support SSL should also be linked with HTTPS. Some sites automatically redirect to HTTPS and some don't. Not everyone forces HTTPS in their browser either. In any case I think it generally would be good practice to require SSL links when available.

Could such a rule/mention be added to `CONTRIBUTING.md`?

brycx:
Fixes #2955.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,9 @@ are stored in folders corresponding to each of those categories in their own
 2. **Use a Nice Icon**: The icon must have a resolution of 32x32. PNG is the
    preferred format. If possible, please also run the image through an optimizing
    utility such as OptiPNG before committing it to the repo.
-3. **Be Awesome**: You need to be awesome. That is all.
+3. **HTTPS links**: All sites that support HTTPS should also be linked with an
+   HTTPS address.
+4. **Be Awesome**: You need to be awesome. That is all.
 
 ## Running Locally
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -149,8 +151,8 @@ The following is an example of a website that *supports* TFA:
 
 If a site does not provide TFA, the `twitter:` field should be included if the site uses
 Twitter. Facebook can also be included using the `facebook` field, as well as Email using
-the `email_address` field. If the website does not use the English language, the `lang:` 
-field should also be included. The fields `sms:`, `email:`, `phone:`, `software:` and 
+the `email_address` field. If the website does not use the English language, the `lang:`
+field should also be included. The fields `sms:`, `email:`, `phone:`, `software:` and
```
mxxcon, line 155: I think want to keep space here since semantically it's `'lang:' field` and not `'lang:'field`? But it's not the end of the world.

Jawshy: The change appears to make no difference once markdown is applied. 

See: [before](https://github.com/brycx/twofactorauth/blob/3de8e300481cbc029ea0bccefc1649aaf0377ea3/CONTRIBUTING.md) _v_ [after](https://github.com/brycx/twofactorauth/blob/dd6c07185e3b834ba5c896f772781ccd4380e005/CONTRIBUTING.md).

mxxcon (approved):

Carlgo11 (approved):

PR #3020, opened 2018-02-17
Remove webmin

stephengroat:
self hosted, doesn't meet criteria

virtualmin is a store selling hosted versions of webmin, but doesn't
offer 2fa, decided not worth it to include


Carlgo11 (approved):

Carlgo11:
Nice catch! 👍

mxxcon (approved):

PR #3021, opened 2018-02-17
Add Linksys Smart Wi-Fi

nagromc:
Linksys Smart Wi-Fi allows users to log in their router without the need to be on the same local network. Typically, from the Internet with a smartphone application.

I though Remote access was the most suitable category for that. Fell free to move it if necessary.

Please note the HTTPS URL does not seem to work at the moment (`SSL_ERROR_UNRECOGNIZED_NAME_ALERT` from my Firefox). I don't know if it's a good idea to switch to HTTP. Maybe HTTPS never has been working.


Carlgo11:
Hello @nagromc,
We don’t list self-hosted services on our site. The service you describe seems to be self-hosted and so I’m afraid it can’t be added.
Sorry about that.

nagromc:
I'm not sure to understand why Synology is listed then.

Besides, for Linksys Smart Wi-Fi, the authentication is made on their website, then it gives access to your own router.

jamcat22:
@Carlgo11 this is not a self-hosted website. It’s similar to using a TeamViewer account to access a Windows computer, or Cisco Meraki’s site to manage a Meraki router.

stephengroat (approved):
thanks for the pr @nagromc

Issue #2993, opened 2018-02-08
Thoughts on adding .travis.yml to list?


kenman345:
https://github.com/2factorauth/twofactorauth/blob/365fc7b491d07dc51d3b5da14df03f0d075c6fb7/_config.yml#L13

Hey, Just wondering if a thought had occurred to include .travis.yml in the exclude tag. Trying to wrestle with doing it myself for our acceptbitcoincash/acceptbitcoincash fork.

mxxcon:
what kind of problems is it causing? 🤔

kenman345:
I just like cleanliness and figured I’d ask

Carlgo11:
Exclude makes Jekyll not include any of said files on the public website. There's no reason for files like CNAME, Gemfile or Rakefile to be on the public facing site and so it's excluded from the build.

Carlgo11:
Whops, I didn't read the entire issue. Sure, the travis file could be added there. I'll make a PR for it now.

Carlgo11:
See #3029

PR #3029, opened 2018-02-21
Exclude .travis.yml from jekyll build 

Carlgo11:
As per #2993


stephengroat:
from [jekyll docs](https://jekyllrb.com/docs/configuration/), dot files are excluded by default

Carlgo11:
haha closing PR and issue then

PR #3038, opened 2018-02-26
Update Epic Games

matt-t-jones:
As of 2/20 it seems EpicGames has implemented a 2FA system. Granted, the authentication code is only required for new sign-ins or after 30 days, so I understand if these changes don't fall in line with the rest of the site. In my opinion, this still qualifies as 2FA because it is authentication for the entire account, not just certain features.

As included in the gaming.yml update, you can read more about it here: http://help.epicgames.com/customer/en/portal/articles/2879127-what-is-two-factor-sign-in-and-how-do-i-opt-in


mxxcon (commented):
I don't think that exception text is necessary. Unfortunately, this is usual for many sites 😞

matt-t-jones:
Shall I remove it?

stephengroat:
@matt-t-jones for consistency, it would probably be best to remove it. thanks for the attention to detail though!

stephengroat (dismissed):
thanks for the pr @matt-t-jones

jamcat22 (dismissed):

jamcat22 (approved):

fpigerre (approved):

PR #3043, opened 2018-03-05
Add Pennsylvania State University

Sunit-Kulkarni:


Carlgo11 (commented):
From looking at http://www.identity.psu.edu/set-up-instructions/#landline it seems like they support phone calls too. Is that right? If so, add `phone: Yes`.

Sunit-Kulkarni:
Good catch. Duo lets you use phone calls.

Carlgo11 (dismissed):

_data/education.yml
```diff
@@ -223,6 +223,16 @@ websites:
...
 
+    - name: Pennsylvania State University
+      url: http://www.psu.edu/
+      img: psu.png
+      tfa: Yes
+      sms: Yes
+      phone: Yes
+      software: Yes
+      hardware: Yes
+      doc: http://www.identity.psu.edu/services/authentication-services/two-factor/self-service-portal/
```
jamcat22: I think we should change the URL to either their overall service topic on 2FA (http://www.identity.psu.edu/services/authentication-services/two-factor/) which links to several articles, including the one about enabling it, or their 2FA enrollment site (http://get2fa.psu.edu/).

jamcat22 (dismissed):

_data/education.yml
Sunit-Kulkarni: I say use the First Link since it is more general.

conorgil: @Sunit-Kulkarni I'm hopping in this thread late, but based on the comments from review it sounds like once you update that URL this PR should be good to go

Sunit-Kulkarni: Alrighty. Just updated it!

stephengroat (approved):

mxxcon (approved):

Issue #3046, opened 2018-03-08
TD Bank US Supports 2FA


elahd:
Supported methods are phone call and SMS.

![td2fa](https://user-images.githubusercontent.com/466460/37169559-38ce3cde-22d6-11e8-8530-7f17a28ad932.PNG)

elahd:
Fixed in #3049

Issue #3045, opened 2018-03-08
Con Edison now supports 2FA


elahd:
Supported methods are phone call, sms, and Google Authenticatior.

![coned2fa](https://user-images.githubusercontent.com/466460/37169351-b3955ce6-22d5-11e8-8d29-abea8d81e1f3.png)

elahd:
Fixed in #3048

PR #3047, opened 2018-03-08
Updated Con Edison and TD Bank US

elahd:
Con Ed and TD Bank US now support 2FA. Updated accordingly.

#3046 #3045


mxxcon (changes requested):
Still must have `tfa: Yes` line for both entries.
Also let's make 2 separate pull requests for each site so that they can be reviewed independently.👍

jamcat22:
Duplicate of #3052, #3049, and #3048

PR #3050, opened 2018-03-09
Update DocuSign to add software token

conorgil:
Evidence is on the existing documentation link:
https://support.docusign.com/en/guides/ndse-user-guide-two-step-verification

Section:
"Use an Authenticator app for two-step verification"


jamcat22 (approved):

stephengroat (approved):
thanks for the pr @conorgil

PR #3061, opened 2018-03-13
Removed executable permissions from some logos

NullPiotrException:
Some .png files had permissions -rwxr-xr-x (755) instead of -rw-r--r-- (644).
Potential vulnerability.

**Edit:** Found them by running:
`~/Documents/twofactorauth$ find . -perm /u=x,g=x,o=x -type f`


mxxcon (dismissed):

stephengroat:
if this is something we're going to enforce, let's do it programatically as well

Carlgo11 (approved):

fpigerre (approved):

PR #3066, opened 2018-03-15
Adds more accurate pattern validations

kenman345:
I've used these for the [acceptbitcoincash/acceptbitcoincash](https://github.com/acceptbitcoincash/acceptbitcoincash) fork to be more precise when validating using the verify.rb script.

Since your verify.rb script validates the img tag uses only .png files, why not just add the pattern here instead of having it in the verify.rb only?

### DIFF
--- a/websites_schema.yml
+++ b/websites_schema.yml
@@ -18,24 +18,26 @@ mapping:
           "img":
             type: str
             required: yes
+            pattern: /\.png$/i
           "tfa":
             type: bool
             required: yes
+            pattern: /(true|false)/
           "software":
             type: bool
-            pattern: /true/
+            pattern: /(true|false)/
           "hardware":
             type: bool
-            pattern: /true/
+            pattern: /(true|false)/
           "sms":
             type: bool
-            pattern: /true/
+            pattern: /(true|false)/
           "phone":
             type: bool
-            pattern: /true/
+            pattern: /(true|false)/
           "email":
             type: bool
-            pattern: /true/
+            pattern: /(true|false)/
           "exceptions":
             type: map
             mapping:


RichJeanes:
The "true|false" on the "tfa" tag seems sensible, but we don't want the types of available 2FA to be listed as False. We assume them to be false if they are missing (since it's a binary choice) instead of cluttering up the .yml files. So if they are present they _should_ only be true.

kenman345:
it feels like you're writing the verify.rb script a little backwards with that logic. See, your Website_schema.yml shouldn't have `required: yes` in it for that tag if its not always supposed to be there. it makes it confusing without reading other context documents in your repository.

stephengroat:
> Since your verify.rb script validates the img tag uses only .png files, why not just add the pattern here instead of having it in the verify.rb only?

@kenman345 we also do other, deeper validation that the img is actually a png beyond simply the file extension [using `FastImage` in `verify.rb` as well](https://github.com/2factorauth/twofactorauth/blob/master/verify.rb#L43). i think centralizing that logic in a separate script (since it's doing more than file extension) with the ability to throws informative errors probably makes sense, right? if (for some crazy reason in the future) someone decides to change the img type (better compression, whatever), that way the change for validation can be made in one place.

one interesting thing that i never got implemented was [kwalify hooks](http://www.kuwata-lab.com/kwalify/ruby/users-guide.03.html#howto-hook). in that case, i think removing the `.png` check from ruby and using kwalify regex might make sense

> See, your Website_schema.yml shouldn't have required: yes in it for that tag if its not always supposed to be there. it makes it confusing without reading other context documents in your repository.

in reference to the `required` question, which tag are you referring to? the `img` tag is always supposed to be there, hence it's required. we don't require (or really want) `false` tfa methods, just adds clutter.

websites_schema.yml
```diff
@@ -18,24 +18,26 @@ mapping:
           "img":
             type: str
             required: yes
+            pattern: /\.png$/i
           "tfa":
             type: bool
             required: yes
+            pattern: /(true|false)/
```
stephengroat, line 25: i'm not sure, is this redundant? `bool` should only be able to be set to `true` or `false`. could you override this with a `nil` or something else?

RichJeanes:
>your Website_schema.yml shouldn't have required: yes in it for that tag if its not always supposed to be there

The "Required" flag is only on the "tfa" tag which is required. I was referring to the types of 2FA (SMS, software, etc.) that should not be listed as false, but rather omitted:
>we don't want the **types** of available 2FA to be listed as False.

websites_schema.yml
kenman345: the validate command in the verify.rb will appreciate this more along with allowing for better handling if you add in more verifications written out manually for those fields.

kenman345:
let me know if you think we need modifications then. Happy to remove the ong pattern if you think thats better for you.

websites_schema.yml
stephengroat: sorry, i'm just especially dense this morning and i'm not understanding.

could you expand on your answer? still don't understand how this isn't redundent

kenman345: looking over things, I cannot find where it was that I was having trouble. Maybe you can ignore this if you think its redundant.

stephengroat:
thanks for the taking the time to make & explain this pr @kenman345 (super impressed by https://acceptbitcoin.cash btw!), but i think we're going to close this for right now

the `img` tag change is non ideal for us (since it splits img type validation between two different files), the `tfa` change _may_ be redundant (not 100% sure), and the additions to different `tfa` forms go against our goal of keeping super clean files.

thanks for thinking of bringing your changes back to us, though! i'm sure there's other great stuff that your doing (or will do) on https://acceptbitcoin.cash that we'd love to see!

kenman345:
@stephengroat you may like #3067 then

PR #3068, opened 2018-03-15
Update sprint

### LINKED ISSUE #2864: Sprint needs updating
Sprint now uses SMS to provide customers 2FA

stephengroat:
fixes #2864

### DIFF
--- a/_data/utilities.yml
+++ b/_data/utilities.yml
@@ -272,10 +272,10 @@ websites:
 
     - name: Sprint
       url: https://www.sprint.com/
-      twitter: sprint
-      facebook: sprint
       img: sprint.png
-      tfa: No
+      tfa: Yes
+      sms: Yes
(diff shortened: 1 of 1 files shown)


fpigerre (dismissed):

mxxcon:
docs site says

> You may also be required to complete the 2 step authentication process if you are making changes to your Sprint.com account, or purchasing a device either online or through Telesales.

That sounds more like 2factor authorization, not authentication?

stephengroat:
I agree, that part is authorization. but the paragraph above seems to be authentication (to me)

>Important: If you have not logged in to your Sprint.com account for more than 6 months, you will be prompted to reset your password, which will require a two step authentication process.  You will receive a code (either by text message or email) which you will need to use to authenticate and complete the password reset process.

jamcat22:
@stephengroat that paragraph to me suggests that another form of authentication is only used for resetting passwords, not for logging in to an account.

Carlgo11:
I also agree with @mxxcon and @jamcat22 on that it looks like authorisation.

PR #3075, opened 2018-03-19
Dockerize your application

jmeridth:
- `make run` will standup the web app in a docker container locally and make the site available via http://localhost:4000

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ Thumbs.db
 ehthumbs.db
 Gemfile.lock
 .sass-cache
+.jekyll-metadata
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM jekyll/jekyll:latest
+
+WORKDIR /tmp
+ADD Gemfile /tmp/
+RUN ls -lah / && whoami && bundle install
+
+COPY . /jekyll
+WORKDIR /jekyll
+
+EXPOSE 4000
+ENTRYPOINT ["jekyll", "serve", "--watch", "--incremental"]
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,17 @@
+.PHONY : help
+help: # Display help
+	@awk -F ':|##' \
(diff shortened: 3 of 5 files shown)


stephengroat:
@jmeridth thanks for the idea!

correct me if i'm wrong, but doesn't this just make the repo dependent on `docker` as well as `ruby`? we have other verification tools and scripts that run in `ruby` ([see the `Rakefile`](https://github.com/2factorauth/twofactorauth/blob/master/Rakefile)) and since the `jekyll` Docker image doesn't have all of our [Gem dependencies](https://github.com/2factorauth/twofactorauth/blob/master/Gemfile), we can only run `jekyll serve` out of the Docker image (and not all of our verification tasks as well).

I'm not sure just splitting off `jekyll serve` into a separate process is worthwhile while still requiring ruby for other verification tasks is worth it, what do you think? (I do like the use of `docker compose` to deal with those nasty docker command line options though, that's slick!)

jmeridth:
@stephengroat definitely valid points.  I did this because I don't like installing dependencies on my local box.  The Dockerfile does run `bundle install` so your gem dependencies are there.  I was able to bring the whole site up locally.  Also the Dockerfile is running with `--incremental` so if changes are made locally, jekyll restarts and the changes are seen immediately.

Please note, this PR is made without full knowledge of your stack end-to-end. :smile:

Glad you like the `docker-compose` usage. :smile:

jmeridth:
@stephengroat I will look at the other Ruby dependencies and see about standing up another ruby container and use multi-stage docker file to bring those into a custom container for this.  Thanks for that heads-up.

stephengroat:
@jmeridth it becomes a maintenance & support issue issue, having to deal with another stack (on top of ruby)

what about adding something to the README.md that points people to the [jekyll docker documentation](https://github.com/envygeeks/jekyll-docker/blob/master/README.md)?

jmeridth:
@stephengroat sounds like a plan.  Will do that instead.  :beers:

PR #3080, opened 2018-03-25
Add California Institute of Technology (Caltech)

Sunit-Kulkarni:


mxxcon (commented):
I looked through the linked site and I don't see where it supports sms or phone call.
I see that sms it needed to activate duo, but it doesn't appear that it's being used for actual 2fa authentication...?

Sunit-Kulkarni:
Good catch. I've been assuming universities that use DUO turn on all the options. But this pictures says otherwise: https://imss.caltech.edu/node/1410. I'll make the change

mxxcon:
I think this also warrants an exception text that sms is required to activate (if that's actually the case).

Sunit-Kulkarni:
I think it's good to go now!

mxxcon (dismissed):

Carlgo11 (changes requested):
I suggest you use this image instead: caltec
It works better in smaller sizes

_data/education.yml
```diff
@@ -34,6 +34,16 @@ websites:
...
       email_address: support@en.babbel.com
 
+    - name: California Institute of Technology
+      url: https://www.caltech.edu
+      img: caltech.png
+      tfa: Yes
+      software: Yes
+      hardware: Yes
+      exceptions:
+          text: "sms is required to activate"
```
Carlgo11: On similar sites we've had the exception text `SMS-capable phone required for initial setup.`. Would that work here?

_data/education.yml
```diff
@@ -34,6 +34,16 @@ websites:
       facebook: babbel.languages
       email_address: support@en.babbel.com
 
+    - name: California Institute of Technology
```
Carlgo11: How about adding the suffix (Caltech) at the end to aid the search feature on our site? I think most people would search for that instead of the full name.

Carlgo11 (changes requested):

mxxcon:
@Carlgo11, from what i understand, i'm not sure if logo is official. it seems like there was some competition to design a new logo for them, but i'm not sure if that logo got approved.

Carlgo11:
@mxxcon but it’s their current favicon

mxxcon:
Then i guess they did approve that logo :)

_data/education.yml
Sunit-Kulkarni: Good Idea. I'll add that in

_data/education.yml
Carlgo11: @Sunit-Kulkarni last thing to change ^

Sunit-Kulkarni: Yup that should work there.

Sunit-Kulkarni: just updated it

mxxcon (approved):

Carlgo11 (approved):

Carlgo11:
Thanks for the changes @Sunit-Kulkarni :smile: 

It could take some time for your contribution to show up as we update periodically. (See [our wiki](https://github.com/2factorauth/twofactorauth/wiki/FAQ-Merge) for more info)

PR #3083, opened 2018-03-26
Minor tweaks to .travis.yml

kenman345:
I've run a lot of Travis builds for github.com/acceptbitcoincash/acceptbitcoincash repository. I've found bundling does not work as expected once you define a directory. So it's recommended to define the folder as well. 

Additionally, when you define `rvm: 2.4.0` you are already getting a ruby language build, so no need to define it.

I decided to not go one step further and change the line `os: linux` to `dist: trusty` as you may want to do more serious validations on that, but essentially its the exact same anyways. I just dont have Travis-CI local environment to validate that so just going off what I know.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 rvm: 2.4.0
 os: linux
+env: NOKOGIRI_USE_SYSTEM_LIBRARIES=true
 cache:
   bundler: true
   directories: 
@@ -16,7 +17,6 @@ matrix:
   include:
     - script: bundle exec rake
     - script: bundle exec rake proof_external
-      env: NOKOGIRI_USE_SYSTEM_LIBRARIES=true
   allow_failures:
     - script: bundle exec rake proof_external
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
 sudo: false
-language: ruby
 rvm: 2.4.0
 os: linux
 cache:
   bundler: true
-  directories: $TRAVIS_BUILD_DIR/tmp/.htmlproofer
+  directories: 
+  - $TRAVIS_BUILD_DIR/tmp/.htmlproofer
+  - vendor/bundle
 addons:
   apt:
     packages:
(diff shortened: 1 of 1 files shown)


kenman345:
Adding the commit message from the second commit inline for you to all see: 

>I forgot about this change too. AB.C cannot use it as we moved to Staged Jobs so we can do deploys as well, but here you go: https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing

I wanted to add to this that currently staged builds on Travis do not recognize that fast_finish option but it is on the roadmap. You are not using staged builds so you do not need to worry, but it is not like I am suggesting something that you will 100% remove if you revised to use the travis-ci staged builds. Its just that currently staged jobs would need that line removed as it doesnt handle it.

kenman345:
BTW, you may wish to move that environment variable up to the top of the travis.yml file due to the fact you are using `check_html: true` in both your rake instances:
https://github.com/gjtorikian/html-proofer#configuration

stephengroat:
just fyi, `dist: trusty`  is unnecessary since [`trusty` is now the default](https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System). `os: linux` should just keep the builds on the tip of whichever ubuntu version is being run

i like the rvm line, one less line is always good!

i'm not seeing any cache misses from using `bundler: true` (we seem to get great performance from that) and there's always the possibility of then doubling up on the cache (which would take long and increase storage costs for travis). do you see any cache misses in our logs?

kenman345:
https://docs.travis-ci.com/user/caching/#Determining-the-bundle-path

I have also looked at a few configurations and bundling performs better when you set the path directly. As I noted, once you use the directory tag, it starts to sometimes miss things and its best to be explicit about it.

stephengroat:
@kenman345 i've never seen the performance promised from `NOKOGIRI_USE_SYSTEM_LIBRARIES=true` (seems to take longer, right?). it does have an effect on external link checking though, so it should remain in that specific job.

also, since we're not using any weird `bundler` arguments, i don't think that explicitly listing the `vendor/bundle` really gives us anything. probably just more maintenance overhead in the future. our cache is really small and seems to take under 30s to down and upload (since the `html-proofer log usually changes and has to get updated), so any performance gain there is almost negibile

kenman345:
Then lets close this and I'll make a new PR just for the fast_finish then

PR #3090, opened 2018-03-28
Move to jquery.lazy for image loading

stephengroat:
maintained project

fixes current slow load time issues

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -6,7 +6,7 @@ $(document).ready(function () {
   }
 
   // Unveil images 50px before they appear
-  $('img').unveil(50);
+  $(function() { $('img').Lazy({visibleOnly: true}); });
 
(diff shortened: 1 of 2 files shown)


js/app.js
```diff
@@ -6,7 +6,7 @@ $(document).ready(function () {
   }
 
   // Unveil images 50px before they appear
```
jamcat22: Does this code comment need to be updated? Does jquery.lazy still use 50px as the padding for lazy loading?

fpigerre (approved):

fpigerre:
I'm fast-tracking this pull request with one review as a result of discussion in #2893, so that we can merge #3015.

fpigerre:
Plugin being used can be found at http://jquery.eisbehr.de/lazy/.

PR #3091, opened 2018-03-29
Partialized the row title div section

kenman345:
This moves the imgs for logos to its own partial to be included from the multiple places it is used. This gives you better revision flexibility in the future.

Also, I removed the If website.img check since img is required. You should never be in a situation where you do not have it.  

However, I could revise it more to the tune of what I did for Accepbitcoincash/acceptbitcoincash here: https://github.com/acceptbitcoincash/acceptbitcoincash/blob/adoption-stats/_includes/segment/row-logo.html

in that version I made placeholder.png our default as we removed the img tag from being required for our site.

### DIFF
--- a/_includes/row-title.html
+++ b/_includes/row-title.html
@@ -1,0 +1,15 @@
+{%- assign website = include.website -%}
+{%- assign img_path = include.section | append: '/' | append: website.img -%}
+<div class="title">
+  <noscript>
+    <img src="/img/{{ img_path }}" class="icon" alt="{{ website.name }}">
+  </noscript>
+  <img src="/img/placeholder.png" data-src="/img/{{ img_path }}" class="icon" alt="{{ website.name }}">
+  <a href="{{ website.url }}" class="name">{{ website.name }}</a>
+  {%- include exception.html website=website -%}
+  {%- if website.status and include.type == 'mobile' -%}
+    <a href="{{ website.status }}" target="_blank">
+      <i class="external url link large icon"></i>
+    </a>
+  {%- endif -%}
+</div>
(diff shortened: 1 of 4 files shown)


kenman345:
let me know if you want me to go a step further and instead of the partial being just the image we could make it the entire `<div class="title">` section since its always the same no matter which style or type of table you're making

stephengroat:
https://github.com/2factorauth/twofactorauth/pull/3091#issuecomment-377236937 sounds like a good idea to me

kenman345:
@stephengroat  done! take a look when you can

stephengroat (dismissed):
lgtm

i'm always a fan of reducing code reuse (as long as it doesn't make the situation more confusing). i don't see this as more confusing, but we'll wait for feedback from others

_includes/row-title.html
```diff
@@ -0,0 +1,15 @@
+{%- assign website = include.website -%}
+{%- assign img_path = include.section | append: '/' | append: website.img -%}
+<div class="title">
+  <noscript>
+    <img src="/img/{{ img_path }}" class="icon" alt="{{ website.name }}">
```
conorgil: nit pick: might make sense to put "/img/" into the img_path variable since its repeated in the body here and on L7

conorgil:
being new to the code base, this looks like a nice improvement in readability. DRY is good!

_includes/row-title.html
kenman345: because on L7 we still have to write out /img/ for the placeholder image it doesnt make that much sense to overly shorten things. It can be done and wouldnt be wrong. I would actually love to play with using the src-set tag to define "/img/" so it gets used for both on Line 7. I just havent messed with it enough to confirm it works as expected

conorgil: I can see someone updating L5 and forgetting L7 or vice versa. The placeholder image path in L7 does also start with "/img/", but doesn't use the variable and is a hardcoded value.

Its definitely a nit pick, but figured I would throw it out there.

kenman345:
@conorgil  made that change, you made a strong case for why it should be updated that I had not thought about. Thanks.

conorgil:
@kenman345 hah cool. Looks good!

fpigerre (approved):

PR #3099, opened 2018-04-01
Add Signal

NinebitX:
* Added signal.png to /img/communication
* Added Signal to /_data/communication.yml


RichJeanes:
Yeah, I guess that does count as 2FA, doesn't it? It's just backwards since the default verification is possession and the second factor is knowledge...

_data/communication.yml
```diff
@@ -179,6 +179,14 @@ websites:
       software: Yes
       doc: https://help.sendloop.com/docs/securing-your-account
 
+    - name: Signal
+      url: https://www.signal.org/
+      img: signal.png
+      tfa: Yes
+      sms: Yes
+      phone: Yes
```
RichJeanes, line 187: In my experience, you cannot verify your number by phone call, only SMS. Can you verify? It's been a while since I set up Signal from scratch.

stephengroat (approved):
thanks for the info @NinebitX! didn't even know this feature was available

RichJeanes (changes requested):

_data/communication.yml
stephengroat: https://support.signal.org/hc/en-us/articles/115005335227

RichJeanes (approved):

_data/communication.yml
RichJeanes: Thanks, @stephengroat

fpigerre (approved):

PR #3102, opened 2018-04-03
Added autofocus attribute

daytonlowell:
So the user can just start typing as soon as the page loads.

### DIFF
--- a/index.html
+++ b/index.html
@@ -58,7 +58,7 @@ hash: SupportTwoFactorAuth
       <i class="search icon"></i>
     </label>
     <input type="search" id="jets-search" placeholder="Search websites" autocomplete="off" spellcheck="false"
-           tabindex="0">
+           tabindex="0" autofocus>
   </div>
 </div>
 


stephengroat (approved):
great idea @daytonlowell, thanks!

fpigerre:
Hello @daytonlowell,

The use of the `autofocus` attribute was discussed a while back in issue #2345, however, for the most part, it became clear that new users would likely prefer not to have their browsers set to autofocus on the search bar when they visited the site for the first time, as well as mobile users.

Thanks,
psgs 🌴

daytonlowell:
Fair.

Google, Yahoo, Bing, and DuckDuckGo all autofocus on desktop. I don’t have an Android device handy to test what they do on mobile.

That said, would you be interested in an autofocus implementation that was desktop-only?

fpigerre:
I'm not quite sure. Perhaps @mxxcon might be interested?
As long as the implementation isn't too JavaScript heavy, it could be useful.

mxxcon:
Google, Yahoo, Bing, and DuckDuckGo do it because there's really nothing else to do on their main pages other than search. 
Here people can browse the list.. ☺️
If it's desktop specific, I think it's a better implementation.

PR #3104, opened 2018-04-04
Update rackspace doc

stephengroat:


mxxcon:
isn't old url more generic which includes other services?

stephengroat:
unneeded now, it was coming back with 404s earlier

PR #3115, opened 2018-04-05
Adding proxycheck.io

proxycheck:


Carlgo11:
Hello @proxycheck,
thank you for your interest in twofactorauth.org!

Due to the size of our lists we try to limit new additions to sites ranked higher than 200k on alexa.com. It appears that your site that criteria and so sadly I'm afraid that we can't merge your PR for the time being.
Hopefully proxycheck will grow bigger 🤞, if that happens we'll certainly take a look at your site again 😃 

Best of luck with your VPN service!
//Carl

proxycheck:
It's actually a proxy and VPN detection / blocking service, not a VPN provider service etc

But I understand your criteria. Thanks anyway.

PR #3117, opened 2018-04-07
Add yulPa hosting

jerome-lecorvaisier:
Wow wow welcome yulPa hosting company :v: :v:


mxxcon:
Thank you for the submission. While it'd be awesome to add all the possible sites out there, we try to stick to alexa top 200k sites, (and particularly the ones that don't support 2fa). You already support it, so job well done. 👍

PR #3119, opened 2018-04-09
Add AirBank

ondrejfuhrer:
Hello guys, adding a Czech bank to the banking list 👍 Thanks for such a great service.


mxxcon (dismissed):
As you can see travis-ci checks are failing. Either move your entry into banking category or move your logo file into finance directory.

ondrejfuhrer:
@mxxcon Ah, my bad 😊

Carlgo11:
@ondrejfuhrer could you provide any documentation supporting this?

ondrejfuhrer:
Hey @Carlgo11 , I was trying to find any documentation regarding that, but unfortunately I was not able to do so. The only thing I found is regarding transaction confirmation (some articles, etc.) but they don't mention there, that the same approach could be also used to login authentication as well.

Theoretically, I can write to that bank support if they could provide me with anything, but that will take time. 

Is that a requirement to have it in this list?

ondrejfuhrer:
Hey @Carlgo11 , spoke with the support, they actually mention that in this FAQ, is that enough? Should it be in the `doc` parameter?

https://www.airbank.cz/co-vas-nejvic-zajima/jake-bezpecnostni-prvky-u-nas-mame

Carlgo11 (commented):
That link you provided looks good. Could you please add it as a `doc: `tag?
Thanks :smiley:

ondrejfuhrer:
@Carlgo11 Added 👍

Carlgo11 (dismissed):

Carlgo11 (approved):

ondrejfuhrer:
I don't want to be pushy, but is there any reason why this is still waiting to be merged? It's been approved for more then a month.

mxxcon (approved):

PR #3120, opened 2018-04-09
Add Caltech (#3080)

dennnath22:
* Added CalTech

* removed SMS and Phone Call options

* Added exception text

* added suffix CalTech

* added new CalTech favicon

* updated exception text


mxxcon:
Project managers will handle merging into `gh-pages`.

PR #3121, opened 2018-04-09
Adding SmartDebit

zeus67:


mxxcon:
Thank you for the pull request. However, for now we try to stick with a somewhat more trafficked sites, such as Alexa 200k+.
It looks like your site already supports 2fa too, so great job there! 👍

PR #3123, opened 2018-04-10
Search by url and website name

### LINKED ISSUE #3118: Not possible to search for domains.
It currently isn't possible to search for a domain directly:
Example: it is possible to search for Electronic arts, but not for ea.com.
Please change this.

kenman345:
Closes #3118 by introducing a new way of defining the search keywords. In this approach, we get the benefit of whatever text we put into the field and since we use this for both mobile and desktop tables, its automatically added to each of them consistently. 

No defaults needed for this as we are using two required fields.

### DIFF
--- a/_includes/row-title.html
+++ b/_includes/row-title.html
@@ -1,6 +1,7 @@
 {%- assign website = include.website -%}
 {%- assign img_path = '/img/' | append: include.section | append: '/' | append: website.img -%}
 <div class="title">
+  <div class="keywords" style="display:none;">{{ website.name }} {{ website.url }}</div>
   <noscript>
     <img src="{{ img_path }}" class="icon" alt="{{ website.name }}">
   </noscript>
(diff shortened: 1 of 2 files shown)


fpigerre (approved):

Carlgo11:
@psgs all done? You can merge if you think it's fine.

Carlgo11:
I get an error when running the site:
`jekyll 3.3.1 | Error:  Liquid syntax error (line 1): Tag '{%- assign website = include.website -%}' was not properly terminated with regexp: /\%\}/`

kenman345:
Run bundle update, Jekyll 3.3.1 is old. GitHub pages is on 3.7.3 and should
not have such error in building

Carlgo11:
That worked.

kenman345:
Yea, the hyphens must be new but it’s rooted in the liquid syntax
standards. Basically defines that ignore white space after this until the
next element. So it can be very useful to ensuring the compiled size is
exactly as you meant it to be

PR #3133, opened 2018-04-15
Reduce json size

stephengroat:
as of build, ~800k to ~200k


mxxcon:
How did you save so much space? by removing white spaces?
If that's the case maybe to preserve readability instead of making it all one giant line, preserve indentation but with only 2 spaces on each?

Also with gzip compression whitepsace compresses extremely well, so it might not give us any advantage?

stephengroat:
json is a serialization format, i don't know if we should be worried about the readability of a serialized data blob. shouldn't that just be max minified? also, there's tons of viewers if there's a need that'll pretty print json

can jekyll gzip something after generation?

mxxcon:
I mean gzip during transfer. If you'll look at https://redbot.org/?uri=https%3A%2F%2Ftwofactorauth.org%2Fdata.json you'll see that it reports gzip saved 92% compared to the original. 855kb file took only 64kb to transfer.

Carlgo11:
I don’t think this is necessary as gzip already compresses it.
Better to preserve readability.

fpigerre:
~~The `data.json` file isn't actually served verbatim to clients. Jekyll processes the contents of `data.json` then serves the resulting data. Whitespace only affects what's seen on GitHub and by Jekyll.~~

Whoops, looks like that was wrong. It turns out the whitespace was affecting the `data.json` output.

stephengroat:
the problem is the after transfer `data.json` gets blown back in memory whenever someone is trying to load it, creating a lot of unnecessary overhead. even if its only 64k on transfer, it's still going to be ballooned in RAM to full file size before it's compressed again by a JSON lib. doesn't seem to make sense to push that overhead if the goal is the have other services programmatically use it

kenman345:
dont you guys use GH Pages? 

This is pointless then since you wont be able to use such a plugin. 

Also, this makes the json file itself hard to manage if you wish to update it in the future. Take a look at what Acceptbitcoincash/acceptbitcoincash did here: https://github.com/acceptbitcoincash/acceptbitcoincash/blob/master/data.json

```liquid
---
layout: 'json'
show_generated_date: true
--- 
{%- assign pages = site.data.sections | group_by:"page" | sort:"name" -%}
{%- assign sortedPages = pages | reverse -%}
{% for tracked_page in sortedPages %} 
	"{{ tracked_page.name | capitalize }}": { {% for section in tracked_page.items %}{% assign section_file = site.data[section.id] %}
		"{{ section.title }}": [ {% for website in section_file.websites %}
			{{ website | jsonify }}{% unless forloop.last %},{% endunless %}{% endfor %}
		]{% unless forloop.last %},{% endunless %}{% endfor %}
	}{% unless forloop.last %},{% endunless %}
{%- endfor -%}
```

basically, by using the hyphens effectively, we reduced the output size of the json by it not containing blank lines anymore. 

Also, compressing the json is 100% fine as if you write a program, then its not going to care and most browsers will pick up .json format and make it pretty for you. Just look at properly formatted JSON url on firefox. If its not making it pretty, then you've got an improperly formatted JSON file and should fix it but otherwise it makes it so you have the sections as collapsed dropdowns/trees you click on to expand.

stephengroat:
> This is pointless then since you wont be able to use such a plugin.

@kenman345 why is this pointless? please don't just make statements about work that other people have done without any proof, it's rude and diminishes their contributions.

also, there's a bunch of unneeded whitespace in `acceptbitcoin.cash/data.json` as well. i compressed it to 217K from 223K. besides, it contains a bunch of newlines as well (while these lines aren't full blank lines, they're completely unnecessary):

```
$ curl -s https://acceptbitcoin.cash/data.json | wc
    1449    2573  227923
```

kenman345:
sorry.

I am still unsure what this is doing. If it relies on JSON lint though to make it smaller, then it will not work with GH Pages the way you have it configured since GH PAges compiles only with its whitelisted plugins.

If you want smaller JSON without it having to be that large or hard to understand when you edit then its probably a good idea to use the hyphen tags to remove the blank lines you create with the liquid tags the way you use them.

stephengroat:
@kenman345 it's not relying on JSON lint to minify the JSON, JSON lint is used to validate the JSON structure

kenman345:
ah okay, then this looks good to me if you see it that way.

Acceptbitcoincash repository uses the new line characters simply because for the small difference in size you just mentioned, its nicer to keep it readable when we output it locally for testing and future changes. GZIP will compress it anyways during transfer so not the biggest deal. Sorry about my misunderstanding of the use of jsonlint

stephengroat:
I understand, but JSON is a serialization format, not a presentation format. When used programmatically, every byte that gets expanded in RAM can really make a different. Formatting JSON for programmer debug readability is adding tons of overhead for (what should be) a tiny fraction of the accesses to the file

the entire website is about presenting the data in a nice format. if there's a desire to show people how to format JSON for readability, link people to something like https://jsonformatter.curiousconcept.com/.   `data.json` was last edited 2 years ago, so i don't think there's a ton of daily work being done on it as well

kenman345:
Do you actually know of something consuming that data?

stephengroat:
https://github.com/pierreprinetti/accountable
https://github.com/ralphtheninja/node-twofactorauth
there's also some chrome plugins that use it as well

kenman345:
Can someone explain why the use of the `raw` and `endraw` tags? I am confused how that is what gives the desired effect.

EDIT: Nevermind, so it makes sure no mistakes for the brackets needed for the JSON. But how does it somehow shorten the whitespaces?

kenman345:
Let me add an additional comment with no other noise around it.

Are you using the fact that you separated the start and end of a tag between lines to minify your JSON? That's the most stupid thing I've ever seen. You can accomplish this much better by making use of Liquid filtering functionality. When you open a tag like `{%` you instead open it like this `{%-` to have it exclude preceding whitespace and newlines from the previous tag. When closing a tag like this `-%}` you are telling the compiler to exclude any whitespace/new lines to the next tag. 

In essence, the way this PR should appear is to make sure each line ends with a tag with the `-` (hyphen) added into the tag beginning and end to make it use as little space as possible while still making it not a ticking time bomb that someone makes an edit and the JSON suddenly compiles to be much larger in size.

conorgil:
@kenman345 

> That's the most stupid thing I've ever seen.

Comments like this are flat out rude and don't further the conversation. I can tell that you disagree with the approach taken for the JSON file, but you've gotta keep comments civil. No one is going to respond well when you call their approach stupid.

conorgil:
@stephengroat Adding the gem for JSON linting makes lots of sense to catch any goofs that break the JSON. Good call.

I do agree with others in this thread that keeping the JSON file human readable re: indents and new lines is a good idea. Developers do read JSON even though it is a serialization format and not a presentation format. I use the JSON from this project in [2FA Notifier](https://github.com/conorgil/2fa-notifier) and read it manually when writing scripts to modify it. Given that the payload is compressed when sent over the wire, the only benefit I can see if the deserialization benefits that you mentioned.

Do you have any performance metrics to justify the change re: formatting?

stephengroat:
see https://github.com/2factorauth/twofactorauth/pull/3133#issuecomment-386348374, which talks about the unneeded RAM overhead that this approach incurs.

makes much more sense to view JSON during debug with a tool like https://codebeautify.org/jsonviewer rather than incur overhead on every single programatic call

conorgil:
@stephengroat yea, I saw that comment. That is what I was referencing when I said:
> the only benefit I can see if the deserialization benefits that you mentioned

Do you have actual metrics to measure how much of a performance benefit this yields? Of course, developers can just put the JSON into a pretty print, but if this change yields a negligible performance improvement, then I don't think it is worth making the change.

kenman345:
@conorgil You're right, its not helpful. But its still a silly method to try and circumvent spacing when the very thing you are using to circumvent spacing happening is also......get this.....able to natively be used to avoid spacing....

Also, this entire thing could be fixed by the simply renaming the `data.json` as it already stands to `data.big.json`. Then making a new file `data.json` and having it only include the relative path to the `data.big.json` file and use the native liquid filter to strip lines 

**Input**:
``` 
{{ "Hello
there" | strip_newlines }}
```

**Output**:
``` 
Hello there
```

conorgil:
@kenman345 notice that my comment didn't mention any of the technical arguments that you put forward. I was just pointing out that communication like that is rude and not welcome. Your call, but I'd shoot @stephengroat a quick apology and then focus on the technical merits of the discussion. You can absolutely 100% disagree with the proposed approach, but can communicate that in a constructive way.

stephengroat:
@conorgil specific performance metrics won't exist. for example, js is a garbage collected language. the RAM overhead created by the additional whitespace (~600k) won't be cleaned at deterministic times. could be around for 10 μs, or 10 minutes, depends on the specific code/implementation and the js engine

conorgil:
@stephengroat Gotcha.

I think there are tons of other low hanging fruit for optimizations on twofactorauth.org that will impact the end-user UX and are worth tackling before worrying about saving 600k of RAM.

I think we'll have to agree to disagree on this.

kenman345:
As I proposed, we may be able to offer an uncompressed version really
easily, and honestly, Firefox will expand compressed JSON into a tree
format automatically and extensions for chrome do the same. I think
compression is a good idea just the way it was done is just a way you would
do it if you didn’t know better, which is why I brought up the other
options.

Carlgo11:
@stephengroat how much of an impact are we talking about?
Would leaving this as it is mean a server could run out of memory just because it curls the data.json file or is this just a 'best practice' thing?

Carlgo11 (approved):
let's do it 🤷‍♂️

brycx (approved):

PR #3136, opened 2018-04-16
Add NextGenUpdate

ghost:


mxxcon:
As as general rule we don't list self-hosted software such as vBulletin here since each site that runs it has a lot of different ways to configure it.

stephengroat (dismissed):

stephengroat:
@mxxcon this isn't self hosted software though, this is a service which implemented the 2fa features of vBulletin. we have a lot of other services that offer hosted versions of their software which we've listed in the past (like Bugzilla@Mozilla)

brycx (changes requested):
The doc link for the 2FA process requires a login. Could you provide screenshots of the process behind enabling 2FA? Remember to redact any visible personal information. Thanks @Exodion.

Exodion:
@brycx Sorry for the late reply. Here is the main authentication page.
ngu 2fa 1

And these are the screenshots from all the options.

Yubikey:
ngu 2fa 2

Email:
ngu 2fa 3

Software Authenticator:
ngu 2fa 4

Hope everything is fine.

brycx (approved):

PR #3137, opened 2018-04-19
Add EveryCloud email spam filtering

AlisaBagrii:
Their webmail unfortunately does not provide 2FA. As far as I am aware the control panel also.


stephengroat:
@AlisaBagrii thanks for the contribution, but [the site doesn't seem popular enough yet to list](https://www.alexa.com/siteinfo/everycloudtech.com) (we try to target ~200k on alexa)

we'll close this for now, but if the situation changes, please let us know and we'll re-evaluate it then!

PR #3144, opened 2018-04-23
Add hardware information from dongleauth.info

alex-nitrokey:
Hi,

sorry, that this is a very big PR, but it contains basically notes regarding the hardware tokens based on dongleauth.info.

I would like to make both repos more up-to-date than it is now (our fault). 

I am quite aware of the [discussions](https://github.com/2factorauth/twofactorauth/issues?utf8=%E2%9C%93&q=dongleauth) around the hardware column. I totally don't want to start this again!

Instead I would like to ask, if it would be an option for you, to merge the "u2f" and "otp" notes without actually using them on twofactorauth.org. This would allow easier transfer between the repos. It would be easier for me to include your changes to dongleauth.info and to submit hardware changes here as well...

If you don't like this idea at all, please let me know, I'll try to find time to adapt PR accordingly.

Kind regards
Alex

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -3,6 +3,8 @@ websites:
     url: https://www.aerofs.com/
     tfa: Yes
     software: Yes
+    hardware: Yes
+    otp: Yes
     img: aerofs.png
     doc: https://support.aerofs.com/hc/en-us/articles/202610424-How-Do-I-Set-Up-Two-Factor-Authentication-
 
@@ -60,6 +62,8 @@ websites:
     sms: Yes
     software: Yes
     hardware: Yes
+    otp: Yes
+    u2f: Yes
     doc: https://www.dropbox.com/en/help/363
 
   - name: Evernote
@@ -68,6 +72,8 @@ websites:
     tfa: Yes
     sms: Yes
     software: Yes
+    hardware: Yes
+    otp: Yes
     doc: https://blog.evernote.com/blog/2013/10/04/two-step-verification-available-to-all-users/
 
   - name: FileThis
@@ -85,6 +91,8 @@ websites:
     phone: Yes
     software: Yes
     hardware: Yes
+    otp: Yes
+    u2f: Yes
     doc: https://www.google.com/intl/en-US/landing/2step/features.html
 
   - name: hubiC
@@ -143,6 +151,8 @@ websites:
     tfa: Yes
     sms: Yes
     software: Yes
+    hardware: Yes
+    otp: Yes
     doc: https://support.microsoft.com/en-us/help/12408/microsoft-account-about-two-step-verification
 
   - name: Pogoplug
@@ -158,6 +168,12 @@ websites:
     email: Yes
     software: Yes
     doc: https://www.qnap.com/i/useng/tutorial/con_show.php?op=showone&cid=151
+    
+  - name: Seafile
+    url: https://www.seafile.com/
+    twitter: seafile
+    img: seafile.png
+    tfa: No
 
   - name: SmartBox
     url: https://www.panterranetworks.com/products/smartbox.php
@@ -196,6 +212,8 @@ websites:
     tfa: Yes
     software: Yes
     email: Yes
+    hardware: Yes
+    otp: Yes
     doc: https://originwww.synology.com/en-us/knowledgebase/tutorials/615#t5
 
   - name: ThisData
@@ -214,6 +232,8 @@ websites:
     phone: Yes
     software: Yes
     email: Yes
+    hardware: Yes
+    otp: Yes
     doc: https://support.tresorit.com/entries/104192996-How-to-set-up-two-step-verification
 
   - name: Yandex.Disk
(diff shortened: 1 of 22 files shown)


mxxcon:
There are a couple of issues with this PR.

- You are adding a new field to our yamls. At this point in time I have no opinion on this. We would need to think and discuss this.

- You are adding new sites, which we'd want to evaluate if they fit with our criteria.

- You are adding a few new sites which do not include logos and they are required.

- I did just a quick spot check on some of the sites you are modifying and I'm not sure how accurate that data is. For example in `_data/task.yml` you are modifying Trello entry to say that it supports hardware 2fa, but looking at https://help.trello.com/article/993-enabling-two-factor-authentication-for-your-trello-account I do not see that.
So either your data is inaccurate or we have different understanding/definition of hardware 2fa...

alex-nitrokey:
> You are adding a few new sites which do not include logos and they are required.

Yes, already realized, sorry for that! Will fix this!

> You are adding new sites, which we'd want to evaluate if they fit with our criteria.

Do you prefer PRs for every new site?

> I did just a quick spot check on some of the sites you are modifying and I'm not sure how accurate that data is. For example in _data/task.yml you are modifying Trello entry to say that it supports hardware 2fa, but looking at https://help.trello.com/article/993-enabling-two-factor-authentication-for-your-trello-account I do not see that.

I used the information I had. So may services that used to have hardware token support, did stop this. Trello was added within the last year, so I guessed that's still true. But I will have a look again.

Please note, that every service which uses OTP as authentification (though advertised as "Google Authenticator") can be used with hardware tokens, which support OTP. Therefore some services may look odd first. Trello (at least did) support OTP.

mxxcon:
>Please note, that every service which uses OTP as authentification (though advertised as "Google Authenticator") can be used with hardware tokens, which support OTP.

While that's true, I think that's kinda defeats the purpose of having 2 distinct categories of `software` and `hardware`. Every software runs on hardware and every hardware needs software.

RichJeanes:
I have to agree with @mxxcon on that software/hardware point. The line has to be drawn somewhere. It's kind of the same issue as sites that support OAuth are not listed as supporting 2FA since it is not a native feature.
While the feat is technically achievable, it is not a native feature of the site. If you're targeting an audience that can understand your distinction you have made (hardware can store TOTP, therefore all sites that support TOTP support hardware), then go for it. But I don't think we necessarily target the same audience. Or at least we make different assumptions about our audience.

(And, personally, as a YubiKey owner myself, I know that it can store TOTP codes, but only a limited number (a fact that they kind of bury on their site...). Therefore, if a site supports a native method of hardware authentication (eg. U2F), I would much prefer that method and so finding that a site supports native hardware 2FA is more useful. TLDR: I find distinguishing between native hardware auth and software-on-hardware auth to be important.)

alex-nitrokey:
Hi,
thanks a lot for your comments!

> Therefore, if a site supports a native method of hardware authentication (eg. U2F), I would much prefer that method and so finding that a site supports native hardware 2FA is more useful.

Does that mean, that "software" mark means necessarily TOTP and "hardware" is necessarily U2F? This would make it easy, to adapt for us. But I guess, there are more software and more hardware options out there and used here :thinking: 

That means one can not know, what exactly is supported by the overview? Or do you have a definition table (I couldn't find one, I am sorry, if it is obvious :smile: ).

alex-nitrokey:
Mh, thinking about it, I guess proprietary software and hardware solutions will fit in software and hardware as well... Okay, I see the problem...

So to conclude: If we mark something as u2f, it will qualify for hardware, if we mark something otp it will qualify for software, but we can not do this the other way around? Okay, then I guess this is not really compatible :cry: Or is there any idea to solve this?

mxxcon:
> That means one can not know, what exactly is supported by the overview? Or do you have a definition table (I couldn't find one, I am sorry, if it is obvious 😄 ).

Earlier iteration of our site used to list specific 2fa implementation types, however we intentionally switched to the current 5 generic broad categories to simplify maintainability and readability of the site.

Having said that, if I remember correctly we also had an idea of somehow indicating sites which support fido/u2f, but at this point as you can see nothing was done about it.

RichJeanes:
Some examples we've seen include:
Software:
- TOTP/HOTP app
- Proprietary app (OTP and/or push notification)

Hardware:
- U2F tokens
- Proprietary/dedicated (T/H)OTP tokens
- Printable letter matrix
- Smart card (sometimes in banking this is a user's debit card)

mxxcon:
Also biometrics I'd consider "hardware".

RichJeanes:
I mean, I've always considered "Hardware" (and "Software") to fall into the "Something you have" factor while biometrics are in the "Something you are" factor. I mean, yes, you need specific hardware to use biometrics, but I would not label it as a "hardware" authentication method.

alex-nitrokey:
Okay, as far as I can see our hardware differentiation won't fit in this logic, I totally understand your decisions!
I will close this PR and try to open a new one with some sites we list, but 2fa don't and a separate PR for some minor changes we did (like better doc links or similar).

PR #3149, opened 2018-04-24
New Whatsapp logo

thibaultmol:
small change, but the gradient is less noticeable than before
has run through optipng


mxxcon:
Consider using this file, it's smaller
whatsapp

mxxcon (approved):

PR #3150, opened 2018-04-24
New Viber logo

thibaultmol:
has run through optipng


mxxcon (commented):
Their favicon and logo seem to be just a purple headset, without the speech bubble

mxxcon (changes requested):
logos must be 32x32

mxxcon (approved):

PR #3159, opened 2018-04-24
Change planethoster domain TLD

NormandLemay:
Planethoster.net change is own primary domain to planethoster.com

### DIFF
--- a/_data/hosting.yml
+++ b/_data/hosting.yml
@@ -313,11 +313,11 @@ websites:
       hardware: Yes
 
     - name: PlanetHoster
-      url: https://planethoster.net
+      url: https://planethoster.com
       img: PlanetHoster.png
       tfa: Yes
       software: Yes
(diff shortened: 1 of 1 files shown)


mxxcon (dismissed):

mxxcon (commented):
@NormandLemay Should the urls be updated to point to English language versions or add `lang: fr`?

NormandLemay:
I add the lang: fr

mxxcon (changes requested):
Actually, I'm a dummy and don't remember our own syntax. 😅
`lang` is needed only if `tfa:no` so that template messages are in proper language. Since this site does support tfa, `lang` is not necessary.
I'm sorry for this confusion.

NormandLemay:
It's correct push again without the lang

mxxcon (approved):

PR #3165, opened 2018-04-29
Update IRS to include SMS and software 2FA

### LINKED ISSUE #3164: IRS supports two-factor auth
https://www.irs.gov/newsroom/irs-launches-more-rigorous-eauthentication-process-and-get-transcript-online: the U.S. internal revenue service's "get transcript" site now requires two-factor auth to sign in.

Supported methods:
- SMS codes
- security code from phone app they provide (IRS2Go: https://www.irs.gov/newsroom/irs2goapp)

Sadly, they don't appear to support U2F.

brycx:
Fixes #3164.

### DIFF
--- a/_data/government.yml
+++ b/_data/government.yml
@@ -52,7 +52,7 @@ websites:
       twitter: dgfip_officiel
       facebook: Direction-générale-des-Finances-publiques-250996308359760
       lang: fr
-      
+
     - name: login.gov
       url: https://login.gov/
       img: login-gov.png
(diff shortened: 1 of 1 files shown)


Carlgo11:
From what I can tell it only protects certain pages. Is this correct? (*Paging any 'mericans*)

mxxcon:
I don't have account with them, but it does look like only some pages are protected

> Currently, the Secure Access authentication process applies to Get Transcript Online, Get an IP PIN, e-Services and online tax account.

But then how does that work? It asks for login multiple times? 🤔

Carlgo11:
If only certain pages are protected, does that fall under two factor **authorization** then?

brycx:
Maybe @jrn from #3164 can shed some light upon this?

jrn:
The IRS is a large agency and probably they have multiple services online.

The only one that matters is Get Transcript, which is protected by two-factor.

mxxcon (approved):

Carlgo11 (approved):

PR #3167, opened 2018-04-29
Update semantic-ui

stephengroat:


_layouts/secondary.html
```diff
@@ -66,7 +66,7 @@
 	integrity="sha384-xBuQ/xzmlsLoJpyjoggmTEz8OWUFM0/RC5BsqQBDX2v5cMvDHcMakNTNrHIW2I5f"
 	crossorigin="anonymous"></script>
 <script src="https://cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.2.13/semantic.min.js"
```
mxxcon: I think you forgot to update version here

kenman345 (changes requested):
https://github.com/2factorauth/twofactorauth/pull/3167/files#diff-f96e55fbd5421f2438a0dbfe7df336e2R19

I think since Semantic 2.3.1 changed `bitcoin` icon to have a circle backdrop, that you would like to instead have the icon be `btc` which is the same as what `bitcoin` looked like in 2.2.7

stephengroat:
@kenman345 great catch!

i actually liked the change to something new, anyone have an opinion on it?

kenman345:
I recommend making my suggested change. the `btc` icon is the one meant to indicate a currency, and not a specific implementation of bitcoin. It makes more sense as a crypto based section to be a bit more agnostic and more towards a symbol recognized for the space youre hoping to cover with that category.

stephengroat:
> the btc icon is the one meant to indicate a currency, and not a specific implementation of bitcoin

are you sure that's right? from googling around, i'm not 100% sure. both seem to stand for bitcoin, it comes down to preference

https://en.bitcoin.it/wiki/Bitcoin_symbol

https://en.wikipedia.org/wiki/Bitcoin

kenman345:
among other reasons, the circle around the icon makes it messier for your site.

I obviously have an opinion on this as I help maintain acceptbitcoin.cash....but you're free to do as you wish, both have their merits.

mxxcon:
I agree with Ken here. 
We'd want to use `btc` icon ₿ as a generic "cryptocurrency" image rather than a specific "bitcoin"-brand icon.

kenman345:
Side note: This PR brings up a glaring issue with your Jekyll setup. Why does each and every notes page have to be adjusted? Why is this not templated as a template or inside an `_include`? 

If you'd like I can take a look at that. Its probably not going to result in smaller page sizes but it should bee able to reduce redundant styling across pages so its easier to make the next upgrade to a newer version of Semantics UI in the future. 

Such a change would best be done once this is merged in so I have these changes to work off of.

stephengroat:
agreed, it's a good point, but let's tackle stuff one pr at a time.

kenman345:
Would love if this got merged in so I can make a PR to consolidate the layouts. This change has already been made on my fork of the site and working wonderfully. I dont foresee this being an issue to your site at all.

fpigerre (approved):

kenman345:
can we merge this? Its been working just fine for a long time on my project, making the same changes you have here. Its also delaying some changes to the notes layout to make it easier to update in the future.

jamcat22 (approved):

jamcat22:
Sorry @kenman345. Didn’t see that @psgs already reviewed the changes.

PR #3168, opened 2018-04-30
method refactor of rakefile

kenman345:
I noticed in my own repo that it makes a bit more sense to call HTMLProofer from a single place, and the overlapping settings be configured as defaults. This makes it a bit easier to manage if you add to your configuration, but also to understand what the expected difference between tests is without having to mentally keep track of what is common between them.

I threw in a small change to move the require for the HTML Proofer to the method itself. By doing this, it gets loaded only if that method gets called, which means rake tasks will not fail for windows machines that commonly do not have libcurl properly installed, making your rake tasks fail because it cannot load up the HTML Proofer library properly even though it installed the ruby library, because the ruby library has a dependency on something annoying and likely not configured for windows based machines. The change however does not any other effect on its ability to run, so its a rather safe change

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,10 +3,12 @@ require 'jekyll'
 
 task default: %w[proof verify rubocop]
 
+dest = './_site'.freeze
+
 task :build do
   config = Jekyll.configuration(
     'source' => './',
-    'destination' => './_site'
+    'destination' => dest
   )
   site = Jekyll::Site.new(config)
   Jekyll::Commands::Build.build site, config
@@ -33,10 +35,9 @@ RuboCop::RakeTask.new
 
 def check_site(options = {})
   require 'html-proofer'
-  dir = './_site'
   defaults = {
     assume_extension: true,
     check_html: true
   }
-  HTMLProofer.check_directory(dir, defaults.merge(options)).run
+  HTMLProofer.check_directory(dest, defaults.merge(options)).run
 end
--- a/Rakefile
+++ b/Rakefile
@@ -1,4 +1,3 @@
-require 'html-proofer'
 require 'rubocop/rake_task'
 require 'jekyll'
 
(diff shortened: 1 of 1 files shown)


stephengroat:
@kenman345 after looking into a bit, this seems like a bad idea https://kevinjalbert.com/defined_methods-in-rake-tasks-you-re-gonna-have-a-bad-time/

i illustrated the point with a code change

kenman345:
youre right about methods in rake files but considering you’re not about to write a ton of methods and tasks to this project anytime soon, it’s fine if you add one method and it’s all still inside one file. 

In the meantime you may decide to enable Cors SRI check for HTML Proofer then you can easily and keep it understandable between where it is and if it’s running on which check. Or if Html proofer adds in new features you wish to use, this organization of methods helps keep that consistency I mention in my original post.

Happy to try and organize this to a class instead though but I think that for such a small repository this would be over complicating the whole process.

stephengroat:
@kenman345 it's a valid point, i'm just not sure saving ~3 lines is worth creating a potential future complication that's hard to debug

there's other potential ways too, like rake tasks with arguments, etc. might be a better option?

stephengroat:
@kenman345 what about a rake task like this? i think it's more `rake`d (like pythonic? i'm trying to coin a phrase!)

```ruby
task :proof, %i[target opts] => 'build' do |_t, args|
  args.with_defaults(target: './_site', opts: '{}')
  opts = { assume_extension: true, \
           check_html: true, \
           disable_external: true, \
           cache: { timeframe: '1w' }, \
           check_sri: true, \
           url_ignore: [
             # see https://github.com/google/fonts/issues/473#issuecomment-331329601
             %r{https:\/\/fonts.googleapis.com\/css\/*}
           ] }.merge!(YAML.safe_load(args.opts, [Symbol]))
  HTMLProofer.check_directory(args.target, opts).run
end
```

it can be called with `bundle exec rake proof['./_site',':disable_external: false']` to override the default `disable_external: true`

PR #3172, opened 2018-05-01
Use html proofer to check SRI

stephengroat:


mxxcon:
What is the purpose of this check?
I understand what it does, but why would we want to do it?

stephengroat:
to ensure that all external elements use sri, preventing possible exploitation via third party imports

mxxcon:
Ah, so it's checking if SRI is present for all external resources and not if existing signatures are valid?

stephengroat:
According to html proofer docs, yup

kenman345:
I honestly feel a bit on ends about this, maybe someone will clarify.

Would we not want to ensure that the google urs for fonts are valid endpoints? With this change you're ignoring those urls because SRI but now we dont check them when we maybe add one and its not valid. 

should we perhaps add an additional check to only test those URLS or just check SRI only in a test and have those urls ignored only in that?

mxxcon:
Can't do SRI with google fonts since they are dynamically generated so hash will always change/always fail.

kenman345:
I understand that 100%, I mean that this url_ignore component means we are not just ignoring it from SRI check but the external URL check to ensure its not an invalid path, which if you change/add in more fonts is something we might have happen.

with the change as it stands, its not going to catch such a mistake.

stephengroat:
to me this sounds like a deficiency with `html-proofer`, maybe we should just try to make the change there

kenman345:
looking at the documentation, they might say write your own HTML-Proofer test to swallow those special case urls or make two separate tests.

https://github.com/gjtorikian/html-proofer#custom-tests

stephengroat:
do you want to take a stab at it?

kenman345:
hell no, I purposely havent added that check to my repo because thats a pain in the butt. lol. I was just making sure you guys were aware

stephengroat:
still loads/presents even if that font is excluded, so i'm not too worried about it

kenman345:
Okay, one last thing then if you're okay with that.

So I do not understand some of the importance in checking the SRI thing and both tests failing from Travis if you have the SRI check in both tests. In my repo, we've removed the HTML check out of the external check since its redundant and external already will take a long time. Maybe only add this SRI check to one of the tests so its still getting checked but doesnt need to run for both when we expect them both to run on anything that would ever hit the `master` branch in the repository.

stephengroat:
i think the eventual idea is to converge on a single check with both sri and external link validity. SRI seems to have a higher importance/precedence (since it deals with the integrity of the page) rather than if a link is valid

stephengroat:
closing in favor of #3420

PR #3183, opened 2018-05-07
Add lichess.org

niklasf:
As of today lichess.org supports two-factor authentication using TOTP.

There is no documentation other than the page where it is configured, https://lichess.org/account/twofactor (requires logged in account).


Carlgo11:
Could you provide any screenshots (with personal information blurred) of either the login or setup of 2FA?

niklasf:
Sure.

image

(Obviously I won't be using this QR code. Also it reveals my username, but that is publicly known already.)

After entering correct username and password:

image

Carlgo11 (approved):
Great. Thanks! :smiley:

PR #3188, opened 2018-05-10
remove use of .each

kenman345:
the use of `.each` is nice because it removes nulls but is not actually needed here. Instead, its cheaper/faster to loop through elements this way and get a small performance increase when performing a search.


fpigerre (approved):

kenman345:
this is approved, why not merged when so many other PRs were merged?

dsoegijono (approved):

PR #3131, opened 2018-04-13
Remove all sites over 200k alexa in security

stephengroat:

PR #3192, opened 2018-05-13
Remove Coin Cafe

spazvt:
Reviewing Alexa traffic stats, I see that coincafe.com is 2.4M, far beyond the 2FA site threshold of 200k. Even during the bitcoin craze in fall 2017, its highest rank was ~600k.
https://www.alexa.com/siteinfo/coincafe.com

Although the website does appear to still be working, I've never seen it mentioned in crypto forums and with such light traffic I'd recommend removing per the 2FA site policies.


spazvt:
per #3131 mods have decided to not remove sites with low traffic.

PR #3202, opened 2018-05-17
Remove Bank of America hardware token support

### LINKED ISSUE #3185: BoA doesn't seem to support hardware token
Following the documentation trail for Bank of America, I could not find any place describing the use of a hardware token.

antsar:
Bank of America no longer supports using SafePass hardware tokens for two-factor auth. The tokens are only used for authorizing high-value transactions, but not login. Fixes #3185.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -94,8 +94,6 @@ websites:
       tfa: Yes
       sms: Yes
       email: Yes
-      hardware: Yes
-      doc: https://www.bankofamerica.com/privacy/online-mobile-banking-privacy/safepass.go
 
     - name: Bank of China (Hong Kong)
       url: http://www.bochk.com


antsar:
Hi, just wondering if this PR is missing something, or if there's anything else I can do to get it merged? Seems silly to give BofA free marketing for a feature they don't actually provide.

mxxcon:
@antsar do they support still support email 2fa or is it now only sms?

antsar:
Yes, they support both email and SMS. Hardware token support was removed.

screenshot from 2018-07-10 12-52-18

mxxcon (approved):

RichJeanes (approved):
*Le sigh* Always sad to see regression.

PR #3206, opened 2018-05-17
Added MeeWee TFA

luposlip:


_data/other.yml
```diff
@@ -330,6 +330,13 @@ websites:
       software: Yes
       doc: https://www.mathworks.com/mw_account/two_step_verification/frequently-asked-questions.html
 
+    - name: MeeWee
+      url: https://app.meewee.com/
+      img: meewee.png
+      tfa: Yes
+      Phone: Yes
```
mxxcon, line 337: `phone` must be lower case.

mxxcon (changes requested):
Is there doc link describing which methods they support?

mxxcon:
And actually looking at https://www.alexa.com/siteinfo/meewee.com unfortunately this site ranks too low for our top 200k cutoff.

PR #3207, opened 2018-05-17
Add Moneybird

rvoortman:
Moneybird is a Dutch bookkeeping software, used by over 100.000 entrepreneurs and small companies. They support 2FA with TTOP (Google Authenticator). The link is the blog article which explains how it is used.


Carlgo11 (approved):
So we normally don't add sites ranked below 200k on Alexa.com just so that we don't have to maintain a list of every single site on the internet, but Moneybird seems like a large enough site to add.
If a hundred thousand customers is right then we certainly should list the site.

Thanks for the contribution! 👍

rvoortman:
@Carlgo11 I'm sorry,  I wasn't aware of that. Moneybird is currently a fastgrowing bookkeeping software, so I thought it would fit in the twofactorauth list. :)

brycx (approved):

PR #3212, opened 2018-05-22
Updated SalesForce to replace dead link, add hardware support

jadchaar:
Updated SalesForce entry to replace the dead doc link with an updated version. Added hardware support.

### DIFF
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -163,7 +163,7 @@ websites:
       software: Yes
       email: Yes
       hardware: Yes
-      doc: https://help.salesforce.com/articleView?id=security_overview_2fa.htm&type=5
+      doc: https://developer.salesforce.com/docs/atlas.en-us.identityImplGuide.meta/identityImplGuide/add_time-based_token.htm
 
     - name: SendGrid
       url: https://sendgrid.com/
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -162,7 +162,8 @@ websites:
       sms: Yes
       software: Yes
       email: Yes
-      doc: https://help.salesforce.com/HTViewHelpDoc?id=security_require_two_factor_authentication.htm
+      hardware: Yes
+      doc: https://help.salesforce.com/articleView?id=security_overview_2fa.htm&type=5
 
     - name: SendGrid
       url: https://sendgrid.com/


_data/communication.yml
```diff
@@ -162,7 +162,8 @@ websites:
       sms: Yes
       software: Yes
       email: Yes
-      doc: https://help.salesforce.com/HTViewHelpDoc?id=security_require_two_factor_authentication.htm
+      hardware: Yes
+      doc: https://help.salesforce.com/articleView?id=security_overview_2fa.htm&type=5
```
mxxcon: Is there a better link for end-users rather than admins? Perhaps https://trust.salesforce.com/en/security/2fa/ ?

jadchaar: Good point. My only problem with the article you linked is that it seems to only promote their proprietary application, rather than an authenticator app in general. Thoughts on that?

jadchaar: The only other relevant link I can find is a developer page, but that is even more cryptic. A bunch of the others just redirect to the admin page.

mxxcon: I personally don't have too much of a problem with them promoting it. People who know the difference will be able to figure it out on their own. Otherwise, it should simplify support issues for Salesforce.

jadchaar: Hi @mxxcon I think I found a good one: https://developer.salesforce.com/docs/atlas.en-us.identityImplGuide.meta/identityImplGuide/add_time-based_token.htm

mxxcon: this link looks better 👍

_data/communication.yml
```diff
@@ -162,7 +162,8 @@ websites:
       sms: Yes
       software: Yes
       email: Yes
-      doc: https://help.salesforce.com/HTViewHelpDoc?id=security_require_two_factor_authentication.htm
+      hardware: Yes
```
mxxcon, line 165: wait a sec, why did you add hardware? we consider mobile authenticator apps to still be "software".
"hardware" are physical fobs or yubikey and similar type devices

Carlgo11: They support U2F

mxxcon (approved):

Carlgo11 (approved):

_data/communication.yml
jadchaar: Yup I saw on that admin page that they supported U2F :)

PR #3214, opened 2018-05-23
Add Atlassian Cloud

ondrejfuhrer:


mxxcon:
We already have Bitbucket and Hipchat listed. Perhaps update all those entries and add them here?

jamcat22:
@mxxcon Bitbucket still has its own 2FA system in use even with Atlassian’s 2FA. Should we keep that entry the same and just update HipChat? Or maybe add an exception?

mxxcon:
🤦If they migrated everything to Atlassian Id accounts, why do they still have separate settings per app...
I'm leaning more towards combining everything into one entry and adding exception text something along the lines of "some 2fa settings are service-specific, not account wide" or some other better wording.
Interestingly, when I went to hipchat, it redirected me to their new Stride service.

jamcat22:
I just enabled Atlassian ID 2FA and it didn't ask me for that code when logging in to Bitbucket. I don't think they're using the same sign on page, just using the same account storage... 😞

ondrejfuhrer:
I guess they have some strange mess there.
Is it possible to have the atlassian cloud entry there and link all the other services (searches on the page) with that? Or you think better to have separate entries for all services?

mxxcon:
We have a couple of open issues that discuss potential approaches to handle such situations where we'd have some metadata which would list underlying services(such all the google products under google entry or all the EA online games under main EA entry). That could potentially help with search/discoverability. If you have ideas on this topic, you are welcome to comment there.

As for this specific entry, I personally think it's better if we have a single Atlassian entry. I think it makes it easier to manage than a bunch of separate records. But I won't block this PR if we keep them separate either.

jamcat22:
I vote to at least keep Bitbucket separate because Atlassian 2FA doesn’t affect Bitbucket’s login, and vice versa. We can remove HipChat if we want since they’re transitioning away from that product.

ondrejfuhrer:
@mxxcon It would be great (seems like you are aware of those) to link all the issues you were talking about together to see all the issues / discussions, so we could think about solving the problem?

My guess, for start, would be to have something like:
```yaml
- name: Atlassian Cloud
  ...
   services:
      - name: Jira
         url: jira.atlassian.net
      - name: Confluence
         url: confluence.atlassian.net
      - name: Bamboo
         url: bamboo.atlassian.net
```

And then when you search it will go through the services as well

ondrejfuhrer:
Hello all, is there any conclusion what we will do with this PR?

mxxcon:
@ondrejfuhrer lets add exception text that 2fa settings are service-specific.

ondrejfuhrer:
@mxxcon Done 👍

mxxcon (approved):

mxxcon (changes requested):
I'm sorry I didn't spot it sooner, but your logo image looks kinda pixely.
Let's use their official file https://atlassian.design/favicon.png 👍

ondrejfuhrer:
@mxxcon Sure, image updated! 👍

mxxcon (approved):

ondrejfuhrer:
Any other reviewer, i.e. @jamcat22 ?

jamcat22 (approved):
Looks good! Thanks @ondrejfuhrer!

PR #3215, opened 2018-05-24
Reverts the value of all doc fields to URLs

conorgil:
# Description
Some of the service entries contain a relative path in
their doc field that points to a local page on
twofactorauth.org. This paradigm is followed for sites
that support 2FA, but do not have good documentation articles
on their websites.

The problem that this commit addresses is for external tools
and services that rely on the doc field to present the user
with a URL to access the 2FA documentation for a service.

For example:
- 1Password
- 2fanotifier.org

These tools will have to specifically address the scenarios
where the doc field is a relative path instead of a full URL.
Updating the source data allows third parties to use the data
without addressing edge cases.

# Context

This change was originally made in  146fa92b79a98f5dbf7b4900bb7a3666390cd989

# Test results

```
$> bundle exec rake
Configuration file: ./_config.yml
            Source: ./
       Destination: ./_site
 Incremental build: disabled. Enable with --incremental
      Generating...
                    done in 2.244 seconds.
Running ["ImageCheck", "ScriptCheck", "LinkCheck", "HtmlCheck"] on ["./_site"] on *.html...


Ran on 8 files!


HTML-Proofer finished successfully.
.rbenv/versions/2.5.0/bin/ruby ./verify.rb
<------------ No errors. You're good to go! ------------>
Running RuboCop...
Inspecting 3 files
...

3 files inspected, no offenses detected
```

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -208,7 +208,7 @@ websites:
       sms: Yes
       phone: Yes
       email: Yes
-      doc: /notes/chase/
+      doc: https://twofactorauth.org/notes/chase/
 
     - name: Chime
       url: https://www.chimebank.com
--- a/_data/domains.yml
+++ b/_data/domains.yml
@@ -258,7 +258,7 @@ websites:
       img: internetbs.png
       tfa: Yes
       software: Yes
-      doc: /notes/internetbs/
+      doc: http://twofactorauth.org/notes/internetbs/
 
     - name: INWX
       url: https://www.inwx.com/en/
--- a/_data/domains.yml
+++ b/_data/domains.yml
@@ -258,7 +258,7 @@ websites:
       img: internetbs.png
       tfa: Yes
       software: Yes
-      doc: http://twofactorauth.org/notes/internetbs/
+      doc: https://twofactorauth.org/notes/internetbs/
 
     - name: INWX
       url: https://www.inwx.com/en/
--- a/_data/gaming.yml
+++ b/_data/gaming.yml
@@ -131,7 +131,7 @@ websites:
       img: gree.png
       tfa: Yes
       software: Yes
-      doc: /notes/gree/
+      doc: https://twofactorauth.org/notes/gree/
 
     - name: Guild Wars 2
       url: https://www.guildwars2.com
--- a/_data/investing.yml
+++ b/_data/investing.yml
@@ -50,7 +50,7 @@ websites:
       hardware: Yes
       exceptions:
           text: "Only available for some account types"
-      doc: /notes/fidelity/
+      doc: https://twofactorauth.org/notes/fidelity/
 
     - name: Flatex
       url: https://www.flatex.de/


Carlgo11:
Doing a hard linking to this would mean forks of the site and local testing versions would link to our site instead of the correct, local, path.
I think it would be better to change the URL on the third party clients instead of in the YML files.
Thoughts on that?

Also, you forgot the HTTPS for the Internetbs URL 😉

kenman345:
This change is addressing the problem indicated in the OP but not the secondary issue @Carlgo11 brings up. Instead, I believe both those services would use the `data.json` file generated when the site compiled by Jekyll. Perhaps you should add such a transform to that file instaead.

The change would simply require adding an if-else positive-negative output of the doc tag. The `if` would be for when the `doc` key value starts with `/` and then it would add in the url from the `_config.yml` which I believe is available at `site.url`. The `else` could leave the value unchanged. 

The end result is we give the `data.json` output the proper data it needs to be useful for multiple services and local testing would be unaffected.

kenman345:
Also of note: https://github.com/2factorauth/twofactorauth/blob/master/websites_schema.yml#L46

The schema indicates we do not require a URL in the docs field for it to be valid. Making a change to make those fields URLs would make its utility different than how its allowed to be used now. If it should only ever be a URL, we should also adjust the schema to ensure that consistency.

conorgil:
@Carlgo11 Fair point on an update like this impacting forks and local testing.

When you say local testing, do you mean manual testing? I ran the automated tests and there were no errors, warnings, etc. Do the automated tests need to be updated to handle this scenario?

I was thinking of the same solution proposed by @kenman345. As the maintainer of [2FA Notifier](https://github.com/conorgil/2fa-notifier), my main focus is the URLs in the `data.json` file. Updating the script that generates the `data.json` file to handle this scenario sounds like a great approach to me.

I am happy to make those changes, but I couldn't find the script where the data.json file is generated. Can you point me to it?

@kenman345 I agree that the schema file should be updated if the field should only contain URLs. Can you point me to where that schema is actually used to validate data? Thanks!

kenman345:
data.json is generated from the root level `data.json` file. [here](https://github.com/2factorauth/twofactorauth/blob/master/data.json) 

Its not generated through a special compiler other than Jekyll. I believe though making edits might not get merged and have to be redone do to #3133 still not being merged yet. I would say you should wait for that PR to get merged (insert rant about how long it takes to merged approved PRs on this repo here) before making additional changes.

the schema is used in the root level `verify.rb` script. It can be found [here](https://github.com/2factorauth/twofactorauth/blob/master/verify.rb).

The `verify.rb` script is run against every PR through Travis-CI which will run it via rake tasks.

Carlgo11:
Fixed in b3f79ac.
Still want this merged?

conorgil:
@Carlgo11 I glanced at https://github.com/2factorauth/twofactorauth/commit/b3f79ac96060909087935fd824475e280755280b super quickly and it looks like the fix in there is good (prepend twofactorauth.org to entries with `/notes`).

I'll close this PR since it is a manual fix and your fix is automated and should catch these same cases.

Carlgo11:
Great!
I estimate that the changes are all done within a week and hopefully it won’t take too long to get it merged after that 🤞

PR #3219, opened 2018-05-25
Add Joomla

itxchandan:


thibaultmol:
@chandanndeep I think it's good, only need to remove joomla.png from the domains folder and Travis will give a green light 👍

mxxcon:
Thank you for this PR, however at this point in time we do not list self-hosted applications.

kilasuit:
I think that's a seriously silly **IDEA**

This site should enable people to see both SaaS and Self hosted applications that allow **ANY** form of MFA to be enabled - please rethink this @mxxcon

mxxcon:
@kilasuit We do list saas. We have salesforce and other products there.
I addressed your other point in the issue you opened.

itxchandan:
I removed the logo from the domains folder. Waiting from your side

PR #3221, opened 2018-05-25
Add Google Hangouts and Digidentity

alex-nitrokey:
This is included on https://www.dongleauth.info and seems to fit in here as well.


brycx (changes requested):
Thanks for the PR @alex-nitrokey. There are two points which need to be addressed:
- Google Hangouts also supports phone and sms for 2FA, so `phone: Yes` and `sms: Yes` needs to be added.
- Due to a limit where a new site needs to be within Alexa top 200k, Digidentity can unfortunately [not be added](https://www.alexa.com/siteinfo/digidentity.eu). This needs to be removed.

The rest looks fine to me.

alex-nitrokey:
> Due to a limit where a new site needs to be within Alexa top 200k, Digidentity can unfortunately not be added. This needs to be removed.

Didn't realized this limitation, good to know.

I adapted accordingly.

brycx (approved):

RichJeanes (approved):

PR #3224, opened 2018-05-27
Add platform.sh

bartlangelaan:
This adds platform.sh as a 2FA supported hosting provider.


phillipplum:
Maybe you should use this link for documentation? https://docs.platform.sh/development/faq.html#do-you-support-two-factor-authentication

bartlangelaan:
I've added the link to the documentation now! Thanks for the suggestion.

brycx:
Thank you for the PR @bartlangelaan. There currently is a limit for new sites, which should be within the Alexa top 200k. Unfortunately this site is [not](https://www.alexa.com/siteinfo/platform.sh), so I'm closing this PR.

PR #3225, opened 2018-05-27
Add exception to Skype

jadchaar:
Since a Microsoft account is needed to enable 2FA (according to the doc), old accounts that were created with a username (and not MS Account) will not have the ability to do this.

# Discussion
I was thinking about how to approach this. I happen to have an old Skype account (pre MS acquisition) and 1Password 7 tells me that Skype supports 2FA, but I do not have it enabled. I tried to figure out how to enable it, and found out that my old Skype account is not associated with my MS Account. Thus, I cannot enable it. I feel like the Skype entry as a whole may be unnecessary and should be removed as Skype is now under the "Microsoft Account" umbrella, and locking down the security of that encompasses Skype. It is a bit misleading to tell people with old Skype accounts that are not associated with a MS Account that 2FA does not work.

Let me know and I will adjust the PR.

### DIFF
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -195,7 +195,7 @@ websites:
       email: Yes
       sms: Yes
       software: Yes
-      exceptions: Yes
+      exceptions:
           text: "Must be associated with a Microsoft Account."
       doc: https://support.microsoft.com/en-us/help/12408/microsoft-account-about-two-step-verification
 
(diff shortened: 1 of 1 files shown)


mxxcon:
Is there a reason why one would not want to update their skype account in this manner?
I think skype is a significant enough service that it'd warrant its own entry...at least until we find a way to better display related services.

Carlgo11 (approved):

brycx (approved):

Issue #3222, opened 2018-05-26
New tags for TOTP and U2F?


Carlgo11:
Hello,
we've seen several people of the past year requesting a way to check if a site supports TOTP([RFC #6238](https://tools.ietf.org/html/rfc6238))(commonly referred to as Google Authenticatior) or if they require proprietary 2FA software.
We've also seen quite a few people requesting a distinction between U2F hardware and other hardware factors.

There are several ways we could implement this.
### 1.
Use the existing `software` and `hardware` tags to specify this.
Such as `software: TOTP` for TOTP authentication and keeping the `software: Yes` for non-TOTP software.
The advantage here is that old contributions don't need to be changed for this to work and so the sites supporting TOTP could over time have their status updated. Furthermore it would keep the length of the YAML files as they are right now.

A disadvantage here is that our current way of allowing third parties to access our data (via data.json) would require them to update their fetching of the data or it could potentially break them.

As of right now there are very few third parties using our data and so making them all implement the changes would probably not be very hard. For the future however we would need some sort of versioning system of the data.json to eliminate this risk.

### 2.
Building on the existing `software` and `hardware` tags, we could do like with the `exception` tag and make it an array. That could allow for a more future-proof setup with multiple choices.
The downside here is that it would require lots more work as the code need to be changed to allow for an array and all existing sites listed here that use the `software` or `hardware` tags would need to be changed.
 
### 3. 
Another way is to simply add a `totp` and `u2f` tag to the YAML parsing.
This would mean less work for the third parties but larger file sizes for us.
<br><br>

Technically speaking, how much work would this mean? (pinging @psgs for UI and @stephengroat for Travis-CI/script changes)
I think the best approach would be to first implement this in data.json and then do the UI changes.
That way there's fewer ways the change could go wrong.

I'd love to hear some feedback from @2factorauth/collaborators along with any third party maintainers (@conorgil @jadchaar)

jadchaar:
It shouldn't be too hard to adjust parsing scripts to the new data format. As long as the changes are beneficial and add useful data points, it will be well worth it.

One thing I wanted to mention is that if you guys decide to go for the array approach for `software` and `hardware` tags, you should add a `duo` option (e.g. support Duo push). Also, since I see that SendGrid uses Authy 7 digit codes, it may be useful to have an `authy` option as well.

jadchaar:
[...]

PR #3231, opened 2018-05-30
Added exceptions for QNAP and Synology

jadchaar:
We got some reports at 1Password regarding QNAP and Synology being listed as 2FA enabled when a user creates a login for a site like QNAP forums (a subdomain at https://forum.qnap.com/), but the other websites do not allow for it. Turns out, this is because the 2FA is associated with a local IP address or custom domain the NAS is hosted on.

Would love some feedback regarding my exception text--is it confusing or unclear?

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -157,6 +157,8 @@ websites:
     tfa: Yes
     email: Yes
     software: Yes
+    exceptions:
+      text: "Only covers NAS system on local IP or custom domains"
     doc: https://www.qnap.com/en-us/how-to/tutorial/article/how-to-enhance-account-security-using-2-step-verification
 
   - name: SmartBox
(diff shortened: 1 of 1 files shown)


jadchaar:
It would be nice to have a specific tag to help detect these types of special cases. For example, tagging certain NAS 2FA cases as local-ip.

Referencing this issue: https://github.com/2factorauth/twofactorauth/issues/3222

jamcat22:
Typically, we do not list self-hosted products on twofactorauth.org. These entries seem to fall into this category, but they each have their own cloud accounts. QNAP uses [QNAP IDs](https://account.qnap.com/) for [myQNAPcloud](https://myqnapcloud.com/), while Synology uses [Synology Accounts](https://account.synology.com/). QNAP requires SMS for 2FA and also allows email with a doc link [here](https://support.myqnapcloud.com/faq/all?category=twostepverification). Synology uses Google Authenticator for 2FA without a public doc link. I think we should modify these entries to reflect these account services.

Synology screenshots:
Synology Account - 2FA Settings
Synology Account - 2FA Login

jadchaar:
@jamcat22 Thanks for the feedback. It seems like QNAP does in fact support 2FA via Google Authenticator: https://www.qnap.com/en-us/how-to/tutorial/article/how-to-enhance-account-security-using-2-step-verification.

PR #3233, opened 2018-05-30
Remove Citibank AU

jimsug:
As per #3232 Citibank AU does not require 2FA upon logging in, only upon performing certain transactions. This does not satisfy the 2FA definition for this site.

(Sorry for the messy commits - I don't `git` much :wink:)

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -272,13 +272,6 @@ websites:
       tfa: No
       lang: de
 
-    - name: Commonwealth Bank of Australia
-      url: https://www.commbank.com.au/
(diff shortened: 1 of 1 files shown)


mxxcon (changes requested):
Instead of deleting this entry, edit it to indicate `tfa:no`

jimsug:
I considered that, and can certainly do so. In that case should I add all other financial institutions which do not support 2fa? Happy to do so, just need to know if that's something we want.

mxxcon:
If they are popular enough, sure.

Issue #3037, opened 2018-02-24
Generic columns headers need explanation/clarification


guntbert:
The two generic headers (*hardware token*, *software token*) need a little explanation.

I suggest (to preserve screen real etstate) attaching an :information_source: icon and on hover showing a popup with a short explaining text and a few examples (non-exhaustive of course).

mxxcon:
what kind of explanation text do you think it should have?

guntbert:
Giving it a try:

- Hardware Token
  > A physical item you have to insert/attach
  > e.g. Yubikey, FIDO, ...
- Software Token
  > A program running on your phone, that will provide an additional code
  > e.g. Google Authenticator, Authy, ...

(The rationale: I had to think a little to understand the meaning of the column titles, and I am not new to security :smile_cat: )

PR #3241, opened 2018-06-01
Add popup explaining tokens as per #3037

jimsug:
I've added popups explaining hardware and software tokens as per #3037, and this closes it if merged :smile:

### DIFF
--- a/_includes/desktop-table.html
+++ b/_includes/desktop-table.html
@@ -10,8 +10,8 @@
       <th class="two wide">SMS</th>
       <th class="two wide">Phone Call</th>
       <th class="two wide">Email</th>
-      <th class="two wide">Hardware Token <div data-content='A physical item you have to insert/attach (e.g. Yubikey, FIDO, etc)' data-position='bottom left' class="icon popup">🛈</div></th>
-      <th class="two wide">Software Token  <div data-content='An app on your phone (e.g. Google Authenticator, Authy)' data-position='bottom left' class="icon popup">🛈</div></th>
+      <th data-content='A physical item you have to insert/attach (e.g. Yubikey, FIDO, RSA SecureID, etc)' data-position='bottom left' class="popup two wide">Hardware Token <i class="info circle icon"></i></th>
+      <th data-content='An app on your phone (e.g. Google Authenticator, Authy)' data-position='bottom left' class="popup two wide">Software Token  <i class="info circle icon"></i></th>
     </tr>
     </thead>
     <tbody class="jets-content">
--- a/_includes/desktop-table.html
+++ b/_includes/desktop-table.html
@@ -10,8 +10,8 @@
       <th class="two wide">SMS</th>
       <th class="two wide">Phone Call</th>
       <th class="two wide">Email</th>
-      <th class="two wide">Hardware Token</th>
-      <th class="two wide">Software Token</th>
+      <th class="two wide">Hardware Token <div data-content='A physical item you have to insert/attach (e.g. Yubikey, FIDO, etc)' data-position='bottom left' class="icon popup">🛈</div></th>
+      <th class="two wide">Software Token  <div data-content='An app on your phone (e.g. Google Authenticator, Authy)' data-position='bottom left' class="icon popup">🛈</div></th>
     </tr>
     </thead>
     <tbody class="jets-content">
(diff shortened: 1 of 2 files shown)


mxxcon:
Hardware is not only that. It's also devices such as RSA SecureID and similar type of dongles.

Also you use `🛈` symbol, however it seems like at least osx doesn't properly display it. It just shows unicode box. 
screen shot 2018-06-01 at 11 35 17 am
Since we use semantic-ui icons, perhaps find a suitable alternative there?

kenman345:
pretty sure you can take all those attributes off the new div and apply them to the `<th>` tag.

jimsug:
Updated to use the icons and added the additional info.

Do you think it's necessary to enumerate all the different kinds of hardware token? Popup allows for HTML so we could list them instead. I was just trying to keep it brief.

RichJeanes:
As @mxxcon said:
>Hardware is not only that. It's also devices such as RSA SecureID and similar type of dongles.

There are dedicated hardware TOTP tokens that do not attach to the computer for authentication, but rather just display a code that the user has to manually enter. This is practically identical to software-based TOTP, but on a dedicated hardware device that does not have the ability to communicate with other devices. 

So the current description is not accurate:
>'A physical item you have to insert/attach (e.g. Yubikey, FIDO, RSA SecureID, etc)'

NilsEnevoldsen:
From many forum discussions, I've noticed that many people (most people?) are unaware that their current password manager already handles software tokens, even though almost all do. Additionally, these people may be more inclined to trust the software (or software vendors) that they already use. Therefore, I wonder if it might be worth specifically listing the top few: LastPass Authenticator, 1Password, Dashlane, Enpass, and Keeper. (I think LastPass is the only one that requires an app installation seperate from the password manager itself.)

Since we should be encouraging everyone to use a password manager anyway, it seems counterproductive to suggest services that _only_ do TOTP, such as Google Authenticator, Microsoft Authenticator, and Authy.

jimsug:
@NilsEnevoldsen Yeah, that's a good point. But I'm concerned that it might confuse people who haven't used it before, and who are more likely to need this additional help text.

There's another issue of weighing up whether it's necessarily a good idea to keep all of one's eggs in the same basket, so to speak. On balance, for most people, using a password manager is good. Using 2FA is great. Protecting both with the same master password for both poses risks and is not something I think we can necessarily take for granted.

This is a broader discussion that _might_ be beyond the scope of this project. For what it's worth though, I _do_ use my password manager for 2FA, even though I'm aware of the risks of doing so - on the other hand, the password manager _itself_ is protected by 2FA, and the means by which I can receive 2FA is also protected by 2FA - so I'm comfortable with that trade-off.

On the other hand, if one were to have the above set up but one of the 2FA methods for the password manager were insecure or less protected, that potentially exposes a pretty crippling SPOF. For example, Dashlane's 2FA is an email and code pushed to the mobile app. If the user's email isn't protected by 2FA, then they're in trouble.

On the _other_ other hand, I guess we're promoting 2FA so I suppose it's not a huge concern? I don't know, I'm a bit conflicted on this. Thoughts?

NilsEnevoldsen:
> There's another issue of weighing up whether it's necessarily a good idea to keep all of one's eggs in the same basket, so to speak. On balance, for most people, using a password manager is good. Using 2FA is great. Protecting both with the same master password for both poses risks and is not something I think we can necessarily take for granted.

I admit I don't know much about other password managers, but 1Password addresses this by having not just a master password or master password + 2FA, but a master password + mandatory secret key (i.e. two-factor *encryption*). If you've compromised my 1Password vault, then you've already compromised my device, so my TOTP tokens are compromised either way.

If other password managers only use a master password or master password + 2FA, then I agree, that seems to open up some additional risk. (Granted, the risk is still small IMHO. Compromising the password manager -- unlikely -- would leave no second line of defense, but compromising a site login -- much more likely -- would still leave 2FA as a second line of defense.)

jimsug:
@NilsEnevoldsen 
> If you've compromised my 1Password vault, then you've already compromised my device, so my TOTP tokens are compromised either way.

I suppose that's my point - if your TOTP tokens are stored with 1Password, and let's say that you have that on more than one device, then yes, compromising your 1Password vault will expose all your TOTP tokens.

However, if you have your 1Password vault on one device, and your TOTP tokens in a dedicated app on another, then compromising one doesn't compromise the other. I guess this is not an issue if you have your password vault on only one device, and that device is also your TOTP device.

PR #3243, opened 2018-06-05
Rename Atlas to TFE, add Vagrant Cloud

justincampbell:
Hashicorp Atlas has been renamed to Terraform Enterprise, and the Vagrant functionality was extracted to Vagrant Cloud.


brycx (commented):
Thanks for the PR @justincampbell. How about using the non-login pages for urls? [Terraform](https://www.terraform.io/)/[Vagrant](https://www.vagrantup.com/). 

Also, this is both self-hosted and a service offered directly by HashiCorp, yes?

justincampbell:
@brycx Those other URLs are for our open-source tools, not for the hosted products which support 2FA.

For Terraform Enterprise, there is no logged-out state, but users could learn more about it at https://www.hashicorp.com/products/terraform. I wanted to add/update this due to [1Password 7 adding support for twofactorauth.org to their Watchtower feature](https://blog.agilebits.com/2018/05/22/1password-7-for-mac-the-best-ever/#watching-out-for-you), which I assume would require matching hostnames for the feature to work, so I think `app.terraform.io` is more accurate.

The URL for Vagrant Cloud in this diff is correct.

justincampbell:
> Also, this is both self-hosted and a service offered directly by HashiCorp, yes?

Terraform Enterprise is available self-hosted, but Vagrant Cloud is only offered as a SaaS.

Thanks!

brycx (approved):

mxxcon (approved):

PR #3244, opened 2018-06-06
Add Charles Schwab Bank

lark047:


mxxcon:
It's already listed under `Investing` category.

brycx:
@mxxcon Does it belong there? They provide both investment options and banking as far as I can tell.

mxxcon:
What is it more known for, banking or investment? To me investment, but that's subjective.

lark047:
The bank is a separate entity. I've been a customer for five years.

mxxcon:
Does investment and banking have separate login systems?
Your `doc:` link has no mention of 2fa at all.

lark047:
As far as I can tell they're the same. All logins are through https://www.schwab.com.

Carlgo11:
If that's the case then this PR should be closed. Could someone confirm that their login systems aren't separate?

PR #3245, opened 2018-06-06
Add Firefox

trockenasche:
Firefox account for Firefox Sync and/or addons.mozilla.org


_data/other.yml
```diff
@@ -222,6 +222,13 @@ websites:
       software: Yes
       doc: https://www.findmyshift.com/help/what-is-two-factor-authentication
 
+    - name: Firefox
+      url: https://accounts.firefox.com/settings/two_step_authentication
```
mxxcon: This url should point either to the main firefox.com or at least accounts.firefox.com, but not settings page.

mxxcon (changes requested):

_data/other.yml
trockenasche: Okay, thanks, I will do so.

mxxcon (approved):

brycx (approved):

PR #3247, opened 2018-06-07
Update gmail.com to mail.google.com

conorgil:
There are many URLs in the data.json data set that
redirect before landing on the final site. This
causes problems for third party clients (e.g. 2FA Notifier)
because they match on the origin URL.

Updating the data to reflect the final URL allows these
third party services to function correctly while also
preserving the functionality of twofactorauth.org
for end-users of the website.

We need to be careful when working with temporary redirects
since they could change in the future, but this PR is
just for gmail, which has a very VERY low likelihood of
redirecting to a new subdomain any time soon.

### DIFF
--- a/_data/email.yml
+++ b/_data/email.yml
@@ -23,7 +23,7 @@ websites:
       facebook: freenet
 
     - name: Gmail
-      url: https://gmail.com
+      url: https://mail.google.com
       img: gmail.png
       tfa: Yes
       sms: Yes


mxxcon:
@conorgil perhaps it makes sense for 3rd party clients to also have a concept of related domains, be it user-specified and/or pre-populated list?
For example LastPass such a feature
screen shot 2018-06-07 at 4 16 25 pm

This way a single entry here would be able to cover multiple domains/products.

conorgil:
@mxxcon thanks for the quick reply.

That is certainly something that third party clients could do. However, that is something that would have to be maintained manually, right? My idea is to write some scripts to help keep data current by automatically fully resolving origins.

Are you against making changes to the URLs in this manner, or just suggesting another way to additionally allow end-users to help maintain the data?

mxxcon:
Maybe not gmail specifically, but overall.
I'm a bit hesitant of us updating URLs specifically for the reason you mentioned previously, where redirects from the canonical domain could change based on site's redesign.

I wonder if there are some existing resources with curated lists of related domains.. 🤔

conorgil:
@mxxcon How would you feel about a PR that resolved only domains that used a permanent redirect?

I agree that 302s et al are trickier to handle, but 301s should be fine to update the URL.

Thoughts?

mxxcon:
Ya, I guess 301 is a good criteria to update.

mxxcon (approved):

brycx (approved):

PR #3254, opened 2018-06-11
Added PCextreme

kevindierkx:


brycx (commented):
Thanks for the PR @kevindierkx. Due to a limit where a site has to be within Alexa top 200k sites, PCextreme cannot be added since [it is not](https://www.alexa.com/siteinfo/pcextreme.com). Also for further PR's, please do not add the same entry in different yml files and instead choose only one.

PR #3256, opened 2018-06-11
Update WordPress exception

jadchaar:
Since we have the sms label, the exception was redundant and unnecessary.

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -277,8 +277,6 @@ websites:
       tfa: Yes
       sms: Yes
       software: Yes
-      exceptions:
-          text: "SMS-capable phone required."
       doc: https://en.support.wordpress.com/security/two-step-authentication/
 
     - name: XING
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -271,12 +271,14 @@ websites:
       img: weibo.png
       tfa: No
 
-    - name: WordPress.com
+    - name: WordPress
       url: https://wordpress.com
       img: wordpress.png
       tfa: Yes
       sms: Yes
       software: Yes
+      exceptions:
+        text: "SMS required for 2FA."
       doc: https://en.support.wordpress.com/security/two-step-authentication/
 
     - name: XING


mxxcon:
I disagree with this change.
Wordpress requires a phone number even if you want to uses TOTP. You can't set it up if you have only cell-less device such as a wifi-only tablet.
If anything that exception should be rephrased.

_data/social.yml
```diff
@@ -277,8 +277,6 @@ websites:
       tfa: Yes
       sms: Yes
       software: Yes
-      exceptions:
-          text: "SMS-capable phone required."
```
jamcat22: Instead of removing this exception, we could rephrase it to say `SMS required for 2FA.` if the current exception is too vague.

jadchaar: Thanks @jamcat22, I agree. @mxxcon thanks for the feedback. I did not realize this was a Facebook style SMS required scenario 😄. Will fix.

jadchaar:
A bit unrelated, but what are the rules for indenting? I see some exception "text:" sections are indented twice, while others are once.

Also, since Gravatar is owned by WordPress (since 2013), should we add a note in the exception saying that Gravatar logins are protected by WordPress (e.g. you can login with your Gravatar credentials but you need to go to WordPress settings to enable 2FA and change passwords, etc)?

mxxcon:
Regarding indents, see `.editorconfig` in the root of the repo.

_data/social.yml
brycx: @jadchaar Actually, Facebook 2FA  can be used without SMS #2937 :grinning:

brycx:
I would leave out the Gravatar part. Both because it can be confusing and because this site is concerned with 2FA. So if WordPress is where 2FA is happening, I would just keep it this way.

brycx (approved):

jadchaar:
Alright fantastic thanks for the feedback @brycx 👍

_data/social.yml
jadchaar: Hah yeah they recently changed it 😄

_data/social.yml
```diff
@@ -271,14 +271,14 @@ websites:
       img: weibo.png
       tfa: No
 
-    - name: WordPress.com
+    - name: WordPress
```
mxxcon, line 274: Wordpress.com is different from Wordpress.org and I want us to be explicit about that, so lets keep ".com" part

mxxcon (changes requested):

jadchaar:
Hi guys, I have moved this PR to here: https://github.com/2factorauth/twofactorauth/pull/3268.

I had to refork the repo as I made a few mistakes (new to contributing to public, open source projects).

PR #3257, opened 2018-06-12
Add SiteHost

simonwelsh:


brycx (commented):
Thanks for the PR @simonwelsh. Unfortunately, due to a recent limitation where a site needs to be within the Alexa top 200k, this site [cannot be added](https://www.alexa.com/siteinfo/sitehost.nz).

PR #3273, opened 2018-06-18
Add ScaleGrid

kunalnagar:
Status: Pending Review

- [ ] Confirm Category (currently cloud)
- [x] Logo


brycx (commented):
Hey @kunalnagar, thanks for the PR! Unfortunately this site cannot be added right now, as we have a limit that is, sites must be within the Alexa top 200k sites. [Scalegrid is not](https://www.alexa.com/siteinfo/scalegrid.io).

PR #3275, opened 2018-06-19
add TierraNet

frenn-sd:


brycx (commented):
Hi @frenn-sd, thanks for the PR! This site cannot be added right now, since we have limit for new sites where they need to be within the [Alexa top 200k](https://www.alexa.com/siteinfo/tierra.net).

PR #3277, opened 2018-06-20
Add SQR.NL to hosting

simonklink:


mxxcon:
Thank you for your pull request. However, for now we are restricting additions to Alexa top 200k sites. https://www.alexa.com/siteinfo/sqr.nl

PR #3280, opened 2018-06-22
Add DMARC Analyzer as a site with TFA setup

rotorsolutions:


brycx (commented):
Thanks for the PR @rotorsolutions. At the moment we have a limitation for new additions that require the site to be within the Alexa top 200k, which it [is not](https://www.alexa.com/siteinfo/dmarcanalyzer.com), so this site cannot be added.

rotorsolutions:
@brycx OK that is too bad..
Can't you make an exception as we're in the 200k for US and this requirement is not really listed anywhere. Otherwise I wouldn't have bothered to create the PR in the first place :)

PR #3285, opened 2018-06-27
added rcodezero.at

MikeAT:


Carlgo11:
Hello @MikeAT,
we've recently started to try and limit new listings to sites ranked at or under 200k on Alexa or with a big social media presence.
This is to try and keep our list small enough that we can maintain it.
Unfortunately this means that some sites don't get listed on twofactorauth.org.

rcodezero.at is ranked above one millions and I can't find any other metric to go by that would suggest that it's used by enough people to warrant a listing here.
I apologize for the inconvenience 😞

PR #3289, opened 2018-07-03
Add Shopify to the list of sites that support 2fa in the retail section

JackMc:
Shopify has supported two factor authentication for many years now so we figured it'd be good to add ourselves to twofactorauth.org!

We don't fit perfectly into the Retail category, since we're a commerce platform and as such more face the merchant than the buyer, but I will leave it up to the maintainers as to if that distinction warrants having its own section. It doesn't appear there are any other commerce platforms in the list. You can learn more about what Shopify is at https://shopify.com.

The other interesting thing here is that our merchants can either log in from myshopify.com or shopify.com, and I wasn't sure which to put. Hopefully the maintainers can advise.


mxxcon:
Is there a difference between myshopify.com or shopify.com? Are they different accounts, different purposes or just using both domains to prevent cybersquatting?

mxxcon (approved):

JackMc:
`myshopify.com` hosts our merchants' stores and the store-specific login page. `shopify.com` hosts a login page that is smart and redirects you to the right store and/or asks for a store URL if you have multiple under that email. But they both use the same sets of credentials. I'm specifically worried about folks who have their login saved in 1Password as `myshopify.com` not getting the alert that's sourced from here.

mxxcon:
This have came up before and 1pass should really look into the concept of domain aliases, not just for 2fa notifications but also for logins. There are some big services out there with many different domains which use the same login system...LastPass supports it and I find it to be very useful.

JackMc:
Yeah, that would definitely make a lot of sense. I guess for now we should just ship this with the place that most merchants will have this saved for, which is shopify.com because of the signup being on that page.

brycx (approved):

PR #3291, opened 2018-07-04
Add Jitbit Helpdesk

maxt3r:


maxt3r:
How to check why it failed?

_data/other.yml
```diff
@@ -304,6 +304,14 @@ websites:
       img: instapaper.png
       tfa: No
 
+    - name: Jitbit Helpdesk
+      url: https://www.jitbit.com/helpdesk/
+      img: jitbit.png
+      tfa: Yes
+      sms: No
```
mxxcon: No need for explicit `no` tag.

mxxcon (changes requested):
At this point in time we do not list self-hosted software since that brings too many variables in how to do display such data.

maxt3r:
I changed the link to the hosted version (we have both)

mxxcon (approved):

brycx (approved):

PR #3294, opened 2018-07-11
Update Wikipedia/Wikimedia entry

AntiCompositeNumber:
Two-factor authentication is available for certain users across all Wikimedia projects. Logo changed from Wikipedia to Wikimedia, and data updated.


mxxcon (changes requested):
I would keep Wikipedia name since it's by far the most well known and probably most used/searched service of all those.
Also indentations are significant in yaml format. Please correct that.

AntiCompositeNumber:
I believe I have fixed the indentation issue. I changed everything back to Wikipedia as well. There are some that would be morally opposed to that, but I am not one of them.

mxxcon (approved):
Why would they be _morally_ opposed to that?

brycx (approved):

PR #3306, opened 2018-07-16
Add Booking.com

Carlgo11:
Can't find any public doc but they do support SMS 2FA now _(although a couple of years too late..._ 🙄<i>)</i>

here's a screenshot of the setup phase:
screenshot

Note that if this is set up it will completely disable any third party integration with your Booking.com account. (Maybe that's worth an exception?)


mxxcon (dismissed):
Maybe it's better for `transport` category?

Carlgo11:
I thought of that but does transport really fit a site primarily for event and hotel booking?
My conclusion was that it doesn't fit for booking.com but maybe for some other sites in the same industry like momondo.com that lay greater importance on the transport factor of the trip.

mxxcon:
Perhaps rename `transport` into `travel`?

Sunit-Kulkarni:
I like that idea. Renaming Transport to Travel

Carlgo11:
Travel makes sense on most of the listings in transport but some feel a bit off. Like Uber, TFL (Transport for London), Lyft, SL, mytaxi and Clipper.
While you technically travel when taking a Uber ride I think most people interpret travel as something you do further than around a couple of blocks.

brycx (approved):

mxxcon (approved):

PR #3309, opened 2018-07-20
Add State Employees Credit Union

gothicman02:
Edited in notepad.  Let me know if there are any issues.


_data/banking.yml
```diff
@@ -1,4 +1,4 @@
-websites:
+﻿websites:
```
mxxcon: Please undo this change.

mxxcon (changes requested):

_data/banking.yml
gothicman02: I believe this may have happened when I accidentally hit delete on the top line when editting, and corrected it, by undoing.  I did a diff and they look identical currently compared to the original.  Also it appears the checks have passed as well, and I double checked other pull requests to make sure the spacing, line placement, etc. is correct.  Am I missing something?

mxxcon: If you edit this file directly on github's website, you'll see there's a character in front of `websites`. I can't tell what exactly it is, but that is what causing this diff. Please delete that char.

gothicman02: I see it.  Looks like an extra space, compared to the original.  It has been removed

Carlgo11 (approved):

mxxcon (approved):

PR #3315, opened 2018-07-23
Add Toontown Rewritten, Wizard101

tonytien:
Added Toontown Rewritten and Wizard101 to Gaming list


_data/gaming.yml
```diff
@@ -284,6 +284,13 @@ websites:
       software: Yes
       doc: http://www.tibia.com/support/?subtopic=faq&question=authenticator
 
+    - name: Toontown Rewritten
+      url: https://www.toontownrewritten.com/
+      img: toontownrewritten.png
+      tfa: Yes
+      software: Yes
```
mxxcon, line 291: What kind of software does this site uses for 2fa?
And overall how does it qualify with our criteria? https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria

tonytien: Toontown Rewritten relies on software authentication through an app like Google Authenticator or Authy for 2FA. I believe Toontown Rewritten would qualify with the criteria under the personal image point as it is a MMORPG, just like other games on the twofactorauth.org list already, such as League of Legends and Tibia. Some game streamers also play this game, and the security of their Toontown Rewritten accounts would certainly play into protecting their personal image.

mxxcon: For now we do not list/indicate sites which use 3rd parties for authentication mechanism. As such toontown would be marked as `tfa:no`.

tonytien: Sorry, I just want to make sure I didn't misunderstand your initial question. Let me try to provide a more clear explanation. Toontown Rewritten's two-factor backend is handled by their own website, but I meant that the user still needs Google Authenticator or some similar app to be able to retrieve that two-factor code (as is the case with many sites). I added a screenshot from toontownrewritten.com of what the two-factor step of login looks like to try to clarify. If the site still doesn't qualify, sorry about the trouble!
screen shot 2018-07-23 at 1 50 37 pm

Carlgo11 (approved):

mxxcon (approved):

PR #3319, opened 2018-07-23
change for netcup

dt-flo:
Corrected domain, the 2fa is available for ccp.netcup.net, the customer control panel.


Carlgo11:
for the `url` tag we wan't the service's main page and not a login portal. As such I'm closing this PR.

PR #3321, opened 2018-07-24
Add Exoscale

ste-m:
Add Exoscale to the cloud list :)


brycx:
Thanks for the PR @ste-m. At the moment we have a limitation for new sites, which require them to be within the Alexa top 200k, and Exoscale [is not](https://www.alexa.com/siteinfo/exoscale.com).

PR #3326, opened 2018-07-25
Add Hetzner Cloud

LKaemmerling:
Hello Guys,

the Hetzner Cloud supports as well as Hetzner itself the complete 2FA Suite as well.

Currently, there is no documentation about this since it is implemented in the customer area.


mxxcon:
Does hetzer cloud use different login system from the other hetzer site?

LKaemmerling:
No, it is the same Single Sign On, but a complete other hosting system special for the Cloud. So there is a difference. Even the complete company (Hetzner Cloud is an indepentend company) use 2fa internal as a requirement.

brycx:
Thanks for the PR @LKDevelopment. Please update the icon to be 32x32 in size. Otherwise the build fails.

LKaemmerling:
@brycx i have changed the icon :) So everything should be green now!

brycx (approved):

Carlgo11:
Sorry for the delay @LKaemmerling!
So you're saying that if I have an account at https://www.hetzner.com/ I can login to https://www.hetzner.com/cloud?
If so then this PR can't be merged.

LKaemmerling:
Yeah this is correct, but the dedicated servers part (hetzner.com) and the cloud part are completely independent.

Carlgo11:
still, if it's the same login system then it's just one listing on our site.

PR #3332, opened 2018-07-29
Move OpenDNS from Security to Domains

TraderStf:
as most DNS services are here


Carlgo11 (dismissed):
You need to actually move the listing and image from security. Not just add it to the domains.yml file.

TraderStf:
Sorry for the mess, did not saw the doc (add a line at the top of the main page), just did based on others sites.
I'll do it asap or you can do it if you want. no problemo.
thanks

brycx (approved):

Carlgo11 (approved):

mxxcon:
I'm not sure we should've made this change...OpenDNS.
From their site
>OpenDNS is a suite of consumer products aimed at making your internet faster, safer, and more reliable.

That does not seem like a registrar service..

Carlgo11:
Well... it doesn’t fit security either. A DNS service handles domain names so even though it’s not a domain regristrar I think domain is the best fit.

PR #3338, opened 2018-07-31
Add Starling Bank

ohheyrj:
Starling Bank is another app based challenger bank in the UK.

Although the main Starling app does not support 2FA, the developers portal does. I'm not sure if I should add in an exception for this or put this in as another site. Feedback would be welcome.


mxxcon (approved):
Why would they implement 2fa for developers but not direct customers?🤦‍♂️

ohheyrj:
I know, I was surprised to find that out. Not as bad as Monzo, there login is just a email link 😱

brycx (approved):

ohheyrj:
So its come to light (after I had to reinstall the app) that they do support 2FA...kind of...
When you go to sign in again it asks if you have your original phone, if you do it asks to scan a QR code from within the app on the old phone. If however you select the 'I dont have my old phone' option, it just asks for a mobile number (doesn't even have to be the one originally registered with the account) and sends a SMS to that number. Would be a much better idea for them to use the mobile number registered to the account. Not sure if anything needs changed in this pull request. I still dont think its true 2FA.

Carlgo11:
@rj175 if it accepts any phone number then no, that wouldn't be a form of authentication.

PR #3341, opened 2018-08-01
Add Discourse forums, incl. F-Droid and Bitwarden

DandelionSprout:
I see that F-Droid's contributor forums supports 2FA, as per the screencap below of that forum's profile settings, so I've attempted to add that site here.
image


DandelionSprout:
By random chance it seems to me that Bitwarden's community forums also support 2-factor authentication, since they're based on the same forum engine. I'll see if I'm able to add that site too to this pull request in a hurry.

DandelionSprout:
What I admit makes this particular pull request a bit odd, is that I find it to be difficult to figure out which exact forum engine that those two forums are using. Finding it out could've helped [me/us] find a shared knowledge base for it (to find articles to use for the _doc:_ tag), and potentially to find furthermore forums that uses that engine and which therefore very likely also have 2FA. For lack of a better idea, I've just as well [just asked the F-Droid forum](https://forum.f-droid.org/t/which-forum-engine-is-the-f-droid-forum-using/3428) about it, so I suppose that that thread could be kept an eye on.

DandelionSprout:
Okay, so I've been told (and can confirm) that the forum engine in question is [Discourse](https://www.discourse.org/), which has [potentially several hundred forums and communities](https://www.discourse.org/customers) under its umbrella (incl. ones for Imgur, Linkin Park, and EVE Online), and who have (as far as I can determine) [rolled out 2FA](https://meta.discourse.org/t/two-factor-local-login-proposal/76689/48) pretty much universally to them.

This escalated quickly. Any ideas on how to proceed and/or build on this now?

mxxcon:
We are not going to list discourse itself since it bumps against our situation with _self-hosted_ software projects.

DandelionSprout:
Sounds fair to me.

Though in that case, which Discourse-hosted forums should be listed on Twofactorauth, and how do we decide which ones? The F-Droid and Bitwarden forums are the only ones that I can 100% confirm first-hand, since those are the two Discourse-based forums that I have accounts with.

mxxcon:
I think logically when we'll know about significantly large forums and they fit our criteria for listing, they'll be added.

PR #3347, opened 2018-08-02
Add Betvictor

nhmarujo:
Added information for Betvictor page


mxxcon:
Is there `docs:` link for this?

nhmarujo:
Not yet. My intention is to submit a new PR with the docs section once they have a page.

mxxcon (dismissed):

Carlgo11 (dismissed):
I don't think betting sites fit under the category entertainment.


Furthermore I feel uneasy with listing gambling sites here.
I do see the value in having a gambling site use 2FA but I'm not sure if we should list those kinds of sites here based of on the nature of them.
Just like if we listed drug or alcohol sites here.

brycx:
I think it would be a good idea to list gambling sites actually. One of the reasons being that you usually have some amount of funds deposited to such gambling sites, so if I wanted to try a online gambling I would much prefer a site that is serious about security and where I would feel somewhat safe having funds deposited. 

I don't think you can compare gambling sites to sites that sell drugs(I assume you mean illegal drugs), since in most places gambling isn't illegal and neither is alcohol. Gambling is entertainment, so I think the category fits as well.

mxxcon:
I don't see an issue with gambling sites being listed here. And I don't see an issue with alcohol retailers being listed here either. We are not the moral police when it comes to that.

beanbaguk:
I also think it makes sense to list gambling sites with 2FA. Entertainment fits well, but if not, perhaps a "Gambling" section would be suitable? 

Amongst others, BetVictor is a tightly regulated, licensed and legal business. Furthermore, it's important that users are made aware of gambling sites that take security seriously and offer 2FA to protect accounts. 

The way I see it; By denying 2factorauth not to display regulated gambling sites that protect users, you are denying legal sites to showcase their security and responsibility to customers, thereby potentially putting users at risk which is exactly what 2factorauth was set-up to avoid.

Just my two cents....

nhmarujo:
@Carlgo11 do you want me to create a new category for this? If so, can you please point me in the right direction?

Carlgo11:
@mxxcon @brycx @beanbaguk the legality of the category entirely depends on what country you're from.
This site would have been illegal in Sweden before ~2016 (I don't know how it is now). The same is true for drug sites/stores (legal medications) and alcohol sites/stores are illegal.

I understand that it must seem odd from a central European or American point of view but these kinds of sites are, in some regions, highly regulated/illegal.

This is actually a lot similar to the sex industry. In some parts of the world prostitution and other sex work is completely illegal and seen as something taboo in culture.
In some parts of the world it's legal with some part of regulation, such as both parties being above 18 years of age.
In some parts of the world it's completely legal and seen as nothing strange.

Therefore I think we must rethink if this is something we'd want to have on our site.

We're not the moral police but I doubt anyone here would, for example, add a site that advertises sex workers that are children, even if it's completely legal in some countries.
Nothing is black and white and sometimes some regulation is necessary.


I'm not saying we should outright ban all gambling sites. Just that we should think harder on what positives and negatives it brings and what cultural biases we have. Basically a dialog.

I look forward to hearing what your views are!

brycx:
@Carlgo11 I see what you mean by cultural biases. On the same account of Betvictor having been illegal in Sweden, back then and maybe now, we still list other things on the 2factorauth site that is illegal in some other countries. For example from this [Wiki](https://en.wikipedia.org/wiki/Legality_of_bitcoin_by_country_or_territory) article, there are some countries that ban all use of cryptocurrency and some only for specific cases. Some of these countries might very well see cryptocurrency as a far greater evil than gambling.

I don't think we can compare it only by legal/illegal, because most things are sure to be illegal in some country. I guess what I'm trying to say is have a poll, because in the end it comes down to what the people behind 2factoauth are willing to list.

beanbaguk:
One could argue WhatsApp, Signal, Telegram and other messaging platforms are illegal in many countries, yet these are still listed on the site. As @brycx mentioned, Crypto is also banned in many places too yet still listed.

Major gambling companies including BetVictor also have very strict age verification and KYC policies in place meaning it's impossible to use an account with them, even if you provide fake data.

Those that choose to gamble, do so legally in a regulated environment. I think choosing not to list a legal business due to personal morals gives a negative precedence for censorship and that defeats the point of this fantastic resource. More-so if it's possible to promote legal sites for driving positive security features.

If the group feels a level of restriction should be put in place, perhaps creating a "Gambling" category with an age verification check (confirming the users DOB) upon opening the category before anything is displayed would be appropriate?

Thanks!

mxxcon:
There's a huge difference between being those sites and simply listing 2fa status of those sites.
We are not endorsing or encouraging any of the sites listed here.
If twofactorauth.org would be illegal in Sweden simply because we listed a gambling site, so then would Google be and every other search engine.
If any of such legal questions would be considered, it would be from the view point of US since that is where GitHub is located and hosted.
I'm sure legit gambling sites allow visitors/accounts only from countries/locations where it is legal.

RichJeanes:
My issue would be if listing those sites would put us on any web filter lists, getting us blocked at work places and schools.

mxxcon:
We could get blocked for cryptocurrency too. Again, we are not such sites, we simply list them.
Do we actually have any reports of this happening with the site? By now filters are smart enough to recognize domains vs content. If simply having a gambling site's name on a page would get us blocked, then so would any news sites would be blocked too.
I think this is a non-issue.

Carlgo11:
The consensus seems to be that gambling/betting sites should be listed here so when this site is moved to a gambling/betting category I'll dismiss my review.

nhmarujo:
@Carlgo11 I've moved the entry to a new section called "betting", as you requested. Can you please approve?

Also @mxxcon can you please see if you are ok with this change?

nhmarujo:
@Carlgo11 / @mxxcon is this ok to be approved now? thanks

(@Carlgo11 thanks for removing that extra space)

nhmarujo:
Hi @Carlgo11 / @mxxcon, can this be merged or do you still think I need to change anything else?

Thanks

mxxcon (approved):

stephengroat (approved):

Issue #1025, opened 2015-03-20
Region Specific Services


RichJeanes:
So we've been trying to figure out how to handle "international" pages, right? Well a thought just struck me. 
What if we separated out region-specific sites? Anyone anywhere (for the most part) can use Google or Facebook, but (for example) the Utilities category that I started to put together only lists services for the States.
What if we had a separate page from the main (similar to how we have a Providers page) for region-specific services? Instead of overall categories, the page would be separated into regions (countries?) and would only have a single list each. Or list all of the regions on the page and make them collapsible (or separate pages, something, I dunno, I'm not much of a UI guy) with all of the categories under each one.
I've realized that being and American working on a project started by an American and working with mostly Americans (plus a few others :wink:), a lot of sites already listed are very American-centric (look at Banking for Pete's sake).
I figure this would let us take a more globally-focused stance in site design and content. Input?

mxxcon:
But then how would you list sites that are in more than one country?

jamcat22:
What about some sort of filter. You could choose what country you were in, and country specific services would show up.
If a website only has 2FA in specific countries, the page would display the status of that site per country. You would still see sites that are in all countries.

Then there could be a way to say either a website has 2FA in all countries except some or it has 2FA in only certain countries and you wouldn't have to add the site to each page.

mxxcon:
@jamcat22 how would that filter work considering we are stuck with static hosting? it would have to be fully client-side.

RichJeanes:
If a site operates independent of location, then it would be on the main page. If it operates in only specific countries, then it would be listed only in those countries. Why would a site be limited to only being listed under a single country? 

I like that idea, too. Couldn't that be handled with Javascript?

mxxcon:
@RichJeanes for example smaller financial institutions operate in a single country or geo region...
I guess all the data could be cooked into a json file(s?) that then would be processed client-side. But I imagine that's not a small task to undertake.

RichJeanes:
Why couldn't we build that statically? I'm not really skilled with Jekyll at all so I have no idea if this is possible, so I'm just throwing ideas out there, but:
[...]

PR #3358, opened 2018-08-06
Add Exact Online

jeohist:
I know the CONTRIBUTING.MD says not to add multiple URLs, but since this is one website hosted on multiple URLs for different countries and we'd really like them to be shown in 1Password's Watchtower functionality... 

I don't think there's really a way for this to be more elegant, other than implementing something like the YAML files in #1025 or straight-up allowing multiple URLs for one website (or Exact Online hosting everything in one domain).


mxxcon:
Are all those sites use the same login system or are they all separate?
If it's all the same then we should not have multiple entries, just a single main one is sufficient.

If you want 1pass' watchtower to display green checkmark or whatever, then you should approach them to implement the concept of related domains. Other password managers have it and it works great. I would not want us to mangle our data to compensate for deficiencies of 1pass.

jeohist:
They are all separate login systems (i.e. they don't share usernames or passwords and the TOTP token is unique for each site).


________________________________

jeohist:
Is anything required from my side at this point in order to get this merged?

Carlgo11 (changes requested):
For the `url` tag we list the homepage of the site and not the login page..
For star.exactonline.* the main page seems to be exact.com so please change the urls to that. Furthermore since they all would have the same url after the edit mentioned above I suggest you only use one entry instead of the 7 you added.

For the `doc` tags we require a url that isn't behind a login-page so if there's no other link that's open to the public then remove the tag.

We'll need some sort of confirmation that they indeed support 2FA so please send a screenshot of the login process or documentation page (with personal information blurred out or painted over)

jeohist:
Thanks for the response @Carlgo11.

start.exactonline.* is the website of the Exact Online application. The only reason exactonline.* redirects to exact.com is a marketing thing. Here's how it looks on the inside: https://i.imgur.com/GKp9MjB.png
Consider it the same situation as GMail :-)

With regards to the documentation pages: Those work fine here without logging on, they are publicly available. I do notice a few of the applications have a broken URL, I'll fix that. 

Login process:
Login page: https://i.imgur.com/fZyaC3k.png
2FA: https://i.imgur.com/DObnyVH.png

Carlgo11:
@jeohist what would you say the main page of the company/services are?

jeohist:
Since #3414 was implemented, I will simplify this PR by only including start.exactonline.nl. The others are not (yet) in the Alexa Top 200k. 

And in the meantime, it'll be a fun pet project to think about #1025 :-)

mxxcon (dismissed):

_data/finance.yml
```diff
@@ -53,6 +53,13 @@ websites:
       img: ditech.png
       tfa: No
 
+    - name: Exact Online
+      url: https://start.exactonline.nl
```
Carlgo11: One last thing, for the url we list the main page of a site. That page can't be a login page and so I suppose the url value would be https://www.exact.com/

Carlgo11 (changes requested):

_data/finance.yml
jeohist: start.exactonline.* is the website of the Exact Online application. The only reason exactonline.* redirects to exact.com is a marketing thing. Here's how it looks on the inside: https://i.imgur.com/GKp9MjB.png
Consider it the same situation as GMail :-)

Carlgo11: Still, we don't list login pages in the `url` field so either make start.exactonline.nl redirect to something else than a login page or change the `url` value to https://www.exact.com/.

jeohist:
I'm part of the team maintaining the website. Literally any request you send to the Exact Online application is sent to start.exactonline.nl. I don't see what the problem is here, it's not just a login page, it's the entire bloody application.

Carlgo11:
@jeohist The `url` tag is designed to list a sites main page. If it's a login page then it's not a main page (or at least not 99.99% of the time.)

If a new user wants to see what a service does then the first thing they'll look at is the main page and not the login page.
If you didn't know anything about Exact Online what would this tell you? My guess is not a lot.
screenshot

_data/finance.yml
```diff
@@ -53,6 +53,13 @@ websites:
       img: ditech.png
       tfa: No
 
+    - name: Exact Online
+      url: https://exactonline.nl
```
Carlgo11: chrome_2018-10-27_21-25-54
And the site does a 301 to exact.com
```suggestion
      url: https://exact.com/
```

Carlgo11 (changes requested):

Carlgo11 (dismissed):

brycx (approved):

Carlgo11 (approved):

PR #3360, opened 2018-08-07
Add Coingi.com

lukocz7:
Add new web Coingi.com


brycx:
Hi @lukocz7, thank you for the PR. We currently have a limitation for new sites, which means they have to be within the Alexa top 200k, but Coingi.com [is not](https://www.alexa.com/siteinfo/coingi.com). I'll have to close this PR.

PR #3362, opened 2018-08-08
Updated Moneybird to the dotcom domain

rvoortman:
The moneybird application lives in the .com domain (https://moneybird.com/login).
Applications like 1Password did not recognize Moneybird correctly because of this reason.

### DIFF
--- a/_data/finance.yml
+++ b/_data/finance.yml
@@ -163,7 +163,7 @@ websites:
       tfa: No
 
     - name: Moneybird
-      url: https://www.moneybird.nl/
+      url: https://www.moneybird.com/
       img: moneybird.png
       tfa: Yes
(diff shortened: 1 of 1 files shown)


brycx (approved):

mxxcon:
What is the primary domain for this site, .com or .nl?

I'm against mangling accurate data of our site just to confirm with 1pass' deficiencies.

rvoortman:
@mxxcon The primary domain for the _application_ is .com. You always login to moneybird.com/login, although  users would probably start at moneybird.nl and click login (on which they will be redirected to the .com domain). 

So I guess the primary domain is .com, because that's where you login.

mxxcon (approved):

PR #3387, opened 2018-08-20
Add Forums and Communities category

ghost:
Added a new category to seggreate online discussion forums from standard social media.


Carlgo11:
I don't think we're able to help the consumers into choosing a 2fa-supporting product/service when it comes to forums. Without any alternatives they're stuck with what they've got.

Our goal is _"to aid consumers who are deciding between alternative services based on the security they offer for their customers. This can also serve as an indicator for the effort a site has put into security in general."_
If we can't do that with a section then the only reason to list it would be to nagg the owners.

ghost:
I understand where you’re coming from, however the reason why I wanted to add forums us because I’m a 1Password user. If I save a username and password to my 1Password vault without two-factor credentials, it checks against this project (specifically the ‘software’ boolean) and warns me that I haven’t set up 2fa on a site that supports it.
This project is a much of a comparison service as well as a catalog of services that support 2fa, as this is what 1Password treats it as.

If you want to see how 1Password presents the information, have a look at https://blog.agilebits.com/2018/05/04/introducing-watchtower-2-0-the-turret-becomes-a-castle/, there’s a screenshot under “A second line of defence”

Carlgo11:
We're aware of their usage of our data and encourage any third parties to make us of our data to fit their needs but as the same time we can't tailor our entire project to fit a specific third party.
Our goal is to help consumers. Their goal is to help their customers enable 2FA on all sites possible.
For the most part those goals work together and sometimes they don't. This seems to be one such case.

PR #3388, opened 2018-08-21
Update WageWorks 2FA status

stevea1:
WageWorks now supports SMS and email 2FA.

### DIFF
--- a/_data/other.yml
+++ b/_data/other.yml
@@ -528,7 +528,6 @@ websites:
       tfa: Yes
       sms: Yes
       email: Yes
-      doc: https://www.wageworks.com/media/567495/terms-of-use.pdf
       exceptions:
           text: "Employer/plan sponsor may opt to not offer 2FA."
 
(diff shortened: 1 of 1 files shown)


_data/other.yml
```diff
@@ -525,8 +525,12 @@ websites:
     - name: WageWorks
       url: https://www.wageworks.com
       img: wageworks.png
-      tfa: No
-      twitter: WageWorksCares
+      tfa: Yes
+      sms: Yes
+      email: Yes
+      doc: https://www.wageworks.com/media/567495/terms-of-use.pdf
```
Carlgo11: I don't think that document fits as a documentation. It just mentions that 2FA may be offered and what kinds of 2FA they support.
For the documentations tag we want something that describes for the user how they can enable 2FA.

Carlgo11 (changes requested):

_data/other.yml
stevea1: Good point. Documentation link removed.

mxxcon (approved):

Carlgo11 (approved):

PR #3398, opened 2018-08-23
Update Liquid Web (icon)

stevea1:
Replace the old Liquid Web icon with their new one


stevea1:
This pull request has some spurious, old commits. I will remove these and then issue another pull request.

PR #3399, opened 2018-08-24
Add Ridgewood Savings Bank

larryv:
The [doc](https://www.ridgewoodbank.com/tob/live/usp-core/static/help.html#login_security) only mentions SMS, but the site does provide an email option.


brycx:
Hi @larryv. Thank you for this PR. Unfortunately we currently have a limit for new additions, so they have to be within the Alexa top 200k, which this site [is not](https://www.alexa.com/siteinfo/ridgewoodbank.com). If it's popularity continues to grow we can probably add it later on.

larryv:
I understand. It would be helpful if CONTRIBUTING.md mentioned this requirement.

PR #3400, opened 2018-08-24
Add HackNotice

trockenasche:
HackNotice alerts you when a site is hacked or your info is leaked


mxxcon:
For now we are concentrating on adding sites which are ranked in alexa top200k. It seems like this site is a little too low trafficed to be added at this time https://www.alexa.com/siteinfo/hacknotice.com

PR #3401, opened 2018-08-25
Update Liquid Web icon

stevea1:
Replaced the old Liquid Web icon with their new one.


Carlgo11 (approved):

mxxcon (approved):

stevea1:
It appears the old logo is still being shown?
https://twofactorauth.org/img/hosting/liquidweb.png

mxxcon:
i see a blue droplet. looks correct to me.

stevea1:
And so it is! Looks like Cloudflare's involved in serving up content so maybe I hit a cache server that hadn't updated yet. Thanks for checking.

PR #3404, opened 2018-08-25
Add PassHub

w3ua:


_data/identity.yml
```diff
@@ -113,6 +113,16 @@ websites:
       software: Yes
       doc: http://www.onelogin.com/product/strong-authentication/two-factor-authentication/
 
+    - name: PassHub
+      url: https://www.passhub.net/
```
Carlgo11: Change the url to `https://passhub.net`

_data/identity.yml
```diff
@@ -113,6 +113,16 @@ websites:
...
+      img: PassHub.png
```
Carlgo11: Everyone else uses lowercase on their images so I suggest you do the same.

_data/identity.yml
```diff
@@ -113,6 +113,16 @@ websites:
...
+      tfa: Yes
+      software: Yes
+      hardware: Yes
+      exceptions:
+          text: "hardware tokens only work on desktop."
```
Carlgo11: If hardware only works on desktops, what happens if someone tries to login via another device?
Also capitalize "Hardware"

Carlgo11 (changes requested):

_data/identity.yml
w3ua: As of today, you can not connect hardware token to a mobile device, so if your account was created with hardware token, it will be accessible only on a PC or Mac with WWPass Security Pack installed, and either NFC reader or USB port. If your account created with phone-based token (WWPass Passkey app from Appstore or Playstore), it can be accessed on PC, Mac, phones, tablets and even coffee machines and refrigerators, as long as they have compatible web browsers, capable of performing client side encryption computations.

_data/identity.yml
w3ua: Actually I did. Initially I used PassHub in the name, but then discovered that everybody else use lowercase, so I renamed the file, deleted the old one, added the new one with all lowercase, and corrected the line in yml

_data/identity.yml
w3ua: Done

Carlgo11 (dismissed):

_data/identity.yml
```diff
@@ -1,4 +1,4 @@
-websites:
+﻿websites:
```
mxxcon: What is going on here? Why is this showing up as changed? 🤔

Carlgo11: I did a diff locally and it didn't show any of those changes. Must be a glitch with GitHub.

w3ua: have no idea why it happened. I didn't do it ;)

mxxcon: I looked at this file using mcedit and it shows that the first char is unicode `FEFF`. It seems like some text editors add it..? @w3ua could you please try to remove it?

w3ua (commented):
Removed first symbol, as suggested.

mxxcon (approved):

w3ua:
Somehow my edit removed Carlgo11's approval, and this merge request still sits here blocked...

mxxcon:
@w3ua that's "normal" in github. When you modify a chunk of code that was previously "reviewed" that vote gets dismissed and has to be re-approved. That's how github works 🤷‍♂️

Carlgo11 (approved):

PR #3408, opened 2018-08-28
Add digitec

jokay:
Add retail website [www.digitec.ch](https://www.digitec.ch/en/) which supports [2FA](https://www.digitec.ch/en/page/new-two-factor-authentication-for-your-user-account-6589).


brycx:
@x-jokay please fix the two errors Travis reports:
```
1. 'lang' should NOT be present when tfa: true.
2. img/retail/digitec.png should not be larger than 2500 bytes. It is currently 4254 bytes.
```
To shrink the image size, use an optimizer like ImageOptim.

jokay:
@brycx ready for review.

brycx:
Thanks @x-jokay. Please add a `doc` link that explains the 2FA setup process or take a screenshot of it, if it's only for registered customers, with personal information blurred out.

x-jokay:
@brycx As already mentionned there is no documentation and therefore I can't set a link.

I can provide 4 printscreens for the setup process.
Where can I upload them and how should the `doc` link look like?

Carlgo11 (approved):
Looking good! 👍

jokay:
@brycx ready for review.

brycx (approved):

PR #3413, opened 2018-08-29
Mobile layout improvements

dsoegijono:
- Move mobile supported methods to 2-column rows instead of a paragraph
- Add `checkmark` icon to supported methods, add `ban` icon to unsupported
- Reduce cell padding

screen shot 2018-08-28 at 22 54 00

### DIFF
--- a/css/base.scss
+++ b/css/base.scss
@@ -188,7 +188,7 @@ input[type=search] {
 
     > div {
       border-bottom: 1px groove rgba(34, 36, 38, 0.14902);
-      padding: 2em;
+      padding: 1em;
     }
 
     .main {
@@ -206,10 +206,34 @@ input[type=search] {
         }
       }
 
+      .methods {
+        @include display-flex();
+        @include flex-direction(row);
+        flex-wrap: wrap;
+        margin: 0 10px;
+
+        > p {
+          margin: 5px 0;
+        }
+      }
+
+      &.positive {
+        .methods {
+          > p {
+            flex-basis: 50%;
+          }
+        }
+      }
+
       .button-group {
         @include display-flex();
         @include flex-direction(row);
         justify-content: space-between;
+        margin-top: 10px;
+      }
+
+      .icon {
+        margin-right: 0 !important;
       }
     }
 
--- a/css/base.scss
+++ b/css/base.scss
@@ -229,6 +229,7 @@ input[type=search] {
       }
 
       .icon {
+        margin-left: 0 !important;
         margin-right: 0 !important;
       }
     }
--- a/css/base.scss
+++ b/css/base.scss
@@ -212,17 +212,13 @@ input[type=search] {
         flex-wrap: wrap;
         margin: 0 10px;
 
-        > p {
+        p {
           margin: 5px 0;
         }
       }
 
-      &.positive {
-        .methods {
-          > p {
-            flex-basis: 50%;
-          }
-        }
+      &.positive .methods p {
+        flex-basis: 50%;
       }
 
       .button-group {
--- a/_includes/mobile-table.html
+++ b/_includes/mobile-table.html
@@ -37,7 +37,7 @@
       </div>
       {% if website.doc %}
       <div class="button-group">
-        <a href="{{ website.doc }}" class="ui twitter mini button"><i class="large book icon"></i></a>
+        <a alt="Documentation" href="{{ website.doc }}" class="ui twitter mini button"><i class="large book icon"></i> Docs</a>
       </div>
       {% endif %}
     </div>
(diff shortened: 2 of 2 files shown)


mxxcon (commented):
I have no commend on the underlying code, but visually it's better than before :)

Carlgo11:
Screenshot from iPhone X

</img>

Looks good but I noticed all categories are opened automatically on mobile devices (also happens on the live site). When was that implemented?

Carlgo11 (dismissed):

dsoegijono:
> When was that implemented?

@Carlgo11 I don't think it was intentional. @jamcat22 mentioned it on Slack as something we should fix but I haven't had the chance to look into it.

css/base.scss
```diff
@@ -206,10 +206,34 @@ input[type=search] {
...
+        @include flex-direction(row);
+        flex-wrap: wrap;
+        margin: 0 10px;
+
+        > p {
+          margin: 5px 0;
+        }
+      }
+
+      &.positive {
```
fpigerre: I would recommend putting all these selectors on one line unless separate declarations are to be inserted at different levels.

css/base.scss
```diff
@@ -206,10 +206,34 @@ input[type=search] {
         }
       }
 
+      .methods {
+        @include display-flex();
+        @include flex-direction(row);
+        flex-wrap: wrap;
+        margin: 0 10px;
+
+        > p {
```
fpigerre: It seems unnecessary to include the `<` in `< p {`, as there seems to be only one `<p>` tag under the "methods" class.

fpigerre (changes requested):
This looks really good! :+1:

Do we know if the categories being opened automatically started occurring before these changes were made?

dsoegijono:
> Do we know if the categories being opened automatically started occurring before these changes were made?

@psgs Yes, it's happening on the live site too.

dsoegijono:
screen shot 2018-08-29 at 22 18 30

Btw, just wondering if you guys have any opinions on the text "No two factor auth". I wasn't sure what to put there.

Carlgo11:
@dsoegijono how about 2FA? That's used on most exception texts so it would probably fit here as well.

dsoegijono:
@Carlgo11 Just "No 2FA" or  "2FA not supported"?

jamcat22:
I suggest "2FA not supported". Also, any way to find out why icons aren't centered on buttons?

And, since the doc button is so large, should we put explainer text on it? Or at the very least, can we add alt text to the doc, Twitter, Facebook, and email buttons so that VoiceOver can describe them as something other than "link"?

dsoegijono:
@jamcat22 [this](https://github.com/2factorauth/twofactorauth/pull/3413/files#diff-99125772e23d9230b4d830624fb79adcR232) should have fixed the button centering. There were some left/right margins.

I'll add alt texts. What kind of explainer text do you think we should add to doc button?

jamcat22:
@dsoegijono maybe just the word "Docs" like the heading on the desktop table?

I tried measuring using the screenshot you posted. Am I just measuring wrong? Or maybe the button size isn't equal or the icon itself isn't centered? 🤔

Twitter button, left side measurement of 156
Twitter button, right side measurement of 165
Facebook button, left side measurement of 156
Facebook button, right side measurement of 165

fpigerre (dismissed):

dsoegijono:
screen shot 2018-09-03 at 11 07 15

@jamcat22 Would "Documentation" be too technical? I was thinking something like "How to enable" but that might be too long for smaller screens.

Carlgo11:
I like Documentation. Seems to fit just fine.

Carlgo11 (dismissed):

jamcat22:
@dsoegijono is “Docs” too ambiguous? Should we consider changing the desktop heading?

dsoegijono:
- Added alt text to links
- Added "Docs" to doc link

Carlgo11:
pssst, @jamcat22 ☝️

jamcat22 (approved):

dsoegijono:
@Carlgo11 / @psgs want to help re-approve? :)

Carlgo11:
I’ll review it in a couple of hours :wink:

Carlgo11 (approved):

Carlgo11:
I'm fine with both "Docs" and "Documentation" however I do think it's easier for non natives to understand 'documentation'. If someone searches for "Define Docs" or similar it shows Google Docs or the .doc format.
But as I said, I'm fine with whatever.

dsoegijono:
I prefer keeping it as "Docs" for now, as it's consistent with desktop version. Do we want to change it on both?

jamcat22:
New design is much better. Thanks @dsoegijono! Now we just have to figure out why all the categories open when loading the page.

PR #3420, opened 2018-08-30
Rake improvements

stephengroat:
enforces mandatory link checking (whitelisting a few that will never work)


kmpoppe:
@stephengroat I just pulled this change and saw that you're working with a cache of the links (which makes total sense). I'm totally missing whether that will work in the GitHub Travis environment? Will the cached data be preserved as long as we need them?
Obviously, through APIv2 (#4015) the verify.rb changed, which now leads to an unresolvable merge conflict in the webeditor.
Would you care to make your changes to the current master again?

Carlgo11:
Closing as we're no longer using Travis CI

PR #3424, opened 2018-09-01
Add GBAtemp

DandelionSprout:
It ranks [fairly high on Alexa](https://www.alexa.com/siteinfo/gbatemp.net), so that won't be a problem for this, at least.

I was however unsure if it this should be in Communication or Gaming, but I felt that the former fits it better.

As has become a recurring trend for my pull requests, I was once again unable to find public documentation of GBAtemp's 2FA support, but those out there who have GBAtemp accounts can go to https://gbatemp.net/account/two-step to verify that they support 2FA. 
image


Carlgo11:
Is it just a forum or does it do anything? (#3387 :confused:)

DandelionSprout:
It's most well-known for its forum, but also maintains a newsfeed, a wiki, and some other things. https://gbatemp.net/

Think of it as a version of XDA for Nintendo consoles. At least if you're familiar with XDA.

jtagcat:
@DandelionSprout the real question is does it really need 2fa? What would 2fa protect?

Carlgo11:
Closing this PR as the need for 2FA is doubtable and the site criteria questionable.

PR #1177, opened 2015-06-13
Remove providers page

smholloway:
This PR addresses [issue 1173](https://github.com/jdavis/twofactorauth/issues/1173) where several contributors agreed that the providers page had outlived its usefulness.

PR #3437, opened 2018-09-06
Add Auth0

lukeocodes:


mxxcon:
Is this for when people integrate auth0 on their sites or if I go to https://auth0.com/ and click on login button?

mxxcon:
auth0
Please use this file. It's 1.8kb

_data/identity.yml
```diff
@@ -6,6 +6,19 @@ websites:
       software: Yes
       doc: https://support.1password.com/two-factor-authentication/
 
+    - name: Auth0
+      url: https://auth0.com/
+      img: auth0.png
+      tfa: Yes
+      sms: Yes
+      email: Yes
+      phone: Yes
```
Carlgo11: Could you point me to where in the doc phone calls are mentioned?

lukeocodes:
This is for people integrating Auth0 into their site.

_data/identity.yml
lukeocodes: Just checking on this, I believe through a 3rd party only.

mxxcon:
If it's for people who integrate auth0 into their site then it should not be listed here. See #1177 for justification why.

lukeocodes:
@mxxcon I'm confused, how do we differ from other identity management providers list here: https://twofactorauth.org/#identity

Carlgo11:
@lukeoliff we don't list 2FA providers here (such as Authy, Duo, Google Authenticatior) so if your service is a 2FA provider then it won't be added. (See #1177 like mxxcon mentioned)
Furthermore, if your service is a login service, like using Google, Twitter, Steam or Facebook to login to a website, then it also can't be added. That's because of our third-party rule.

However if your service is like 1Password and Lastpass, allowing the user to store their login information for different things on your service then it's fine.


Are we misinterpreting what your service does?

mxxcon:
@lukeoliff auth0 would be listed in the information you provided was relevant for me going to https://auth0.auth0.com/login and trying to login into auth0 and these are the 2fa methods I could be using.
But not if I'm a webmaster somewhere and want to implement auth0 on my site.

Issue #2891, opened 2017-12-14
Capital One 360 software 2FA is weak


scovich:
Capital One 360 claims a push-based 2FA mechanism, but in order to use it you have to log into their mobil app on your device using normal username password. The mobile device thus remains as a weak link because anyone who compromises it can just access your account directly; why bother messing with the other 2FA-protected devices at that point?

alecbw:
The explainer video on their splash [page](https://www.capitalone.com/identity-protection/swiftid/) implies 2FA is capable of being enabled on the webapp, though neither the instructions nor a service rep could describe how to actually set it up. At this time it appears 2FA only actually works for the mobile apps via sending an email code auth.

ghost:
Agreed with @alecbw.  Spent 30 minutes on the phone with tech support and SwiftID would not work. Additionally the help desk, SwiftID is not available. The 2FA software is absolutely worthless if SMS and email codes cannot be disabled. It is up to the individual logging in how they want to receive the code, so of course an attacker would pick whichever they could easily control. I disagree with @scovich to an extent, the mobile device is MUCH better than SMS/Email (i.e. Google's 2FA push actually works and lets you disable all other 2FA methods after initial setup)

ryancwalsh:
I agree. https://www.capitalone.com/identity-protection/swiftid/ feels like security theater. Probably useless. Very disappointing for a major bank to not take security seriously.

Carlgo11:
>What about these verification methods? They seem to be used for logins as well.
https://www.capitalone.com/support-center/bank/what-is-otp
https://www.capitalone.com/digital/mobile-app-verification/
 \- @phallobst https://github.com/2factorauth/twofactorauth/issues/6170#issuecomment-936680085

PR #3460, opened 2018-09-15
Add Capital One note

ghost:
Add notes about Capital One's SwiftID, the 2FA mobile software application which accepts and responds to push notifications. 

reference to #2891


pdg137:
There's no information about SwiftID or two-factor available in [CapitalOne's help page](https://www.capitalone.com/support-center/search/).

ghost:
Not sure what having an entry in Capital One’s site has to do with the
change. The location of their documentation is in a hyperlink off of the
notes page. The current capital one row links to that same SwiftID page and
I kept that link. The main reason for the update was due to the
shortcomings of trying to get that to work with their technical support and
statements about it not always available.

pdg137:
Right, I am agreeing with you that SwiftID does not seem to be a well-supported feature, since they don't link to it from their help page.  The term "SwiftID" is not actually mentioned anywhere in the Android app, either.

ghost:
Ok. Thanks. Obviously I hadn’t understood.

DavidCao22 (dismissed):

DavidCao22:
I don't have any official authority or write access, just wanted to get this merged. Spent a while looking for this info. It would've been nice to have merged in, so I'm trying to help it along and save time for others in the future.

Some layout changes need to be made since the UI has changed since the PR was put up, but I could take care of that after the merge.

kmpoppe:
The question is, if this project is the right place to put these critics about Capital One. Through APIv2 (#4015) we can now add `proprietary` (software) as a 2fa-method, which would mean SwiftID.
Information and the way Capital One handles these software problems with change on short notice, so we would be adding information that could be out of date relatively soon.
I'd think that extending the exception with something like "If SwiftID doesn't work, you should use SMS instead" would be sufficient. @Carlgo11 ?

Carlgo11:
I agree with @kmpoppe. The amount of information in this PR could easily be compressed into an exception text. Furthermore, what happened to the documentation link? I can't find anything about SwiftID on their site any longer 🤔

Carlgo11:
Closing as the source is gone

PR #3471, opened 2018-09-22
Mobile: stop opening all categories

dsoegijono:
`.mobile-table` is the table for each category, so this block actually displays all table.


Carlgo11 (approved):
Looks good!

jamcat22 (approved):
:+1:

PR #3481, opened 2018-10-01
Add Passwordstate

arnydo:


stephengroat:
@arnydo thanks for the addition, but as you can see in the [doc](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md), the alexa ranking of the site must be better than 200k.

also, as a general rule of thumb, we only list hosted services and not self hosted software (since the 2fa features can always be turned off)

if you have any other ideas on sites to be listed, please make another request! since this one isn't right for the site, i'm closing it for now

PR #3483, opened 2018-10-01
Add Sophos Central

arnydo:


stephengroat (dismissed):
@arnydo thanks for the pr!

_data/security.yml
```diff
@@ -237,6 +237,14 @@ websites:
       twitter: SaferVPN
       facebook: SaferVPNCom
 
+    - name: Sophos Central
+      url: https://central.sophos.com
```
Carlgo11: The main url seems to be `https://www.sophos.com`. Please change the value of the `url` tag to that.

Carlgo11 (changes requested):

_data/security.yml
arnydo: I can certainly change to the requested URL, however, they have several products and not all support MFA... Sophos Central is one that Does support MFA.

_data/security.yml
```diff
@@ -237,6 +237,14 @@ websites:
       twitter: SaferVPN
       facebook: SaferVPNCom
 
+    - name: Sophos Central
+      url: https://www.sophos.com
+      img: sophoscentral.png
+      tfa: yes
```
Carlgo11: Please capitalize all `Yes` values

Carlgo11 (changes requested):

arnydo (commented):
Updated the "yes" values. Capitalized each.

Carlgo11:
For the email part of the authentication, is that instead of a password?

arnydo:
> 
> 
> For the email part of the authentication, is that instead of a password?

No, it is in addition to the password. The email contains a OTP (4 digit code) as an alternative to Google Authenticator.

brycx (approved):

Carlgo11 (approved):

carlylemiii:
This should have included "Exceptions" like done for Wikipedia and OpenDNS, for example, and been named just Sophos if the URL wasn't going to be specific to Sophos Central. 🙁

Carlgo11:
I’m not sure I understand what you mean @carlylemiii.
The listing is for Sophos Centeal and not any other products.

Our guidlines require the url tag to link to the service’s main page. It also can’t be linked to a login page or similar.

The name should make it obvious that it’s only for Sophos Central and not any other product.

PR #3495, opened 2018-10-10
Rename Bank Zachodni WBK to Santander PL

NullPiotrException:
Santander PL acquired Bank Zachodni WBK.
Not long before that, BZWBK also added software 2FA via mobile signatures: you confirm login for desktop session in your smartphone app with a 8 digit PIN that doesn't change. Inferior to TOTP but still software 2FA. Checked and it works.

If you google "Santander Polska" you'll see on their offical page the title: "Santander | Santander Bank Polska (dawniej BZWBK)" (formerly BZWBK)


_data/banking.yml
```diff
@@ -780,6 +773,16 @@ websites:
...
 
+    - name: Santander PL
+      url: https://santander.pl
+      img: santander.png
+      tfa: Yes
+      software: Yes
+      sms: Yes
+      doc: https://santander.pl/firmy/bankowosc-elektroniczna/bzwbk24-mini-firma/metody-autoryzacji/metody-autoryzacji.html
+      exceptions:
+       text: "Software authentication requires bank's mobile app."
```
Carlgo11: Is the app "iBiznes24 mobile"? If so, change the exception text to "Software authentication only available via iBiznes24 mobile".

NullPiotrException:
Done. The app is actually [Santander mobile](https://play.google.com/store/apps/details?id=pl.bzwbk.bzwbk24) but it has the same developer as [iBiznes24 mobile](https://play.google.com/store/apps/details?id=pl.bzwbk.ibiznes24) so you were close.

Carlgo11 (approved):

mxxcon:
If it's a never-changing pin, would that be even considered a second factor? Wouldn't this still be the knowledge factor such as password? Once I know that pin, I don't need to have that mobile app to login, I can just input that pin from memory..

Carlgo11:
@mxxcon but it requires the use of the app to authenticate so access to the actual hardware (something you have) is required then.
I we should view the pin as something redundant. Like if a hardware token would require a pin to unlock.

That's my view of it at least.

mxxcon:
Perhaps @einsteinsfool can clarify how it works.
To me this sounds like if you have an account and everything is setup, and if I find out your login info and this pin I can login on desktop into your account without having access to the mobile app. If all it does is just store a permanent pin, once it gets out, it's no longer another factor. It's just a very expensive sticky note with your bank pin.
It becomes kinda no different to a security question, which we don't consider 2nd factor.

NullPiotrException:
Yes, it's like a 2nd password that can only be entered in the mobile app. So if the attacker knows the ID, password and the PIN they can log in from their app.

It's definitely inferior to TOTP and I suggested changin it but geted downvoted because people prefer free money to security. Still, it's just as 2FA as email 2FA is. Somebody can run the app in Android emulator [on their desktop] and log in with these 3 pieces of data but the same is correct with email 2FA - the [never-changing] email password being the 2nd factor.

mxxcon:
> It's definitely inferior to TOTP and I suggested changin it but geted downvoted because people prefer free money to security. Still, it's just as 2FA as email 2FA is. Somebody can run the app in Android emulator [on their desktop] and log in with these 3 pieces of data but the same is correct with email 2FA - the [never-changing] email password being the 2nd factor.

You are assuming that email account doesn't have 2nd factor. But that's beside the point.

So in order to login on the desktop one must have the mobile app, which will pop up asking you to enter the pin. What does it to take to get the mobile app setup?

NullPiotrException:
> You are assuming that email account doesn't have 2nd factor. But that's beside the point.

Yes, my bad. If email has TOTP then it acts like TOTP. But then for emails without TOTP there should be a warning that it's not 2FA according to your definition.

> So in order to login on the desktop one must have the mobile app, which will pop up asking you to enter the pin. What does it to take to get the mobile app setup?

Only the ID and password if I recall correctly.

Perhaps there's a mechanism in the bank app for remembering the device and one can't use the PIN on another device when it's already been seted up with one. Will check it today and let you know.

NullPiotrException:
I checked and was pleasantly surprised that it wouldn't let me input my PIN in a freshly installed Santander app on somebody else's mobile. It telled me to disable "mobile signature" (as it's called) after logging in on my mobile device or desktop and then add mobile signature to the new device.

So the 2nd factor is my phone and noone else's (assuming they implemented the hardware detection well that is).

mxxcon (approved):

brycx (approved):

Issue #3447, opened 2018-09-08
Vanguard only allows YubiKey devices for 2FA


nelsonjchen:
You can't use anything else for Hardware Token. It prompts you to allow Vanguard to see the make and model and won't proceed if it's not a:

* Yubikey FIDO U2F Security Key
* Yubikey NEO
* Yubikey 4 NANO
* Yubikey 4

For that matter, I heard Google's 401Ks are at Vanguard. I would guess they don't allow those fancy new Titan keys either. 

https://gist.github.com/nelsonjchen/e9e28a58165f1a6db9acb1a756045003

Carlgo11:
Are you sure? It looks like they also support SMS and phone calls.
[![screenshot](https://tempfiles.carlgo11.com/download/D5B947858079F0/?p=czaH3EuIyU19XM)](https://investornews.vanguard/sign-up-for-these-3-security-features/)

nelsonjchen:
Oh! Yeah, sorry for the inaccurate title. I meant for the hardware token option. Let me edit my title and post.

nelsonjchen:
Oof, I suppose this isn't actually a big enough problem yet nor a concern for this project. For instance, Blizzard, and I'm sure some banks are Vasco based which means a fob for every service. Also you still enter in a code from the device itself. Blizzard's doesn't have any differentiator on its listing.

I was hoping to use SoftU2F with Vanguard. No dice. 

Just not looking forward to the day where I have a pile of keyfob-size-like things on my keychain.

mxxcon:
You could submit a PR which adds exception text with a note on this restriction that it requires only specific yubikey models.

nelsonjchen:
I don't know. It's not really an "exception" either. Almost all sites on the list supporting hardware tokens usually support one technology or vendor to implement their hardware token. Vanguard isn't really exceptional here. If the project adds listings for the type of hardware tokens, then this would be indeed exceptional for the U2F type.

Carlgo11:
@nelsonjchen there's a fork of this project over at https://www.dongleauth.info/ that is entirely about U2F.

PR #3499, opened 2018-10-16
Add Vanguard Exception

rjoffe:


Carlgo11:
Is this the same as #3447?

rjoffe:
Yes, same as #3447, just adding the exception

_data/investing.yml
```diff
@@ -202,6 +202,8 @@ websites:
       sms: Yes
       phone: Yes
       hardware: Yes
+      exceptions:
+        text: "Only support: Yubikey FIDO U2F Security Key, Yubikey NEO, Yubikey 4 NANO, Yubikey 4"
```
Carlgo11: `Yubikey FIDO U2F Security Key` Seems like they just call it "Security Key" so you could probably compress it to just `Yubikey Security Key`.
You don't list any Yubikeys from the 5 series. Are they not supported?
What's the underlying technology that's needed in the keys to login?
If it's U2F you could just say `"Hardware 2FA requires a Yubikey with U2F support"`

rjoffe: The Vanguard site specifically lists those 4 devices only.

I've tried others, and it will not accept the keys, therefore not utilizing true universal U2F.

Carlgo11: Just got a reply from them on Twitter. Turns out they don't support the 5 series. Only Yubikey 4

mxxcon (approved):

Carlgo11 (approved):

PR #3503, opened 2018-10-17
Add Rabobank

### LINKED ISSUE #3370: Add Rabobank
Please add Rabobank internet banking, they support a hardware token, their own Rabo Scanner device.

https://bankieren.rabobank.nl/klanten

Carlgo11:
Closes #3370


brycx:
I see no mention of 2FA in the docs? If it is the thing called Rabo Scanner, do we need an exception for this? Could be some proprietary solution.

Carlgo11:
@brycx the 2FA method is a smartcard reader often used by banks (at least over here).
The reader queries the bankcard chip, the chip responds with a signed copy of the query and that in turn is used to generate a OTP for the account holder to authenticate with.

I don't think a exception is necessary as every user will receive a reader and card upon registering with the bank.

brycx:
That makes sense.

brycx (approved):

stephengroat (approved):

PR #3504, opened 2018-10-17
Add Cigna

tresni:
Logged into to Cigna today and found out they support 2FA via email and SMS.  Figured I'd share that information.


Carlgo11:
Hi Brian, 👋 
thanks for the PR!

We don't list documentation links that requires the user to login to view it. As such, could you remove the doc tag?
Also, for us to verify that the site indeed has two factor authentication, could you post a screenshot of the setup page or login page __(with any personal information blurred out or painted over)__?

Thanks!
//Carl 😄

tresni:
Is there a place where it would be appropriate to put that link?  My account happened to prompt me to enable 2FA, but if it wasn't for that, I'm not sure I would have known where to look (or even that it existed.)  I haven't found any public documentation other then what appears to be a completely unrelated or possible internal Multi Factor Auth location.

tresni:
mycigna - two-step authentication 2018-10-17 16-04-51

Updated w/o the doc link and here's the requested screenshot

Carlgo11 (approved):
Thank you! 👍

stephengroat (approved):

PR #3507, opened 2018-10-23
Add Mint doc

mlindgren:
MFA for Mint must be configured via another Intuit site, since Mint itself doesn't support setting it up. I added a link to the TurboTax documentation which describes how to do this. Unfortunately this is rather confusing so it would be nice to add an extra note describing the process, but the current schema doesn't seem to allow that.

### DIFF
--- a/_data/finance.yml
+++ b/_data/finance.yml
@@ -163,7 +163,7 @@ websites:
       sms: Yes
       email: Yes
       software: Yes
-      doc: https://ttlc.intuit.com/questions/2902682-what-is-two-step-verification
+      doc: https://ttlc.intuit.com/questions/2905185
 
     - name: money by Envestnet | Yodlee
       url: https://money.yodlee.com/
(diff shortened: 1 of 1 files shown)


Carlgo11:
@mlindgren does Mint provide their own form of authentication other than the third party login via Intuit?

mlindgren:
@Carlgo11 No, Mint authentication is always through Intuit now. Mint is owned by Intuit (and always was, I think?)

_data/finance.yml
```diff
@@ -163,6 +163,7 @@ websites:
       sms: Yes
       email: Yes
       software: Yes
+      doc: https://ttlc.intuit.com/questions/2902682-what-is-two-step-verification
```
Carlgo11: I think this url `https://ttlc.intuit.com/questions/2905185` explains the process better than that one.

Carlgo11 (changes requested):

_data/finance.yml
mlindgren: Sure, looks good to me.

Carlgo11 (approved):

stephengroat (approved):

PR #3528, opened 2018-11-01
Remove Scottrade

stephengroat:
dead


Carlgo11 (approved):
Apparently it was bought up by TD Bank in 2017. https://en.wikipedia.org/wiki/Scottrade#History

brycx (approved):

PR #3531, opened 2018-11-01
Add Authy

### LINKED ISSUE #3376: Add dashboard.authy.com
The Authy Dashboard can be added in the security category. It supports SMS and Software tokens.

stephengroat:
closes #3376


Carlgo11:
Does it make sense from a consumer perspective to add Authy's dashboard?
If we want to aid consumers make a educated choice in what services and websites they use it's not like anyone is going to switch from a 2FA provider to another 2FA provider just because the new provider uses 2FA.
I mean, what 2FA provider isn't supporting 2FA?

stephengroat:
so no listing of any 2fa providers? should we add that to the `CONTRIBUTING` list?

we currently list https://www.onelogin.com/product/multi-factor-authentication

Carlgo11:
Sounds like a good idea. Especially since we've already removed them from twofactorauth.org/providers

brycx:
I opened PRs for both things.

stephengroat:
closing due to new Guidelines https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#guidelines

PR #3534, opened 2018-11-03
Remove DediServe

stephengroat:
blog entry deleted


brycx:
Do they even support 2FA anymore? I had a look through their new blog and knowledge-base. Not a single mention of 2FA.

Carlgo11:
If we're not sure then we could probably just delete it since https://www.alexa.com/siteinfo/dediserve.com.
Better not to list it than to give false information.

brycx:
> If we're not sure then we could probably just delete it since [...]

Agreed.

brycx (dismissed):

brycx (approved):

Carlgo11:
`git push -f` is a godsend 😅

mxxcon:
https://manage.dediserve.com/cart/ shows that you can setup 2fa during the signup process..

Carlgo11 (approved):

PR #3542, opened 2018-11-06
Add Ubuntu & WSL note

Carlgo11:


brycx (approved):

mxxcon (commented):
Which packages require these deps? Why would we need to install gcc?🤨

Carlgo11:
@mxxcon I ran up a server with a clean Ubuntu version and tried to run the site. I then wrote down all the dependencies that caused the build to fail.
I don't remember which specific gem required what dependency, just that they're all required to run. (Also did the same thing for WSL and the deps matched)

On Mac all dependencies are installed by Xcode.

mxxcon (approved):

PR #3546, opened 2018-11-08
Add Nextcloud

rugk:
Same as https://github.com/Nitrokey/dongleauth/pull/167


Carlgo11:
@rugk we don't use SVG files here. Only PNG files. See our [guidelines](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#guidelines) for more information.

rugk:
Yeah, no problem, but why do you close the PR, because of that? That's totally unnecessary and I've already adjusted it.
So here is the new PR: https://github.com/2factorauth/twofactorauth/pull/3547

Carlgo11:
No, I closed the PR because it used a SVG instead of PNG and it’s a self hosted service. We don’t list self hosted sites/services.

rugk:
How senseless…

mxxcon:
@rugk It's not senseless. It's a long-standing problem that nobody seems to be willing to address but only complain about. It's been discussed multiple times in the past.
If you can come up with a way for us to accurately yet easily represent 2fa status of self-hosted services in all the possible different combinations that can arise from the use of external plugins and custom configurations then we'll be more than happy to list self-hosted apps.

PR #3547, opened 2018-11-09
Add Nextcloud

rugk:
Follow-up of https://github.com/2factorauth/twofactorauth/pull/3546
Same as Nitrokey#167

Issue #3476, opened 2018-09-26
Inclusion of Self-hosted Software


RichJeanes:
Currently, we have a blanket exclusion of self-hosted software on the list. I believe this restriction should be lifted with specific requirements.
- Self-hosted software that is accessible to the public (anyone can register for an account, or that has a public interest) can be listed so long as it meets the other requirements for listing a site (eg. Alexa Top 200k)
- The software must have an independent authentication database (eg. you cannot log in to the site using credentials created on another domain)
- The software must support 2FA natively, not through extensions or plug-ins
  - If a first-party or "official" extension created by the original developers is available to enable 2FA, should that be allowed?
  - Third-party extensions and add-ons do not qualify

conorgil:
Is this restriction on "self-hosted software" written down anywhere that I can read? Mind linking?

1. What is the definition of "self-hosted software"? If something is hosted and accessible to the public where anyone can create an account, how is that different than any other website on the internet?
2. Can you please explain this more? How would you propose to identify whether this criteria was satisfied without knowing how the backend architecture and software are actually built?
3. Same as bullet 3. How could you know by looking at a website whether they are using plugins or extensions to provide 2FA functionality? Perhaps I am misunderstanding here.

rugk:
> If a first-party or "official" extension created by the original developers is available to enable 2FA, should that be allowed?

IMHO, yes. Modular software architecture is good and should not be punished. Obviously, also for closed-source (not self-hosted) sites, they could have a similar architecture and you would not know. Or consider closed-source self-hosted software, like forum software (e.g. https://github.com/2factorauth/twofactorauth/issues/1296) belongs to this category, too.

> Is this restriction on "self-hosted software" written down anywhere that I can read? Mind linking?

Despite being claimed that this is done, it actually is not. See https://github.com/2factorauth/twofactorauth/pull/3550 for a potential fix.

> If something is hosted and accessible to the public where anyone can create an account, how is that different than any other website on the internet?

Actually, that is the big point why it made no sense to _not_ include these in the first place. For the end-user they are just usual websites…

> How would you propose to identify whether this criteria was satisfied without knowing how the backend architecture and software are actually built?

Well… you usually know. I mean we all have an intuitive understanding of the term "self-hosted", I think. Apart from that, I would however argue that it does not make any difference, because _it actually does not_. GitLab is e.g. self-hostable and already included, so what…?
So you do not need to see it…

[...]

Issue #3234, opened 2018-05-30
Allow Self hosted as well as SaaS products


kilasuit:
Reason being these are still valid products and the whole point behind a site like this should be to include EVERY possible application as a single 1 stop place to be able to look at for what allows MFA

mxxcon:
We do list saas products.

The problem with self-hosted entries are situations where multiple external modules/plugins/packages/addons/whatevers provide different/conflicting features. How would such be represented?
As an easy example: wordpress.org has 17 different 2fa plugins...

kilasuit:
Well that's easily fixed by saying there additional installs needed using the below like I had to add to my PR #3210 

```exceptions:
       text:  "M/TFA is only available with additional Plugin installs"
```

It wouldn't add that much more but would be a reasonable expectation especially as a need for this site is to know **Where** MFA can be used not **How** its implemented - that can be found by other means and would negate declining PR's like #3219

mxxcon:
What about situations where mutually exclusive plugins provide different features? What methods do we indicate? Indicating all of them could be misleading for any given plugin..

kilasuit:
We don't indicate/suggest individual plugins - but more inform the users that they can find available plugins that will enable MFA to be used on which ever system it is - perferably linking to the underlying software provider's help pages (aka Wordpress)

Carlgo11:
As of right now our stance is that we don't list self hosted sites and we don't list third party login systems.
We'll take note of your request and maybe in the future, if enough people request it, we'll add self hosted services.
For the moment however, I think a fork of the site would be better suited to list self hosted services.

Issue #3030, opened 2018-02-21
Add software/service category


drohhyn:
Hi,

as I understand the project is dedicated to list popular websites which provide (or should provide) 2FA. Maybe it would be interesting to add an own category like "software" to list software which provides 2FA. For example [nextcloud](https://nextcloud.com/) is a data sync suite (like Google Drive, dropbox etc.) which provides 2FA (via plugin) for your private/company installation. 2FA for Atlassian's Jira is possible via third party module etc.

What do you think?

Kind regards,
  drohhyn

Carlgo11:
Hello drohhyn,
We used to list 2FA providers but because of all the sales bs the companies used we don’t do this anymore.

Furthermore we don’t list self-hosted services such as ownCloud and nextCloud.
As for third party systems we only list the site’s native authentication system.

Sorry I couldn’t be of any more help!

//Carl

drohhyn:
Thanks for your comment. I fully understand the reasons, also it is sad that it was not suitable for this awesome project.

mfc:
> Furthermore we don’t list self-hosted services such as ownCloud and nextCloud.

i see this statement but it's not clear to me how it is related to the [Contributing guidelines](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md), as ownCloud and nextCloud (as examples) clearly fit into `Data` in the [site criteria](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md).

it sounds like you are opposed to listing 3rd party 2FA plugins which is understandable, but including software that comes with 2FA natively seems to fit well within twofactorauth's aim.

If the issue is that they can be self-hosted, there are obviously [many providers](https://nextcloud.com/providers/) willing to receive user money to host it for the user. so i don't see that as a strong argument.

i'd like to be able to link folks to twofactorauth so they can have a more complete sense of the software ecosystem for a given need, but not having the FOSS options there makes it much less useful than it could be.

mxxcon:
If we will be listing self-hosted providers with builtin 2fa, there will be questions why preferential treatment is given to such projects, but not the ones that decided to outsource authentication to external modules or sources.
This is also similar why we don't list sites which use federated authentication.

If you have a suggestion how we can list such entries where there can be cases of mutually exclusive authentication methods and all the other nuances that arise from that, please do share with us.

mfc:
[...]

PR #3550, opened 2018-11-13
Be upfront about what you do not support

rugk:
You've declined https://github.com/2factorauth/twofactorauth/pull/3547 and in https://github.com/2factorauth/twofactorauth/pull/3546 you've explained:
> We don’t list self hosted sites/services.

However, that was actually **stated nowhere in your guidelines or so**! I mean this is your website/your rules, so it's totally your right to say what to support, but then you should be honest about what you not want to list here.

Without any indication of that you are essentially **wasting the time** of contributors. If they can read in the Contributing.md that self-hosted services are not supported, okay, they won't bother to create a PR here. Otherwise they'll do and this results in things like https://github.com/2factorauth/twofactorauth/pull/3546.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ are stored in folders corresponding to each of those categories in their own
 4. **Alexa top 200k**: A new site, that is not already listed, has to be within the
    Alexa top 200k ranking. You can check the ranking of a site [here](https://www.alexa.com/siteinfo).
 5. **No 2FA providers**: We do not list 2FA providers, such as [Authy](https://authy.com/), [Duo](https://duo.com/) or [Google Authenticator](https://github.com/google/google-authenticator).
-6. **No self-hosted software:**: We only support websites, not software you can self-host like Nextcloud, Mastodon etc.
+6. **No self-hosted software:** We only support centralized websites/web services, not (website) software you can self-host like Nextcloud, Mastodon etc.
 7. **Be Awesome**: You need to be awesome. That is all.
 
 ## Running Locally
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ are stored in folders corresponding to each of those categories in their own
 4. **Alexa top 200k**: A new site, that is not already listed, has to be within the
    Alexa top 200k ranking. You can check the ranking of a site [here](https://www.alexa.com/siteinfo).
 5. **No 2FA providers**: We do not list 2FA providers, such as [Authy](https://authy.com/), [Duo](https://duo.com/) or [Google Authenticator](https://github.com/google/google-authenticator).
-6. **No self-hosted software:** We only support centralized websites/web services, not (website) software you can self-host like Nextcloud, Mastodon etc.
+6. **No self-hosted software:** We only support centralized websites/web services, not (website) software you can self-host like [Nextcloud](https://nextcloud.com/), [Mastodon](https://joinmastodon.org/) etc.
 7. **Be Awesome**: You need to be awesome. That is all.
 
 ## Running Locally
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,8 @@ are stored in folders corresponding to each of those categories in their own
 4. **Alexa top 200k**: A new site, that is not already listed, has to be within the
    Alexa top 200k ranking. You can check the ranking of a site [here](https://www.alexa.com/siteinfo).
 5. **No 2FA providers**: We do not list 2FA providers, such as [Authy](https://authy.com/), [Duo](https://duo.com/) or [Google Authenticator](https://github.com/google/google-authenticator).
-6. **Be Awesome**: You need to be awesome. That is all.
+6. **No self-hosted software:**: We only support websites, not software you can self-host like Nextcloud, Mastodon etc.
+7. **Be Awesome**: You need to be awesome. That is all.
 
 ## Running Locally
 


rugk:
And before you try to argue your "Alexa top 200k" already covers that: It [does not](https://www.alexa.com/siteinfo/mastodon.social) as the largest Mastodon instance (/cc @gargron) e.g. does easily fulfill this criteria. And this is just one – theoretically you have to sum up all different instances and their rank. (respectively not sum up, but consider them all – you get what i mean :wink:)

Gargron:
Not listing self-hosted software seems odd. If it's popular, fulfills the same function as centralized alternatives, and has two-factor authentication, why should it not be listed? If you can't list the software in general, how about listing individual Mastodon servers, since they are websites? E.g. mastodon.social like @rugk points out above.

rugk:
> how about listing individual Mastodon servers, since they are websites? E.g. mastodon.social like @rugk points out above.

However, from a decentralized point of view (or "federation" view) that also seems a little unfair/bad, as obviously _all_ instances/servers that run this software (can) support it.
Also, in other cases like Nextcloud there is no "one big server" (at least not that I know of), but I'd guess all instances together easily would be in the 200k Alexa list.

jomo:
What's the definition of self-hosted? Aren't Bugzilla or GitLab self-hosted?

rugk:
Any news here?

FYI [this fork of this project has merged Nextcloud](https://github.com/Nitrokey/dongleauth/pull/167#issuecomment-4399274799). Just so you know…

kinafu:
For example Nextcloud:

You do not have to host the service by yourself. There are companies/people providing instances to the public. You'll just need to register.

There is _no difference_  for a user between accessing Nextcloud or Dropbox. No additional challenges for the user.

The services vary in their licence and the dev. process (open source), but that's another layer of view.
The user just wants to know about 2fa when visiting this web page.

Self-hosted? "I want cloud and 2fa" :smile:

Gargron:
>You do not have to host the service by yourself. There are companies/people providing instances to the public. You'll just need to register.

The same is true for Mastodon.

rugk:
**TL;DR:** Upvote https://github.com/2factorauth/twofactorauth/issues/3476. :wink:

Any news here? Could not you at least _explain_ why you don't want to include self-hosted services…?

Or answer the questions that have been asked here several times e.g.:
* Why no self-hosted?
* Where do you draw the line, i.e.:
* "What's the definition of self-hosted? Aren't Bugzilla or GitLab self-hosted?"
* Why not include e.g. mastodon.social?
* etc.

On that topic here also saw https://github.com/2factorauth/twofactorauth/issues/3030 (by @drohhyn and @mfc – the latter also having raised concerns about the documentation here).
Also saw https://github.com/2factorauth/twofactorauth/issues/1592 about this topic, where some discussion sparked, but I also did not see any conclusion or so there…
The issue has been raised by @kilasuit in https://github.com/2factorauth/twofactorauth/issues/3234 and I really found this old issue by myself (https://github.com/2factorauth/twofactorauth/issues/1296) about a closed-source self-hosted forum software.

Also there is always the saying "we do not list self-hosted software", and sometimes you even point to the Contributing.md, which makes no sense, as this actually _is not_ documented in that file. You can merge this PR, of course, to change that – or re-evaluate the whole decision.

---

Finally, seeking through all these scattered issues, I found https://github.com/2factorauth/twofactorauth/issues/3476 (I hope I've cross-linked enough to make it easy to find them now :wink:), which looks promising, but has not been received many replies.

Carlgo11:
@rugk I haven't reviewed your PR yet as I'm working on listing all excluded categories in EXCLUSION.md.
Depending on how that turns it it might be better to have this text over there instead.  

Regarding your question why list self-hosted sites aren't listed; It began as a way to keep our lists somewhat small and manageable. There are billions of self-hosted services that pop up and just as quickly disappear.
I suppose this is somewhat fixed by the 200k alexa rule as mentioned in #3476.
Another reason is that TwoFactorAuth.org is targeted towards the consumer and not webmasters. (See our readme)
It's very unlikely that a general consumer would quit a social media platform because said platform don't support 2FA and instead host their own Mastodon server, or host their own ownCloud server because their cloud storage provider don't support 2FA.  


I use self hosted stuff a lot. I have my own mail server and cloud storage using self-hosted services but I'm not a general consumer. And I'm guessing neither are you.

If you grabbed someone randomly off the streets and explained to them about the benefits of 2FA and the risks of not having it I'm very skeptic that you'd manage to get them to host their own things.

Anyways, I might be wrong here. @RichJeanes suggested we'd remove the self-hosting exception in his issue over at #3476. You should vote on it!  

By the way, I'm genuinely curious were all of you came from. Every time a self-hosted thing is mentioned here, there's always a bunch of new people commenting. Like a bot army just waiting to pitch in.
@rugk is the only one here that's not new so I'm very curious how everyone else found this PR.

//Carl

jancborchardt:
@Carlgo11 well, I am designer on Nextcloud and @Gargron works on Mastodon, so we are definitely not bots. :wink: (Sure, we are biased too. ;)

Some of our points are:
- You are essentially running a site for awareness about security and privacy. _Excluding_ things like Mastodon and Nextcloud which seem to fit the bill perfectly just because there are multiple providers seems odd.
- Both Mastodon and Nextcloud, while also possible to self-host, have easy ways to sign up with providers. Mastodon at https://joinmastodon.org/#getting-started and Nextcloud at https://nextcloud.com/signup/
- There are a whole lot of Cryptocurrencies listed, so you could argue that’s also quite nerdy and not easily understandable if you "grabbed someone randomly off the streets". ;) Same for "Cloud computing" and the "Developer" categories.
- You already list open source apps like Wordpress (with their .com instance) and Minds (with their .com instance).

Hope that makes it more understandable why it feels strange that among these rules, Mastodon and Nextcloud are excluded even though they are quite popular, and arguably less nerdy and/or easier to sign up to for a "general consumer" than a lot of other services on this list.

Gargron:
I understand it's frustrating when a lot of people who have never appeared in your repository before come in with opinions. Anyway, we're here because this issue was discussed on Mastodon, a decentralized social network with close to 2 million users spread over around 2,000 servers.

mxxcon (commented):
@rugk I take a problem with the title of your pull-request. Your implication that we some how lie about what gets listed is extremely inaccurate and offensive if not to all the contributors, then at least to me!

It has been a long-standing problem that nobody seems to  be willing to help addressing but only complain about.
It's been discussed multiple times in the past. If you can come up with a way for us to accurately yet easily represent 2fa status of self-hosted projects in all the possible different combinations that can arise from the use of external plugins and custom configurations then we'll be more than happy to re-evaluate the current guideline.

rugk:
You may not have lied, and I don't want to claim you did, but you just miss the information this pull requests adds. If you think, this is a way of lying, then this is your interpretation. In any case, you can see this PR just adds (missing) information.
If you don't think so, try to explain how one would see the requirement of "not include self-hosted sites" without the PR.

---

As for the whole discussion about the way you may include 2FA sites, I think https://github.com/2factorauth/twofactorauth/issues/3476 is the way to go. I've written down ideas, there.

jancborchardt:
@rugk just to be a bit more friendly, maybe change the "honest" in the title to "upfront". :slightly_smiling_face:

kilasuit:
@mxxcon - you've been in the past quite against this being opened up to state that self hosted items like Wordpress can *BE* configured to use MFA by **insisting** that it needs to give **Exact** Guidelines as how to do this.

Whereas I would prefer it suggest that these types of sites/applications/etc can be configured to allow use of MFA services and have that show as a potential link away from this site to the providers help pages (if given directly by the provider)

Which I've suggested prior as a solution as per #3234

mxxcon:
> 
> 
> @mxxcon - you've been in the past quite against this being opened up to state that self hosted items like Wordpress can _BE_ configured to use MFA by **insisting** that it needs to give **Exact** Guidelines as how to do this.

I have not insisted that it needs to give exact guidelines _how_ to enable 2fa, but rather that we need to have a way to accurately represent for all the information that would arise from listing such entries.

For example, if we list Mastodon project, do we list the project proper itself or a ton of different specific servers running it? If we list the project proper, but the specific instance doesn't have 2fa enable, then the info on this site is not accurate for that user. If we list specific implementations, many of them will be under 200k alexa rank and won't be listed anyway.

rugk:
> If we list the project proper, but the specific instance doesn't have 2fa enable, then the info on this site is not accurate for that user.

You can have a "note" for that. Or in your example, that's not a concern, because Mastodon enables 2FA by default. (I don't even know if you can disable it – can you, @Gargron?)

IMHO, BTW, yes do list the "project itself".

And, as said thousand of times in this thread already, you already list some of them, i.e. Bugzilla and GitLab. You can also self-host them.

kmpoppe:
Closing this PR in favor of #3632.

PR #3566, opened 2018-11-29
Revise header.html for easier readability for future updates

kenman345:
Saw this and thought it was a bit hard to read for no good reason.

Also, I revised the bottom section in the same way as the top even though it never is shown/rendered right now. Didnt know enough to determine if we should put it back in or remove the code block entirely.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -4,27 +4,6 @@
     Home
   </a>
 
-    {% if include.page == 'home' %}
-    <div class="ui dropdown item">
-      Categories <i class="dropdown icon"></i>
-      <div class="menu">
-        {% for section in site.data.main.sections %}
-        {% if forloop.first %}
-        <a href="#" class="item">
-          <i class="{{ section.icon }} icon"></i>
-          {{ section.title }}
-        </a>
-        {% else %}
-        <a href="#{{ section.id }}" class="item">
-          <i class="{{ section.icon }} icon"></i>
-          {{ section.title }}
-        </a>
-        {% endif %}
-        {% endfor %}
-      </div>
-    </div>
-    {% endif %}
-
     <div class="right menu">
       <a href="{{ site.repo }}" class="item">
         <i class="github alternate icon"></i>
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -1,12 +1,8 @@
 <div class="ui teal inverted menu">
-  {% if include.page == 'home' %}
-  <a href="/" class="active item">
-    {% else %}
-    <a href="/" class="item">
-      {% endif %}
-      <i class="home icon"></i>
-      Home
-    </a>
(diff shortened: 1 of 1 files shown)


fpigerre (dismissed):

kenman345:
I know this is approved, and the top of the page has value but in another PR we mentioned removing dead code. Can we get this merged so I can remove the bottom half out cleanly?

_includes/header.html
```diff
@@ -1,12 +1,8 @@
 <div class="ui teal inverted menu">
-  {% if include.page == 'home' %}
-  <a href="/" class="active item">
-    {% else %}
-    <a href="/" class="item">
-      {% endif %}
-      <i class="home icon"></i>
-      Home
-    </a>
+  <a href="/" class="{% if include.page == 'home' %}active {% endif %}item">
```
Carlgo11: Is this even used?

_includes/header.html
```diff
@@ -1,12 +1,8 @@
...
+    <i class="home icon"></i>
+    Home
+  </a>
 
     {% if include.page == 'home' %}
```
Carlgo11: If I'm not mistaken 'home' is the old main page so there's not really any reason for keeping this section.

_includes/header.html
```diff
@@ -38,14 +34,10 @@
 
     {% if 1 == 0 %}
     <!-- Don't show for right now -->
-    {% if include.page == 'restrictions' %}
```
Carlgo11, line 41: There's no restrictions page so this part can also be removed.

Carlgo11 (commented):
Maybe @psgs can confirm my comments before we remove anything.

Other than that everything looks good. 👍

_includes/header.html
fpigerre: Good catch, I completely forgot we removed this :+1:

fpigerre (dismissed):

kenman345:
this was appproved by one person Jan 17th, what the hold up?

Carlgo11:
https://github.com/2factorauth/twofactorauth/pull/3566/files#discussion_r247767107

_includes/header.html
kenman345: not sure what part you are saying to remove. Wanna give some line numbers? havent looked at this in a while and github showing things a bit weird for me.

Carlgo11: L7 to L26.

kenman345:
should be all set now!

Carlgo11 (approved):

jtagcat:
Is this dead?

kenman345:
Why dead? I revised this as recently as last week to make it in alignment
of what was asked. It’s ready to get a second approval and then merged

kmpoppe (approved):
Looks good

PR #3574, opened 2018-12-07
Allow both 644 and 664 permissions on images

Carlgo11:
Due to our latest Travis builds failing I think it would be a good idea to also allow 664 octal permissions.
By comparing the default permissions on different operative systems we've seen that some use 644 and some use 664.
The difference between 644 and 664 is that the group of the owner can edit the image. The group is usually just the owner so it shouldn't really matter here.

The result of this change caused rubocop to complain that the Assignment Branch Condition size was too high. I disabled the check for the method in question as I didn't want to redo the entire method or split it into two functions.

I also made a variable for the permissions for easier editing.

### DIFF
--- a/verify.rb
+++ b/verify.rb
@@ -45,6 +45,7 @@ def test_img(img, name, imgs)
 end
 
 # rubocop:disable Metrics/AbcSize
+# rubocop:disable Metrics/MethodLength
 def test_img_file(img)
   # Check image file extension and type
   error("#{img} is not using the #{@img_extension} format.")\
@@ -59,10 +60,15 @@ def test_img_file(img)
 
   # Check image permissions
   perms = File.stat(img).mode.to_s(8).split(//).last(3).join
-  error("#{img} permissions must be one of: #{@img_permissions.join(',')}. It is currently #{perms}.")\
+  # rubocop:disable Style/GuardClause
   unless @img_permissions.include?(perms)
+    error("#{img} permissions must be one of: #{@img_permissions.join(',')}. "\
+    "It is currently #{perms}.")
+  end
+  # rubocop:enable Style/GuardClause
 end
 # rubocop:enable Metrics/AbcSize
+# rubocop:enable Metrics/MethodLength
 
 # Load each section, check for errors such as invalid syntax
 # as well as if an image is missing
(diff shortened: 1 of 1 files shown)


fpigerre (dismissed):

stephengroat (dismissed):
while this temporarily solves the issue, i don't think this is a good idea. there's no obvious reason that this is happening (maybe a travis issue? not sure). probably should be looked into to determine an actual root cause

Carlgo11:
The root cause is that some OS use 664 as their default permissions for files and some use 644. Travis have changed to full VMs and apparently the new linux version they use have 664 as the default perms. See slack for their response.

verify.rb
```diff
@@ -54,10 +58,11 @@ def test_img_file(img)
   end
 
   # Check image permissions
-  perms = File.stat(img).mode
-  error("#{img} is not set 644. It is currently #{perms.to_s(8)}")\
-    unless perms.to_s(8) == '100644'
+  perms = File.stat(img).mode.to_s(8).split(//).last(3).join
+  error("#{img} is not set 644 or 664. It is currently #{perms}")\
```
braxtone: For future-proofing, this should be changed to use the new instance variable's values:
`error("#{img} permissions must be one of: #{@img_permissions.join(',')}. It is currently #{perms}") `

stephengroat:
> The root cause is that some OS use 664 as their default permissions for files and some use 644. Travis have changed to full VMs and apparently the new linux version they use have 664 as the default perms. See slack for their response.

@Carlgo11 could you post the slack response here? can't get into that service right now

since git tracks file permissions, if travis isn't respecting the file permissions in the repo, maybe we just shouldn't check for them at all?

Carlgo11:
@stephengroat it's not only the email but the conversation around it so it's a bit hard to post that here.

The thing with the permission checking is that we don't have files to be 777. git seems to ignore the difference between 664 and 644 and treat it as either one depending on the system.

email: https://twofactorauth.slack.com/archives/C09K577CK/p1544201333069700

discussion: https://twofactorauth.slack.com/archives/C09K577CK/p1544203837080100 (and below)

kenman345:
can someone please help me understand why such a check of permissions is necessary?

mxxcon:
> can someone please help me understand why such a check of permissions is necessary?

I could be mistaken and misremembering things, but I think GitHub would refuse to generate pages if permissions weren't to its liking...? 🤔

stephengroat (approved):

fpigerre (approved):

PR #3576, opened 2018-12-11
Remove indicator that USAA supports hardware 2FA

braxtone:
They don't support hardware tokens currently per the linked documentation.
https://www.usaa.com/inet/pages/security_token_logon_options

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -1051,7 +1051,6 @@ websites:
       img: usaa.png
       tfa: Yes
       sms: Yes
-      hardware: Yes
       email: Yes
       software: Yes
       doc: https://www.usaa.com/inet/pages/security_token_logon_options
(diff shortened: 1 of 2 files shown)


braxtone:
After all that fun, I just saw #3574 and we did similar things to solve the build problem. Happy to back those last two commits out if #3574 gets merged first.

jamcat22:
USAA does support hardware tokens through their brand name “CyberCode Token”. While they’ve changed instructions to show how to use Symantec VIP Access as a mobile app, you can still enroll VIP Access [hardware tokens](
https://www.amazon.com/dp/B06XY422T9) (serial numbers commonly beginning in VSST or BCHT) through the service.

braxtone:
Huh, well today I learned... By the slimmest of margins and oldest of solutions they do. Thanks for catching that @jamcat22

PR #3578, opened 2018-12-13
Add ICDSoft to hosting

korikori:
Hi, we would like to add ICDSoft to the Hosting category. The site is just short of Alexa 200k, but we do host over 60k sites, and we do support TFA, so I believe there is merit to including us. Thanks in advance.


brycx:
Hi @korikori. Thank you for the PR, but the site [does not](https://www.alexa.com/siteinfo/icdsoft.com) meet the 4th criteria of our contributing guidelines, so I'll have to close this. Feel free re-open the PR if it does in the future.

PR #3582, opened 2018-12-18
Add nectar without 2fa

mrjonstrong:
Add nectar to retail, no 2fa.


_data/retail.yml
```diff
@@ -423,4 +430,3 @@ websites:
       tfa: No
       twitter: zooplusDE
       facebook: zooplusde
-
```
mxxcon, line 426: Please preserve end of file line break

mxxcon (changes requested):

PR #3601, opened 2018-12-26
Capital One has no 2FA solution

puzzletron:
I'm a customer. There is no 2FA solution. They started something 2FA-like in 2015 but you can't get anything but username and password as a current customer.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -202,11 +202,7 @@ websites:
     - name: Capital One
       url: https://www.capitalone.com/
       img: capitalone.png
-      tfa: Yes
-      sms: Yes
-      email: Yes
-      software: Yes
-      doc: https://www.capitalone.com/identity-protection/swiftid/
+      tfa: No
 
     - name: CDC Federal Credit Union
       url: https://www.cdcfcu.com


dsoegijono (approved):

Carlgo11:
@puzzletron is [this page](https://github.com/2factorauth/twofactorauth/blob/ba66dbc6e6cd0f57589ae8c7b32a4ef013e8b348/notes/capitalone.html) invalid then?

puzzletron:
Yes. I also agree with the past comments from others. Thus far they seem to have missed the mark on 2FA.

mxxcon:
@puzzletron Doesn't #3460 address this?

puzzletron:
No, I do not think so. Whether one uses that custom SwiftID mobile application or not, if they detect "something amiss" -- see [https://www.capitalone.com/applications/identity-protection/swiftid/](https://www.capitalone.com/applications/identity-protection/swiftid/) for their exact wording -- then and only then will they attempt to automatically verify your identity another way. If you don't have the mobile app then they will text you. As a customer I can't set it up to always use 2FA on login, which is what I would think is required to claim 2FA on this project page.

mxxcon:
> 
> 
> As a customer I can't set it up to always use 2FA on login, which is what I would think is required to claim 2FA on this project page.

It is not a requirement. Unfortunately, too many sites act like that. Fortunately, it's most often are simply cookies, which you can do delete after each session and you'll be asked to provide 2fa.

ghost:
I would say this update does not match my experience. As stated by @mxxcon, 2FA doesn't have to be requested every time. I did see that they would enforce 2FA (SMS or email) when a new device was connected. At least on iPhone within a few days of registration deleting and reinstalling the application was not good enough to cause a re-prompt for 2FA. A wipe of the phone would cause the re-prompt for 2FA.

puzzletron:
I questioned the implementation because as a customer I can't set 2FA to be always-on. It is a secret sauce algorithm as to when one might be challenged to provide more than just a password to authenticate when the extra security is activated. If creative implementations are OK and this is the case elsewhere on this site then I understand, just reject the proposal.

PR #3613, opened 2019-01-08
Add Growing.IO

aiyanbo:


Carlgo11:
@aiyanbo if you still want to add growing.io to the list you just need to move it under the `Go Fund Me` listing and change the logo to 32x32 pixels (and maybe compress it depending on if it's larger than 2.5Kb or not)

PR #3637, opened 2019-01-17
Add Community America and First Community Credit Unions

bschwedler:


dsoegijono:
Hi @bschwedler, thanks for the PR. Can you please add documentation links for these entries?

[This](https://www.communityamerica.com/search/index.php?findtext=How+secure+is+online+banking%3F) seems like a good link for Community America.

I couldn't find one for First Community; do you know if there's one?

bschwedler:
I added docs that are linked off the online banking site.  The links are publicly accessible but I was only able to find them after logging in.  They seemed like the best resources.

mxxcon:
It would be best if you added these sites as 2 separate pull requests. That way issues with one won't delay merging of the other.

bschwedler:
Closing in favor of #3645 and #3646 which break out the individual sites.

PR #3656, opened 2019-01-24
Add .ruby-version to match travis version

Firefishy:
Add a .ruby-version to set local ruby version for rvm and rbenv users.


stephengroat:
@Firefishy thanks for the addition, great idea!

where did you get that `github` is running `2.4.5`? do you have a link to the source of that info? from https://github.com/github/pages-gem, it seems like its actually `2.5.3`

Firefishy:
@stephengroat Updated ruby to `2.5.3` now to match version across all 3: Local dev, TravisCI and github.

.travis.yml
```diff
@@ -1,5 +1,5 @@
 sudo: false
-rvm: 2.4.0
+rvm: 2.5.3
```
stephengroat:
```suggestion
```
have travis use `.ruby-version`
https://docs.travis-ci.com/user/languages/ruby/#using-ruby-version

stephengroat (approved):
👍

.ruby-version
```diff
@@ -0,0 +1 @@
+2.5.3
```
Carlgo11, line 1: What are the benefits of having the ruby version in `.ruby-version` as opposed to .travis.yml and Gemfile?

stephengroat: travis doesn't recognize `Gemfile` ruby version, `rvm` and travis will use `.ruby_version`

Carlgo11 (approved):

Issue #3644, opened 2019-01-18
Wells Fargo only accepts RSA SecurID hardware tokens (and not Yubikey, or anything else)


zoltan57:
For Wells Fargo I suggest replacing the check mark under "Hardware Tokens" with "RSA SecurID only".

zoltan57:
And Charles Schwab only accepts Symantec Hard Tokens and software tokens from the Symantec VIP Access app. I think it is important to push these vendors to a common standard where we consumers can choose the hardware token or MFA app of our choice, and indicating the limitations on your very informative website would be good start.

RichJeanes:
We have left the columns as generic as possible for a reason. Having a separate column for each various 2FA implementation would quickly balloon and we never really settled on a good way to represent specific solutions within each column.

For the moment that leaves adding an exception to the site or implementing a solution that allows for a good, clean representation of specific implementations within columns. Discussion or Pull Requests on both ideas (or others) is absolutely welcome.

PR #3667, opened 2019-01-28
Update Token exceptions

kmpoppe:
Regarding Issue #3644

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -1119,6 +1119,8 @@ websites:
       phone: Yes
       hardware: Yes
       doc: https://www.wellsfargo.com/privacy-security/advanced-access
+      exceptions:
+       text: "Hardware token: specific vendor only"
 
     - name: Wescom Credit Union
       url: https://www.wescom.org/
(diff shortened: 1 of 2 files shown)


mxxcon:
I'm not sure if this is necessary.
For Charles Schwab we have a whole dedicated page where it can be explained.
And for both just because they use some specific vendor I'm not sure we need to list that.
`exception` attribute is more for situations where 2fa is available only to paid accounts or only in certain countries...

But i won't block merging this if others think otherwise. 🤷‍♂️

kmpoppe:
I see what you mean. I'll also submit to what others think. If we don't need these exceptions, I'd be inclined to say that #3644 should be closed for good for the time being, right?

Carlgo11:
As of right now the hardware section is for every time of authentication that's neither software or phone related with no specification of what types of hardware types are supported. The exception to this is if, for example, only a certain types of Yubikeys are accepted and there's no good documentation on that. See [investing.yml#L200](https://github.com/2factorauth/twofactorauth/blob/8c292d6cb957914dca43036a4450a75ce840ee73/_data/investing.yml#L200)

#3222 could potentially change that.

PR #3668, opened 2019-01-28
Update Rebel facebook handle

stephengroat:


kmpoppe:
Too bad I'm not allowed to approve/review ;-)
I find it interesting that on rebel.com the facebook-Link still goes to f/RebelCom but then redirects to the changed alias.

mxxcon (approved):

Carlgo11 (approved):

PR #3671, opened 2019-01-28
Add geocaching.com

kmpoppe:
Global Alexa Rank at time for checkin: 8,886.
No 2FA support.


mxxcon:
How would it fit into our https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria ?

kmpoppe:
@mxxcon 1 - Personal Image. Given the profile of a user that can be viewed once you signed up, you can roughly see the dates (and sometimes times of day) where a user has been, including relatively specific geo-coordinates.

mxxcon:
Isn't that the point of geocaching?

kmpoppe:
Well. The point of geocaching is to "find tupperware hidden in the woods" - or words to that end. Anyway, users mark geocaches as found. So this can be traced back to the person itself which is essentially a sensitve information. 
Additionally, the users are the ones that hide the geocaches and give the coordinates to the caches or to subpoints of a multi-station cache. 
An intruder into a user's profile could either change the visited caches or change coordinates of hidden caches at will, which could be used to illude people to visit a place they endanger themselves more than by the rules of the "game" allowed.

Again, if you think the site does not fit into that category, please close the PR, this will not ruin my day ;-)

PR #3676, opened 2019-01-29
Ampersand correction

kmpoppe:
When a site name contains an ampersand (&) the normal HTML-notation would be to use &amp;. Yet, this &amp; will show up in the JSON-export and not the &.
Other examples in the data show, that the ampersand can be used without the HTML notation in the name when it's in single quotation marks (') - thus also resulting in a correct JSON-export.
All names with ampersands are now in single quotes.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -134,7 +134,7 @@ websites:
       phone: Yes
       email: Yes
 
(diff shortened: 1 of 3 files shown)


kmpoppe:
There's something wrong in my line of reasoning ;-)

PR #3694, opened 2019-02-05
Update Category icons

Carlgo11:
I went over the category logos we use. As Semantic have added more icons I replaced those that didn't explain what the category was about in favor of a new icon.
There are some that I'm uncertain about.
* Backup and Sync - The new one is of a cloud since it's cloud storage. I don't know which way the arrow should point. Would it be better if it pointed upwards instead?
* Entertainment - The music icon of the old one looked uncentered due to the way the music note is flipped diagonally. The new one looks more centered (although not 100% to my liking) and it fits a broader range of services listed under the entertainment category.
* Investing - There are currently 3 currency icons (Banking, Finance & Investing) and so I thought it would be better if at least one of them changed icon. As investments and stocks are somewhat similar I chose a chart icon.
* IoT - The old plug icon doesn't describe the Internet part of "Internet of Things" so I oped for a WiFi symbol instead.
* Identity Management - The old one was too generic for my taste and so I oped for an ID card instead as I think that better describes what the manager actually handles.
* Logistics - The Logistics category isn't in this PR and accidentally got mixed in when the screenshot was taken. Disregard this category.
* Other - the old "Lab" icon didn't really fit the category so I opted for something a little more generic. I'm not certain this is the best icon for the category but I couldn't find anything better. Suggestions are welcome 😄 
* Retail - Just updated the icon name.
* Security - To make the icon look less like a hand bag I chose the halfway opened lock.
* VPN - The "Link" icon is more associated with WWW than VPN services and so I picked an eye icon. I also thought about the "exchange" icon as it's a two way communication stream. Thoughts? 

Here is the current page:
old versoin
Here are the suggested changes:
new version

### DIFF
--- a/_data/sections.yml
+++ b/_data/sections.yml
@@ -1,6 +1,6 @@
 - id: backup
   title: Backup and Sync
-  icon: hdd outline
+  icon: cloud download
 
 - id: banking
   title: Banking
@@ -40,7 +40,7 @@
 
 - id: entertainment
   title: Entertainment
-  icon: music
+  icon: play
 
 - id: finance
   title: Finance
@@ -48,7 +48,7 @@
 
 - id: food
   title: Food
-  icon: coffee
+  icon: utensils
 
 - id: gaming
   title: Gaming
@@ -60,7 +60,7 @@
 
 - id: health
   title: Health
-  icon: medkit
+  icon: heartbeat
 
 - id: hosting
   title: Hosting/VPS
@@ -68,15 +68,15 @@
 
 - id: identity
   title: Identity Management
-  icon: user
+  icon: id card
 
 - id: investing
   title: Investing
-  icon: basic money
+  icon: chart line
 
 - id: iot
   title: IoT
-  icon: plug
+  icon: wifi
 
 - id: legal
   title: Legal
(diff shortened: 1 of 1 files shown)


mxxcon:
All suggestions are good except for the VPN icon. I understand the logic behind picking it, but I'm not sure it's the right association. Perhaps `shield alternate` would work better?

Other ideas for Backup category could be `hdd` icon.

kmpoppe:
I'd also opt for `shield alternate` for VPN. When I set up that category I wasn't aware of the actual icon set that was used, now knowing the Semantic UI is used, that fits way better.
If backup only contains cloud backups, I'd go for `cloud upload`.
I'm not particularly happy with the entertainment icon, but what I had in mind for that (a theatre mask) isn't available, so the play button would be ok.

kmpoppe (approved):

kenman345:
This looks pretty good to me, I copied some of the icon changes for my repository. We had already made some of the same changes as well so it made sense for things like IOT to be wifi instead of plug. I'll be monitoring here for any other suggested changes, but

mxxcon (approved):

dsoegijono (approved):

PR #3704, opened 2019-02-07
Added Apple iCloud Email to email section

jwishnie:
Added Apple iCloud Email to email section.

Note: Apples Two-Factor auto via 'trusted devices' is a bit unique, but functions like an Authenticator app generating an OTP that the user types into the iCloud web page as their second-factor.

It is a slightly different UI but closest in user experience (and security level) to Software Token


mxxcon (dismissed):

jwishnie:
added second commit to update/correct iCloud entry in backup.yml

Carlgo11:
Is this any different to the `Apple iCloud` entry?
If not then this PR is a duplicate.

Many sites fit under several categories and so we choose the best fitting one. Apple iCloud could fit in Backup and Sync, Communication, Developer, Email, Entertainment, Gaming, Health, Identity Management, IoT, Payments, Remote Access, Retail, Security and Task Management but making separate entries for every site that fits several categories would be ludicrous.

jwishnie:
@Carlgo11 there are two commits in the request: 
- The first adds iCloud to `email.yml`
- The second updates the iCloud entry in `backup.yml` to account for Apple's updated 2FA system (which supports Phone now), and corrects a broken image

Two your point about sites fitting in many categories, I'm happy to argue that they SHOULD appear in all relevant categories because site viewers looking to sign up for a service may choose to select a category to see what services support strong 2FA.

In fact, that's what prompted me to make these changes: I selected 'email' to see what services support Hardware Tokens and noticed that iCloud was not on the list.

Users will not necessarily think to also search and see if a service appears in another category.

Multiple listings is a bit of a quick-fix.

I've suggested a couple of info design changes that could be longer term design improvements to address this: #3707 #3708 

I added the duplicate entry _specifically_ because I am using the site to help educate email users on email services that offer the strongest 2FA by sharing the anchored link: https://twofactorauth.org/#email

And the first iCloud user I shared this with asked where iCloud was!

jtagcat (changes requested):
I'd suggest not adding the email entry, just updating iCloud.

Carlgo11:
As iCloud already is listed in the backup category and as iCloud is primarily cloud storage I'm closing this PR.
If we, in the future, add a way to list a site in several different categories then I think me.com would be a great fit in the email category.
Until then, I'm closing this PR.

PR #3706, opened 2019-02-07
Update Apple's 2fa

mxxcon:
Seems like there's a difference between Apple's "2-step verification" and "2-factor authentication", so this update points to the newer (presumably more secure) one. It also doesn't seem to be listing any restricted carriers like previous one so removing `exceptions` block.

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -4,10 +4,9 @@ websites:
     img: icloud.png
     tfa: Yes
     sms: Yes
+    phone: Yes
     software: Yes
-    exceptions:
-        text: "See https://support.apple.com/HT202656 for a list of supported SMS carriers."
-    doc: https://support.apple.com/HT204152
+    doc: https://support.apple.com/en-us/HT204915
 
   - name: Backblaze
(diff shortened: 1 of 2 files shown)


_data/backup.yml
```diff
@@ -4,10 +4,9 @@ websites:
     img: icloud.png
     tfa: Yes
     sms: Yes
+    phone: Yes
     software: Yes
-    exceptions:
-        text: "See https://support.apple.com/HT202656 for a list of supported SMS carriers."
```
Carlgo11, line 9: The exception still applies to SMS based 2FA.

Carlgo11 (changes requested):

_data/backup.yml
mxxcon: I just spent 20min on the phone with Apple support to clarify this and he did not see any explicit mention 2FA is restricted to some carriers the way 2SV is.

Carlgo11 (approved):

kmpoppe (approved):

PR #3709, opened 2019-02-07
Update Comcast software token support & mobile app exception

jakejarvis:
Comcast has added software token support, but it requires their own [Xfinity Authenticator](https://xfinity.com/support/articles/multi-factor-authentication-xfinity-authenticator-setup) app.

### DIFF
--- a/_data/utilities.yml
+++ b/_data/utilities.yml
@@ -71,8 +71,9 @@ websites:
       tfa: Yes
       sms: Yes
       email: Yes
+      software: Yes
       exceptions:
-          text: "SMS-capable phone and a non-Comcast.net email required for 2FA."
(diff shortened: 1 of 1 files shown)


Carlgo11 (commented):
So it's not based on RFC-6238?

jakejarvis:
Not sure if it follows TOTP standards behind the scenes (you definitely know more about this distinguishment than I do) but there's no way to get a seed to use with any 2FA app. The app is tied to my Xfinity account login credentials, not a QR code/TOTP secret.

I skimmed other services that are listed here which use their own proprietary app and there seems to be an exception on those. If the exception is overkill and the link to the docs where it mentions this is sufficient, I can remove that and just add the software checkmark. :)

Carlgo11 (approved):

mxxcon (approved):

PR #3710, opened 2019-02-07
Update USAA exception with Symantec VIP app requirement

jakejarvis:
USAA annoyingly requires a Symantec app ([Symantec VIP](https://vip.symantec.com/)) to get the software token.

Info here: https://www.usaa.com/inet/pages/security_token_logon_options (CyberCode Token tab)

PR #3711, opened 2019-02-07
Update Twitch exception with Authy app requirement

jakejarvis:
One more! Twitch software authentication requires the use of Authy.

See "Why Authy?" section of this blog post: https://blog.twitch.tv/two-factor-authentication-now-available-on-your-twitch-account-b03300862ba7

### DIFF
--- a/_data/entertainment.yml
+++ b/_data/entertainment.yml
@@ -173,7 +173,7 @@ websites:
       software: Yes
       doc: http://blog.twitch.tv/2015/11/two-factor-authentication-now-available-on-your-twitch-account/
       exceptions:
-          text: "To activate two factor authentication, you must provide a mobile phone number."
+          text: "To activate two factor authentication, you must provide a mobile phone number. Software authentication requires the Authy app."
 
     - name: Vimeo
       url: https://vimeo.com/


kmpoppe:
Twitch 2FA works totally fine with a qr code scanned in 1password, Authy is
not required IMHO

Am Do., 7. Feb. 2019, 19:19 hat Jake Jarvis <notifications@github.com>
geschrieben:

jakejarvis:
Hmm, I just tried disabling and re-enabling 2FA  on my account and that doesn't seem to be possible. In that blog post they mention an "old method" so maybe this is a recent change?

screen shot 2019-02-07 at 1 28 27 pm

screen shot 2019-02-07 at 1 29 11 pm

kmpoppe:
I enabled it about 2 weeks ago. Strange. Would need to check when I return
home Saturday. Could we postpone that PR untl then?

Am Do., 7. Feb. 2019, 19:29 hat Jake Jarvis <notifications@github.com>
geschrieben:

jakejarvis:
Sounds good to me!

kmpoppe (approved):
Checked again, I was just wrong. Authy is the only software token that'll work with twitch.

jakejarvis:
Just asked in #3710 whether use of specific/proprietary apps should be noted. With so many apps exclusively requiring Authy (unfortuantely), we should probably declare a universal rule on whether this rises to the point of adding an exception to them rather than just linking to the documentation that mentions Authy.

I think the requirement of a vendor-specific app (like Comcast, Blizzard, Steam, etc. have) should definitely be noted, but Authy is a little different than those.

Just a thought.

Carlgo11 (approved):

dsoegijono (approved):

Issue #3643, opened 2019-01-17
Open Links in a New Tab


OLLI-S:
Please open all links in a new browser tab.
It is annoying when I open a documentation to set up 2FA and have to press "back" several times to get back to your Two Factor Auth List.

RichJeanes:
This has been suggested before and I believe the current mentality is to leave that option to the user instead of forcing it to a new tab. If you middle click the link, it will open in a new tab (alternatively, right click -> Open in New Tab).

OLLI-S:
OK, then close this issue.
But normally users click with the left mouse-button and the new page should normally open in a new tab.
I work as Usability-Engineer and UX expert.

kmpoppe:
@OLLI-S I'm also always annoyed by websites linking away from themselves - twofactorauth.org seems to have no new-window-links at all (neither to GitHub, Josh's site and the other websites). I suggest you close this Issue, maybe there'll be a more clever idea to decide whether the user wants a new window or not in the future!

OLLI-S:
Closed

PR #3712, opened 2019-02-07
Make documentation links open in new tab

jakejarvis:
The icon used for documentation links (the box with an arrow pointing out) [generally means a link will open with a new tab](https://www.google.com/search?q=open+in+new+tab+icon&num=20&client=firefox-b-1-d&source=lnms&tbm=isch&sa=X). It always throws me off and makes it harder to return to the twofactorauth website.

(Use of the "noopener" attribute is [explained here](https://mathiasbynens.github.io/rel-noopener/), FYI.)

### DIFF
--- a/_includes/desktop-table.html
+++ b/_includes/desktop-table.html
@@ -45,7 +45,7 @@
 
         <td class="positive icon">
           {% if website.doc %}
-          <a href="{{ website.doc }}"><i class="external url link large icon"></i></a>
+          <a href="{{ website.doc }}" target="_blank" rel="noopener"><i class="external url link large icon"></i></a>
           {% endif %}
         </td>
 
--- a/_includes/mobile-table.html
+++ b/_includes/mobile-table.html
@@ -37,7 +37,7 @@
       </div>
       {% if website.doc %}
       <div class="button-group">
-        <a alt="Documentation" href="{{ website.doc }}" class="ui twitter mini button"><i class="large book icon"></i> Docs</a>
+        <a alt="Documentation" href="{{ website.doc }}" target="_blank" rel="noopener" class="ui twitter mini button"><i class="large book icon"></i> Docs</a>
       </div>
       {% endif %}
     </div>


Carlgo11:
This has been discussed previously in #3643 and the decisions was to keep it as it is.

jakejarvis:
Maybe we can change the icon then?

Carlgo11:
What icon replacement do you have in mind?

jakejarvis:
These from Semantic UI would make more sense to me, but I'll leave it totally up to you guys. :)

screen shot 2019-02-07 at 1 52 25 pm screen shot 2019-02-07 at 1 52 30 pm screen shot 2019-02-07 at 1 52 14 pm screen shot 2019-02-07 at 1 55 11 pm screen shot 2019-02-07 at 1 51 02 pm

Maybe even this one, which is pretty close to the current one but the arrow stays inside the box:

screen shot 2019-02-07 at 1 51 49 pm

I know I'm being _super_ nitpicky but the meaning of the current icon == "I don't need to hold the command/control key when clicking this" in my mind. I might be weird and the only person with that engraved in muscle memory but I'm curious to hear what others think!

jakejarvis:
Pinging @OLLI-S from that #3643 issue – are there any other icons from here that would make more sense to you? https://semantic-ui.com/elements/icon.html

OLLI-S:
I would prefer to open the links in a new tab and keep the current icon.
If you just change the icon, then many users will still click the link and wonder why the Two Factor Auth List is no longer opened.
Just my opinion.

But if you still don't want to change the behavior of these links and want to exchange the icons, then I would use the following icons from https://semantic-ui.com/elements/icon.html:

- **Book** Icon (because it represents a manual / instructions that you should red)
- **Info** Icon (tells you more information how to set 2FA up)

I personally wold prefer the **Book** icon.

jakejarvis:
@OLLI-S Good find, I like the book icon a lot too. 👍  

screen shot 2019-02-07 at 2 37 29 pm

(But definitely still prefer them opening in a new tab if possible. When I first found this site I wanted to enable 2FA on as many services as possible and went back and forth between the site/documentation as I'm sure many visitors do.)

kmpoppe:
I'm not OLLI but I voiced my Op to #3643 as well.

If @carlgo11 and other members are open for an icon change my favourite would be `sticky note outline`, the current icon really makes me think we open a new window.

jakejarvis:
For the record, I agree with what was said in #3643:

> the current mentality is to leave that option to the user instead of forcing it to a new tab

...but when giving users a choice, the default choice that we make for them should be as obvious and intuitive as possible.

Any chance of re-opening this or #3643 to get some more thoughts? I respect any final decision but getting more input is always valuable. :)

mxxcon:
https://ux.stackexchange.com/questions/104154/should-external-link-open-in-same-tab-or-new-tab
And I saw multiple other articles stating to leave default behavior to the browser and let user decide.

jamcat22:
I think it makes sense to keep the link opening in the current tab, but change the icon to the book icon. We currently use that icon on the mobile site, and I agree that the current icon suggests the link being opened in a new tab from a symbolism perspective.
Image of mobile TwoFactorAuth.org site showing book icon in docs button

PR #3713, opened 2019-02-07
Add StackPath

scotwells:


stephengroat:
@scotwells love the pr! think this might fit better under `cloud`? i'm not 100% familiar with their offerings, but from a quick, cursory glance, they seem to be like DigitalOcean 

(just saw that you work there, please feel free to correct me about whatever i'm wrong about, you're clearly the expert 😄 )

mxxcon (dismissed):

PR #3743, opened 2019-02-20
Add Finnair

jerr0328:
Finnair does not provide official documentation about this, but they sent out an email with a direct link to it, and only other mention is in their Privacy Policy.


Carlgo11 (approved):
I couldn't find any documentation but for reference, here's a screenshot of the setup page screenshot

dsoegijono (approved):

PR #3754, opened 2019-03-12
Update 1Password 2FA options

### LINKED ISSUE #3753: 1Password does support hardware 2FA
1Password advertises hardware 2FA support

- <https://support.1password.com/two-factor-authentication/>
- <https://support.1password.com/yubikey/>
- <https://1password.com/yubikey/>
- <https://www.yubico.com/works-with-yubikey/catalog/1password/>

06kellyjac:
Added Hardware 2FA to 1Password

This resolves #3753

### DIFF
--- a/_data/identity.yml
+++ b/_data/identity.yml
@@ -4,6 +4,8 @@ websites:
       img: 1password.png
       tfa: Yes
       software: Yes
+      exceptions:
+        text: "1Password advertises hardware 2FA on Yubikey but it's just software TOTP. Teams using Duo with 1Password can't enable additional 2FA as it's already handled by Duo"
       doc: https://support.1password.com/two-factor-authentication/
 
     - name: Bitium


Carlgo11 (dismissed):

jamcat22 (dismissed):
I’m not sure we should list 1Password as supporting hardware-based 2FA, as they rely on using YubiKey Authenticator to allow RFC 6238-based (aka Google Authenticator) TOTP secrets to be stored within the hardware key. Typically, when we list services, we consider RFC 6238-based 2FA to be software, not hardware, otherwise almost all of the software-based 2FA services we list which use this method would also be listed as supporting hardware-based 2FA. Further discussion on this topic has been covered in #3144.

We could make an exception since 1Password provides specific documentation oriented towards YubiKey support (even though they don’t support YubiKey OTP or FIDO-based 2FA). What do y’all think?

RichJeanes:
Yeah, if we allow that as "Hardware", we'd have to apply that to every site that has TOTP. I use my YubiKey for TOTP instead of an app, but it's certainly not something I think would be helpful to list for the lay-audience .

Honestly, that seems rather deceptive to me for 1Password and YubiCo to both be advertising is that way...

06kellyjac:
I see. I'd at least like some way to show twofactorauth.org know's there is a "hardware" option but say that it doesn't really count, because I saw the page and thought it was a mistake/needed an update (hence the PR).

Carlgo11:
Whoops, didn't mean to close the PR.
As @jamcat22 said, an exception would be better as they advertise hardware support. Otherwise we'll get a lot more issues and PRs about this.

RichJeanes:
But that exception would apply to EVERY site that supports TOTP. I don't think this is a special case just because they call out the YubiKey in their documentation. The YubiKey isn't even the only TOTP hardware solution out there (and arguably isn't the best hardware TOTP token, either).

Carlgo11:
@RichJeanes the exception to every other site is that they (deceptively?) advertise support for something which they don't support. I don't think @jamcat22 had `hardware: Yes` in mind, but rather something like
```YAML
exception:
  - text: "1Password also advertises hardware 2FA support through Yubikey software implementation."
```

RichJeanes:
Ah, yes, adding an "exception" text, not actually marking the Hardware column as an exception to the standard. Gotchya. Yes, I think that would be good to address the point for anyone who goes looking for it.

06kellyjac:
I've removed `hardware: Yes` and extended the exception to detail the "hardware" 2FA is just software TOTP

kmpoppe (approved):

Carlgo11 (approved):

PR #3762, opened 2019-03-21
Add Wasabi

radeksimko:
I'm aware there was a PR for Wasabi https://github.com/2factorauth/twofactorauth/pull/2998 which was closed mainly because Wasabi wasn't so popular back then.

According to current Alexa rank the site became a lot more popular in the past year and now fits within the "200k or less rank" bracket.

https://www.alexa.com/siteinfo/wasabi.com


Carlgo11:
If I'm understanding the documentation page correctly 2FA is only available for a "root user".
If that's the case then an exception text would be good.
Something like 
```YAML
exception:
  - text: "2FA only available for the Root User."
```

radeksimko:
@Carlgo11 To my knowledge MFA is available for users too:

Screenshot 2019-03-25 at 11 14 57

but I wasn't able to find any publicly-linkable documentation for it unfortunately 😞

Carlgo11:
Ah okay. That’s fine then

jtagcat (changes requested):
Image is not below 1.5kb.

jtagcat (changes requested):
wasabi
png could be 60% smaller, use this

radeksimko:
@vc7ee @Carlgo11 PTAL

jtagcat (approved):

kmpoppe (approved):

Carlgo11 (approved):

PR #3768, opened 2019-03-27
Update TD Waterhouse/Direct Investing

ghost:
updated td Waterhouse 2fa

### DIFF
--- a/_data/investing.yml
+++ b/_data/investing.yml
@@ -172,9 +172,11 @@ websites:
 
     - name: TD Waterhouse
       url: https://www.tdwaterhouse.ca
-      twitter: TD_Canada
+      twitter: TD_DirectInvest
       img: tdameritrade.png
-      tfa: No
+      tfa: Yes
+      sms: Yes
+      phone: Yes
 
     - name: TIAA-CREF
       url: https://www.tiaa-cref.org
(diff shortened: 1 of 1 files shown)


ghost (commented):
added doc

Carlgo11 (dismissed):
It looks like you're trying to add a branch of TD Canada called TD Direct Investing. Is this correct?
It also looks like TD Canada have separate logins for banking and investing.

I suggest you add TD Direct Investing as a separate entity instead of modifying the existing TD Canada entity.

_data/investing.yml
```diff
@@ -172,9 +172,12 @@ websites:
 
     - name: TD Waterhouse
```
Carlgo11, line 173: Looks like they're not using "Waterhouse" in their name any longer.
```suggestion
    - name: TD Direct Investing
```

_data/investing.yml
```diff
@@ -172,9 +172,12 @@ websites:
...
       url: https://www.tdwaterhouse.ca
```
Carlgo11, line 174:
```suggestion
       url: https://tddirectinvesting.ca
```

_data/investing.yml
```diff
@@ -172,9 +172,12 @@ websites:
...
-      twitter: TD_Canada
+      twitter: TD_DirectInvest
```
Carlgo11, line 175: Twitter handles aren't used when the site supports 2FA.
```suggestion
```

Carlgo11 (changes requested):
Just a few things that needs to be changed.

PR #3780, opened 2019-04-05
Update mailbox.org doc link

kmpoppe:

PR #3788, opened 2019-04-07
Update doc link for mailbox.org

rscharfe:
The old one leads to a 404 error page.


kmpoppe (changes requested):
Identical with #3780 by me ;-)

PR #3794, opened 2019-04-10
LHV

jtagcat:
"LHV is the biggest domestic financial group" (in Estonia)


kmpoppe (changes requested):
Hello @RasmusKallas !
The image you supplied is larger than 32x32 pixels and 2500 bytes, which is not allowed under the current rules. Please use https://tinypng.com/ to reduce the size of the image and change it in your branch by a new checkin. Additionally the entry you added is not in alphabetical order within the YML-file. Would you please correct this as well?
Thanks for your contribution!

jtagcat:
Issues fixed.

kmpoppe:
Hi Rasmus, you checked in the Migadu change in the same branch as LHV.

Would you please separate that change into a different branch and pull request?
Kai

jtagcat:
Sorry, I don't see a way to undo the last change :/

kmpoppe:
See here:
https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-a-pull-request

What you want to do is checkout the branch, revert the commit id and push
it back.

Kai

Rasmus <notifications@github.com> schrieb am Mi., 10. Apr. 2019, 18:20:

jtagcat:
Thanks, tough it could be done from the web..

PR #3803, opened 2019-04-10
Add Floatplane

jtagcat:


Carlgo11 (approved):

kmpoppe (approved):

ukanuk:
Floatplane does have 2FA and has had it for awhile. Maybe it didn't back in April when this issue was originally created? The "authenticator app" uses the standard QR code which can be scanned with any regular 2FA app like Aegis/Authy/FreeOTP/etc. 

image

PR #3807, opened 2019-04-13
Namecheap now supports hardware 2FA

golf1052:
See https://www.namecheap.com/support/knowledgebase/article.aspx/10102/45/how-can-i-use-the-u2f-method-for-twofactor-authentication

### DIFF
--- a/_data/domains.yml
+++ b/_data/domains.yml
@@ -343,7 +343,6 @@ websites:
       img: namecheap.png
       tfa: Yes
       sms: Yes
-      phone: Yes
       software: Yes
       hardware: Yes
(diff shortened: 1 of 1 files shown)


kmpoppe (changes requested):
Hi @golf1052 ,
Going through the documentation, it looks like the `phone`-option (an automated phone call to give you the two factor code) does not exist (anymore). Would you please remove that option from the file while you're at it?

kmpoppe (approved):
Thanks!

jtagcat (approved):

RichJeanes (approved):

PR #3810, opened 2019-04-14
Add Ceska Sporitelna (bank)

radeksimko:


jtagcat:
csas-cz
I suggest you reduce the png size with this image.

kmpoppe:
@radeksimko Is there any way to get an English description of how 2FA works at Ceska Sporitelna?

Carlgo11 (changes requested):
As parts of the documentation text can't be translated. Please explain a bit how the software and sms 2fa works, when it's applied and who it's applied to.

Also change the logo to this one:
csas

radeksimko:
@Carlgo11 logo changed. Thanks for the suggestion. I was actually looking for this simple logo on Google Images but couldn't find one.

Although the site itself (including the e-banking) has english version there doesn't seem to be english version of that linked page with as much detail. I will therefore explain how it works below.

By default user (a bank account holder) is forced to enable SMS-based 2FA for logging into the account _and_ confirming any money transfer or changing information about their accounts and themselves.

The software based 2FA involves downloading & installing a custom application (iOS/Android) which needs to be paired up with the account (via SMS) and it automatically disables SMS-based 2FA once enabled. From there user just receives notification via the app to their phone and can confirm by entering a static PIN or scanning a fingerprint and pressing "Confirm" button.

Let me know if you need further explanation.

Carlgo11:
@radeksimko I actually just used their favicon converted to png format. (https://www.csas.cz/content/dam/cz/csas/www_csas_cz/obrazky/ikony/favicon.ico) 😄 

Thanks for the clarification, Could you add an exception note with the text "Software implementation requires use of the X app."? With X being the name of their app.

radeksimko:
@Carlgo11 Note added. PTAL.

Carlgo11 (approved):

kmpoppe (approved):

PR #3811, opened 2019-04-14
Update eBay - documentation link

kmpoppe:
This community-link describes where to go to activate 2FA - it seems there's no "offical" entry in eBay's help for this 🙄

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -131,6 +131,7 @@ websites:
       img: ebay.png
       tfa: Yes
       sms: Yes
+      doc: https://community.ebay.com/t5/Getting-Started/HOW-TO-add-Two-Factor-Authentication-2FA-when-signing-in-on-EBay/td-p/27465876
 
     - name: ePRICE
       url: https://www.eprice.it/


dsoegijono:
Is this entry specifically for the US site? I just tried on the eBay app in AU, and I could turn on 2fa from settings, but it only gives me the option to verify using the app instead of text. Might worth noting this?

kmpoppe:
Welp, for my German account I followed the description to go to the personal account information, security setting and so forth - given using the website from a desktop browser.
In the German eBay app, I'm unable to change any account settings apart from the delivery address.
It nags me, that eBay completely misses out on describing what to do :-/

kmpoppe:
@dsoegijono I wrote to eBay Germany on Twitter if they have a better help page, yet, as of today it still seems that the link provided in the last entry (https://community.ebay.com/t5/Getting-Started/HOW-TO-add-Two-Factor-Authentication-2FA-when-signing-in-on-EBay/m-p/29076326/highlight/true#M20521) is still the ultimate solution.

kmpoppe:
grafik
Translation: Hi and thank you for your question, we are currently changing the process for 2FA and there will be a corresponding help-page mid-May. ;-) ~ Steffi

We should be seeing a help page soon, which then makes it possible to check these between country-sites.

LadySmith:
FYI: eBay mobile app can be used alternative to SMS authentication. But enabling it automatically disable SMS authentication.

kmpoppe:
Replaced by #3859

PR #3817, opened 2019-04-19
Update TransIP

### LINKED ISSUE #3755: TransIP doesn't support hardware tokens

kmpoppe:
Closes #3755.

While the information in the #3755 are quite sparse, I read from the (now new) documentation-link for software-2FA and https://www.transip.nl/knowledgebase/artikel/163 for SMS-2FA that these are the only methods available.

### DIFF
--- a/_data/hosting.yml
+++ b/_data/hosting.yml
@@ -465,7 +465,6 @@ websites:
       url: https://www.transip.nl/
       img: transip.png
       tfa: Yes
-      sms: Yes
       software: Yes
       doc: https://www.transip.nl/knowledgebase/artikel/162
 
--- a/_data/hosting.yml
+++ b/_data/hosting.yml
@@ -467,8 +467,7 @@ websites:
       tfa: Yes
       sms: Yes
       software: Yes
-      hardware: Yes
-      doc: https://www.transip.nl/vragen/574-hoe-stel-two-factor-authentication/
+      doc: https://www.transip.nl/knowledgebase/artikel/162
 
     - name: UKFast
       url: https://www.ukfast.co.uk/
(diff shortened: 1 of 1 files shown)


Carlgo11 (dismissed):

_data/hosting.yml
```diff
@@ -436,8 +436,7 @@ websites:
       tfa: Yes
       sms: Yes
```
jamcat22: Because 2FA requires setting up a software method and SMS is only used as a backup method, I'd categorize this site as not supporting SMS 2FA.
```suggestion
```

_data/hosting.yml
```diff
@@ -436,8 +436,7 @@ websites:
...
       software: Yes
-      hardware: Yes
-      doc: https://www.transip.nl/vragen/574-hoe-stel-two-factor-authentication/
+      doc: https://www.transip.nl/knowledgebase/artikel/162
```
jamcat22: Based on their doc for [logging in to their STACK product](https://www.transip.nl/knowledgebase/artikel/234-de-webinterface-van-stack/) and for [setting up 2FA on STACK](https://www.transip.nl/knowledgebase/artikel/1531-two-factor-authentication-2fa-voor-stack/), I think we should add an exception that enabling 2FA on one doesn't enable it on the other. This is similar to how we handle Andrews & Arnold and Hostiso.
```suggestion
      exceptions:
          text: "While both the main control panel and STACK support 2FA, each system is separate and requires its own setup."
```

jamcat22 (dismissed):

Carlgo11 (dismissed):

Carlgo11 (dismissed):

Carlgo11 (approved):

PR #3830, opened 2019-04-28
Update multiple http to https links

kmpoppe:
Changed a bunch of http to https links, as already proposed by @einsteinsfool in #3134, yet the PR got a whole lot of conflicts, which is why we agreed to close it.

Also, where possible, URLs where shortened (when `.../<someid>/some-google-optimzed-text` yielded the same page as `.../<someid>`).

PR #3836, opened 2019-05-02
Update banking.yml

step7:
Edited info regarding security in Intesa San Paolo link


kmpoppe:
Thanks for your contribution @step7 !

I have incorporated your change into my PR #3830, because it would otherwise conflict. I'll therefore close this PR now.

// Kai

PR #3838, opened 2019-05-03
Update PayPal

Jawshy:
- PayPal does not indicate that it supports hardware-based 2FA. 
- Updated [documentation link](https://www.paypal.com/us/smarthelp/article/what-is-2-step-verification-faq4057) includes steps to setup 2FA. 
- Minor grammar fixes.


RichJeanes:
Do we have confirmation that Software Authentication is as region limited as SMS?

Carlgo11 (approved):
I think this is a good revision for now. If it happens to be the case that software 2FA is available globally we could always change the phrasing later 🤷‍♂

kmpoppe (approved):

PR #3849, opened 2019-05-13
Update Zurich Cantonal Bank (ZKB)

### LINKED ISSUE #3848: 2fa that isn't any of the categories?
I have a question concerning one specific bank: Zurich Cantonal Bank (ZKB).

As far as I can tell most of the data is wrong. While they require every customer to use 2fa for online banking, it's neither through SMS, Phone Call, Email, Hardware Token or Software Token (at least the way I know it).

Basically how the system works:
1. Login uname & pwd
2. Scan a "photoTan" on the screen with an authenticated device (mobile phone or dedicated device).
3. Enter the TAN (transaction authentication number) in the ebanking.

If you have done all of the above succesfully, you are then logged in.

For me the definition of software token would probably be the closest, but the system is completely different from most other software tokens I know. And I assume it would also mess with 1Password, since they except software tokens to be TOTP.

What should we do here?

olikami:
* Remove the phone, sms & email options
* Removed whitespace in the image line

This implements and closes #3848

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -1232,4 +1232,6 @@ websites:
       tfa: Yes
       software: Yes
       hardware: Yes
+      exceptions:
+          text: "Software & Hardware authentication requires the ZKB photoTAN app or PhotoTAN device."
       doc: https://www.zkb.ch/de/pr/pk/efinance/ebanking/login-verfahren.html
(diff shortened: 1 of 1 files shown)


kmpoppe (dismissed):

Carlgo11 (changes requested):
Please add an exception text. Something like 'Software & Hardware authentication requires the ZKB app or PhotoTAN device.' (Or whatever the app and device is called. I only studied German for two weeks in elementary :wink:)

Carlgo11 (approved):

kmpoppe (approved):

PR #3851, opened 2019-05-13
Update Black Desert Online

LadySmith:
Game had 2fa since 2018.

PR #3857, opened 2019-05-16
Add Bohemia Interactive

LadySmith:
Alexa Rank: 34,730
Seems like document is missing. However, it is not difficult to activate 2fa. It does mention on account settings page.

I hope two different pulls on same file doesn't cause an error. (github illiterate here)
#3851


LadySmith:
- The site does include personal info with billing info.
- Deals with money but doesn't store credit card info. Uses an external website to check out.

Jawshy (approved):
Thanks for submitting a new site, @LadySmith. 

I also couldn't find documentation for enabling 2FA, but I can confirm that TOTP 2FA can be enabled. Bohemia Interactive also uses the *.bistudio.com domain.

kmpoppe (approved):

PR #3867, opened 2019-06-01
Add Segment

Rowno:


kmpoppe:
Hello @Rowno and thank you for your contribution. You've added Segment to the developer section, yet I fail to understand what they really do as a product. Could you please elaborate?
// Kai

Rowno:
Segment is a service for routing user analytics data from multiple sources to multiple destinations. For example, using the Javascript library you could track some user events on the frontend and Segment will route the track events to all the third party tools you use (like Google Analytics, Mixpanel, Amplitude, Mailchimp, data warehouses, etc) without you having to go through all the complexities of integrating with each third party API.

dsoegijono (dismissed):

Carlgo11 (changes requested):
While it's not a perfect fit, I think hosting fits Segment better than developer.
The developer category is for things like Visual Code Studio, Jetbrains, Unity, GitHub etc.

Rowno:
Done!

Carlgo11:
Looks like 2FA is only available for business accounts with third party authentication (such as G-suite, Okta etc). Is that correct?

Rowno:
Nope, MFA (OTP and SMS) is available for all accounts. Only SSO (single sign on via a third party) is restricted to business tier accounts.

Carlgo11 (approved):

kmpoppe (approved):

PR #3874, opened 2019-06-07
Add Pornhub

DandelionSprout:
This one doesn't really need much of an explanation nor a speech to warrant its inclusion, due to its extreme popularity and the existing presence of similar video sites like YouTube and Niconico.

One thing to note, however: On their official documentation page, their link to turn on 2FA is broken for non-professional accounts, and will only lead to https://www.pornhub.com/partners/mpp. The correct settings page for non-professionals (as far as I can personally determine) is https://www.pornhub.com/user/security.


Carlgo11:
Hi @DandelionSprout,
thanks for your pull request! Unfortunately we don't add pornographic sites to the list that wouldn't really be fair to us members having to maintain the list and go through that kind of stuff 😕 and it could cause some enterprise firewalls to block our site.

A revision of the CONTRIBUTING.md/EXCLUSION.md is in the works. Sorry for the inconvenience 😞

DandelionSprout:
Fair enough, as long as the guideline revisions happen reasonably quickly.

PR #3875, opened 2019-06-08
Add Espresso House

Carlgo11:
Espresso House is the largest coffee chain in Sweden with operations in all of the Nordic countries.

Their app, which acts as the main payment method, is currently only using SMS codes to authenticate.


kmpoppe (approved):
Watch me checking whether a site added by a collaborator is really within the 200k Alexa ranking *facepalm*.

PR #3880, opened 2019-06-11
Add creditkarma.com doc

jonade:


_data/finance.yml
```diff
@@ -70,6 +70,7 @@ websites:
       tfa: Yes
       sms: Yes
       phone: Yes
+      doc: https://help.creditkarma.com/hc/en-us/articles/360001472366-Two-Factor-Authentication-FAQ
```
kmpoppe:
```suggestion
      doc: https://help.creditkarma.com/hc/en-us/articles/360001472366
```
URL can be shortened, which is what we like here ;-)
// Kai

kmpoppe (approved):

Carlgo11 (approved):

PR #3882, opened 2019-06-12
Update Vultr

seonwoolee:
Vultr's docs say they support Yubikey, but if you go to add a Yubkey, it states they're only supported in OTP mode, not U2F mode. Unfortunately their docs do not clarify this point.

### DIFF
--- a/_data/cloud.yml
+++ b/_data/cloud.yml
@@ -233,5 +233,5 @@ websites:
       software: Yes
       hardware: Yes
       exceptions:
-        text: "While the documentation makes no mention of this, when you go to add a Yubikey (the only hardware token supported), it says U2F Yubikeys are not supported. Yubikeys are only supported in OTP mode, not U2F mode."
+        text: "Hardware 2FA only supports YubiKeys with Yubico OTP credentials."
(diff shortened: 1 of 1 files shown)


_data/cloud.yml
```diff
@@ -232,4 +232,6 @@ websites:
       tfa: Yes
       software: Yes
       hardware: Yes
+      exceptions:
+        text: "While the documentation makes no mention of this, when you go to add a Yubikey (the only hardware token supported), it says U2F Yubikeys are not supported. Yubikeys are only supported in OTP mode, not U2F mode."
```
Carlgo11:
```suggestion
        text: "Hardware 2FA only supports YubiKeys with Yubico OTP credentials."
```

Carlgo11 (changes requested):
I suggest this exception message instead. It's more in line with our other 'proprietary 2FA'-exceptions.

_data/cloud.yml
seonwoolee: Sounds good.

Carlgo11 (approved):

kmpoppe (approved):

PR #3883, opened 2019-06-12
Update Scaleway

kdeldycke:
Scaleway now support for a while mail-based password-less authentication by the way of a magic link.

See: https://www.scaleway.com/en/docs/console-magic-link-authentication/


Carlgo11:
Looks like they just moved from one authentication type to another. If they had password authentication with email authentication it would classify as two factor authentication.

PR #3920, opened 2019-07-04
Add RealDebrid

theoricher:
RealDebrid Website : https://real-debrid.com
Email : Yes
Software : Yes
Hardware : No
SMS : No
Documentation : NO


Carlgo11 (commented):
So it's like TeamViewer?
That one is placed under 'Remote Access' so if the two are similar I suggest you move RealDebrid from other to remote.

theoricher:
> So it's like TeamViewer?
> That one is placed under 'Remote Access' so if the two are similar I suggest you move RealDebrid from other to remote.

Is an unrestricted downloader that allows you to quickly download files hosted on the Internet or instantly stream them into an innovative web player

Issue #3345, opened 2018-08-02
USAA supports email 2FA


coleslaugh:
Consider updating the list to reflect that USAA supports OTP via email.

pmocek:
Note that this is a rather obscure option.  The linked [USAA MFA page](https://www.usaa.com/inet/pages/security_token_logon_options) describes second factors SMS (as "CyberCode Text") and software token, specifically Symantec Validation & ID Protection (as "CyberCode Token"), but the only mention of e-mail I see is in a note explaining why an e-mail address is required in order to use their SMS MFA system: "If you need to log on and don't have your phone, we can send a one-time passcode to your email address."

coleslaugh:
Good point. It looks like their documentation is out dated. Here is a screen shot of the 2fa screen when I log in. The first three bullets are the phone numbers I have with them and the last two are email addresses. I’m able to get a OTP to either email. I’m a CyberCode Text user. It looks like they added email support to the CyberCode Text option.

![4238fff2-b1bc-4f20-b9e9-8ed33a9fc597](https://user-images.githubusercontent.com/42038245/43607327-874c03e0-9663-11e8-9b54-db6ad8349108.jpeg)

PR #3921, opened 2019-07-05
USAA does not support hardware tokens or email

pconwell:
USAA only supports three 'levels' of security: normal username and password, username and password with an SMS OTP, or username with PIN+OTP generated by a symantec app.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -1160,8 +1160,8 @@ websites:
       img: usaa.png
       tfa: Yes
       sms: Yes
-      hardware: Yes
-      email: Yes
+      hardware: No
+      email: No
       software: Yes
       doc: https://www.usaa.com/inet/pages/security_token_logon_options
 
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -1160,8 +1160,7 @@ websites:
       img: usaa.png
       tfa: Yes
       sms: Yes
-      hardware: No
-      email: No
+      email: Yes
       software: Yes
       doc: https://www.usaa.com/inet/pages/security_token_logon_options
 


Carlgo11:
@pconwell according to #822 and #3345 they do support hardware and email 2FA. Are you saying that that's changed?

pconwell:
In #822, they are referring to Symantec VIP tokens which are *not* hardware tokens. They are software tokens identical to standard OTP tokens that happen to run on Symantec "hardware". The "hardware tokens" they are identical to a smart phone OTP app, except instead of running on a smart phone they run on a Symantec VIP token. Additionally, Symantec no longer offers the actual physical tokens and the tokens were replaced by a smart phone app anyway - so I would not consider it a hardware token. At one point in time, Symantec *did* have physical software tokens, but 1) even back then it was still a software token by all intents and purposes, and 2) since they have switched to smart phone apps, it is a software token in every sense  of the word today.

As far the email OTP, their website does not mention email and even calls it "CyberCode Text" - however @coleslaugh suggests that they now include email in the 'text' codes. I haven't tested it, but I can if needed.

In summary - USAA has never supported hardware tokens, and USAA *might* support email OTP codes.

Carlgo11 (changes requested):
Alright, thanks for the clarification @pconwell. 
Could you remove the lines with `No` as the value? Values are false by default and so it's not necessary to specify (plus it also breaks the build).

pconwell:
I've edited the pull request (I think... I'm not an expert on git). I also confirmed that USAA *does* support email 2FA as long as the user explicitly enables email as a backup recovery method, so I made that change as well. So the only actual change is removing the hardware token from USAA.

Carlgo11 (approved):

RichJeanes (approved):

PR #3946, opened 2019-07-23
Sticky table header

olback:
Give thead `position: sticky` to make it easier to see what category the checkmark belongs to.

This feature does not yet work in Chromium based browsers. See [https://caniuse.com/#feat=css-sticky](https://caniuse.com/#feat=css-sticky) for more details. However, the change is non-breaking.

### DIFF
--- a/css/base.scss
+++ b/css/base.scss
@@ -40,6 +40,14 @@ p {
   margin: 1em 0;
 }
 
+table {
+  thead {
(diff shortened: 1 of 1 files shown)


Carlgo11 (approved):
Works on Firefox & Safari 👍

kmpoppe (approved):

PR #3950, opened 2019-07-30
Future proof data.json by creating different versions

Carlgo11:
As of right now, the `data.json` prints out the value of the YML tags regardless of the content. If we, in the future rework, let's say the `tfa`-tag, it would mean that third party scripts relying on the `data.json` would break.

Therefore I created two versions of the data.json:
* __Version 2__ is our current `data.json`. It spits out every tag and data.
* __Version 1__ is a little different. For tfa tags it returns true unless the data of the tag is false. This would allow us to in the future compress the YMLs or add new tags such as U2F and WebAuthN tags without breaking third party scripts.

In this PR I haven't removed the `data.json` in the root directory since that would break scripts if the PR got merged. However, if/when this PR gets merged we can redirect requests for `twofactorauth.org/data.json` to the `data.json` inside `api/v1` using CloudFlare.

### DIFF
--- a/api/v1/data.json
+++ b/api/v1/data.json
@@ -1,0 +1,30 @@
+---
+---
+{
+{%-  assign tfa_tags = "tfa, sms, phone, software, hardware" | split: ', ' -%}
+
+    {%- for section in site.data.sections -%}
+    {%- assign section_file = site.data[section.id] -%}
+    "{{ section.title }}": {
+                        {%- for website in section_file.websites -%}
+                                "{{ website.name }}": {
+                                    {%-for tag in website -%}
+                                      {%- unless tfa_tags contains tag[0] -%}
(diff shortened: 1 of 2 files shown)


Carlgo11:
This is also a requirement for future work on #3222.

jdavis:
Shouldn't V1 be the original and V2 be the new and improved one? If it needs to be changed in the future, then what would we do? Swap V1 to V2 and V2 to V3? We should count up to avoid this.

> However, if/when this PR gets merged we can redirect requests for twofactorauth.org/data.json to the data.json inside api/v1 using CloudFlare.

Going off of your versions and not what I proposed, wouldn't `data.json` need to point to `api/v2`?

~~Also, what are the proposed changes to the data YAML files to support U2F and WebAuthN? Is there a PR for that?~~

Just saw you linked to #3222. I’ll look at that and respond

Carlgo11:
@jdavis I figured that the versioning should be based on the API output and not the actual backend code. If a tag is changed in the YAML files it shouldn't break an older version. Therefore I modified the data.json to display the older syntax in v1 regardless of tfa tag changes in the YAML.

PR #3960, opened 2019-08-08
Add dmarcian

brightball:
Add dmarcian.com to the Security list. We've just added 2FA for all accounts using both software token and FIDO2 U2F via Webauthn.


Carlgo11:
From the log:
> img/security/dmarcian.png should not be larger than 2500 bytes. It is currently 4103 bytes.

brightball:
I'll get that corrected.

brightball:
@Carlgo11 the image has been updated and everything seems to be in order now.

Carlgo11 (approved):

kmpoppe (approved):

PR #2962, opened 2018-01-22
Investing - Vanguard poor use of Yubikey (FIDO) noted

PeterTheobald:
Updated Vanguard entry to note that use of FIDO device like Yubikey is trivially bypassed by simply checking a box for "I forgot my device please use SMS instead". SMS cannot be disabled, thus the security of Yubikey is not actually enforced. Tested as of Jan 2018.

PR #3963, opened 2019-08-11
Update Vanguard

kmpoppe:
Changed exception wording according to #2962.

### DIFF
--- a/_data/investing.yml
+++ b/_data/investing.yml
@@ -185,7 +185,7 @@ websites:
       phone: Yes
       hardware: Yes
       exceptions:
-        text: "Hardware 2FA requires using a Chrome browser and only supports the following hardware keys: Yubikey 4 Series, YubiKey 5 Series, Yubikey Security Key Series."
+        text: "SMS/Phone require for 2FA. Hardware 2FA requires using a Chrome browser and only supports the following hardware keys: Yubikey 4 Series, YubiKey 5 Series, Yubikey Security Key Series."
       doc: https://investor.vanguard.com/security/security-keys
 
     - name: Wealthfront
--- a/_data/investing.yml
+++ b/_data/investing.yml
@@ -185,7 +185,7 @@ websites:
       phone: Yes
       hardware: Yes
       exceptions:
-        text: "SMS/Phone 2FA activation required before hardware 2FA can be used. Hardware 2FA requires using a Chrome browser and only supports the following hardware keys: Yubikey 4 Series, YubiKey 5 Series, Yubikey Security Key Series."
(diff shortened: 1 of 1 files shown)


Carlgo11:
Do you mean that a phone number is required for 2FA?

Carlgo11:
@kmpoppe ^

kmpoppe:
@Carlgo11 I just took what was in #2962 and put it in a new PR because the other one was so garbled. As vanguard.com always seems to be in maintenance mode when I want to check back on the documentation I can't get any further :-/

kmpoppe:
@Carlgo11 After the site maintenance (looks like saturday nigth in the US) is over, I've got further information:
- From https://investor.vanguard.com/security/security-keys
	- Do I need to register for **security codes** if I have a security key?
		- **Yes**, because if you don't have your key or if you're logging on from a device that isn't compatible, we'll need to send you a security code instead.

- From https://investor.vanguard.com/security/security-codes
	- What are **security codes** and how do they work?
		- A **security code** is a unique, single-use, 6-digit number you receive by **text** or **automated phone** call that you'll enter when you log on to your Vanguard account from a device we don't recognize as one you regularly use.

_data/investing.yml
```diff
@@ -185,7 +185,7 @@ websites:
       phone: Yes
       hardware: Yes
       exceptions:
-        text: "Hardware 2FA requires using a Chrome browser and only supports the following hardware keys: Yubikey 4 Series, YubiKey 5 Series, Yubikey Security Key Series."
+        text: "SMS/Phone require for 2FA. Hardware 2FA requires using a Chrome browser and only supports the following hardware keys: Yubikey 4 Series, YubiKey 5 Series, Yubikey Security Key Series."
```
kmpoppe:
```suggestion
        text: "SMS/Phone 2FA activation required before hardware 2FA can be used. Hardware 2FA requires using a Chrome browser and only supports the following hardware keys: Yubikey 4 Series, YubiKey 5 Series, Yubikey Security Key Series."
```

_data/investing.yml
```diff
@@ -185,7 +185,7 @@ websites:
       phone: Yes
       hardware: Yes
       exceptions:
-        text: "Hardware 2FA requires using a Chrome browser and only supports the following hardware keys: Yubikey 4 Series, YubiKey 5 Series, Yubikey Security Key Series."
+        text: "SMS/Phone 2FA activation required before hardware 2FA can be used. Hardware 2FA requires using a Chrome browser and only supports the following hardware keys: Yubikey 4 Series, YubiKey 5 Series, Yubikey Security Key Series."
```
Carlgo11:
```suggestion
        text: "Hardware based 2FA requires SMS / Phone call 2FA as a backup. Hardware 2FA is only supported on Chrome browsers and works with the following hardware keys: Yubikey 4 Series, YubiKey 5 Series, Yubikey Security Key Series."
```

Carlgo11 (changes requested):

Carlgo11 (dismissed):

Carlgo11 (approved):

PR #3967, opened 2019-08-17
Update sparkasse and comdirect

MartinThoma:


_data/banking.yml
```diff
@@ -285,7 +285,9 @@ websites:
       twitter: comdirect
       facebook: comdirect
       img: comdirect.png
-      tfa: No
+      tfa: Yes
+      software: Yes
+      doc: https://www.comdirect.de/konto/photo-tan.html#Vorteile
       lang: de
```
```suggestion
```

_data/banking.yml
```diff
@@ -995,7 +997,9 @@ websites:
       facebook: sparkasse
       email_address: sparkasse@sparkassen-finanzportal.de
       img: sparkasse.png
-      tfa: No
+      tfa: Yes
+      hardware: Yes
+      doc: https://www.sparkasse.de/service/sicherheit-im-internet/ihre-daten-bei-der-sparkasse.html
       lang: de
```
```suggestion
```

_data/banking.yml
kmpoppe: As per rake: 'lang' should NOT be present when tfa: true.

_data/banking.yml
kmpoppe: As per rake: 'lang' should NOT be present when tfa: true.

kmpoppe (changes requested):
See single line comments.

kmpoppe:
Hey @MartinThoma - as you set the commit to be only changeable by you, I'd kindly ask you to remove the lang: tags because they are no longer needed, when tfa is already true. 
Thank you!
// Kai

MartinThoma:
Ok, I removed them :+1:

kmpoppe:
@MartinThoma Thank you for your contribution. I just now looked at the pages you provided, and even though they speak of "Zwei-Faktor-Authentifizierung" (two-factor authentication), what they really mean is authorization. Two Factor Authentication, as we see it for this project, means that you are directly challenged to provide a second factor upon login, not just for a sensitive transaction. This is also described in our Contribution Guidelines: https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions

Therefore I think that this pull request can not be merged. If I perceived any information wrongly, please do not hesitate to tell me so. Otherwise, I'd be glad, if you closed this PR.

Thanks again for taking the time to contribute to our project!

// Kai

MartinThoma:
> I just now looked at the pages you provided, and even though they speak of "Zwei-Faktor-Authentifizierung" (two-factor authentication), what they really mean is authorization. Two Factor Authentication, as we see it for this project, means that you are directly challenged to provide a second factor upon login, not just for a sensitive transaction.

Ok, I didn't know that. In this case you are right, the PR should not be merged. I close it.

PR #3993, opened 2019-09-05
Adds URL to configure 2FA for Wyzed devices.

aphive:
I checked in with Wyze and they provided the link to instructions for enabling 2FA.

### DIFF
--- a/_data/iot.yml
+++ b/_data/iot.yml
@@ -79,6 +79,5 @@ websites:
     - name: Wyze
       url: https://www.wyzecam.com
       img: wyze.png
-      tfa: No
-      twitter: WyzeCam
-      facebook: WyzeCam
+      tfa: Yes
+      doc: https://support.wyzecam.com/hc/en-us/articles/360026994652-Two-Factor-Authentication


Carlgo11:
>Wyze utilizes SMS 2FA solely through the Wyze app, and it isn't needed for use with wyze.com.

If it's only supported on certain devices then they're not fully supporting two factor authentication.

aphive:
I'll check back with them for clarification and will update here once I have more information.

kmpoppe:
@MrTechBot did you get any further information that you can update us with?
It seems that the doc url is not available publicly, could you check whether
https://support.wyzecam.com/hc/en-us/articles/360026994652
also leads to the same page (and thus reducing the length of the documentation URL in your PR?)

aphive:
Sorry, I was offline for a while and have not heard back as yet. I will follow up with them. Interesting that the url is now taking you to a login page when it wasn't in the past. I will check in on that as well.

aphive:
Interesting, I just tried it again and both URLs work just fine.

aphive:
Wyze 2FA

Carlgo11:
They've moved URL to https://support.wyzecam.com/hc/en-us/articles/360024402052.

The box at the bottom seems to describe the process pretty good. 2FA works on the app but is bypassed on the website.
That just gives users a false sense of security.

Closing for now. If, in the future, they start enforcing 2FA, make a comment and we'll reopen the PR.

PR #3995, opened 2019-09-06
Add CDON.COM

nybogard:
Added 2FA info for the Nordic shopping site CDON.com.


nybogard:
Got this from Travis:
> 1. img/retail/cdoncom.png should not be larger than 2500 bytes. It is currently 4155 bytes.

I've juggled around with the png a bit, getting it down to 2600 bytes using indexed 1-bit color mode and applying optipng. I don't know what else to do.

kmpoppe:
Please use tinypng.com - that might get the original picture down to below
2.5kB :-)

// Kai

nybogard:
> Please use tinypng.com - that might get the original picture down to below 2.5kB :-) // Kai

That did it, thank you!

PR #4001, opened 2019-09-08
Add OTP banka

anotherus3r:


kmpoppe:
@anotherus3r My Croatian is a little rusty (reads: non-existent), but despite the name "OTP token", which seems to be a PIN generator, it looks like OTP banka uses it for authorization rather than authentication. Maybe I'm reading that wrong, could you enlighten me?

anotherus3r:
@kmpoppe It's used for both - to log in and to authorize transactions. When you want to log in, you have to enter a PIN into the token to get a one-time password with which you can log into the website.

Carlgo11:
@anotherus3r I don't see anything mentioning software based 2fa 🤔

anotherus3r:
@Carlgo11 The documentation for it is here: https://www.otpbanka.hr/hr/upute/otp-m-token-upute-za-koristenje

Carlgo11 (approved):

kmpoppe (approved):

PR #4005, opened 2019-09-10
Add VirusTotal

jakejarvis:
A pretty popular tool in the infosec world (and owned by Alphabet). https://www.virustotal.com/

Can't find any docs unfortunately (except a blog post from a few months ago) but it's pretty self-explanatory in their settings:

Screen Shot 2019-09-10 at 11 56 24 AM

Thanks! :)


Carlgo11 (changes requested):
The image is a bit larger than what I should be.
Here's a better one:
virustotal

jakejarvis:
Yup, that one is definitely much smaller! 👍

Carlgo11 (approved):

kmpoppe (approved):

PR #4009, opened 2019-09-12
Add ING-DiBa Deutschland

ASMag:
Added ING-DiBa Deutschland in banking.yml.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -580,7 +580,6 @@ websites:
       tfa: Yes
       sms: Yes
       doc: https://www.ing.de/kundenservice/sicherheit/psd2/
-      twitter: ING_Deutschland
       facebook: ing.deutschland
 
     - name: Intesa Sanpaolo
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -580,7 +580,6 @@ websites:
       tfa: Yes
       sms: Yes
       doc: https://www.ing.de/kundenservice/sicherheit/psd2/
-      lange: de
       email_address: info@ing.de
       twitter: ING_Deutschland
       facebook: ing.deutschland
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -580,7 +580,6 @@ websites:
       tfa: Yes
       sms: Yes
       doc: https://www.ing.de/kundenservice/sicherheit/psd2/
-      facebook: ing.deutschland
 
     - name: Intesa Sanpaolo
       url: https://www.intesasanpaolo.com
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -580,7 +580,6 @@ websites:
       tfa: Yes
       sms: Yes
       doc: https://www.ing.de/kundenservice/sicherheit/psd2/
-      email_address: info@ing.de
       twitter: ING_Deutschland
       facebook: ing.deutschland
 
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -574,6 +574,17 @@ websites:
       tfa: No
       lang: de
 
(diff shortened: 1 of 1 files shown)


_data/banking.yml
```diff
@@ -574,6 +574,17 @@ websites:
       tfa: No
       lang: de
 
+    - name: ING-DiBa Deutschland
+      url: https://www.ing.de
+      img: ing.png
+      tfa: Yes
+      sms: Yes
+      doc: https://www.ing.de/kundenservice/sicherheit/psd2/
+      lange: de
```
dsoegijono:
```suggestion
      lang: de
```

_data/banking.yml
```diff
@@ -574,6 +574,17 @@ websites:
...
+      email_address: info@ing.de
```
dsoegijono:
```suggestion
```
We don't need social info if tfa is supported

_data/banking.yml
```diff
@@ -574,6 +574,17 @@ websites:
...
+      twitter: ING_Deutschland
```
dsoegijono:
```suggestion
```

_data/banking.yml
```diff
@@ -574,6 +574,17 @@ websites:
...
+      facebook: ing.deutschland
```
dsoegijono:
```suggestion
```

dsoegijono (dismissed):
Could use someone else's help to verify the doc linked

_data/banking.yml
Carlgo11: lang tags aren't actually used for sites with 2FA enabled.
```suggestion
```

Carlgo11 (changes requested):

ASMag (commented):
This seems to be the right changes for adding ING-DiBa Deutschland.

kmpoppe:
@ASMag If you might be wondering why the approval takes it's time (apart from obvious reasons with spare time of volunteers), I'm currently unsure as to how we will list all this proprietary stuff that the PSD2 regulation forced on the European banks and got implementated better or worse 🙄.
We will come back to this later though.
// Kai

_data/banking.yml
```diff
@@ -574,6 +574,13 @@ websites:
       tfa: No
       lang: de
 
+    - name: ING-DiBa Deutschland
```
Carlgo11, line 577: Shouldn't it be `ING Deutschland`?

Carlgo11 (changes requested):
According to https://youtu.be/f8r3dcI9BKw they support software 2FA.
I've also found [photoTAN](https://www.ing.de/kundenservice/konto-und-depot-verwalten/phototan/) which would be hardware 2FA.
I can't find anything about SMS based 2FA. Could you give me a link about it?

kmpoppe:
> According to https://youtu.be/f8r3dcI9BKw they support software 2FA.

The video says "we will also ask you at every login", so it's really _authentication_ not _authorization_ only. I stand corrected

_data/banking.yml
kmpoppe: Yes. We do have ING (Luxembourg), ING Australia but then IBG-DiBa Austria and ING-DiBa Deutschland.
Due to the alphabetical order restriction in Travis, we cannot rename ING-DiBa Deutschland to ING Deutschland without also renaming ING-DiBa Austria - which would be OK since _all_ product now go by "ING" rather than "ING DiBa" even though the bank in regulatory terms calls itself ING DiBa AG, at least in Frankfurt.
I'd go for renaming both banks in this PR before merge.

Carlgo11 (approved):

kmpoppe (approved):

PR #4018, opened 2019-09-17
Remove www. from Synology.com

bensamjaeger:
2FA is supported by account.synology.com but did not appear on 1Password


Carlgo11:
https://synology.com/ redirects to https://www.synology.com/ so the main URL of the site is https://www.synology.com.

PR #4021, opened 2019-09-19
Update YNAB

adambowles:
YNAB added support for TOTP tokens today

### DIFF
--- a/_data/finance.yml
+++ b/_data/finance.yml
@@ -409,4 +409,6 @@ websites:
       twitter: ynab
       facebook: iYNAB
       img: ynab.png
-      tfa: No
+      tfa: Yes
+      software: Yes
+      doc: https://docs.youneedabudget.com/article/1364-two-step-verification
--- a/_data/finance.yml
+++ b/_data/finance.yml
@@ -406,7 +406,6 @@ websites:
 
     - name: You Need A Budget (YNAB)
       url: https://www.youneedabudget.com/
-      twitter: ynab
       img: ynab.png
       tfa: Yes
       software: Yes
--- a/_data/finance.yml
+++ b/_data/finance.yml
@@ -407,7 +407,6 @@ websites:
     - name: You Need A Budget (YNAB)
       url: https://www.youneedabudget.com/
       twitter: ynab
-      facebook: iYNAB
       img: ynab.png
       tfa: Yes
       software: Yes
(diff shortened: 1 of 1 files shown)


_data/finance.yml
```diff
@@ -409,4 +409,6 @@ websites:
       twitter: ynab
```
kmpoppe:
```suggestion
```

_data/finance.yml
```diff
@@ -409,4 +409,6 @@ websites:
...
       facebook: iYNAB
```
kmpoppe:
```suggestion
```

kmpoppe (changes requested):
Social Media links should not appear when `tfa: yes` is set.

kmpoppe (dismissed):

_data/finance.yml
```diff
@@ -406,7 +406,7 @@ websites:
 
     - name: You Need A Budget (YNAB)
       url: https://www.youneedabudget.com/
-      twitter: ynab
-      facebook: iYNAB
       img: ynab.png
-      tfa: No
+      tfa: Yes
+      software: Yes
+      doc: https://docs.youneedabudget.com/article/1364-two-step-verification
```
Carlgo11:
```suggestion
      doc: https://docs.youneedabudget.com/article/1364--
```

Carlgo11 (changes requested):

_data/finance.yml
adambowles: What's the purpose of removing the article slug?

I chose the address with slug as it's what is suggested via YNAB's docs search function:
image

adambowles:
YNAB have added a note to the docs article that 2FA has not been fully rolled out to all customers yet: 
image
(https://docs.youneedabudget.com/article/1364-two-step-verification)

Should this PR wait until full release?

_data/finance.yml
kmpoppe: > What's the purpose of removing the article slug?

We have removed the slug from every HelpCenter and HelpScout addresses for size reduction reasons and because "it works". Also, we had the assurance from HelpScout that they were working on removing the need for a slug at all in the coming months.

kmpoppe:
> Should this PR wait until full release?

There's always the possibility to add an "exception text" that says it might not be available to everyone now, the question most likely would be how long that "Beta" state will hold. Can you try to find that out from YNAB?

Carlgo11 (approved):

kmpoppe (approved):

PR #4030, opened 2019-09-21
Exceptions and Restrictions currently text only

### LINKED ISSUE #3116: Your Schema doesnt account for everything your HTML does
https://github.com/2factorauth/twofactorauth/blob/44a27551a18669f3adc5b37ac57270268db89bc6/websites_schema.yml#L42

It would appear that you have the ability to also account for a mapping "link" for an excption, and even where its not either of those....https://github.com/2factorauth/twofactorauth/blob/44a27551a18669f3adc5b37ac57270268db89bc6/_includes/exception.html#L11

kmpoppe:
Exceptions and Restrictions are currently used text only, thus removing exceptions.link from exception.html.
Closes #3116.

### DIFF
--- a/_includes/exception.html
+++ b/_includes/exception.html
@@ -7,28 +7,16 @@
 <div class='content'>
     {% if include.website.exceptions.text %}
         {{ include.website.exceptions.text }}
-    {% else %}
-        {% if include.website.exceptions.link %}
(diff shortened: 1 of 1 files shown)


Carlgo11:
Fixed in 70aed0008ab066268d19f290472fd1eb9f060ed5

Issue #3707, opened 2019-02-07
Extend schema to allow entries in multiple categories


jwishnie:
Many service providers offer a _suite_ of services behind a single login.

e.g. Google and Microsoft accounts provide email, chat, backup, etc...

Rather than recreating entries for each service in each separate category, we could extend the schema for entries to contain a list of categories in which the entry should appear.

kmpoppe:
Second that.

mxxcon:
This was previously discussed already, but nobody stepped up with ideas on the shcema for this and other implementation intricacies.

jakejarvis:
I literally just had the same thought a few minutes ago, too – currently, if Microsoft suddenly announced support for hardware tokens, it would require a change to _nine_ files!

But agreed that this would be a pretty major undertaking... This would mean a complete restructuring of the `_data` directory from category-based to service-based, to the point where each service would need its own file and its category/categories listed within it, no?

pm17788:
This almost feels like each "entry" (say, Microsoft as an example) should simply have multiple tags indicating categories, and either at load time or some sort of a macro thing would generate data.
JSON-ish view, I suppose:
```json
{
  "Companies": {
    "Microsoft": {
      "Services": [
        "email",
        "backups",
        "productivity"
      ],
      "2FA Methods":[
        "SMS",
        "HW Token",
        "SW Token",
        "Scratch-n-Sniff"
      ],
      "URL": "https://www.microsoft.com",
      "PrettyLogoPicturePath": "images/Microsoft/"
    }
  }
}
```

So, an entry like this would "show up" under `emails`, `backups`, and `productivity`, but it's one entry, with the common dataset.

I'm _certain_ I'm missing something in the intricacies, and would be glad to bounce ideas back and forth.

jwishnie:
@pm17788 that's exactly the sort of thing I was thinking about. Essentially instead of separate yml's for each category, there's a single list of login services ('identity providers') and each entry lists all the services it provides access to...

@jakejarvis MS _has_ announced they will be supporting security keys visa Windows Hello, and _currently_ support keys on personal MS accounts...

pm17788:
@jwishnie: So, what's our next step on this? :) How can I help move this forward?

kmpoppe:
Coming back to this.

Combining what @jwishnie and @pm17788 said, I'd go for:

1. A new "invisible" category (e.g. `Identity providers`), that just says which services it provides (sms, software, ...) including documentation.
[...]

PR #3812, opened 2019-04-14
Add 'Attention' message to desktop- and mobile-table and update mobile-table button layout

kmpoppe:
Closes #1450.

PR #4034, opened 2019-09-21
Allow entries in multiple categories

kmpoppe:
This Pull Request is a first try on solving #3707.

To create an entry, that is supposed to be listed in multiple categories, those are NOT to be listed in their respective categories but in the new file `multi.yml`. Subsequently, their images/logos are also to be places in the directory `img/multi`.

As I said in https://github.com/2factorauth/twofactorauth/issues/3707#issuecomment-484370254, those three entries shall be listed in Domains _and_ Hosting/VPS, so the multi.yml would look like this:

```websites:
    - name: Rackspace
      url: https://www.rackspace.com/
      img: rackspace.png
      tfa: Yes
      sms: Yes
      doc: https://support.rackspace.com/how-to/myrackspace-multi-factor-authentication/
      categories:
        - hosting
        - domains

    - name: UpCloud
      url: https://www.upcloud.com
      img: upcloud.png
      tfa: Yes
      software: Yes
      exceptions:
          text: "SMS is required for password recovery."
      doc: https://www.upcloud.com/support/two-factor-authentication/
      categories:
        - hosting
        - domains

    - name: VentraIP
      url: https://ventraip.com.au/
      img: ventraip.png
      tfa: Yes
      software: Yes
      sms: Yes
      email: Yes
      doc: https://ventraip.com.au/faq/article/two-factor-authentication-faq-vipcontrol/
      categories:
        - hosting
        - domains
```

See how this works here: http://2fa.poppe.work:4001.

Comments are very welcome!

// Kai

### DIFF
--- a/_includes/desktop-table.html
+++ b/_includes/desktop-table.html
@@ -1,6 +1,7 @@
 {% assign section_id = include.id-param %}
 {% assign section_title = include.title-param %}
 {% assign section_file = site.data[section_id] %}
+{% assign multi_section_file = site.data["multi"] %}
 <div class="website-table desktop-table {{ section_id }}-table" id="{{ section_id }}-desktoptable">
   <table class="ui celled table content-wrapper">
     <thead>
@@ -15,8 +16,9 @@
     </tr>
     </thead>
     <tbody class="jets-content">
-    {% for website in section_file.websites %}
-    {% if website.lang and site.data.languages[website.lang] %}
+    {% assign multi_sites = multi_section_file.websites | where: "categories", section_id | concat: section_file.websites | sort: "name" %}
+    {% for website in multi_sites %}
+      {% if website.lang and site.data.languages[website.lang] %}
       {% assign progress_tweet = site.data.languages[website.lang].progress_tweet %}
       {% assign workonit_tweet = site.data.languages[website.lang].work_tweet %}
     {% else %}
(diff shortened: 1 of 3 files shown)


Carlgo11:
Is this something that we perhaps want to wait with until the new YAML scheme is in place?

kmpoppe:
Indeed, that would be sensible (I didn't update the data.json, so that's still to do anyway).
An idea on whether my idea itself would work from an administrative/user point of view would be welcomed though.

kmpoppe:
Assigned to @Carlgo11 for the same reason as #3812

phallobst:
Moving/merging entries as a whole into a new file only works for those that are identical. Others share a substantial amount of information, possibly with slight deviations, e.g. the various Google or Amazon services. Maybe something like this would be possible:

`backup.yml`:
```
- name: Google Drive
  url: https://drive.google.com
  img: drive.png
  include: Google
```
`cloud.yml`:
```
- name: Google Cloud Platform
  url: https://cloud.google.com
  img: googlecloud.png
  include: Google
```
new file `include.yml`:
```
- name: Google
  tfa:
    - sms
    - phone
    - totp
    - proprietary
    - u2f
  doc: https://www.google.com/intl/en-US/landing/2step/features.html
```

That way, the entries would keep their respective categories with URLs and icons, but could share the other information (which would need to be maintained at one place only). The entry in the category should be able to overwrite stuff from the include, just in case.

Carlgo11:
Closing the PR and putting it in the todo list instead.

PR #4047, opened 2019-09-29
Fix dead link for Commonwealth Bank of Australia

mhansen:
Previous one was redirecting to an unrelated Lenders Mortgage Insurance page.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -312,7 +312,7 @@ websites:
       tfa: Yes
       sms: Yes
       hardware: Yes
-      doc: https://www.commbank.com.au/support.digital-banking.explain-netcode-token.html
+      doc: https://www.commbank.com.au/support/security/netcode.html
 
     - name: Community America Credit Union
       url: https://www.communityamerica.com/


kmpoppe (approved):
Hi @mhansen and thanks for your contribution!
The fixed link looks correct, yet to me it feels like the whole two-factor-process at this bank is "authorization" and not "authentication" (because it says "codes for certain transactions"), which would then not qualify to be listed here.
Do you have experiences with the usage of NetCode? Can a NetCode also be used for authenticating a user that tries to log in from an unknown device for example?
// Kai

mhansen:
I haven't used it myself yet, can give it a go and get back to you perhaps.
But your reading is probably correct FWIW

kmpoppe:
Take all the time you need, thanks for giving it a go!
// Kai

mhansen:
I just managed to login to the web portal for the bank without being
challenged for a second factor.

In their account settings it says:

"When will I need to use NetCode?
You'll be asked to get and enter a NetCode to:

Register for the CommBank app on your mobile or tablet
Authorise certain NetBank transactions like paying someone new, updating
your personal information or getting your log on details
Confirm some online Credit Card and Debit MasterCard purchases and when
using CommBank Checkout."

So yeah, I think you're spot-on: this code is only used when authorizing
high-risk things, not your usual 2FA.

Carlgo11:
Closing as it's not 2FA according to our [definition](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions).

kmpoppe:
@Carlgo11, would we not have to have a tfa-no-PR then?

PR #4049, opened 2019-09-30
Add Hypo Tirol, Hypo Vorarlberg

simon04:


kmpoppe:
Hello @simon04 and thank you for your contribution.

Travis CI check has failed, because you specified a Facebook page even though `tfa: Yes`.

According to our Contribution Guidelines we'd like to have every change in on separate PR (and thus different commit). Even though these two banks sound vaguely similar and I guess they might be two branches but then again they have different URLs.

Would you please be so kind as to split the two changes in to two PRs with no Facebook page in the YAML?

Thanks!

PR #4050, opened 2019-09-30
LastPass: add exceptions text

simon04:


Carlgo11:
We don't actually state that they support Yubikeys or U2F. Only that they provide hardware 2FA, and some hardware 2FA solutions are available for free accounts as well.

PR #4074, opened 2019-10-07
Add Seedrs.com

StarkDevelopments:


StarkDevelopments:
Alexa rank on 07.10.2019 12:38:47 UTC: 67,239

Carlgo11 (commented):
> If you enable 2FA, your mobile phone will receive an authentication codevia SMS when a withdrawal is initiated in your Investment Account.

That would qualify it as two factor __authorization__, not authentication.

StarkDevelopments:
> 
> 
> > If you enable 2FA, your mobile phone will receive an authentication codevia SMS when a withdrawal is initiated in your Investment Account.
> 
> That would qualify it as two factor **authorization**, not authentication.

Very good point. Would you like me to cancel the pull request ? Thanks.

PR #4094, opened 2019-10-11
Add Yelp

Rfosu2k2:
Add Yelp as not having 2FA.


Carlgo11 (changes requested):
I think it's better to make the red background cover the entire image so that it doesn't look weird on our site.
yelp logo

Rfosu2k2:
There is an alternate approved image to use as well, but I will let you pick between the two.  Let me know so I can change to the correct image.
yelp

Carlgo11:
Your image is fine 👍

kmpoppe (approved):

Carlgo11 (approved):

Issue #4073, opened 2019-10-06
Fidelity now supports 2FA by text or call


VictorYing:
https://www.fidelity.com/security/extra-security-login

The previous feature to use a smartphone app from Symantec is still available, but the ability to set up 2FA for logins without downloading a special app or needing to call Fidelity is new.

Also, it's not clear to me but the note on notes/fidelity/ about the old "transaction security" feature might be out of date. I'm not able to find any clear documentation on any still-existing distinction in how to set up 2FA for logins and identity verification for sensitive transactions at Fidelity.

kmpoppe:
Hi @VictorYing, there's already a Pull Request that addresses Fidelity (#4098), yet it only speaks of SMS, not calls. Could you clarify?

VictorYing:
Thanks for the ping, I made some comments at #4098

PR #4098, opened 2019-10-12
Update Fidelity documentation

mathstuf:
Not sure if there still exist account types which are not supported or
not. SMS-only though when through this account type though.


notes/fidelity.html
```diff
@@ -12,12 +12,7 @@ <h2 class="ui header">
         For <em>authentication</em>, Fidelity uses either a physical token provided by Fidelity
         or the Symantec VIP Access smartphone app.
```
VictorYing: The first sentence up here needs updating.

notes/fidelity.html
```diff
@@ -71,6 +66,11 @@ <h2 class="ui header">
         <li>
             Call Fidelity Account Key Support at +1 800-544-6666.
         </li>
+        <li>
+            Once logged in, navigate to <b>Profile</b> &gt; <b>Security
+            Center</b> &gt; <b> 2-factor authentication</b>. This process only
+            supports SMS.
```
VictorYing: I have an ordinary personal account at Fidelity.com (not set up through an employer). I signed up for 2FA with my phone number. When I log in via an incognito/private browser I get prompted after my password for whether I'd like a text message or automated voice call to receive a security code, so it seems it isn't true that this only supports SMS.

notes/fidelity.html
```diff
@@ -12,12 +12,7 @@ <h2 class="ui header">
...
-        Fidelity only offers 2FA on certain account types. For example, they
-        do not support it for institutional retirement accounts of the kind
-        that might be set up through an employer.
-        <br>
-        It appears there is no information about this service provided
-        online and it must be enabled and managed by phone.
+        Fidelity may only offer 2FA on certain account types.
     </p>
 
     <p>
```
VictorYing: I believe everything below here about "As of October 2016..." and "That this is different than the weaker Transaction Security..." is outdated. The two links below still link to pages that exist, but the page content has changed and the quoted text is no longer there, and Fidelity is no longer advertising Transaction Security as a feature, instead advertising 2FA at login. I'd recommend replacing all this outdated text with a simple link to Fidelity's newer page documenting 2FA: https://www.fidelity.com/security/extra-security-login

notes/fidelity.html
```diff
@@ -71,6 +66,11 @@ <h2 class="ui header">
         <li>
             Call Fidelity Account Key Support at +1 800-544-6666.
         </li>
+        <li>
```
VictorYing: The organization of these nested lists is rather confusing. As documented at https://www.fidelity.com/security/extra-security-login, Fidelity has two separate routes to two different classes of 2FA for logins. You can either set up the Symantec VIP Access proprietary 2FA on logins as before (using either a physical token or the proprietary apps) which still requires a phone call to Fidelity's support line and cannot be set up entirely online, OR you can set up 2FA via a phone number (SMS or automated voice call) which should be set up through the online security center and does not require a proprietary app or physical token.

It's not obvious to me whether it's better to try to document all these details here, or to simply link to (or quote directly from?) the page at Fidelity.com which seems to cover it pretty well: https://www.fidelity.com/security/extra-security-login

VictorYing (commented):
It seems much of the information here is outdated and no longer matches my recent experience with 2FA on my Fidelity.com account.

kmpoppe:
@VictorYing thank you for your comments.
From what I read it would seem that the usefulness of our local documentation has ended and we can just link the document of fidelity, maybe with adding some exception text.
I'll look into this later.
Thank you again for your contribution!
// Kai

_data/investing.yml
```diff
@@ -50,7 +50,8 @@ websites:
     url: https://www.fidelity.com
     img: fidelity.png
     tfa:
-      - software
+      - proprietary
+      - sms
       - hardware
     doc: "/notes/fidelity/"
```
kmpoppe:
```suggestion
    doc: https://www.fidelity.com/security/how-two-factor-authentication-works
```

kmpoppe: @mathstuf Ben,
please replace this link to the page of fidelity and delete the `notes/fidelity.html` files.
Thanks!
// Kai

_data/investing.yml
```diff
@@ -50,7 +50,8 @@ websites:
     url: https://www.fidelity.com
     img: fidelity.png
     tfa:
-      - software
+      - proprietary
+      - sms
       - hardware
```
kmpoppe, line 56:
```suggestion
      - phone
      - hardware
```

kmpoppe: As 2FA pins can be obtained by text or call, this should be included as well.

kmpoppe (changes requested):
See inline comments

mathstuf:
Thanks @VictorYing and @kmpoppe. I've updated it and left an exception noting that support may vary between different account types (at least I didn't see anything supporting hardware for my account type).

kmpoppe (approved):

kmpoppe:
Closes #4073

Carlgo11 (approved):

PR #4106, opened 2019-10-15
Update Deutsche Kreditbank (DKB)

phallobst:


kmpoppe (approved):

kmpoppe:
@phallobst What happened? We were like inches away from merging this PR...

Carlgo11 (approved):

phallobst:
Did I delete the branch too early? I havent't really mastered the github web interface, I suppose...
I have restored the branch now, I guess. Is that sufficient?

Carlgo11:
You did. Thanks for restoring it 👍 
We're all done now and so you can delete the branch if you'd like.

Braintelligence:
Hello @phallobst - can you please elaborate on what exactly this "Hardware Token" is that DKB is allegedly supporting? I can't find any trace on DKB about supporting hardware tokens for 2FA.

phallobst:
That procedure is called "chipTAN", and it's using your debit card in combination with a TAN generator.
See https://www.youtube.com/watch?v=tC9QDI2j7Ec for a demo.

Braintelligence:
I see, I wasn't aware that chipTAN would be considered a hardware token for the purpose of twofactorauth.org since it's a proprietary method only observed in online banking and unusable for any other service than the one your debit card was issued for. So just reading "hardware token - check" in the table left me pretty uninformed.

chipTAN comfort also being very susceptible to MITM-attacks gives it a false flair of security unbefitting of what hardware tokens stand for (to me at least).

For example with Twitter it also says hardware token with a checkmark but in fact it supports FIDO.

So to me, either the list needs a lot more differentiation, so a person actually can tell if they can use FIDO with a service or what other type of hardware token would be needed for 2FA with the specific service.

phallobst:
The 2FA types are explained in the [Wiki](/2factorauth/twofactorauth/wiki/FAQ-2FA-Types). `hardware` includes TAN generators like the ones used here.

We **do** distinguish between `u2f` keys and (non-u2f) `hardware`. The Twitter entry erroneously shows `hardware` instead of `u2f`, I suppose?

Braintelligence:
I see, thanks for clarifying.

I guess having the checkmark instead of specifying "U2F" is because of this portion:

"""
Exceptions & Restrictions
Mobile apps do not support Hardware 2FA. Hardware 2FA requires enabling back-up method. Cannot add multiple security keys. SMS only available on select providers.
"""

But it looks like this currently:
image

Thank you for your time.

PR #4123, opened 2019-10-16
Add sciebo.de

schwedenmut:
Sciebo is a non-commercial cloud storage service for research, studying and teaching for [participating institutions ](https://sciebo.de/en/login/index.html).


Carlgo11:
Our goal with this project is to aid consumers in deciding between alternative services based on the security they offer for their customers.

If a site is limited to only a few universities, I don't see how listing sciebo.de would aid the general consumer.

schwedenmut:
@Carlgo11 thanks for the clarification, which I didn't find that clear in the contribution "guidelines". The general consumer will not benefit from this, only the students of those enlisted universities. So I am closing this PR.

schwedenmut:
Btw. does this "general consumer" approach target also PRs such as  #4125 and others ?

PR #4131, opened 2019-10-17
Update Crypto Currencies category

Carlgo11:
I went through all listings under the crypto currency category and updated all sites I could find documentation for (or screenshots).

I also removed the following sites:
* **ANXPRO** - >600K on Alexa and on the documentation page they advertise 128-bit encryption and their own hashing algorithms. I think it would be immoral for us to list the site if their security is this bad. We've taken this decision several times before when sites have very lacking security.
* **Bitcoins Norway** - Site no longer exists


_data/cryptocurrencies.yml
```diff
@@ -147,108 +134,107 @@ websites:
...
+      - totp
     doc: https://coindeal.com/security
 
   - name: CoinFalcon
     url: https://coinfalcon.com/
     img: coinfalcon.png
     tfa:
-      - software
+      - totp
     doc: https://support.coinfalcon.com/en-us/article/how-to-setup-two-factor-authentication-2fa-nkjppd/
```
kmpoppe:
```suggestion
    doc: https://support.coinfalcon.com/en-us/article/-nkjppd/
```
Shorter URL possible.

_data/cryptocurrencies.yml
```diff
@@ -147,108 +134,107 @@ websites:
...
       - hardware
     doc: https://www.coinfloor.co.uk/security
 
   - name: CoinGate
     url: https://coingate.com
     img: coingate.png
     tfa:
-      - software
+      - totp
     doc: https://support.coingate.com/en/65/how-do-i-set-up-2-factor-authentication-2fa-on-my-account
```
kmpoppe:
```suggestion
    doc: https://support.coingate.com/en/65/-
```
Shorter URL possible.

kmpoppe (changes requested):
See inline comments

kmpoppe (approved):

PR #4133, opened 2019-10-17
Update Cloud Computing category

Carlgo11:
Also removed ProfitBricks as it seems they've either been acquired or gone out of business.


_data/cloud.yml
```diff
@@ -62,15 +64,17 @@ websites:
     tfa:
       - sms
       - phone
-      - software
+      - totp
+      - proprietary
+      - u2f
       - hardware
```
kmpoppe: Is the need to keep `hardware` given as you added `u2f` - because you remove `software`?

kmpoppe (changes requested):
`hardware` / `u2f` ?

kmpoppe (approved):

PR #4138, opened 2019-10-18
Add HypoVereinsbank

phallobst:


Carlgo11 (changes requested):
How about this image instead?
It's smaller and transparent.
hypovereinsbank

phallobst:
No objections, looks good to me.

Carlgo11 (approved):

PR #4144, opened 2019-10-20
Add Deutsche Bank

phallobst:


Jawshy (commented):
After [translating](https://www.deepl.com/translator), I can see that SMS 2FA is by using mobileTAN and proprietary 2FA is by using the photoTAN app. However, I couldn't see any hardware 2FA.

phallobst:
Using the photoTAN method is not only possible via the app, but also with a dedicated hardware device. My guess is that the bank is trying to guide the customer to use the app, so it is rather hidden. There is a manual (in German) available:
https://www.deutsche-bank.de/content/dam/deutschebank/de/shared/pdf/pk-digital-banking-photo-tan-bedienungsanleitung-lesegeraet.pdf

Jawshy:
Thanks for the link. Do you know if the photoTAN reader is readily available to customers?

phallobst:
Yes, you can order them via the bank at a low price (or even for free if you have accounts of a certain type).

Jawshy (approved):

_data/banking.yml
```diff
@@ -394,6 +394,15 @@ websites:
...
 
+  - name: Deutsche Bank
+    url: https://www.deutsche-bank.de/
+    img: deutschebank.png
+    tfa:
+      - sms
+      - proprietary
+      - hardware
+    doc: https://www.deutsche-bank.de/pk/lp/psd2.html
+
```
Carlgo11, line 453:
```suggestion
    regions:
      - de

```

Carlgo11 (approved):

PR #4161, opened 2019-10-22
Update N26

phallobst:


_data/banking.yml
```diff
@@ -650,8 +650,10 @@ websites:
   - name: N26
     url: https://n26.com
     img: n26.png
-    twitter: n26
-    facebook: N26
+    tfa:
+      - sms
```
Carlgo11, line 654: SMS seems to only be a backup method.
```suggestion
```

Carlgo11 (changes requested):

_data/banking.yml
phallobst: I don't think so. The documentation says (under "What if my phone doesn’t support the latest app version?") that SMS verification is possible for logins, i.e. authentication.

phallobst: Right, but that is _authorization_, not _authentication_, or am I mistaken here?

Carlgo11: No you're right. Good catch :+1:

Carlgo11 (approved):

_data/banking.yml
```diff
@@ -650,8 +650,10 @@ websites:
...
+      - proprietary
+    doc: https://support.n26.com/en-de/getting-started/psd2/psd2-strong-customer-authentication
```
```suggestion
    doc: https://support.n26.com/en-eu/getting-started/psd2/psd2-strong-customer-authentication
```
When I looked at the `doc`-link I was asking myself: since when is `en-de` a valid [locale code](https://www.science.co.il/language/Locale-codes.php)?
Switching the location in the bottom left corner reveals that N26 seems to think that it's ok to use the second part of the identifier as their location.
So `en-de` stands for English in the German N26 department 🙄.
As the PSD2-page is not available in the United States (because PSD2 is a European thing) I'd suggest we use `en-eu` in the link.

kmpoppe (changes requested):
See inline comment.

// Kai

_data/banking.yml
phallobst: That is strange indeed. However, looking around in the support area, I found an ever better page (I believe):
https://support.n26.com/en-de/getting-started/psd2/how-to-log-in-with-two-factor-authentication
Weirdly, that one is not available with "en-eu" -- but it is with "en-gb" for example.

Does it really matter which location the server thinks we're in? As long as the documentation is in English, we're in the clear.

Carlgo11: Go with https://support.n26.com/en-gb/getting-started/psd2/how-to-log-in-with-two-factor-authentication

Carlgo11 (approved):

kmpoppe (approved):

PR #4163, opened 2019-10-22
Add Fidor

phallobst:


kmpoppe:
@phallobst Reading what Fidor says on the PSD2-page you linked, it looks to me that the _authentication_ part of what they did is **mTAN** only and therefore would only qualify as `sms` for the method? Because they say the **pushTAN** is only applied for transactions?

phallobst:
That might be true. However, I'm not quite sure how trustworthy their documentation actually is (see https://www.fidor.de/blog/sicheres-online-banking):
> Eine Sache an dieser Stelle: wir als Fidor Bank werden dich niemals nach einer mTAN zum Login in dein Konto fragen.

:man_facepalming:

kmpoppe:
Yeah sure ... that is a very contradiction in terms. My oh my. I'll plague them through their social media channels 😈

Carlgo11:
@kmpoppe  uhm... did they respond?

PR #4178, opened 2019-10-23
Update+Move Baden-Württembergische Bank

phallobst:


_data/banking.yml
```diff
@@ -84,6 +84,14 @@ websites:
       - hardware
     doc: https://www.avanza.se/kundservice.html/1335/tvafaktorsinloggning/d264248e-bea3-48a0-bf2d-157309734649
 
+  - name: Baden-Württembergische Bank
+    url: https://www.bw-bank.de/
+    img: bw-bank.png
+    tfa:
+      - proprietary
+      - hardware
```
Carlgo11, line 92:
```suggestion
      - hardware
      - sms
```

_data/banking.yml
```diff
@@ -84,6 +84,14 @@ websites:
       - hardware
     doc: https://www.avanza.se/kundservice.html/1335/tvafaktorsinloggning/d264248e-bea3-48a0-bf2d-157309734649
 
+  - name: Baden-Württembergische Bank
```
Carlgo11, line 87: Is the correct name `Baden-Württembergische Bank` or `BW Bank`?

Carlgo11 (changes requested):

_data/banking.yml
phallobst: SMS is actually _not_ used for authentication, but only for authorisation of credit card payments via 3D-Secure.
The available 2FA methods for login are chipTAN-QR (= hardware) and pushTAN (= proprietary).

_data/banking.yml
phallobst: Their legal info (https://www.bw-bank.de/de/home/toolbar/impressum.html) says:
> Die Baden-Württembergische Bank (BW-Bank) ist ...

The long name is a bit of a mouthful, so they shortened it to BW-Bank. However, I would leave it in so that users searching for the long name will still find it. Maybe even "Baden-Württembergische Bank (BW-Bank)"?

Carlgo11: That works :+1:

phallobst: Since the search apparently also looks at the URL, we don't even have to include "BW-Bank" here. I think the original commit should be good to go then.

Carlgo11 (approved):

kmpoppe (approved):

PR #4180, opened 2019-10-23
Add Allianz

phallobst:


_data/finance.yml
```diff
@@ -9,6 +9,13 @@ websites:
     facebook: AutomaticDataProcessing
     img: adp.png
 
+  - name: Allianz
+    url: https://www.allianz.de/
+    img: allianz.png
+    tfa:
+      - sms
```
Carlgo11, line 16: Also supports iTAN
```suggestion
      - sms
      - hardware
```

Carlgo11 (changes requested):

_data/finance.yml
phallobst: I'm not quite sure about this, the documentation is kind of dodgy there. One paragraph only talks about mTAN (= SMS) as the second factor, but another mentions iTAN for authentication.
I don't have an Allianz account myself, but I could send them a mail for clarification.

Anyway, since iTAN is only a paper-based list of codes, this doesn't even qualify as a second authentication factor in my opinion. It falls under "knowledge" as well, doesn't it?

Carlgo11: Paper codes are hardware 2FA. At least to our definition.

phallobst: Okay, I will contact Allianz support to provide clarification in this matter then. To be honest, I would be surprised if they really used paper codes for login purposes -- they would be used up pretty quickily.

phallobst: Still no reply by Allianz support to my mail from 2019-10-26... :frowning_face:
I wonder if they monitor their email at all.

Carlgo11: Meh, let's skip this then. Thanks for trying :+1:

Carlgo11 (approved):

kmpoppe (approved):

_data/finance.yml
phallobst: Just to bring closure to this cliffhanger:
The Allianz support finally replied and confirmed that only SMS is supported for 2FA.

PR #4181, opened 2019-10-23
Add Caixadirecta

esteveste:


_data/banking.yml
```diff
@@ -195,6 +195,14 @@ websites:
     img: caisse_epargne.png
     lang: fr
 
+  - name: Caixa Directa
+    url: https://www.cgd.pt/
+    tfa: 
+      - sms
+    doc: https://www.cgd.pt/Ajuda/Seguranca/Pages/Autenticacao-Dupla-SMS-Token.aspx
+    img: caixa_directa.png
+    lang: pt
```
```suggestion
```
`lang`-field must not be present, when one or more tfa-methods are available (it's only used for i18n-tweet-texts).

kmpoppe: Resolved by 
c6ab66f

Carlgo11 (approved):

Jawshy (commented):
Looks good to me, but I think `name` 'Caixa Directa' should be 'Caixadirecta'.

PR #4185, opened 2019-10-24
Add Saxo Bank

niclashedam:
This adds Saxo Bank to the list.
Saxo Bank provides different platforms depending on the client type. This represents the platform used by private persons.


Carlgo11 (changes requested):
You've taken a WEBP image and renamed it to PNG. You need to provide a proper PNG image that's 32x32 pixels and less than 2.5KB

niclashedam:
Sorry for the mistake. I was a little too quick.
I have replaced it with a compressed PNG that is within the boundaries.

kmpoppe:
Source branch and fork of 2fa is gone.

PR #4191, opened 2019-10-25
Update Schwab to reflect support of SMS and better notes.

PBorocz:
Based on first-hand experience and verbal confirmation with Schwab today.

### DIFF
--- a/_data/investing.yml
+++ b/_data/investing.yml
@@ -29,6 +29,7 @@ websites:
     url: https://www.schwab.com/
     img: charlesschwab.png
     tfa:
+      - sms
       - software
       - hardware
     doc: "/notes/schwab/"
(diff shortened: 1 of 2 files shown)


_data/investing.yml
```diff
@@ -29,6 +29,7 @@ websites:
     url: https://www.schwab.com/
     img: charlesschwab.png
     tfa:
+      - sms
```
Carlgo11, line 32: I don't see any mentions of SMS in the note?

PBorocz: Thanks @Carlgo11. Agreed...I added the SMS flag at the last moment after noticing that it wasn't there (yet having used it personally for several years)...thus, didn't think about adding it to the note. I have done so now _and_ explicitly confirmed Symantec support by Fidelity.

Carlgo11 (approved):

kmpoppe (approved):

PR #4193, opened 2019-10-26
Update DHL Deutschland

phallobst:
Doc is for the web interface. Using SMS for the app 2FA is first-hand information :)

### DIFF
--- a/_data/other.yml
+++ b/_data/other.yml
@@ -172,6 +172,9 @@ websites:
     img: dhl.png
     tfa:
       - email
+      - sms
+    doc: https://www.dhl.de/de/toolbar/footer/datenschutz.html#kundenkontoundempfaengerservices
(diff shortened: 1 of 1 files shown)


kmpoppe:
@phallobst I can't approve this PR because I don't know what App you're referencing. "DHL Paket"? I never found any way to make that App ask for a second factor when logging in?

Could you please provide screenshots of what exactly you are referring to (and maybe the second factor setup so that I can check this myself)?

// Kai

Goldin83 (approved):

phallobst:
I was wrong here, the "DHL Paket" app only requires an SMS verification at _setup_, not at every _login_.

kmpoppe (approved):

Carlgo11 (approved):

PR #4196, opened 2019-10-26
Add AOK

phallobst:


kmpoppe:
The `doc`, for me, just leads to a login-wall. As I'm not an AOK customer, that seems legit. Yet, I believe, that we still have to decide whether doc pages behind logins should be in the listing or not, right @Carlgo11 ?

phallobst:
Oops, I didn't realize that. Maybe https://meine.aok.de/landingpage/faq is sufficient for documentation purposes?

kmpoppe:
Yep, we can take that URL. It describes how mTAN works (albeit not the setup but ... ¯\_(ツ)_/¯).
Please put that URL in the doc :)
Thanks!

Carlgo11 (approved):

jamcat22 (approved):

PR #4199, opened 2019-10-27
Update Booking.com

phallobst:


Carlgo11:
The notes are meant for anything that's out of the ordinary but where the site doesn't offer proper documentation on the setup.

I don't see anything unusual with the setup process for 2FA on booking.com.
I suggest removing the note but leaving the exception text.

Carlgo11 (approved):

kmpoppe (approved):

PR #4200, opened 2019-10-27
Update verify.rb

kmpoppe:
Updates to `verify.rb` that is used with Travis and GitHub Actions:

* Wording for Image Size and Tags check: **should** => **must**, because it's a breaking error, not just a warning.
* Wording for Tags checked: `tfa: true/false` to `tfa is present/not present` 
* `test_img_file` function now accepts the website name to display it in error messages.
* All error messages now show the website name the error originated from to make them easier to find.

### DIFF
--- a/verify.rb
+++ b/verify.rb
@@ -57,16 +57,16 @@ def test_img_file(img, name)
   # Check image file size
   img_size = File.size(img)
   unless img_size <= @img_max_size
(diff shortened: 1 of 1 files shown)


Carlgo11:
#owned by rubocop 😄

kmpoppe:
> #owned by rubocop 😄

😝

Carlgo11 (approved):

PR #4212, opened 2019-10-28
Update Zoho Mail

Jawshy:
- Update 2FA: 
  - proprietary (Zoho OneAuth)
  - totp (Authenticator apps (like Google Authenticator))
  - u2f (Yubikey)
- New documentation page that lists all MFA options
- Exception

### DIFF
--- a/_data/email.yml
+++ b/_data/email.yml
@@ -262,6 +262,8 @@ websites:
     img: zoho.png
     tfa:
       - sms
-      - software
-      - hardware
-    doc: https://adminconsole.wiki.zoho.com/mail/Two-Factor-Authentication.html
+      - proprietary
+      - totp
+      - u2f
+    doc: https://help.zoho.com/portal/kb/articles/zohoone-configure-mfa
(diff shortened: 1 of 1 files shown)


kmpoppe (approved):

_data/email.yml
```diff
@@ -262,5 +262,8 @@ websites:
     img: zoho.png
     tfa:
       - sms
-      - software
-    doc: https://adminconsole.wiki.zoho.com/mail/Two-Factor-Authentication.html
+      - proprietary
+      - totp
+      - u2f
```
Carlgo11, line 267: If only Yubikeys are supported then it's not authentication through the FIDO U2F standard
```suggestion
      - hardware
```

Carlgo11 (changes requested):

_data/email.yml
Jawshy: @Carlgo11, I have investigated this further and it appears that Zoho is using FIDO U2F. [This help page](https://help.zoho.com/portal/kb/articles/mfa-yubikey) says 'Zoho uses U2F-based YubiKeys...' and sign in with YubiKey does not work when `security.webauth.u2f` is `false` in Firefox 70. It is odd, however, that YubiKey is explicitly mentioned when other U2F options exist.

Carlgo11 (approved):

PR #4213, opened 2019-10-28
Add nodebb

bdharrington7:
Adds NodeBB as a communication software that supports 2FA


Carlgo11:
We don't currently list self-hosted services or forums. See CONTRIBUTING and EXCLUSION.

PR #4231, opened 2019-10-30
Add Swizzonic

phallobst:
Alexa rank: 142,948


_data/hosting.yml
```diff
@@ -451,6 +451,13 @@ websites:
       - totp
     doc: https://support.stackpath.com/hc/en-us/articles/360022992631
 
+  - name: switchplus
```
Jawshy:
```suggestion
  - name: Swizzonic (formerly switchplus)
```
They have since changed names.

_data/hosting.yml
```diff
@@ -451,6 +451,13 @@ websites:
...
+    url: https://switchplus.ch/
+    img: switchplus.png
+    tfa:
+      - sms
+    doc: https://switchie.ch/2fa/
```
Jawshy: This link redirects to https://switchplus.ch/
After a brief search, I couldn't find any documentation.

Jawshy (changes requested):
In addition to their name, they have also changed their logo. 

As they also offer domain registration, should they also be listed under the domains category? We've done this with other sites like VentraIP.

_data/hosting.yml
phallobst: The link works for me... Strange. How about this one in English?
https://switchie.ch/en/2-factor-authentication-2fa/

_data/hosting.yml
phallobst: Do we really want include the former name?

Jawshy: I included their old name for the edge case that people may search the site use that name.

_data/hosting.yml
Jawshy: For me, that link also redirects to https://switchplus.ch/
I can read the contents preview [using a search engine](https://duckduckgo.com/?q=2+factor+authentication+site%3Aswitchie.ch&t=ffab&ia=web).

_data/hosting.yml
phallobst: The URL is included in the search terms, so using _Swizzonic_ as the name will be sufficient.

_data/hosting.yml
phallobst: I can reproduce that with a VPN connection via North America. I'd say we use the link to the English page anyway.

phallobst:
Handling multiple categories for one website is addressed in #4034. I'd say we wait for the result of that discussion and update Swizzonic later if necessary.

Jawshy (approved):
👍

Carlgo11 (approved):

PR #4238, opened 2019-10-31
Update Airbnb

phallobst:


kmpoppe:
The 2FA implementation is meh. I didn't manage to get asked for a second code. Strange.

kmpoppe:
@phallobst Please rebase the change to current master, because of AirBnb now moved to hotels.yml.
Thanks
// Kai

kmpoppe:
It feels like that didn't go as planned, because we've got a ton of commits in this PR now?

phallobst:
Aren't these the commits that were merged into my branch? According to the [branch comparison](https://github.com/2factorauth/twofactorauth/compare/master...phallobst:airbnb), there is only the update for the Airbnb entry left.

If that still is not correct, I could easily re-branch and submit a new PR.

PR #4240, opened 2019-10-31
Update ING-DiBa Austria

phallobst:


kmpoppe (approved):

Jawshy:
`SMS` is [mTAN](https://en.wikipedia.org/wiki/Transaction_authentication_number#Mobile_TAN_(mTAN)) and `proprietary` is the smart code. Is that correct?

phallobst:
That's correct.

Carlgo11 (approved):

PR #4261, opened 2019-11-02
Translate email subject

phallobst:
The email subject may be translated (similar to the tweets).

Furthermore, all translations are now optional, with English as the fallback.
Adding a language does not necessarily require the translation of all strings at once.


_includes/desktop-table.html
```diff
@@ -16,12 +16,19 @@
     </thead>
     <tbody class="jets-content">
     {% for website in section_file.websites %}
+    {% assign email_subject  = page.email_subject  %}
+    {% assign progress_tweet = page.progress_tweet %}
+    {% assign workonit_tweet = page.workonit_tweet %}
```
kmpoppe: I'd go for using `site.data.language["en"].` for this three lines (and for the same in the mobile-table.html), because that would eliminate the need to have the same variables added in index.html and thus remove the need to change them at two different places.
And because we need them in the `languages`-data we could strip them from index.html.
// Kai

kmpoppe (dismissed):

kmpoppe:
Looks good @phallobst!
Could you rebase your branch to the current master? That would add the Hebrew language from #4274 to your languages.yml and you could add the email-subject translation from that same PR to this one, would that be OK?
Thanks!
// Kai

phallobst:
There was some funky business going on with the right-to-left writing, but I think it is correct.

kmpoppe:
Hi  @phallobst Michael, I rebased your changes to your repo's current `HEAD` and cherry picked the changes that were really needed (with the hebrew language added to the base).
Thanks for your understading!

kmpoppe (approved):

dsoegijono (approved):

PR #4268, opened 2019-11-04
Add Park.io

dleshem:
Alexa rank: [57,838](https://www.alexa.com/siteinfo/park.io)


_data/domains.yml
```diff
@@ -438,6 +438,13 @@ websites:
       - software
     doc: https://www.pairdomains.com/kb/posts/345
 
+  - name: Park.io
+    url: https://park.io/
+    img: parkio.png
+    tfa:
+      - software
+    doc: https://park.io/support
```
Carlgo11, line 446:
```suggestion
```

_data/domains.yml
```diff
@@ -438,6 +438,13 @@ websites:
       - software
     doc: https://www.pairdomains.com/kb/posts/345
 
+  - name: Park.io
+    url: https://park.io/
+    img: parkio.png
+    tfa:
+      - software
```
Carlgo11: Please don't use the software tag. Use proprietary or totp depending what the site supports.
See [CONTRIBUTING.md](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#new-sites)

Carlgo11 (changes requested):
It's always a good idea to see if the image can be compressed better even if it's below 2.5KB
park.io

_data/domains.yml
dleshem: This link points to a FAQ page with the following question: "Can I use multi-factor authentication to add security to my Park.io account?".

Unfortunately there's no way to link directly to this question (no anchor in HTML).

I suggest we keep the link - it's better than nothing...

_data/domains.yml
dleshem: Park.io supports Authy accounts (7-digit codes). It won't work with, say, Google Authenticator.

Do you consider this proprietary or standard TOTP?

Carlgo11: That's a proprietary system (aka non RFC-6238).

phallobst:
Since the support page mentions "Authy", the tfa method should be "totp" instead, shouldn't it?

Carlgo11:
@phallobst no, Authy uses a proprietary system.

phallobst:
So the method itself is time-base one time passwords (aka "totp"), but since Authy is not initialized by scanning a QR code (or entering the secret manually), you cannot replace it by any other "totp" application, thus it counts as "proprietary". Did I understand that correctly?

Carlgo11:
That’s correct.
The totp tag really means that the site uses the RFC-6238 standard but that would be too complex for most people to understand.

phallobst:
They say that Authy is compliant, just not using the defaults:
https://www.twilio.com/docs/authy/api/one-time-passwords
https://twitter.com/authy/status/498244613766139904

kmpoppe:
@phallobst It's far too early for me to read through the RFC and see if they **really** are RFC compliant, but your assessment (no other Applications can read an Authy code, thus you **must** use the Authy App) is quite correct, that's the literal definition of proprietary.

_data/domains.yml
dleshem: @Carlgo11 Can you confirm leaving the doc link as-is, lacking a better option?

Carlgo11: Yeah it’s fine.

Carlgo11 (approved):

kmpoppe (approved):

PR #4272, opened 2019-11-04
Change "Email" icon

phallobst:
The @ symbol is one of the most recognizable icons in the world and should therefore be used for the "Email" category.
This also frees up the envelope icon for other (future) uses, e.g. postal delivery services.

### DIFF
--- a/_data/sections.yml
+++ b/_data/sections.yml
@@ -36,7 +36,7 @@
 
 - id: email
   title: Email
-  icon: mail
+  icon: at
 
 - id: entertainment
   title: Entertainment


Carlgo11:
I feel like the `@` is more associated with usernames than email addresses nowadays.
Just did a quick google image search on "email" and the top images are of envelopes.
https://www.google.com/search?q=email&tbm=isch

RichJeanes:
Yeah, maybe pre-social media that was true, but I agree with Carlgo on this one.

PR #4277, opened 2019-11-07
Add Semantic-ui website for icon usage

Rfosu2k2:
This will add the semantic-ui website under the "New Categories"  section for the proper naming convention to be used for the `icon-class`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,6 +96,7 @@ and follow the template below:
   title: Category Name
   icon: icon-class
 ```
+The `icon-class` value needs to be chosen from [Semantic-Ui][semantic-ui]https://semantic-ui.com/elements/icon.html.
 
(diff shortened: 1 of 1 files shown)


kmpoppe (approved):
Makes total sense, I was going to do that sooner or later ;-)

CONTRIBUTING.md
```diff
@@ -96,6 +96,7 @@ and follow the template below:
   title: Category Name
   icon: icon-class
 ```
+The `icon-class` value needs to be chosen from [Semantic-Ui][semantic-ui]https://semantic-ui.com/elements/icon.html.
```
Carlgo11: Why do you display it twice?

Rfosu2k2: I meant to remove the link after adding it in the footer.  Great catch.  I have removed the accidental duplication.

Carlgo11 (approved):

PR #4285, opened 2019-11-08
New category "Crowdfunding" with some entries from "Other"

### LINKED ISSUE #4273: Entries for "Crowd Funding" sites
I had a look at the "Other" category and found eight entries for crowd funding sites (there may be more, maybe even in different categories).

- Should these go to "Social" instead?
- Do they deserve a category of their own? To be honest, I haven't found a decent icon yet (that wasn't already taken).

phallobst:
Fixes #4273


kmpoppe (changes requested):
Hey @phallobst, afaik the `software` entries would be `totp`, would you be so kind and change these?

phallobst:
Changed `software` to `totp`.

Also, the exception for "GoFundMe" was probably a misunderstanding of the explanation on the support page, so I removed that as well.

kmpoppe:
Could your please also check out #4312 for the Kickstarter exception, if that is correct? 
Thanks a bunch!
// Kai

phallobst:
It is required to supply a phone number to set up 2FA, but do we really want to document this as an exception? This is used as the backup method for the TOTP code generator. To be honest, I expect every website to have some kind of fallback in case you cannot access your TOTP software, so it's not much of an exception in my eyes.

phallobst:
Since SMS is required and always available, it's probably actually better to state it in the exception.

kmpoppe (approved):

PR #4286, opened 2019-11-08
Move Wikipedia

phallobst:


kmpoppe:
Hey @phallobst Michael, why do you think that Wikipedia would be better off in the Social category?

phallobst:
In my opinion, "Social" should include those websites which engages their users to interact with each other in any particular way (so not just the "Social Media" sites). In the case of Wikipedia, they are working together to create _the_ most comprehensive encyclopedia ever.

("Quora" is similar in this regard I guess, and it's also in that category.)

jamiehowarth0 (approved):
Looks good

PR #4290, opened 2019-11-09
Move Other/Wikia to Social/Fandom

phallobst:
[Wikia](https://wikia.com) redirects to [Fandom](https://fandom.com/explore)


Carlgo11 (changes requested):
I don't think Fandom should be in the social media category.

phallobst:
The category is called "Social", not "Social Media", so I think it fits. The argument is the same as for [Wikipedia](https://github.com/2factorauth/twofactorauth/pull/4286#issuecomment-552074433).

kmpoppe (approved):

jamcat22:
Closing since the rebranding part is addressed in #4869. If we move Wikipedia in #4286, then we can also move FANDOM as well. If this PR switches to just moving FANDOM, feel free to reopen it @phallobst.

PR #4296, opened 2019-11-11
Update Web.de

### LINKED ISSUE #4295: Web.de 2FA only after adding mobile number
Web.de only allows Authenticatior App after adding the mobile number. See [Web.de FAQ](https://hilfe.web.de/sicherheit/2fa/einrichten.html#indexlink_help_sicherheit_2fa)

phallobst:
Fixes #4295

### DIFF
--- a/_data/email.yml
+++ b/_data/email.yml
@@ -230,8 +230,9 @@ websites:
     url: https://web.de
     img: webde.png
     tfa:
-      - software
+      - totp
     doc: https://hilfe.web.de/sicherheit/2fa/index.html
+    exception: "Setup requires SMS verification."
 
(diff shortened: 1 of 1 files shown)


Carlgo11:
So they don't support SMS 2FA but they still require an SMS code to setup 2FA?
Sounds a bit strange.

phallobst:
They require an SMS enabled phone to ensure there is a verified way to restore the password in case it's needed. So it is not actually necessary to set up 2FA, but you have to verify it _during_ the setup process.

Carlgo11 (approved):

kmpoppe (approved):

PR #4299, opened 2019-11-11
Add EVENTIM

phallobst:
Alexa rank: 6,568


_data/retail.yml
```diff
@@ -153,6 +153,14 @@ websites:
       - software
     doc: https://help.etsy.com/hc/en-us/articles/115015569567
 
+  - name: EVENTIM
+    url: https://www.eventim.de/
+    twitter: eventimDE
+    facebook: eventimDE
+    email_address: info@eventim.de
```
Carlgo11, line 160: Have you verified that the email address works and is monitored?

_data/retail.yml
```diff
@@ -153,6 +153,14 @@ websites:
       - software
     doc: https://help.etsy.com/hc/en-us/articles/115015569567
 
+  - name: EVENTIM
```
Carlgo11, line 156:
```suggestion
  - name: Eventim (DE)
```

Carlgo11 (dismissed):

_data/retail.yml
phallobst: I have not actually tried it, but it is listed on the imprint page: https://www.eventim.de/help/imprint/

_data/retail.yml
phallobst: They use the all-capitals notation throughout, with the exception of the logo, where it's all lowercase. :man_shrugging:

I'm okay with the country marker, though. How about `EVENTIM (DE)`? Or is that too loud?

Carlgo11: On twitter and facebook they use lowercase.
http://facebook.com/eventimDE
https://twitter.com/eventimDE
Also, the title on their page is `Eventim: [...]`

phallobst: Neither the facebook nor the twitter handle are the official company name, and the page title seems to be a mistake.

Although it may seem ugly, the all-capital spelling is the correct one (for Germany). Many of the [international branches](https://www.eventim.de/campaign/international/) use it as well.

kmpoppe: I checked whether it was possible to have your name begin with a lowercase letter on facebook (I know twitter has no problems with that) and it is, see f/edX. 
We've used the imprint of sites before to determine their legal name, which in this case is
CTS EVENTIM AG & Co. KGaA (company) and EVENTIM Management AG (shareholder).
Having the logo in all lowercase letters where the company name is all capitals might be the outright the stupidest marketing decision one could take but here we are. 
I'd, therefore, root for `EVENTIM`. I could also perfectly live with `eventim` that would be consistent with the logo. But I don't think `Eventim` "looks" right.

kmpoppe (approved):

Carlgo11 (approved):

PR #4300, opened 2019-11-11
Reduced icon file sizes

phallobst:
No further modifications other than passing through TinyPNG.
These were the largest icons in terms of file size.


Carlgo11 (changes requested):
`img/gaming/swtor.png` and `img/gaming/tibia.png` are only reduced by <100 bytes and since tinypng uses lossy compression I think it's better to leave those as they were.

`img/retail/speedcubecomau.png` contains text that I perceive as worse looking after the recompression so please undo that change as well.

phallobst:
I zoomed in on all three icon pairs at 500% and flipped back and forth, but couldn't really see any difference.

The "speedcubecomau.png" is a terrible logo to begin with (especially against the black background). Even without the text, the cube picture itself (from https://cdn.shopify.com/s/files/1/0646/0891/t/12/assets/logo.png) doesn't really look good at 32x32.

Let's revert these three then.

Carlgo11 (approved):
:+1:

kmpoppe (approved):

PR #4303, opened 2019-11-12
Add IT Services Category & move sites into category

### LINKED ISSUE #2130: Please add an IT Services category
The IT Industry is largely security conscious and would do well to seek out sites which protect their accounts using 2fa.

`Backup and Sync` and `Remote Access` are both of interest, but many other services in the `other` category ( Aerohive, CiscoMeraki, Kayako, Monkey Box, RepairShopr, Watchman Monitoring, and Zendesk, etc) don't relate to `Remote Access`

Would `IT Services` be category to add?  Perhaps with the Font Awesome "ticket" icon...

Rfosu2k2:
Fixes #2130

I chose the "microchip" icon as it seemed to be the most appropriate since all IT-Services require the use of a microchip in some capacity.  If this is not a suitable icon, please suggest other options.

I am sure there are other sites to be added from other.yml, but I only added what was originally mentioned.  Feel free to list any others needed in this category and I can move over.


Carlgo11:
IT Services describes almost all listings on the site. I suggest naming it something more specific.

Rfosu2k2:
> IT Services describes almost all listings on the site. I suggest naming it something more specific.

I found this image that breaks out IT-Services into 31 categories.  There are already some of these combined and listed.  There probably needs to be more than one category added to group sites listed in [other.yml](https://github.com/2factorauth/twofactorauth/blob/master/_data/other.yml) into their appropriate category.  Any suggestions based on this image?

IT_Services

Rfosu2k2:
How about `IT Support`?

Issue #4306, opened 2019-11-13
Websites with multiple 2FA policies


phallobst:
Some websites have various 2FA policies, for example depending on the user type (commercial vs. private, domestic vs. foreign, left-handed vs. right-handed). I'm sure we all agree that this is basically a terrible idea. However, the question is how to address these websites. There are several ways to handle them:

1. Document as _not_ supporting 2FA at all
2. Document as _not_ supporting 2FA, but with an exception text: "2FA is only available for right-handed people."
3. Document as supporting 2FA, but with an exception text: "2FA is not available for left-handed people."
4. Do _not_ include the site until their policy is sorted out
5. Add two entries with different info: "Initech.com (left-handed people)", "Initech.com (right-handed people)"

Currently. once the 2FA information has been added, the contact channels (twitter, facebook, email) are removed, which means there is no information how to pester them to roll out 2FA for everybody.

PR #4310, opened 2019-11-15
Add Vacasa & Vrbo to H & A

Rfosu2k2:


phallobst:
The [Support Pages](https://help.homeaway.com/category/Two_Factor_Authentication) suggest that 2FA (sms, phone) is actually supported by Vrbo. Those pages are hosted by homeaway.com, but that's because Vrbo was acquired by HomeAway and is just used as a brand name, as it seems.

Rfosu2k2:
> The [Support Pages](https://help.homeaway.com/category/Two_Factor_Authentication) suggest that 2FA (sms, phone) is actually supported by Vrbo. Those pages are hosted by homeaway.com, but that's because Vrbo was acquired by HomeAway and is just used as a brand name, as it seems.

I have checked after downloading both homeaway and Vrbo mobile apps, and there is no option for the(Traveler) to enable any sort of 2FA.  The support page you linked only states an option for the (Owner) website/mobile app.  I have confirmed that you can enable 2FA (sms,phone) via the (Owner) account, but even linking both the traveler and owner account the 2FA only works for the owner login.  

So this is in fact enabling 2FA, but the problem is that even with a linked account that only the owner portion is enabled.  This to me is only a partial implementation, and to me doesn't fully qualify as having strong 2FA enabled.

How should this be declared that documents it's only available for the owner account?  Also should we be listing both Vrbo & HomeAway, or just the parent company HomeAway?

Vrbo_HomeAway_2FA

phallobst:
1. Just two days ago I opened an issue regarding the documentation of multiple 2FA policies (#4306). Maybe a decision will be made there.
2. I think the entry might be called something like `HomeAway (Vrbo)` or `HomeAway / Vrbo`, so that searching for Vrbo would still find it. But since HomeAway is the parent company, the listing should be named after it and also point to its website. Using two entries is not very helpful, I guess.

PR #4314, opened 2019-11-16
Move PushCoin to Payments

phallobst:
As a payment platform (albeit targeted at schools/students), it does not belong in the Education category.


kmpoppe:
Already moved to `Finance` by #4327

phallobst:
Could you elaborate why this belongs to "Finance" rather than "Payments"? It seems to be just an e-wallet, with no additional "finance-y" features (i.e. loans, interest).

PR #4315, opened 2019-11-17
Move entries from Other to suitable categories

phallobst:


Carlgo11 (approved):

kmpoppe (dismissed):
I think Postcrossing is more suitable for `Social` than `Communication`, given the other entries in that category.

kmpoppe (approved):

PR #4319, opened 2019-11-17
Update Airbnb

phallobst:
Supersedes #4238

### DIFF
--- a/_data/hotels.yml
+++ b/_data/hotels.yml
@@ -1,9 +1,11 @@
 websites:
   - name: Airbnb
     url: https://www.airbnb.com
-    twitter: Airbnb
-    facebook: airbnb
     img: airbnb.png
+    tfa:
+      - email
+      - sms
+    doc: https://www.airbnb.com/help/article/501
 
   - name: Booking.com
     url: https://www.booking.com/


_data/hotels.yml
```diff
@@ -1,9 +1,11 @@
 websites:
   - name: Airbnb
     url: https://www.airbnb.com
-    twitter: Airbnb
-    facebook: airbnb
     img: airbnb.png
+    tfa:
+      - email
+      - sms
+    doc: https://www.airbnb.com/help/article/501
```
Carlgo11, line 8: The doc link doesn't describe how to enable 2FA.

phallobst: The doc page (kind of) suggests that it is enabled by default. There is a [reddit comment](https://www.reddit.com/r/AirBnB/comments/865zj2/how_to_enable_2fa/) by a "Verified Airbnb" user that would confirm this.

In the original PR #4238, @kmpoppe noted that he didn't get any verification message. Maybe this is dependent on the user's region?

Rfosu2k2:
I just checked on my account, and while they verify your email/phone number, there is still no option to enable 2FA.  They send you an email and sms to verify that you own those, but there is no option to use those for 2FA.  Per the documented link to MFA that airbnb uses, it only will send an email/sms when you logon from a new device.  So given that the MFA they use is only on new devices maybe there needs to be an exception added "Only sends codes when logging in from new device, uses the trusted device model."

Carlgo11 (approved):

kmpoppe (approved):

PR #4327, opened 2019-11-20
Move Pushcoin

Rfosu2k2:
Move Pushcoin to Finance as it is not actually an educational institute, it deals with the financing of students accounts for use at school.

-minor html edit line #19


Carlgo11 (approved):

kmpoppe (changes requested):
Would you please revert the ampersand changes to Amundo so that we keep one PR limited to one site only? Thanks!

Rfosu2k2:
> Would you please revert the ampersand changes to Amundo so that we keep one PR limited to one site only? Thanks!

Sure, I can open an additional PR for that change.  I have reverted the ampersand.

kmpoppe (approved):

PR #4342, opened 2019-11-24
Update Australia and New Zealand Banking Group

phallobst:


Carlgo11:
Looks more like 2SV

phallobst:
I guess it all boils down to what they mean with "accessing your account". Watching the video on the given page suggests that it's actually 2FA.
When the kid tries to log in, his mom gets a notification on her phone and the login screen asks for a code, leaving the criminal masterming hopeless and defeated, contemplating his life choices... :smile:

dsoegijono:
Tried it on my account and I can't find 2fa options. According to the doc, there are 2 options:
> Activate two layers of security by using the 2FA option whenever it is offered including Personal & Business banking – ANZ Voice ID and ANZ Shield

ANZ Voice ID can be used to increase transaction limit

ANZ Shield is also transactional [see doc](https://www.anz.com.au/security/account-protection/anz-shield/):
> Once you've registered for ANZ Shield, you'll be prompted to generate a Shield Code when making internet banking transactions or any changes to your account.

I don't think this is 2fa

phallobst:
Oh well, down the digital drain it is, then.

PR #4343, opened 2019-11-24
Update CIBC

phallobst:


Carlgo11:
> Sign on from a different location

That's the only mention of the verification process being required when logging in.
The wording makes me skeptic to if it's really 2FA or 2SV 🤔 

If they said "Sign on from a different device" it'd be like most banking sites on our list classified as supporting 2FA but "location" makes me think it might be based on geo ip data or ip-address based which I wouldn't say can be classified as 2FA.

phallobst:
On the [doc page](https://www.cibc.com/en/privacy-security/two-step-verification-b.html), it says right under the top picure:
> Starting November 6, you will not be able to sign on if you have not set up your account to receive one-time verification codes.

That sounds like 2FA is mandatory for signing (logging) in. Whether they use cookies to determine that you use a trusted device/browser should not be of our concern, I think.
It would be very surprising for me if they actually used geo location data, but if we really want be sure, either someone with a CIBC account or their support would have to clarify.

Carlgo11 (approved):

_data/banking.yml
```diff
@@ -246,9 +246,12 @@ websites:
 
   - name: CIBC
     url: https://www.cibc.ca
```
jamcat22:
```suggestion
    url: https://www.cibc.com
```
Permanent redirect, even from Canadian IPs.

jamcat22 (changes requested):

jamcat22 (approved):

jamcat22:
Twitter handle check issue already resolved in #4478.

PR #4357, opened 2019-11-27
Call to Action email: Adding body proposal

kmpoppe:
After the email subject was made translatable in #4261, this adds the call-to-action-tweet-text as the email body when clicking on the link.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,7 @@
 
 source 'https://rubygems.org'
 
-# Dependencies are bundled with the github-pages gem
-gem 'github-pages', group: :jekyll_plugins
+gem 'jekyll', group: :jekyll_plugins
 
 group :test do
   gem 'diff-lcs', platforms: :mswin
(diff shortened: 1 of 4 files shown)


phallobst:
Some testing (under Linux, with Opera/Chrome/Firefox/Vivaldi/Brave vs. Thunderbird) showed that most of time, the body was just
> Security is important

Only Firefox behaved as expected. The comma in the message seems to be treated specially. After I changed the code to `|uri_escape|replace:',','%2C'`, it worked all of time.

Is it a bug/shortcoming in `uri_escape` (that Firefox still handles correctly), or is it just something with my local setup?

kmpoppe:
> Is it a bug/shortcoming in `uri_escape` (that Firefox still handles correctly), or is it just something with my local setup?

Thanks for testing @phallobst! Obviously, I really tested using Firefox, where my "local" environment was also on Linux. I'll look into that further, thanks!
// Kai

phallobst:
The same problem (in theory) probably applies to the email subject as well. Currently there might not be any additional characters that need to be escaped, but in the future (or even in translations) there may be.

kmpoppe:
I built a plugin based on [this](https://stackoverflow.com/a/17769287/1838858) answer on StackOverflow.
@phallobst would you check if this now works locally with your browser/os?
@Carlgo11 I changed the Gemfile to accommodate that plugin. The Jekyll Action tests checked out and I didn't see anything wrong with it in my local env. As this site is no longer hosted on gh-pages, this should be working in master as well, right?

phallobst:
Yes, this solution works with all browsers on my side. :+1:

Carlgo11:
@kmpoppe the site is run on GitHub Pages so the gem is needed.

kmpoppe:
@carlgo11, where did I take a wrong turn then? Why did we kill the gh-pages branch?

Carlgo11:
The branch gh-pages isn’t needed for hosting on GitHub Pages. The master branch works just as well

kmpoppe:
Ah ffs. So I gotta figure out something new 🙄.
Bloody inconsistent program behaviour.

PR #4358, opened 2019-11-28
Add maccabi4u.co.il

dleshem:
Alexa rank: [16,649](https://www.alexa.com/siteinfo/maccabi4u.co.il)


phallobst (changes requested):
The image is 128x128 in size, but should be 32x32. There was a check in place to catch that, but it didn't run here, I guess.

dleshem:
> The image is 128x128 in size, but should be 32x32.

Done.

_data/health.yml
```diff
@@ -98,6 +98,13 @@ websites:
       - totp
     doc: https://support.hint.com/en/articles/2360627
 
+  - name: Maccabi Healthcare Services
+    url: https://www.maccabi4u.co.il/
+    twitter: MaccabiBriut
```
Carlgo11: Doesn't look like they monitor their Twitter account at all.
```suggestion
```

Carlgo11 (changes requested):

_data/health.yml
dleshem: @Carlgo11 This nitpicking frustrates me, and takes the fun out of contributing to this otherwise great project.
1. If you think this change is a blocker for merging this PR - do it yourself, and merge. Asking me to do it just adds back-and-forth bureaucracy. I can always add another PR if I care enough.
2. More generally, if there are guidelines for when a social channel should be added and when it shouldn't - write them down. Right now there are none, so as far as I'm concerned all official social channels should be added, regardless to anything.

I know this comes from good intentions, but you're not making it easy and fun to contribute.

kmpoppe: @dleshem 
grafik
We couldn't "do it ourselfes" b/c Github sometimes, even if you tick the "allow edits from maintainers" box outright refuses to allow these changes via the web interface (alas, one would always need to suggest the change in the first place before being able to use this mechanism to change the PR).
That means the only way to do so is adding your branch as a remote one, making the change, comitting and than removing the remote info to keep your own repo cleaan and not going crazy from having all the remote stuff lying around - which makes it all the more time consuming, which, as this is all purely volountarly and done besides our normal day jobs, can be a bit of a strain sometimes.
I'd therefore ask for your understanding, that it's sometimes easier to just suggest a change and move on to the next PR than going about correcting everything ourselfes.

Secondly, "write them down" is a very good point regarding exclusions and rules. Given the time this has taken for some parts of our guidelines, as you always need a bucketeload of time to find consensus on how to write stuff - sometimes down to the last word - this might be easier said than done. 
I will, obviously, take it on the agenda to find a wording for how we'd like to see social media be handled in PRs.

Meanwhile, again, I'd ask for your understanding for these points, while I thank you for your contributions to this project.!

// Kai

Carlgo11 (approved):

kmpoppe (approved):

PR #4362, opened 2019-11-28
Add Barnes & Noble

phallobst:
Alexa rank: 2,008


Carlgo11:
@phallobst you don't actually need to write the Alexa ranking. We'll check it either way.

Carlgo11 (approved):

kmpoppe:
@Carlgo11 I think @phallobst just adopted that from me adding the rank whenever I came across a PR to make reviewing easier. As we now have automated checks in place for the ranking we can now stop writing that down manually :-)

kmpoppe (approved):

PR #4363, opened 2019-11-28
Add Cloudbet

dleshem:
Alexa rank: [31,948](https://www.alexa.com/siteinfo/cloudbet.com)


phallobst:
The icon looks rather strange. Was it upscaled from 16px? Maybe there is a better one available.

dleshem:
> The icon looks rather strange. Was it upscaled from 16px? Maybe there is a better one available.

Yes, didn't find a high-res version with transparent background. How critical is this?

phallobst:
Not actually critical, but how about this one? It is used on Twitter and Facebook as well.
cloudbet

dleshem:
> Not actually critical, but how about this one? It is used on Twitter and Facebook as well.
> cloudbet

Thanks, done!

Jawshy (changes requested):
Cloudbet appears twice in betting. Also, you've got the old icon for Poalim as a new commit.

dleshem:
Damn, this got all screwed up. I'll just create a new PR.

PR #4364, opened 2019-11-28
Change Education icon

phallobst:
The "book" is a bit too generic here (and may be used otherwise).
The "graduation cap" fits much better.


kmpoppe (dismissed):

Carlgo11:
Really? I thought that the graduation cap was a UK- US-only thing.
The Education category contains school systems for kindergarten to universities plus apps like Duolingo and Pluralsight.

To me, the book is the perfect symbol for the very broad category "Education".

phallobst:
The graduation cap may be US/UK-centric, but IMHO an icon works in the other direction: You see the image and immediately derive its purposed meaning. That's why the car icon works for the "Transport" category as well, although that one contains mostly airlines and public transport systems.

When I see a book, I think of fiction or non-fiction writing, but not primarily of education. The graduation cap however is tightly related to it.

And if I may use your weapon of choice from #4272 against you: An image search for "education icon" shows quite a substantial number of graduation caps... :stuck_out_tongue_winking_eye:

Carlgo11:
You can use my “weapon of choice” but only if you do it correctly. Search for “education” and not “education icon” 😝

kmpoppe:
@phallobst @Carlgo11 Could you two get back to discussing which icon we take, please? :-)

PR #4365, opened 2019-11-28
Add craigslist

dleshem:
Alexa rank: [150](https://www.alexa.com/siteinfo/craigslist.org)


Carlgo11:
@dleshem the build failed because your fork is too old. You should probably pull from upstream before you make your next PR

_data/other.yml
```diff
@@ -115,6 +115,12 @@ websites:
       - totp
     doc: https://www.contentful.com/faq/two-factor-authentication/
 
+  - name: Craigslist
+    url: https://craigslist.org/
+    img: craigslist.png
+    twitter: craigslist
+    facebook: Craigslist-74561112489
```
Carlgo11: Page returns 404 for me.

Carlgo11 (changes requested):

_data/other.yml
dleshem: Twitter: https://twitter.com/craigslist
Facebook: https://www.facebook.com/Craigslist-74561112489

Both work for me.

kmpoppe:
```suggestion
    facebook: officialcraigslist
```
That page doesn't 404 for me and it looks nothing less official than the twitter account with it's last activity in 2009.
I really wonder if there is any point in adding those social media channels as they don't seem to care about them at all.

dleshem: I don't mind changing, but the page I linked to has 451,611 likes, while this one has 1,140. Let me know.

dleshem: You can also try this URL (which is the one Facebook's HTML marks as canonical): https://www.facebook.com/pages/category/Advertising-Marketing/Craigslist-74561112489/

dleshem: OK, I give up :( Let's go with the Facebook page that has 0.25% the number of likes.

PR #4367, opened 2019-11-29
Add World4you

aschneider-w4y:


_data/hosting.yml
```diff
@@ -539,7 +539,14 @@ websites:
     twitter: WixHelp
     facebook: wix
     img: wix.png
-
+    
+  - name: World4You
+    url: https://www.world4you.com
+    img: world4you.png
+    tfa:
+      - software
```
Carlgo11:
```suggestion
      - totp
```

Carlgo11 (changes requested):

_data/hosting.yml
phallobst: The doc page says:
> The app is compatible with the free Time-Based One-Time (TOTP) standard...

They even refer to Microsoft Authenticator as an alternative. So this should just be "totp".

Carlgo11: Nice catch @phallobst!

Carlgo11 (approved):

kmpoppe (approved):

PR #4369, opened 2019-11-30
Add Verimi

phallobst:


Carlgo11 (commented):
I see password (knowledge), pin (knowledge) and approval some a trusted device (no idea if that's even an auth type).

Am I missing something?

phallobst:
The authentication is done via a proprietary app on your smartphone by confirming a push notification. The app itself is protected by fingerprint.

phallobst:
Big oof on the Alexa front (~200k -> ~600k). I guess I'll close this PR for now, that will take some time to recover...

PR #4374, opened 2019-12-01
Add SCHUFA

phallobst:
(Name in all capital letters is deliberate)


Carlgo11 (approved):
Another German site... 🙄

kmpoppe (approved):

PR #4387, opened 2019-12-05
Add DomainFactory

szepeviktor:
See https://forum.df.eu/forum/allgemein/datensicherheit-und-privatsph%C3%A4re/306-zwei-faktor-authentifizierung


Carlgo11 (changes requested):
Hello @szepeviktor,
thanks for the pull request!

As DomainFactory first and foremost advertise themselves as a domain registrar I think the domain category is better suited for them.
Multiple categories is something we're looking into but for now we try to list companies under their main service.

szepeviktor:
Okay.
Starting migration....

szepeviktor:
# 🛑 

Critical Error :) https://github.com/2factorauth/twofactorauth/blob/master/_data/domains.yml#L91-L96

Carlgo11:
Sorry for the inconvenience 😅

PR #4392, opened 2019-12-08
Change Questrade

obscurans:


_data/investing.yml
```diff
@@ -135,6 +135,8 @@ websites:
     tfa:
       - sms
       - email
+    doc: https://questrade-support.secure.force.com/mylearning/view/h/Account/Account-security-centre/
+    exception: "SMS must be used as primary, and email must be used as secondary. SMS number and email must be the ones on the account profile."
```
Carlgo11:
```suggestion
```

_data/investing.yml
```diff
@@ -135,6 +135,8 @@ websites:
     tfa:
       - sms
       - email
```
Carlgo11: Thanks for the doc link! If email 2FA only is used as a backup method then it shouldn't be listed here at all.
```suggestion
```

Carlgo11 (changes requested):

Carlgo11 (approved):

kmpoppe (approved):

PR #4394, opened 2019-12-08
Add MyFRITZ!

phallobst:
Although the doc page refers to FRITZ!Box 7590, the MyFRITZ! login is device-independent.


_data/remote.yml
```diff
@@ -85,6 +85,14 @@ websites:
       - totp
     doc: http://www.cc-onlinehelp.com/en/#index/administrati/users/authenticati/authenticati.html
 
+  - name: MyFRITZ!
+    url: https://sso.myfritz.net/
+    img: myfritz.png
+    tfa:
+      - totp
+      - hardware
+    doc: https://en.avm.de/service/fritzbox/fritzbox-7590/knowledge-base/publication/show/3427
```
Carlgo11, line 95: Is there any other doc page (maybe in German?) that better explains how to enable 2FA?

phallobst:
I haven't found any better documentation, but here's a screenshot of the setup process:

tfa-myfritz

The "Proof of Personal Identity" also supports using a U2F key, so I added that as well. I suspect that fingerprint and face recognition works on iOS devices only, so I could not check those.

_data/remote.yml
phallobst: (see below)

DeanR97 (approved):

Carlgo11 (approved):

_data/remote.yml
```diff
@@ -85,6 +85,15 @@ websites:
...
     doc: http://www.cc-onlinehelp.com/en/#index/administrati/users/authenticati/authenticati.html
 
+  - name: MyFRITZ!
+    url: https://sso.myfritz.net/
+    img: myfritz.png
+    tfa:
+      - totp
+      - hardware
+      - u2f
+    doc: https://en.avm.de/service/fritzbox/fritzbox-7590/knowledge-base/publication/show/3427
```
ApCoder123, line 95: I think that this doc page might be slightly more useful - it mentions both U2F and TOTP at the bottom of the page and has better details on how to enable
```suggestion
    doc: https://en.avm.de/service/fritzbox/fritzbox-7590/knowledge-base/publication/show/3602
```

ApCoder123 (changes requested):

phallobst:
Superseded by #5703

PR #4397, opened 2019-12-09
Add Degussa Bank

phallobst:


kmpoppe:
@phallobst Degussa uses a slightly different image as their favicon and I find that much nicer to look at, albeit it not showing the text. What do you think?
grafik

phallobst:
That may be true, but even on their Twitter and Facebook posts they use that down-scaled logo with the (hardly readable) text. It's not our job to brush up their rubbish corporate identity skills... :man_shrugging:

Carlgo11:
I like your idea @kmpoppe but it might be legally problematic to edit their logo to that extent.
Aside from that you Germans don't seem to have any fair use laws which makes adding sites here technically illegal 🤷‍♂

kmpoppe:
What do you lot think about upscaling https://www.degussa-bank.de/degussa-bank-theme/images/degussa-logo-mini.png to 32*32?

phallobst:
The original dimensions of 24x29 do not allow for a graceful upscaling, but they are close enough to 32x32 to just resize (and center) it.
degussa
However, I haven't seen this logo in use on their website or social media channels. It may be outdated.

kmpoppe:
¯\\__(ツ)__/¯ So be it, let's take the one from your PR then - and hope for a better one at a later date.

kmpoppe (approved):

kmpoppe:
> I like your idea @kmpoppe but it might be legally problematic to edit their logo to that extent.
Yes, Copyright Law in Germany and stuff. Don't change a Logo even in size ...

> Aside from that you Germans don't seem to have any fair use laws which makes adding sites here technically illegal 🤷‍♂

There has been a court case in 2011 that ruled it to be illegal, to use the logo from a car manufacturer for advertising the same services as they do. The "Markengesetz" (Law on Brands) §23 __explicitly__ allows the use of a brand from a company if you refer to their services (and at the same time don't offer the same services). So there is actually as paragraph allowing our use and not one prohibiting it with exceptions or stuff.

Carlgo11 (approved):

PR #4401, opened 2019-12-10
Add River Financial

sampeiomichi:
Although the www.river.com site currently does not meet Guideline #4 (Alexa top 200K), the company River Financial is a service that should be listed on twofactorauth.org due to the nature of its business. Under Site Criteria, it would fall under Money (#3) as River Financial is a brokerage service for buying and selling Bitcoin.


Carlgo11:
When a site is close to 200K but with a major social media following we can sometimes make an exception for that but since the site you want to add is ranked above 2M I’m afraid it will have to get a much better ranking before we can accept it.

sampeiomichi:
We actually just bought a new domain and rebranded our product. Our old domain was alto.financial and our Twitter is https://twitter.com/RiverFinancial. We previously ranked about 1M on Alexa. Would this suffice?

sampeiomichi:
> When a site is close to 200K but with a major social media following we can sometimes make an exception for that but since the site you want to add is ranked above 2M I’m afraid it will have to get a much better ranking before we can accept it.

We actually just bought a new domain and rebranded our product. Our old domain was alto.financial and our Twitter is https://twitter.com/RiverFinancial. We previously ranked about 1M on Alexa. Would this suffice?

Carlgo11:
No but something around 250K would be.

Your product is simply not used by enough people to be listed here.

Sorry about the inconvenience.

PR #4404, opened 2019-12-12
Move+Update Parsec

phallobst:


phallobst:
Although the doc page does not explicitly say so, the TFA method is actually regular TOTP.

Carlgo11 (approved):

kmpoppe (approved):
How in the world did that entry end up in `remote`? Well then...

PR #4409, opened 2019-12-14
Show "Thank them" tweet button only if Twitter handle is available

### LINKED ISSUE #4407: "In Progress" assumes twitter handle to be set
Example: Developer / Read the Docs

There is no twitter handle, but the "Thank them..." twitter link is displayed anyway, yet with an empty @ recipient.

kmpoppe:
Fixes #4407

### DIFF
--- a/_includes/desktop-table.html
+++ b/_includes/desktop-table.html
@@ -41,9 +41,13 @@
           </div>
         </td>
         <td class="twitter progress" colspan="6">
+          {% if website.twitter %}
           <a class="ui twitter mini button"
              href="https://twitter.com/share?url={{site.url|cgi_escape}}&amp;text={{progress_tweet|replace:'TWITTERHANDLE',website.twitter|cgi_escape}}&amp;hashtags={{page.hash|cgi_escape}}"
              target="_blank"><i class="twitter icon"></i> {{page.link_progress}}</a>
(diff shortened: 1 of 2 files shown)


_includes/desktop-table.html
```diff
@@ -41,9 +41,13 @@
           </div>
         </td>
         <td class="twitter progress" colspan="6">
+          {% if website.twitter %}
           <a class="ui twitter mini button"
              href="https://twitter.com/share?url={{site.url|cgi_escape}}&amp;text={{progress_tweet|replace:'TWITTERHANDLE',website.twitter|cgi_escape}}&amp;hashtags={{page.hash|cgi_escape}}"
              target="_blank"><i class="twitter icon"></i> {{page.link_progress}}</a>
+          {% else %}
+          &nbsp;
```
Carlgo11, line 49: why the space?

kmpoppe: Since forever, some browsers go haywire when a table cell is empty (read regex: `<td>\s*</td>`) also see [here](https://www.emailonacid.com/blog/article/email-development/emailology_empty_table_cells/), so the space makes it display empty while not being considered empty by the browser.

Carlgo11 (approved):

PR #4411, opened 2019-12-14
Update JetBrains

windmueller:
JetBrains use their own Hub software for managing logins to their services, e.g. YouTrack. The mentioned documentation is actually from that product.


phallobst:
Do they really offer U2F on their own website? I created a test account, and the 2FA setup process only greeted me with this:
>**What is two-factor authentication?** Two-factor authentication is an extra layer of security for your JetBrains Account. After you turn it on, signing into your account will require both your regular password and a 6-digit one-time verification code.
>
>**A mobile device is required.** Using this security feature requires a mobile device capable of scanning a QR code using a downloadable authenticator application.
>
>**Recovery your access.** We will provide you with a list of recovery codes that you can use to sign in if you have lost or can’t access your mobile device.

So for me it's still just TOTP.

windmueller:
The explanation you are seeing is from the account used to manage your licenses.
I was referring to the Hub account which is used for their products, e.g. https://youtrack.jetbrains.com/

phallobst:
Then the exception should rather say "U2F only available for Hub account...", since TOTP is available for the main account.

dsoegijono (approved):

phallobst (approved):

Issue #4410, opened 2019-12-14
Update software tags


Carlgo11:
All the categories needs to have their software 2FA tokens replaced with our API 2 software tags.
So far these categories have been updated:
- [x] Backup
- [x] Banking
- [x] Cloud
- [x] Communication
- [x] Cryptocurrencies
- [x] Developer
- [x] Domains
- [x] Education
- [x] Email
- [x] Entertainment
- [x] Finance
- [x] ~Food~
- [x] Gaming
- [x] Government
- [x] Health
- [x] Hosting
- [x] ~Hotels~
- [x] Identity
- [x] Investing
- [x] ~IoT~
- [x] Legal
- [x] Other
- [x] Payments
- [x] Remote
- [x] Retail
- [x] Security
- [x] Social
- [x] Task
- [x] Transport
- [x] Utilities
- [x] VPN

phallobst:
In the works:
- [x] banking
- [x] betting
- [ ] developer
  - aTech Media has insufficient documentation
  - Phacility is lacking documentation
- [x] domains
- [x] education
- [x] email
- [x] entertainment
- [x] finance
- [x] gaming
- [x] government
- [x] health
- [ ] hosting
  - Hostnet has insufficient documentation
- [x] identity
- [x] investing
- [x] legal
- [x] other
- [x] payments
- [x] remote
- [x] retail
- [x] security
- [x] social
- [x] task
- [x] transport
- [x] utilities
- [x] vpn

phallobst:
So far, I have packed all the categories in to one PR, but by now I think it might become a bit crowded in there. Shall I rather open one PR per category?

Carlgo11:
One per category please!

PR #4418, opened 2019-12-15
Update "software" method in Task Management

phallobst:
See #4410

### DIFF
--- a/_data/task.yml
+++ b/_data/task.yml
@@ -9,7 +9,7 @@ websites:
     img: airtable.png
     tfa:
       - sms
-      - software
+      - totp
     doc: https://support.airtable.com/hc/en-us/articles/219409917
 
   - name: Any.do
@@ -60,8 +60,8 @@ websites:
     url: https://www.meistertask.com
     img: meistertask.png
     tfa:
-      - software
-    doc: https://support.meistertask.com/hc/en-us/articles/207318670
+      - totp
+    doc: https://support.mindmeister.com/hc/en-us/articles/218901367
 
   - name: Microsoft To-Do
     url: https://todo.microsoft.com/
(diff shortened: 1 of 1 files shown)


Carlgo11 (approved):

_data/task.yml
```diff
@@ -60,23 +60,23 @@ websites:
     url: https://www.meistertask.com
     img: meistertask.png
     tfa:
-      - software
-    doc: https://support.meistertask.com/hc/en-us/articles/207318670
+      - totp
+    doc: https://support.mindmeister.com/hc/en-us/articles/218901367
```
kmpoppe, line 64: I guess you choose the MindMeister URL because it details the TOTP Apps?

phallobst: MeisterTask uses the MindMeister account, so the tfa method(s) should be the same.

_data/task.yml
```diff
@@ -60,23 +60,23 @@ websites:
...
 
   - name: Microsoft To-Do
     url: https://todo.microsoft.com/
     img: microsofttodo.png
     tfa:
       - sms
       - email
-      - software
+      - totp
```
Jawshy, line 72: In additional to TOTP, Microsoft has their own authenticator app called [Microsoft Authenticator](https://play.google.com/store/apps/details?id=com.azure.authenticator&hl=en) so `proprietary` could also be included under supported 2FA options.

Jawshy (commented):
Bar one comment, looks good to me.

_data/task.yml
phallobst: The _Microsoft Authenticator_ app is covered by the `totp` method, since it is [RFC-6238](https://tools.ietf.org/html/rfc6238) compliant and can be replaced by other apps, such as _Google Authenticator_ or _andOTP_.

Jawshy: Excellent. Thanks for the information.

Jawshy (approved):

PR #4433, opened 2019-12-16
Update "software" method in Health

phallobst:
See #4410

### DIFF
--- a/_data/health.yml
+++ b/_data/health.yml
@@ -88,6 +88,7 @@ websites:
       - sms
       - phone
       - totp
+      - proprietary
       - u2f
     doc: https://www.google.com/intl/en-US/landing/2step/features.html
 
--- a/_data/health.yml
+++ b/_data/health.yml
@@ -3,7 +3,7 @@ websites:
     url: https://www.23andme.com/
     img: 23andme.png
     tfa:
-      - software
+      - totp
     doc: https://customercare.23andme.com/hc/en-us/articles/360034119874
 
   - name: Australian Red Cross Blood Service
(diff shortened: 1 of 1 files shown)


_data/health.yml
```diff
@@ -87,7 +87,7 @@ websites:
     tfa:
       - sms
       - phone
-      - software
+      - totp
```
Carlgo11, line 90:
```suggestion
      - totp
      - proprietary
```

_data/health.yml
```diff
@@ -87,7 +87,7 @@ websites:
...
       - hardware
```
Carlgo11:
```suggestion
      - u2f
```

Carlgo11 (changes requested):

_data/health.yml
phallobst: Someone will have to explain to me which hardware constitutes "u2f" and which does not.

_data/health.yml
phallobst: Which proprietary software (that doesn't just support regular totp, like Authy) do you mean?

Edit: I just noticed that other Google entries have this as well, so for the sake of consistency, let's add it here.

Carlgo11 (approved):

_data/health.yml
kmpoppe: @phallobst You can use the "Google App" on your phone to sign in to your account, which is `proprietary` - because you don't enter a TOTP but just answer the push message on the phone to log in. I think that's a fair use of prop ;)

kmpoppe (approved):

_data/health.yml
phallobst: I should have known that, since I also use push notifications for Google 2FA. Oh well...

PR #4441, opened 2019-12-17
Update "software" method in Developer

phallobst:
See #4410

- aTech Media has insufficient documentation
- Phacility is lacking documentation

### DIFF
--- a/_data/developer.yml
+++ b/_data/developer.yml
@@ -7,7 +7,7 @@ websites:
       - phone
       - proprietary
       - hardware
-    doc: https://support.aha.io/hc/en-us/articles/202000957
+    doc: https://support.aha.io/hc/en-us/articles/360031319232-Account-settings#SSO
 
   - name: Airbrake
     url: https://airbrake.io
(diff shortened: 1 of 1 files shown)


_data/developer.yml
```diff
@@ -129,45 +131,44 @@ websites:
     img: cmtelecom.png
     tfa:
       - sms
-      - phone
-      - software
-    doc: https://help.onlinesmsgateway.com/en/accounts-and-payments/how-do-i-enable-two-factor-authentication
+      - proprietary
```
Carlgo11, line 135: I don't see any mentions of any app.

Carlgo11 (approved):

_data/developer.yml
phallobst: I tried this by myself:
cmtelecom-tfa
The website did not accept my phone number (for whatever reason), so I could not check whether they use regular TOTP. However, the "CM Authenticator" app was not able to scan the TOTP QR codes that I already have, so I would classify this as "proprietary".

Jawshy (approved):

PR #4442, opened 2019-12-17
Proper number formatting

phallobst:
This works for big numbers (>1M) as well.
Some whitespace fixes, while we're at it...

### DIFF
--- a/.tests/alexa.sh
+++ b/.tests/alexa.sh
@@ -1,13 +1,18 @@
 #!/bin/bash
 
+export LC_NUMERIC=en_US.UTF-8
+format_number () {
+  numfmt --grouping "$@"
+}
+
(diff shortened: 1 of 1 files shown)


.tests/alexa.sh
```diff
@@ -33,14 +38,14 @@ check_rank () {
     fi
 
     # Check if the rank is at or below 200K
-		if [ "$alexa_rank" -gt "$max_ranking" ]; then
```
Carlgo11, line 36: I was just about to comment that you've used tabs instead of spaces which is against the editorconfig settings but then I realized that I'm the one who used it... 🙄 \*cough\* \*cough\*

Carlgo11 (approved):
All checks out 👍

kmpoppe (approved):

PR #4444, opened 2019-12-17
Add Western Union

phallobst:


Jawshy (commented):
Does Western Union explicitly say that 2FA is only available for business accounts? I can only infer so based on the sub-domain for the documentation. 

Their third 2FA option is Authy application which looks like SMS or phone with extra steps. If customers could scan a QR code for the shared secret it would probably be TOTP instead of proprietary.

phallobst:
I have created a non-business WU account, and there is no sight of 2FA whatsoever. I think there was some kind of support page that explicitly stated that 2FA is only available for business accounts, but I cannot find it anymore. Maybe I can dig it up.

_Authy_ does use a TOTP algorithm, but does not qualify as such for the purpose of this list, since it cannot replaced by some other application. In this case, the setup is done not by scanning a QR code, but by linking the app via the mobile phone number that has been verified before.
(see https://business.westernunion.com/en-us/product-documents/tfa-overview/how-to-install-Authy)

Carlgo11 (approved):

Jawshy (approved):
OK. So I called Western Union on four numbers to get the answer. They do not offer 2FA for non-business accounts; only for business accounts. 

PR approved.

PR #3951, opened 2019-07-30
Add SoFi

jlw:
SoFi offers a high-interest online checking account, investment accounts, and several loan products.

PR #4446, opened 2019-12-18
Update "software" method in Banking

phallobst:
See #4410

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -9,7 +9,7 @@ websites:
     url: https://www.addiko.hr/
     img: addikobank.png
     tfa:
-      - software
+      - totp
       - hardware
     doc: https://www.addiko.hr/static/uploads/mtoken-uputa-za-koristenje.pdf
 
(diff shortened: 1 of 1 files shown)


_data/banking.yml
```diff
@@ -998,7 +995,6 @@ websites:
     img: sofi.png
     tfa:
       - sms
-      - software
```
Carlgo11, line 1001: https://github.com/2factorauth/twofactorauth/pull/3951/files#r309195970
```suggestion
      - totp
```

Carlgo11 (changes requested):

_data/banking.yml
phallobst: Is that information still accurate? The doc only mentions SMS.

Carlgo11: 🤷‍♂ yeah you're right. Let's keep it as SMS only for now.

Carlgo11 (approved):

kmpoppe (approved):

PR #4449, opened 2019-12-18
Update Evernote

### LINKED ISSUE #4382: Evernote Update
(1) Even on the free account, SMS 2FA can be used by selecting "I need help getting a verification code." when prompted for an authenticator app code.

(2) As a corollary to the above, SMS can be used to bypass software method.

To cover both, I suggest updating the exceptions/restrictions text: "Software method can be bypassed with SMS."

phallobst:
Closes #4382

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -57,9 +57,10 @@ websites:
     img: evernote.png
     tfa:
       - sms
+      - phone
       - totp
     doc: https://help.evernote.com/hc/en-us/articles/208314238
-    exception: "SMS authentication only available for Evernote Premium users"
+    exception: "SMS and phone call are always enabled as backup methods for TOTP."
 
   - name: Files.com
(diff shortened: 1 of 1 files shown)


_data/backup.yml
```diff
@@ -57,9 +57,10 @@ websites:
     img: evernote.png
     tfa:
       - sms
+      - phone
```
Carlgo11, line 60: Is this from personal experience or is there some documentation that they support 2FA through phone calls which I can't find?

phallobst: This is from personal experience. After clicking the "I need help getting a verification code" link, the following screen appears for me:
evernote-tfa
I don't know whether voice call is supported everywhere, or why they didn't include it in their docs. But it did work.

Carlgo11 (approved):

kmpoppe (approved):

PR #4451, opened 2019-12-18
Update Hostiso

phallobst:


phallobst:
hostiso

_data/cloud.yml
```diff
@@ -80,7 +80,8 @@ websites:
     url: https://hostiso.com/
     img: hostiso.png
     tfa:
-      - software
+      - totp
+    doc: https://github.com/2factorauth/twofactorauth/pull/4451#issuecomment-567219085
```
Carlgo11: doc links are designed to show the user how to enable 2FA on the site. What you've added is more of proof that they use 2FA.
```suggestion
```

Carlgo11 (changes requested):

_data/cloud.yml
phallobst: Well, it shows that Hostiso uses TOTP instead of some proprietary solution. Some users might be interested in that, as long as there is no official documentation. So it's better than nothing...

Carlgo11: That's what Git is for.

phallobst: Okay, I don't mind.

Carlgo11 (approved):

kmpoppe (approved):

PR #4458, opened 2019-12-21
Update build status image source

Carlgo11:
Updated the build status image since we no longer use Travis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 TwoFactorAuth.org
 =================
 
-[![Build Status](https://img.shields.io/travis/2factorauth/twofactorauth/master?style=for-the-badge)](https://travis-ci.org/2factorauth/twofactorauth)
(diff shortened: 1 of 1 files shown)


phallobst:
Can this be something other than "passing" if each PR has to pass the check before it is merged?

Carlgo11:
@phallobst uhm not really. I can’t change the status messages

phallobst:
I think you misunderstood. What I meant was:

Is it possible to have a situation where the build status is something other than "passing", given that each PR individually has to pass the checks before merging? If not, that status is not really useful.

Carlgo11:
Ah no this build status is for the master branch. So unless we merge a PR that breaks our site it shouldn’t fail.

phallobst (approved):

dsoegijono (approved):

PR #4473, opened 2019-12-28
Add Bank Of Scotland

rfadam:
Bank Of Scotland is a sister brand to Halifax & Lloyds bank in the UK and has no 2FA as per the siblings. Listed as Alexa siterank 16,652


Carlgo11:
Are you sure they don't provide any 2FA? EU banks are required to have mandatory 2FA for online banking as of this fall.

rfadam:
So they have a secondary password required sometimes (e.g. Logging in from a new device or activating Biometrics on their mobile app) but No 2FA as expected where a code is generated or sent to you. As I referenced in the opening comment, is the same group as Halifax & Lloyds, so I would expect all to get 2FA at the same time if it was supported.

Carlgo11:
What about https://www.bankofscotland.co.uk/aboutonline/changes-to-internet-banking.html?WT.ac=common/promotion/r3pr/serv/til/s/rl/BoSSevHpTl and "Changes to banking online"?

rfadam:
So I just tried following those instructions on a completely different PC to one I usually use and I was not prompted nor received a code when logging in - but would do if I added a new Payee.

So the question now is - what should be classified as 2FA? As I believe you can login to Internet banking without 2FA but would be prompted for a code sent to you when doing certain actions. Whatever the result is, I would be happy to either update this PR plus also update Lloyds / Halifax

_data/banking.yml
```diff
@@ -119,6 +119,12 @@ websites:
       - hardware
     doc: https://www.bochk.com/en/security/2factorauthentication.html
 
+  - name: Bank of Scotland
+    url: https://www.bankofscotland.co.uk
+    img: bankofscotland.png
+    twitter: askbankofscot
```
Carlgo11:
```suggestion
```

_data/banking.yml
```diff
@@ -119,6 +119,12 @@ websites:
...
+    facebook: bankofscotland
```
Carlgo11:
```suggestion
    tfa:
      - proprietary
      - phone
      - sms
```

Carlgo11 (changes requested):
As they're required to do this on login I'm rather confident that it's not just for certain actions. It might be that they're doing some weird stuff when checking for 'known devices' and that could be why you didn't get asked to enter any codes.

_data/banking.yml
phallobst: Are you sure about "totp"? The confirmation via their app should rather be "proprietary". I see no sign of TOTP.

Carlgo11: @phallobst you're right. My bad

_data/banking.yml
rfadam: Will update PR to remove this

_data/banking.yml
rfadam: Ok - Thanks for suggestions. I'll update the PR to include this plus also Lloyds & Halifax to the same if they are not already

Carlgo11: @rfadam try to keep one site listing per PR.

Carlgo11 (approved):

kmpoppe (approved):

PR #4481, opened 2019-12-30
Fix bundler cache & speed up html-proofer

Carlgo11:
In this PR I've:
* Sped up the bundler dependency installation
* Fixed the bundler gem cache
* Made html-proofer use 5 threads instead of 1 (~5s improvement)
* Moved html-proofer to the end of the script to make it more likely that the other parts of the script finds the error in a bad build and therefore fails the build faster.

html-proofer takes up the majority of the time for our PR checks to run and I can't find any reason for it to be this slow. We might need to search for a replacement in the future.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -18,11 +18,11 @@ end
 
 task :proof do
   HTMLProofer.check_directory(
-    './_site', \
-    assume_extension: true, \
-    check_html: true, \
-    disable_external: true, \
-    cache: { timeframe: '2d', storage_dir: '/tmp/html-proofer' }
+    './_site',
+    assume_extension: true,
+    check_html: true,
+    disable_external: true,
+    parallel: { in_threads: 5 }
   ).run
 end
 
--- a/.github/workflows/jekyll.yml
+++ b/.github/workflows/jekyll.yml
@@ -9,18 +9,18 @@ jobs:
     steps:
     - uses: actions/checkout@v1
 
-    - name: Get Cache
-      id: cache
-      uses: actions/cache@v1
+    - uses: actions/cache@v1
       with:
         path: vendor/bundle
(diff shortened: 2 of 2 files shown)


Rakefile
```diff
@@ -18,11 +16,11 @@ end
...
-    './_site', \
-    assume_extension: true, \
-    check_html: true, \
-    disable_external: true, \
-    cache: { timeframe: '2d', storage_dir: '/tmp/html-proofer' }
+    './_site',
+    assume_extension: true,
+    check_html: true,
+    disable_external: true,
+	parallel: { in_threads: 5 }
```
phallobst: The "parallel" line is indented by a tab instead of 4 spaces, so "rubocop" fails:

> Offenses:
> 
> Rakefile:23:1: C: Layout/Tab: Tab detected.
> 	parallel: { in_threads: 5 }
> ^
> Rakefile:23:2: C: Layout/AlignHash: Align the elements of a hash literal if they span more than one line.
> 	parallel: { in_threads: 5 }
>  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 
> 2 files inspected, 2 offenses detected
> RuboCop failed!
>

Rakefile
```diff
@@ -5,8 +5,6 @@ require 'rubocop/rake_task'
 require 'jekyll'
 require 'jsonlint/rake_task'
 
-task default: %w[proof verify jsonlint rubocop]
-
```
phallobst: This breaks "bundle exec rake" (as given in `CONTRIBUTING.md`) since there are no default tasks:
>rake aborted!
>Don't know how to build task 'default' (see --tasks)

phallobst (changes requested):
see inline comments

PR #4485, opened 2020-01-01
Add MyHeritage

phallobst:


phallobst:
@Carlgo11 Previously, the check logs showed this message:
> NOTE: Inheriting Faraday::Error::ClientError is deprecated; use Faraday::ClientError instead. It will be removed in or after version 1.0

which is just what happened here:
> Fetching faraday 1.0.0
> Installing faraday 1.0.0

I don't know if the dependencies are wrong, or if the default version has been bumped to 1.0.0 -- whatever the reason is, I can hardly do anything about it.

Carlgo11 (approved):

kmpoppe (approved):

PR #4491, opened 2020-01-02
Update Wave to add update and change name

aidenmitchell:


_data/finance.yml
```diff
@@ -383,11 +383,12 @@ websites:
     facebook: VitalityUK
     img: vitality.png
 
-  - name: WaveApps
+  - name: Wave Accounting
```
Carlgo11: Wave Accounting seems to be one of the Waveapps products while the login system is for all the products.
Therefore either leave it as it is (`WaveApps`), use their legal name (`Wave Financial`)  or just `Wave`.

_data/finance.yml
```diff
@@ -383,11 +383,12 @@ websites:
...
     url: https://www.waveapps.com
     img: wave.png
     twitter: WaveHQ
     facebook: waveHQ
+    status: https://community.waveapps.com/discussion/3424/add-in-option-for-two-factor-authentication-2fa
```
Carlgo11: Do you see something which I don't? Because to me it looks like they don't have any intentions of adding 2FA let alone are actively working on it.

Carlgo11 (changes requested):

_data/finance.yml
aidenmitchell: Didn't realize that, thanks.
```suggestion
  - name: Wave
```

_data/finance.yml
aidenmitchell: Screen Shot 2020-01-03 at 3 20 31 PM

aidenmitchell: I looked for a way to directly link to that specific comment, but it doesn't look possible.

Carlgo11: That's from March 14, Then in August and October they write this:

aidenmitchell: Found a direct link, changing.

```suggestion
    status: https://community.waveapps.com/discussion/3424/add-in-option-for-two-factor-authentication-2fa#Comment_15005
```

aidenmitchell: Oh, I didn't look that far. That's too bad. Sorry for the nuisance.
```suggestion
```

Carlgo11 (approved):

kmpoppe (approved):

PR #4501, opened 2020-01-06
Update NexusMod

apollo2k4:
Nexusmod enable software tokens


Carlgo11 (changes requested):
Please revert the doc link change. They both work but the old one is shorter.

apollo2k4:
Revert to the old link. Thanks for the comment.

Carlgo11 (approved):
Thanks!

kmpoppe (approved):

PR #4118, opened 2019-10-16
Add Fanatical.com to Gaming

webdevian:

PR #4502, opened 2020-01-07
Add Fanatical.com

webdevian:
Including suggested change from https://github.com/2factorauth/twofactorauth/pull/4118


webdevian:
Not sure why Circle is failing, looks like a ruby gem error

Jawshy (approved):

Carlgo11 (dismissed):
I think the site would be better fit under the retail section. Please move it.

phallobst:
"Humble Bundle" is also under _Gaming_ - these two should be in the same category.

webdevian:
Yes, it should definitely be in the same category as Humble Bundle. The product range is very similar

jamcat22 (approved):

Issue #4515, opened 2020-01-11
Tutanota does not support hardware 2FA on desktop/mobile apps


leggomyinfo:
Add exceptions & restrictions: "Hardware 2FA not supported for the desktop & mobile apps."

Cite: https://www.reddit.com/r/tutanota/comments/e5dti7/u2f_on_linux_app_security_key_not_registered_for/

PR #4516, opened 2020-01-11
Add Tutanota exception

leggomyinfo:
Tutanota doesn't currently support U2F through their desktop or mobile apps. It is only supported when logging in through a browser. [According to Tutanota,](https://www.reddit.com/r/tutanota/comments/e5dti7/u2f_on_linux_app_security_key_not_registered_for/) support for this is coming "soon" as of a month ago. Since it's on their roadmap, I included "yet" in the exception language to differentiate them from companies who have not indicated eventual support.

Closes out #4515.

### DIFF
--- a/_data/email.yml
+++ b/_data/email.yml
@@ -213,6 +213,7 @@ websites:
       - totp
       - hardware
     doc: https://tutanota.com/howto/#2fa
+    exception: "Hardware 2FA not supported for the desktop & mobile apps."
 
   - name: Virgin Media E-Mail
     url: https://www.virginmedia.com/
--- a/_data/email.yml
+++ b/_data/email.yml
@@ -211,7 +211,7 @@ websites:
     img: tutanota.png
     tfa:
       - totp
-      - hardware
+      - u2f
     doc: https://tutanota.com/howto/#2fa
     exception: "Hardware 2FA not yet supported for their desktop & mobile apps."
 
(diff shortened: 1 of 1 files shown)


_data/email.yml
```diff
@@ -213,6 +213,7 @@ websites:
       - totp
       - hardware
```
Carlgo11: Let's fix the hardware tag as well while we're at it.
```suggestion
      - u2f
```

Carlgo11 (changes requested):

_data/email.yml
Carlgo11: firefox_PFbhu3qndV
There should be a "Commit suggestion" button available to you. Sadly I can't use it because of a GitHub bug 🙄

leggomyinfo: Should it have a leading space in front of the dash like " - hardware" does? By clicking "Commit changes," I am approving your additional change, right?

Carlgo11: The leading space is another GitHub bug hehehe 🙁  (It's not really there)
Yes the "Commit changes" incorporates my suggestion into the PR.

Carlgo11 (approved):
Great! Your first pull request 🥳 🎈 

Once another maintainer also approves it, your change will go live.

kmpoppe (approved):

PR #4517, opened 2020-01-11
Update GoDaddy

### LINKED ISSUE #4514: GoDaddy restriction missing
"If hardware 2FA enabled, cannot call customer support without also enabling software or SMS 2FA."

leggomyinfo:
- added U2F tag (works with Yubikey)
- added restriction (GoDaddy presents this warning when enabling hardware 2FA)

- addresses #4514


_data/domains.yml
```diff
@@ -213,7 +213,9 @@ websites:
       - sms
       - software
       - hardware
+      - U2F
```
Carlgo11:
```suggestion
      - u2f
```

_data/domains.yml
```diff
@@ -213,7 +213,9 @@ websites:
       - sms
       - software
       - hardware
```
Carlgo11: Looks like they don't support any hardware methods other than U2F.
```suggestion
```

_data/domains.yml
```diff
@@ -213,7 +213,9 @@ websites:
...
+      - U2F
     doc: https://www.godaddy.com/help/enable-two-step-verification-7502
+    exception: "If hardware 2FA enabled, customer support unavailable by phone without also enabling software or SMS 2FA."
```
Carlgo11: I feel like this is more straight forward. What do you think?
```suggestion
    exception: "U2F 2FA requires another backup 2FA method when calling GoDaddy support."
```

Carlgo11 (changes requested):

_data/domains.yml
leggomyinfo: So we only add specific tags when a site supports more than one method?

_data/domains.yml
leggomyinfo: Prefer your brevity but I think some clarity is lost. What about "If hardware 2FA enabled, customer support by phone requires enabling back-up method."

_data/domains.yml
Carlgo11: the hardware tag is a catch all for every single hardware method. Since you added the U2F tag already, and they don't provide anything other than U2F, then the hardware tag is redundant.
Does that make sense?

leggomyinfo: @Carlgo11 Got it. I don't think I'm a believer in the approach (confusing mobile UX) but I'll defer to those who've been working on this much longer than me!

_data/domains.yml
dsoegijono: I agree with @djfredalert's suggestion, it's easier to read/understand.

leggomyinfo:
@Carlgo11 I tried to resolve the conflict... doesn't look it worked.

Carlgo11:
It’s fixed now. That happens when your branch is too outdated.

leggomyinfo:
> It’s fixed now. That happens when your branch is too outdated.

So did changing "software" to "totp" fix it? Or you/something else fixed it?

Carlgo11:
Because I updated your branch. (Merged master with it)

jamcat22 (approved):

Carlgo11 (approved):

PR #4522, opened 2020-01-12
Append "(Google account)" and add exception to Google services

### LINKED ISSUE #4510: Connection between Google products unclear
It makes sense for each Google product (Drive, Gmail, etc) to appear on its own to make it easy to find for those looking for it, but it is also confusing because it suggests that each has its own independent 2FA settings. To address the latter shortcoming, I would suggest the following:

(1) Append " (Google account)" to the name column of every site which draws from one's Google account settings for 2FA.
(2) Add "Google account" as a row in the "Other" section.

This would also address the problem that certain Google products don't show up when you search Google (e.g. YouTube, even though "Hangouts" shows up despite also not having "Google" in the name).

### LINKED ISSUE #4511: Google does not support NFC security keys
Google does not support NFC security keys (https://support.google.com/accounts/thread/13082621?hl=en). As a result, you cannot use YubiKeys to log onto anything Google-connected through iOS (except through a workaround like the Yubico Authenticator app).

Suggested solve - add "Exceptions & Restrictions" for all Google products: "NFC security keys not supported"

Would apply to Google Drive, Google Cloud Platform, Hangouts, Google Domains, Gmail, Google Play, Google Fit, Google Pay, Google Fi, Google Fiber, YouTube (there may also be others)

leggomyinfo:
Addresses #4510 and #4511.


leggomyinfo:
@Carlgo11 Figured out how to edit the utilities yml which I initially couldn't find. Can I edit the commit to update Google Fi and Google Fiber or do I need to create a different pull request/commit in this same branch?

Carlgo11:
Just make the edit inside the branch you used to create this PR and it should show up here automatically.

phallobst:
I am not a fan of the "(Google account)" addition to the website names. This does not really add relevant information for the user. If I look for "Google Domains" 2FA for example, I don't care whether the same options apply to "Google Play" as well.

From a developer/contributor view, an [include mechanism](https://github.com/2factorauth/twofactorauth/pull/4034#issuecomment-566500921) would probably help more, but that needs some additional effort.

BTW: Was the _Dropbox_ change included by mistake?

leggomyinfo:
> I am not a fan of the "(Google account)" addition to the website names. This does not really add relevant information for the user. If I look for "Google Domains" 2FA for example, I don't care whether the same options apply to "Google Play" as well.

@phallobst I agree that it may not add much value in this use case. It is more relevant for a user scanning through each section, or for someone searching "Google" so they can see everything covered (e.g. YouTube).


> From a developer/contributor view, an [include mechanism](https://github.com/2factorauth/twofactorauth/pull/4034#issuecomment-566500921) would probably help more, but that needs some additional effort.

If there's a more elegant solution, happy for that to be captured and for this to be deleted.


> BTW: Was the _Dropbox_ change included by mistake?

Yeah, just removed it from this branch (I think).

PR #4523, opened 2020-01-12
Update Twitter

### LINKED ISSUE #4468: Update Twitter exceptions/restrictions
Exceptions & Restrictions: Hardware 2FA requires also enabling Software or SMS 2FA as an alternative login option.

I don't know what the current language ("SMS only available on select providers.") refers to which is the only reason I omitted it above. I'm not suggesting that be deleted.

leggomyinfo:
- Cannot add more than one security key
- Mobile apps do not support hardware keys

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -253,7 +253,7 @@ websites:
       - software
       - hardware
     doc: https://support.twitter.com/articles/20170388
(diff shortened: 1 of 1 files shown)


Jawshy (approved):
All three exceptions ✔️

phallobst (approved):

PR #4527, opened 2020-01-12
Update ConEd

leggomyinfo:
- the documentation link was broken; what I found is not ideal but I couldn't find anything else
- 2FA can be bypassed by customer service
- phone number required to set up any method

### DIFF
--- a/_data/utilities.yml
+++ b/_data/utilities.yml
@@ -88,7 +88,7 @@ websites:
       - totp
       - proprietary
     doc: https://www.coned.com/-/media/files/coned/documents/small-medium-large-businesses/building-project-center/contractor-resources/user-registration-guide.pdf?la=en
-    exception: "2FA method can be bypassed by customer support." 
+    exception: "Requires phone # to set up. Can be bypassed by customer support." 
 
   - name: Cox
     url: https://www.cox.com/
--- a/_data/utilities.yml
+++ b/_data/utilities.yml
@@ -87,7 +87,8 @@ websites:
       - phone
       - totp
       - proprietary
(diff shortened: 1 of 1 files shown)


phallobst:
Does the phrase "Can be bypassed by customer support." really merit to be an exception here? I'm aware that it opens up social engineering options, but having the support helping the user to log in would apply to quite a number of sites, I guess.

leggomyinfo:
> Does the phrase "Can be bypassed by customer support." really merit to be an exception here? I'm aware that it opens up social engineering options, but having the support helping the user to log in would apply to quite a number of sites, I guess.

I think that's a fair point. Con Ed states it explicitly during the login process which is why I thought it was warranted here, if not elsewhere.

_data/utilities.yml
```diff
@@ -87,7 +87,8 @@ websites:
       - phone
       - totp
       - proprietary
-    doc: https://www.coned.com/en/accounts-billing/new-my-account-faqs
+    doc: https://www.coned.com/-/media/files/coned/documents/small-medium-large-businesses/building-project-center/contractor-resources/user-registration-guide.pdf?la=en
+    exception: "Requires phone # to set up. Can be bypassed by customer support."
```
Carlgo11: Proper English without slang is preferred when writing exceptions. That way more people and translation services can understand it.
```suggestion
    exception: "Requires phone number to set up. Can be bypassed by customer support." 
```

Carlgo11 (changes requested):

Carlgo11 (approved):

kmpoppe (approved):

PR #4528, opened 2020-01-12
Update Tresorit

leggomyinfo:
Tresorit requires enabling at least two methods.

### DIFF
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -225,7 +225,7 @@ websites:
       - email
       - totp
     doc: https://support.tresorit.com/hc/en-us/articles/216114527
-    exception: Requires enabling at least two methods.
+    exception: "Requires enabling at least two 2FA methods."
 
   - name: Wasabi
     url: https://wasabi.com/
--- a/_data/backup.yml
+++ b/_data/backup.yml
@@ -225,6 +225,7 @@ websites:
       - email
       - totp
     doc: https://support.tresorit.com/hc/en-us/articles/216114527
+    exception: Requires enabling at least two methods.
 
   - name: Wasabi
     url: https://wasabi.com/


phallobst:
According to the doc page, this is not actually required. It says
> If you have multiple verification methods set up, ...
> ...if you set up a secondary option as a backup.

Is the documentation out of date, i.e. do you have first-hand information here?

leggomyinfo:
> According to the doc page, this is not actually required. It says
> 
> > If you have multiple verification methods set up, ...
> > ...if you set up a secondary option as a backup.
> 
> Is the documentation out of date, i.e. do you have first-hand information here?

@phallobst It's actually in the documentation, albeit hard to spot: [https://support.tresorit.com/hc/article_attachments/360003584614/primary_2x.gif](https://support.tresorit.com/hc/article_attachments/360003584614/primary_2x.gif)

phallobst:
I did see that, but it's still contradicting the text. Either the text or the GIF is out of date, I guess.

Edit: I tried it out myself. The user actually needs to select at least two methods, so the GIF is up-to-date and the text is not. Removing the second-to-last option disables 2FA altogether.

leggomyinfo:
The text is out-of-date. I have a Tresorit account and confirmed two are required.

_data/backup.yml
```diff
@@ -225,6 +225,7 @@ websites:
       - email
       - totp
     doc: https://support.tresorit.com/hc/en-us/articles/216114527
+    exception: Requires enabling at least two methods.
```
Carlgo11:
```suggestion
    exception: "Requires enabling at least two 2FA methods."
```

Carlgo11 (changes requested):

Carlgo11 (approved):

kmpoppe (approved):

PR #2662, opened 2017-07-27
added additional methods now available in bitwarden

kspearrin:
bitwarden recently released additional 2FA support for email, YubiKeys, FIDO U2F, and Duo (which supports push, sms and phone calls). see https://help.bitwarden.com/article/setup-two-step-login/ for all details

PR #4529, opened 2020-01-12
Bitwarden update

leggomyinfo:
Does not support SMS or phone call

### DIFF
--- a/_data/identity.yml
+++ b/_data/identity.yml
@@ -20,8 +20,6 @@ websites:
     url: https://bitwarden.com
     img: bitwarden.png
     tfa:
-      - sms
-      - phone
       - email
       - software
       - hardware


Jawshy:
Bitwarden premium users can use Duo for phone, SMS, and push notifications ([support page](https://help.bitwarden.com/article/setup-two-step-login/) and [original PR](https://github.com/2factorauth/twofactorauth/pull/2662)).

leggomyinfo:
> Bitwarden premium users can use Duo for phone, SMS, and push notifications ([support page](https://help.bitwarden.com/article/setup-two-step-login/) and [original PR](https://github.com/2factorauth/twofactorauth/pull/2662)).

Good catch. My bad.

RichJeanes:
An exception wouldn't be out of order to note that a subscription is required if you would like to add that.

PR #4427, opened 2019-12-15
Update "software" method in Identity Management

phallobst:
See #4410

PR #4532, opened 2020-01-12
Update "software" method in Identity Management

Carlgo11:
Continuation of #4427 as it won't accept commits any longer 🙄


Carlgo11:
Merging as @Jawshy and I have approved the previous PR.

PR #4541, opened 2020-01-18
Fixing security flaw CWE-1022

aidenmitchell:
https://cwe.mitre.org/data/definitions/1022.html
https://lgtm.com/rules/1806963085/


Carlgo11:
I don't see the issue here. We're a static page so there shouldn't be anything sensitive in the DOM.

aidenmitchell:
Just a precaution, I'll close it if this is unnecessary.

Carlgo11:
Yeah there's no risk factor in leaking the DOM so for us it's unnecessary.

PR #4555, opened 2020-01-26
Add HelloBank! and LCL

moviuro:


Carlgo11:
Due to the PSD2 EU law I'm pretty sure they use some sort of 2FA.

moviuro:
As client of both establishments, I can confirm that there is absolutely none at the moment. Even with PSD2.

I don't use LCL's mobile app, and on the web it never asked me for a 2FA challenge; even when accessing my bank account from work in an incognito window.

*Maybe* HelloBank! does some sort of partial proprietary 2FA with their ["finger key"](https://www.hellobank.fr/faq/comment-activer-la-cle-digitale.html); though most operations on the web from my workplace don't require any 2FA.

phallobst:
I found these pages regarding PSD2 (= DSP2 in French):
* https://www.hellobank.fr/fr/client/dsp2
  presumably SMS and proprietary app (Digital Key / Clé Digitale)
* https://www.lcl.fr/dsp2-af
  presumably SMS 

I did have to rely on online translation services here, but I think these are strong indications that 2FA is in use, as mandated by EU regulations.

It should be noted however that banks are allowed to only ask for a strong customer authentication every 90 days. This may explain the perceived absence of 2FA.

moviuro:
@phallobst I have *never* received an SMS message to open my LCL account so far. The only texts I received are for authenticating payment (3DSecure), which don't use the same info (card number instead of customer/contract number).

As for HelloBank, they don't check it either: never once had to use the "finger key" --"clef digitale"-- to open my session on a browser.

Also, depending on how you interpret PSD2, it's a weak, half-measure that just adds a false sense of security:

1. I try to open my session at home
2. I use my 2FA token (SMS or "finger key"), and successfully pass the challenge
3. My session is open on my computer
3. I logout
4. One hour later someone in another part of the world uses my leaked credentials to open my session on their machine
5. Does PSD2 mandate this person solve a challenge?

- If not, I got pwned and PSD2 is not 2FA
- If yes, then PSD2 *is* legit 2FA -- but *so far* experiment shows that it does not happen

As a security consultant myself, I would be *very* careful about labelling banks as having 2FA simply because of PSD2. How 2factorauth decides to label what is valid 2FA though, is a choice that should be clearly stated (see e.g. my scenario).

phallobst:
Okay, this is my state of knowledge right now:

First of all, the PSD2 (to be precise: the supplementing EU directive [2018/389](https://eur-lex.europa.eu/legal-content/EN/TXT/HTML/?uri=CELEX:32018R0389)) mandates _Strong Customer Authentication_ (SCA) in the common 2FA sense: two (or more) elements of _knowledge_, _possession_ and _inherence_.

However, a grace period of 18 months (ending December 2020) has been granted to fully implement SCA. Apparently, this is especially utilized in France and the UK. This would explain the current behaviour of "Hello Bank!" and "LCL".

In short: PSD2 _is_ 2FA, but it's not in place everywhere right now.

(Maybe we need a proper lawyer on our team...)

Carlgo11:
But then it isn’t really worth listing this bank yet as we know the entry will change within a couple of months.

moviuro:
@phallobst I am currently working at a bank, where directors, UX-experts and the like chose the following interpretation with no/little input from security:

> If the customer did not do 2FA in the last 90 days (*[Chapter III](https://eur-lex.europa.eu/legal-content/EN/TXT/HTML/?uri=CELEX:32018R0389#d1e653-23-1)*)
> then do 2FA
> else do usual login.

In that case, PSD2 does not amount to 2FA (see my [scenario](https://github.com/2factorauth/twofactorauth/pull/4555#issuecomment-593308086)). My take on the issue because I'm an insider is, for the project: *don't blindly add banking sites as 2FA-compliant simply because of DSP2*.

@Carlgo11 that might be wise. I'll report back whenever the situation changes for either or both.

phallobst:
I think this boils down to how the implementation of the login process looks like, and how "the customer" is identified. If the bank (as is common) stores a cookie in the browser after a successful 2FA login, then the cookie itself could act as the second factor for subsequent logins, i.e. something you have. In that scenario, someone at the other end of the world would not be able to circumvent 2FA.

Carlgo11:
I'm closing this PR for now. We can come back to in later in Q3/Q4

moviuro:
@Carlgo11 @phallobst  Just tested on both LCL and HelloBank:

1. Open my session from my own personal computer
2. Correctly pass the ~2FA "challenge" (Press the finger key for HelloBank, SMS OTP for LCL)
3. Logout
4. Open my session from my work machine using its VPN
5. *no challenge*
6. Logout

phallobst:
Okay, so they seem to have implemented only the bare minimum regarding PSD2 (or SCA for that matter). Once you log in with 2FA confirmation, the account is only protected by password for the following 90 day period. Security by hoping for the best IMHO.

kmpoppe:
Security through obscurity is NOT 2FA 😜

Carlgo11 (approved):
Alright let's add them

moviuro:
For the record, though: don't blindly add European banking sites to the 2FA list because of PSD2, unless someone can confirm they really do it (right).

jamcat22 (approved):
CircleCI is failing because of an already-removed Twitter handle. This issue is fixed in later versions of the build script.

PR #4571, opened 2020-02-03
Twitch requires Authy for software 2fa

Stevoisiak:


Carlgo11:
We've moved away from listing these types of things as exceptions. Instead we use the proprietary tag.

PR #4573, opened 2020-02-05
Add Aetna

miamimanni:


Carlgo11 (changes requested):
The health category is for health apps and services. Insurance is supposed to be in the finance category.

miamimanni:
> The health category is for health apps and services. Insurance is supposed to be in the finance category.

Ah gotcha! Fixed

Carlgo11 (approved):

Jawshy (approved):

PR #4546, opened 2020-01-19
Update Internet.bs

phallobst:

PR #4585, opened 2020-02-11
Update Internet.bs

phallobst:
The update from #4546 didn't last long...

### DIFF
--- a/_data/domains.yml
+++ b/_data/domains.yml
@@ -261,7 +261,6 @@ websites:
     img: internetbs.png
     tfa:
       - totp
-      - proprietary
     doc: "/notes/internetbs/"
 
   - name: INWX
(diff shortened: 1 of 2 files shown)


phallobst:
From a mail to their customers:
> Please take notice that on the 1st of March 2020, we will be removing the CryptoPhoto Two Factor authentication service from our offering. At this time, any accounts still using CryptoPhoto will no longer have their account secured via Two Factor authentication.
>
> If you wish to continue to make use of Two Factor Authentication please update your account to make use of Google Authenticator. You can update your account to this option by logging into your account then navigating to the "My Profile" page. Here you will see the option to "Manage" your Two-step login verification settings.

Carlgo11 (approved):

Jawshy (approved):

Issue #4230, opened 2019-10-30
Missing translations for "call to action" tweets/emails


kmpoppe:
We are currently missing translations for the following languages for our "call to action tweets/emails":
- [ ] Bengali
- [ ] Croatian
- [ ] Czech
- [ ] Estonian
- [ ] Finnish
- [ ] Hebrew
- [ ] Norwegian
- [ ] Polish
- [ ] Spanish

The English texts are as follows:
* "Call to Action" tweet:
  `Security is important, @TWITTERHANDLE. We'd like it if you supported two factor authentication.`
* "Call to Action" email subject:
  `Support Two Factor Authentication`

If you want to contribute, please just open a Pull Request and please reference this issue **without** the `closes`, `fixes` or likewise keywords so that this Issue stays open.

If you want to add any other language that we are missing, feel free to do this as well :-)

Thanks in advance to anyone willing to contribute!

// Kai

kmpoppe:
All translations exist now

PR #4593, opened 2020-02-15
Add Banco Galicia (AR)

fainpablo:


phallobst:
Could you also use https://tinypng.com to compress the icon (on this and future PRs)? This often reduces the file size considerably.

fainpablo:
> Could you also use https://tinypng.com to compress the icon (on this and future PRs)? This often reduces the file size considerably.

Sure thing. Latest file should be 1 KB.

phallobst:
Even for seemingly small files it does matter. In this case, TinyPNG produced a virtually indistinguishable icon of only 524 bytes:
bancogalicia
Given that the twofactorauth.org list includes almost 1400 icons, this actually makes a difference in the long run.

fainpablo:
@phallobst Makes sense. I've just uploaded the compressed file. Thanks again.

phallobst (approved):

phallobst:
May I ask you to have a look at Issue #4230 as well? We are still looking for Spanish translations of the two tweets and the email subject that are in use, and you can certainly help us out there. I'm not sure whether there are significant differences for Argentinian Spanish (if that's a thing...), but it would certainly be better than nothing.

fainpablo:
> May I ask you to have a look at Issue #4230 as well? We are still looking for Spanish translations of the two tweets that are in use, and you can certainly help us out there. I'm not sure whether there are significant differences for Argentinian Spanish (if that's a thing...), but it would certainly be better than nothing.

Absolutely. I've just submitted PR #4594. In spanish we use "two step" instead of "two factor", as you can see [here](https://www.google.com/landing/2step/?hl=es) and [here](https://www.facebook.com/help/148233965247823). While they can be considered different things, I believe it will be better to keep using "step" if we take into account the recipients of the tweets. If you prefer to use the word "factor", just let me know and I'll make the changes.

Carlgo11 (approved):

PR #4596, opened 2020-02-15
Add HSBC Argentina (AR)

fainpablo:


fainpablo:
HSBC Argentina no longer supports 2FA for accessing the online banking although on their website is still stated that the token is required for authentication too. Now they only require 2FA for high risk transactions, so they idea is to allow their customers to request this feature back.

phallobst:
So they previously required 2FA for login, but not anymore? This sounds rather strange, and quite frankly I wouldn't expect that from such a major player in the banking industry.

Could it be that the user does not have to provide a code/token for **each** login, but only once every 90 days or so? This is quite common with other banks as well.

fainpablo:
That was my first assumption, but asked a friend who has an account with them if he could show me the login flow browsing in Incognito mode and he was totally right. The Central Bank only requires users to be challenged for high risk transactions (authorization) rather than authentication, so for some reason they must have decided to dump this feature. I'll try to get a formal response from them because the documentation seems outdated.

fainpablo:
So... I had a conversation with the HSBC guys on Twitter, and they confirmed that 2FA through their mobile app is still available for accessing the Online Banking BUT as a replace of the password.

hsbc_ar-twitter

There are two available login flows:

1. Username and password. No token required. You'll be asked for token only when trying to execute a high risk transaction (authorization).
2. Username and token. No password required. You won't be asked to enter the token for any transaction, because you've used it for authentication.

hsbc-login-flow-second-screen

I would say the second option should not be considered 2FA because they're asking for the username, which can be basically anything (your name and last name, for example), and it's used for identification not authentication. Any thoughts?

Carlgo11 (approved):

phallobst (approved):

PR #4597, opened 2020-02-15
Add Mi Argentina (AR)

fainpablo:


_data/government.yml
```diff
@@ -55,6 +55,13 @@ websites:
       - phone
     doc: https://www.gov.uk/government/publications/genuine-hmrc-contact-and-recognising-phishing-emails/genuine-hmrc-contact-and-recognising-phishing-emails#hmrc-short-message-service-text-messages
 
+  - name: id.argentina.gob.ar
```
phallobst: I think "Mi Argentina" would be a better name.

phallobst:
```suggestion
  - name: Mi Argentina
```

phallobst (changes requested):

_data/government.yml
phallobst: Renaming would also mean to move the entry to the correct alphabetical position. It would be best to rename the icon accordingly as well.

phallobst (approved):

_data/government.yml
```diff
@@ -72,6 +72,12 @@ websites:
       - hardware
     doc: https://www.login.gov/help/creating-an-account/two-factor-authentication/
 
+  - name: Mi Argentina
+    url: https://id.argentina.gob.ar
```
Carlgo11: The url tag shouldn't link to a login page. Is there a main page or similar that you can use instead?

Carlgo11 (commented):
What's the deal with the DNI phone stuff? Is it like an [eID](https://en.wikipedia.org/wiki/Electronic_identification)?
If so, that could technically be an auth factor which would make it a [provider](https://twofactorauth.org/providers/).

_data/government.yml
fainpablo: The main page is https://mi.argentina.gob.ar/, but it redirects to https://id.argentina.gob.ar. As soon as you log in, you'll see your available options.

fainpablo: Regarding the DNI on your phone, the idea of the government is that you don't need to carry your physical ID but instead use the "picture" of your ID that's inside "Mi Argentina" app. Banks and government offices just started to accepting this "digital" ID. And this is one of the reasons why I've been telling them to support a 2FA mechanism.

If anyone manages to get my username and password, that's it. They can get my ID, my drivers license, my medical information, and some other sensitive stuff.

Not sure if that clarifies your question. Let me know!

phallobst: Maybe link to https://www.argentina.gob.ar instead? There is a prominent "Mi Argentina" button at the top.

fainpablo: Ha! That button is totally new. Makes sense, then. I'm pushing the updated URL.

Carlgo11 (approved):

PR #4598, opened 2020-02-15
Add AFIP (AR)

fainpablo:


phallobst:
This logo is a better match, I think:
afip

phallobst:
The [deepl](https://www.deepl.com/translator) translation of the first paragraph on the doc page is as follows:
> As of the first working day of January 2019, in place of the "Hardware-Token", the "Software-Token" must be used to access and use the computer services of this Agency that require a fiscal code with security level 4.

So if this is correct:
1. "hardware" 2FA is not supported anymore, but only "proprietary".
2. The exception text should rather say "2FA is required for services with security level 4."

Can you confirm that, or is either the translation or my understanding of it wrong?

fainpablo:
> This logo is a better match, I think:
> afip

I grabbed the original logo from one of their iOS apps, but the suggested one looks good. I'll push the new logo.

fainpablo:
The exception text must contain "Clave Fiscal" since this is the [name of the security level](https://www.afip.gob.ar/claveFiscal/). If you refer to 'security level 4' it could be possible that some people doesn't understand what it means. How about "2FA is required for users with Clave Fiscal (security level) 4" ?

fainpablo:
Regarding the hardware token it will be deprecated, but not yet. Users who has a hardware token can continue using it, as stated on this [memo](http://biblioteca.afip.gob.ar/dcp/REAG01004368_2018_12_19), which is translated below. Not sure what's the position of this site about methods that are still valid but will be deprecated anytime soon. I can keep an eye on this deprecation roadmap and update again once deprecated, if you're ok.

**Spanish**
Que sin perjuicio de lo establecido en el considerando precedente, quienes a la fecha de entrada en vigencia de la presente resolución general cuenten con un dispositivo “Hardware-Token” podrán continuar utilizándolo hasta tanto esta Administración Federal disponga su desafectación.

**English**
Without prejudice to the provisions of the preceding, those who, at the effective date of this general resolution, have a “Hardware-Token” device may continue to use it until this Federal Administration disposes of its disinfection.

phallobst:
Okay, we leave "hardware" in as long as that token is still supported.

Regarding "Clave Fiscal": Does this really refer to the _person_ who is using a service, or the _service_ itself? This is not really clear to me.

fainpablo:
Let's suppose you're a citizen trying to submit your taxes declaration. You need **Clave Fiscal** level 2 or above, which is associated to your tax ID. Now if you need to submit a customs declaration, you're gonna need **Clave Fiscal** level 4.

Depending on the service you're trying to use, the security level you'll need. But yes, it is associated to the person.

phallobst:
Okay, "2FA is required for users with Clave Fiscal (security level) 4" sounds good then.

fainpablo:
Update pushed. Thank you!

phallobst (approved):

_data/government.yml
```diff
@@ -1,4 +1,13 @@
 websites:
+  - name: AFIP
+    url: https://www.afip.gob.ar
+    img: afip.png
+    tfa:
+      - hardware
+      - proprietary
+    doc: https://serviciosweb.afip.gob.ar/genericos/GuiaDeTramites/VerGuia.aspx?tr=38
+    exception: "2FA is required for users with Clave Fiscal (security level) 4"
```
Carlgo11: >is required

But hardware and software tokens can still be used for security levels <4?

fainpablo: No, to request a hardware token or be able to use the software one you must be on level 4. Levels 1, 2, and 3 are not allowed to use 2FA.

Carlgo11:
```suggestion
    exception: "2FA is only available for users with security level (Clave Fiscal) 4"
```

Carlgo11 (approved):

PR #4604, opened 2020-02-17
Update 1Password

fainpablo:
Added 'u2f' to 1Password and an exception about hardware tokens.

### DIFF
--- a/_data/identity.yml
+++ b/_data/identity.yml
@@ -4,9 +4,10 @@ websites:
     img: 1password.png
     tfa:
       - totp
+      - u2f
       - hardware
     doc: https://support.1password.com/two-factor-authentication/
-    exception: "Teams using Duo with 1Password can't enable additional 2FA as it's already handled by Duo"
+    exception: "Teams using Duo with 1Password can't enable additional 2FA as it's already handled by Duo. Hardware tokens (not U2F) are only available through Duo."
 
   - name: bitwarden
     url: https://bitwarden.com


_data/identity.yml
```diff
@@ -4,9 +4,10 @@ websites:
     img: 1password.png
     tfa:
       - totp
+      - u2f
       - hardware
     doc: https://support.1password.com/two-factor-authentication/
-    exception: "Teams using Duo with 1Password can't enable additional 2FA as it's already handled by Duo"
+    exception: "Teams using Duo with 1Password can't enable additional 2FA as it's already handled by Duo. Hardware tokens (not U2F) are only available through Duo."
```
Jawshy, line 10: What are the non-U2F hardware tokens that Duo offers?

Jawshy (commented):
Does 1Password support SMS, phone, and Duo Push 2FA through Duo? If yes, then those [options](https://github.com/2factorauth/twofactorauth/wiki/FAQ-New-2FA-Supported-Site#i-want-to-add-a-site-using-two-factor-authentication-to-the-list) should also be included.

_data/identity.yml
fainpablo: From https://duo.com/product/multi-factor-authentication-mfa/authentication-methods/tokens-and-passcodes.

"Duo supports third-party hardware tokens, like Yubico’s YubiKeys, **or any OATH HOTP-compatible tokens**. Third-party hardware tokens can be imported into the system by an administrator.

**Duo’s D-100** tokens have an expected minimum battery lifetime of 2 years. They are only available in increments of 10. You can purchase tokens from the Duo Admin Panel or through your Account Team."

fainpablo: From https://duo.com/product/multi-factor-authentication-mfa/authentication-methods/tokens-and-passcodes.

"Duo supports third-party hardware tokens, like Yubico’s YubiKeys, **or any OATH HOTP-compatible tokens**. Third-party hardware tokens can be imported into the system by an administrator.

**Duo’s D-100** tokens have an expected minimum battery lifetime of 2 years. They are only available in increments of 10. You can purchase tokens from the Duo Admin Panel or through your Account Team."

fainpablo (commented):
> Does 1Password support SMS, phone, and Duo Push 2FA through Duo? If yes, then those [options](https://github.com/2factorauth/twofactorauth/wiki/FAQ-New-2FA-Supported-Site#i-want-to-add-a-site-using-two-factor-authentication-to-the-list) should also be included.

That will depend on each organization's configuration. If the org decides to allow tokens to be sent through SMS and/or phone, then yes. Default is yes. Should we include these methods as well and the exception?

Carlgo11 (approved):

Jawshy (approved):

PR #4605, opened 2020-02-17
Update Amazon and AWS

fainpablo:
Added 'phone' to Amazon.

amazon-2fa-options

### DIFF
--- a/_data/cloud.yml
+++ b/_data/cloud.yml
@@ -3,7 +3,6 @@ websites:
     url: https://aws.amazon.com
     img: aws.png
     tfa:
-      - sms
       - totp
       - hardware
       - u2f
--- a/_data/cloud.yml
+++ b/_data/cloud.yml
@@ -3,8 +3,6 @@ websites:
     url: https://aws.amazon.com
     img: aws.png
     tfa:
-      - sms
-      - phone
       - totp
       - hardware
       - u2f
--- a/_data/cloud.yml
+++ b/_data/cloud.yml
@@ -3,6 +3,8 @@ websites:
     url: https://aws.amazon.com
     img: aws.png
     tfa:
+      - sms
+      - phone
       - totp
       - hardware
       - u2f
--- a/_data/payments.yml
+++ b/_data/payments.yml
@@ -4,6 +4,7 @@ websites:
     img: amazonpay.png
     tfa:
       - sms
+      - phone
       - totp
     doc: https://www.amazon.com/gp/help/customer/display.html?nodeId=201962420
     exception: "SMS or phone call required to enable 2FA. Enabling on Amazon.com activates 2FA on other regional Amazon sites, such as UK and DE."
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -30,6 +30,7 @@ websites:
     img: amazon.png
     tfa:
       - sms
+      - phone
       - totp
     doc: https://www.amazon.com/gp/help/customer/display.html?nodeId=201596330
     exception: "SMS or phone call required to enable 2FA. Enabling on Amazon.com activates 2FA on other regional Amazon sites, such as UK and DE."


Carlgo11 (approved):

phallobst:
Does this apply only to the "Retail" entry of Amazon?

Carlgo11:
@phallobst no it applies to any oauth login for Amazon (including AWS)

fainpablo:
I'm updating Amazon Pay now. Phone does not apply to AWS nor AWS Workspaces.

I've double checked on my AWS account and confirmed they accept both U2F and other hardware tokens (one at a time), however they're dropping SMS support (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_mfa_enable_sms.html). 

Should we remove SMS from AWS?

phallobst:
> AWS will soon end support for SMS

_Valve "soon" or normal "soon"?_ (:copyright: @Carlgo11)

I'd say remove the "sms" method from the AWS entry.

Carlgo11:
AWS actually uses two different 2FA steps. The first one is for authenticating the Amazon account and the second is any AWS specific 2FA.

So if you setup 2FA for your amazon account and then at the account settings in AWS you’ll have to authenticate with two TOTP codes.

Thereby all Amazon 2FA options apply to AWS but not all AWS options apply to your Amazon account.

phallobst:
So they discontinue SMS as a 2FA method for AWS, probably because of the security implications, but are fine with using it for the retail market? What a considerate company!

fainpablo:
LOL Yes. Not to mention you can only setup one 2FA method at a time in AWS. If you add a security key and lose it, you'll in trouble.

So I've added "phone" and maintained "sms" from AWS, considering users need to go through Amazon 2FA challenge before getting into AWS.

_data/cloud.yml
```diff
@@ -4,6 +4,7 @@ websites:
     img: aws.png
     tfa:
       - sms
+      - phone
       - totp
```
jamcat22:
```suggestion
      - totp
```

jamcat22 (changes requested):
I'd suggest this be handled similar to Atlassian Cloud and Bitbucket, where Atlassian lists the methods for Atlassian's login and Bitbucket lists the methods for Bitbucket's login, even though Bitbucket [uses Atlassian Accounts for login](https://confluence.atlassian.com/bitbucket/upgrade-to-atlassian-account-829056056.html).

Bitbucket works similar to AWS, where you first log in to your Atlassian Account with 2FA if enabled, then provide 2FA for Bitbucket if enabled. While Atlassian supports SMS and TOTP, Bitbucket instead supports U2F and TOTP. When enabling Atlassian 2FA, you inherently protect the main access method to your Bitbucket account through your Atlassian web login. However, unless 2FA is enabled on the Bitbucket account itself, Atlassian 2FA doesn't protect basic authentication when using the API. Bitbucket 2FA must be enabled separately to force the use of [application-specific passwords](https://confluence.atlassian.com/bitbucket/app-passwords-828781300.html) to protect basic authentication logins on the API. It's also possible to enable Bitbucket 2FA without enabling Atlassian 2FA (for example, if you're in a Bitbucket team that [enforces 2FA](https://confluence.atlassian.com/bitbucket/control-access-to-your-team-s-private-content-915147488.html#Controlaccesstoyourteam'sprivatecontent-Requiringtwo-stepverification)).

Amazon and AWS security both work similarly. When enabling Amazon 2FA, you inherently protect the main access method to your AWS account, though your Amazon login. However, in the same vein as Bitbucket, you can leave basic authentication methods unprotected by Amazon 2FA. Also, AWS 2FA becomes much more important when using [Cross-Account Access](https://aws.amazon.com/blogs/security/how-do-i-protect-cross-account-access-using-mfa-2/) and [IAM Users](https://docs.aws.amazon.com/IAM/latest/UserGuide/id.html#id_iam-users), users that login to AWS directly, **without** going through Amazon's login system. For IAM Users, because authentication does not go through Amazon's system first, they are not able to use the methods available to main Amazon account logins and must use TOTP, hardware, or U2F with their AWS specific 2FA login.

fainpablo:
PR replaced by #4736, #4737, and #4738.

PR #4609, opened 2020-02-17
Add BBVA Argentina and update BBVA Compass

fainpablo:


Carlgo11 (changes requested):
I think this one works better for the small 32x32 size that we use.
logo

fainpablo:
You bet! I just pushed the updated icon.

Carlgo11 (approved):

phallobst:
Should "BBVA Compass" and "BBVA Argentina" use the same icon? Maybe we could substitute these two by a single "bbva.png".

fainpablo:
Good catch. Looks like they're using the same branding worldwide now. I've just updated the logo and the reference on both BBVA Argentina and Compass.

phallobst (approved):

PR #4611, opened 2020-02-17
Add Mercado Pago

fainpablo:


phallobst:
Is there no public documentation available?

fainpablo:
Same case as Mercado Libre. They only require you to add your phone number to your account and, if you have Authy installed, you'll start generating tokens. The only reference I found is this one, where they explain what to do in case you changed your phone number. https://www.mercadolibre.com.ar/ayuda/Cambié-el-numero-de-celular_2401

fainpablo:
Not sure this is the best option, but we may use the Authy's guide: https://authy.com/guides/mercadolibre/.

Carlgo11 (approved):
I think the Mercado Libre doc would just be confusing.
This is fine

phallobst (approved):

PR #4619, opened 2020-02-19
Add NIC Argentina

fainpablo:
For the purpose of this site, is it correct to say that 2FA is only available through AFIP login, and list the same methods available for AFIP?


phallobst:
Maybe I do not understand the process correctly, but it seems to me that nic.ar does not have a login procedure on its own, but is handled by https://tramitesadistancia.gob.ar instead. Should that website be listed here rather than nic.ar?

fainpablo:
That's right. And if you take a look at the site header, you'll see a couple options.  
Screen Shot 2020-03-01 at 01 42 04

We already have "Mi Argentina", "AFIP" and "Anses" (which is waiting for approval). I think we can get rid of the other two options as I heard they're going to probably get removed.

Should we list all those options here instead of https://nic.ar and https://tramitesadistancia.gob.ar? I'm wondering because it's the same case as with "Mercado Pago" vs "Mercado Libre" or "Gmail" vs "Google Drive". They don't handle the login process themself but they're still listed.

fainpablo:
That's right @phallobst. And if you take a look at the site header, you'll see a couple options.  
Screen Shot 2020-03-01 at 01 42 04

We already have "Mi Argentina", "AFIP" and "Anses" (which is waiting for approval). I think we can get rid of the other two options as I heard they're going to probably get removed.

Should we list all those options here instead of https://nic.ar and https://tramitesadistancia.gob.ar? I'm wondering because it's the same case as with "Mercado Pago" vs "Mercado Libre" or "Gmail" vs "Google Drive". They don't handle the login process themself but they're still listed.

Carlgo11:
@fainpablo the rule is that if a site doesn't provide it's own authentication but instead uses a third party (like 'Sign in with Facebook') then that doesn't get listed.
To answer your Gmail question: Since Gmail is a part of Google it gets a listing. Same with Skype (that uses MS accounts for authentication).

That said there are always edge cases like sites using Duo for their authentication still get listed here as long as they don't provide Duo & their own authentication options.

PR #4624, opened 2020-02-19
Add Jumbo Argentina

fainpablo:
Retail store. I believe it's a good idea they support 2FA since their allow you to save your payment methods for future purchases.


phallobst:
I think this icon looks better:
jumbo

fainpablo:
I like it! Thanks.

Carlgo11 (approved):

phallobst (approved):

PR #4629, opened 2020-02-20
Add Neolo

fainpablo:
While they don't specify the available methods in the documentation, they confirmed over email that any TOTP generator app is compatible. No hardware accepted yet.


phallobst:
This page actually references _Google Authenticator_, but is about the "cPanel" account. I have no idea if this might be a better doc link.
https://c.neolo.com/knowledgebase/166/iComo-configurar-la-autenticacion-de-dos-factores.html

fainpablo:
That's for accessing cPanel, as you mentioned. The one I'm talking about is the client area, where you can add or remove services to/from your account, manage payment methods, launch cPanel, etc.

I'm not quite sure whether we should add cPanel or not. I would say no because we may need to do the same for other hosting providers (like GoDaddy, Namecheap, etc.), and the URL of the cPanel is usually tied to the user domain or the server itself. Example: https://a2plcpnl0321.prod.iad2.secureserver.net:2083

What do you think?

phallobst:
I think we can be reasonably sure that Neolo uses TOTP. Leave the doc page as it is, then.

phallobst:
Maybe change the icon to this one? That is what they use as their favicon as well.
neolo

fainpablo:
Done! ;)

phallobst (approved):

Carlgo11 (approved):

PR #4630, opened 2020-02-21
Update Chime 2FA

markrainer:


phallobst:
Can you elaborate on what you mean with "Selectively used"?

markrainer:
A 3rd party is used based on the client connecting and determines the risk.

phallobst:
That sounds more like _authorization_ than _authentication_ to me.

Carlgo11 (commented):
@markrainer Is it like a cookie based 'remember me' feature? Many banks on our list use systems that only require users to enter their 2FA codes on new devices (via cookies)

If it's based on something like the IP-address reputation or machine learning then it would be authorization and not authentication according to our definition (and thus be listed as not accepting 2FA)

markrainer:
@Carlgo11 : Chime will 2FA any new device, however cookies aren't used. I'm curious as how banks set the "remember me" securely and allow you to bypass MFA? That doesn't sound very secure.

Carlgo11 (approved):

Carlgo11:
@markrainer encrypt the user-agent and IP, store it as a secure cookie and every time the user does an HTTP request, compare the encrypted data with the user-agent and IP of the client that initiated the request.

jamcat22 (approved):

PR #4631, opened 2020-02-22
Add Alexa cache & Clean up all test scripts

Carlgo11:
Amazon has a 10K API request per month limit on the Alexa API. We haven't reached that limit any month yet but its likely we will sometime in the future. By adding a cache for the `alexa.rb` script we can be reasonably sure that we won't reach that limit.

I also made the following changes:
* Removed `twitter.sh` as it can be merged with `twitter.rb`.
* Check git diff for Twitter handle changes before installing Twitter gem in CircleCI.
* Cleaned up `alexa.rb` and fixed Rubycop errors/warnings inside the file.
* Implemented @phallobst's suggestion on Alexa rank formatting.
* Fixed Rubocop warning in `verify.rb`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,9 +9,7 @@ jobs:
 
       - run:
           name: Check Twitter handles
-          command: |
-            gem i twitter --no-post-install-message --no-suggestions --minimal-deps --no-verbose -N -q
-            bash twitter.sh
+          command: bash twitter.sh
           working_directory: ./.tests
 
       - run:
@@ -19,10 +17,19 @@ jobs:
          command: bash facebook.sh
          working_directory: ./.tests
 
+      - restore_cache:
+          keys:
+            - awis-cache
+
       - run:
           name: Check Alexa ranking
           command: bash alexa.sh
           working_directory: ./.tests
+
+      - save_cache:
+          key: awis-cache
+          paths:
+            - "/tmp/alexa/*"
 
 workflows:
   version: 2
--- a/.tests/alexa.rb
+++ b/.tests/alexa.rb
@@ -20,7 +20,9 @@ if File.exist?("/tmp/alexa/#{site}.txt")
   # Prettify rank
   rank = rank_i.to_s.reverse.scan(/\d{1,3}/).join(',').reverse
 else
-  url = URI("https://awis.api.alexa.com/api?Action=UrlInfo&ResponseGroup=Rank&Output=json&Url=#{site}")
+  url = URI('https://awis.api.alexa.com/api' \
+  "?Action=UrlInfo&ResponseGroup=Rank&Output=json&Url=#{site}")
+
   https = Net::HTTP.new(url.host, url.port)
   https.use_ssl = true
 
--- a/.tests/alexa.rb
+++ b/.tests/alexa.rb
@@ -1,86 +1,57 @@
-#/usr/bin/ruby
-require "cgi"
-require "base64"
-require "openssl"
-require "uri"
-require "net/https"
-require "rexml/document"
-require "time"
+# frozen_string_literal: true
 
-access_key_id = ENV["alexa_access_key"]
-secret_access_key = ENV["alexa_access_secret"]
-#
-# Sample request to Alexa Web Information Service
-#
+require 'uri'
+require 'net/http'
+require 'json'
+require 'fileutils'
 
-if ARGV.length < 1
-  $stderr.puts "Usage: urlinfo.rb site"
+max_ranking = 200_000
+max_ranking_string = max_ranking.to_s.reverse.scan(/\d{1,3}/).join(' ').reverse
+
+if ARGV.empty?
+  warn 'Usage: alexa.rb url'
   exit(-1)
+end
+
+site = ARGV[0]
+
+if File.exist?("/tmp/alexa/#{site}.txt")
+  rank = File.open("/tmp/alexa/#{site}.txt", &:readline).delete(' ')
+  # Prettify rank
+  rank = rank.to_i.to_s.reverse.scan(/\d{1,3}/).join(' ').reverse
 else
-  site = ARGV[0]
+  url = URI("https://awis.api.alexa.com/api?Action=UrlInfo&ResponseGroup=Rank&Output=json&Url=#{site}")
+  https = Net::HTTP.new(url.host, url.port)
+  https.use_ssl = true
+
+  request = Net::HTTP::Get.new(url)
+  request['x-api-key'] = ENV['alexa_access_key']
+  request['Accept'] = 'application/json'
+  response = https.request(request)
+
+  unless response.code == '200'
+    raise("Request failed. Check URL & API key. (#{response.code})")
+  end
+
+  # Parse response
+  body = JSON.parse(response.body)
+  body = body['Awis']['Results']['Result']['Alexa']['TrafficData']
+
+  # Prettify body['Rank'] output
+  rank = body['Rank'].to_s.reverse.scan(/\d{1,3}/).join(' ').reverse
+
+  # Create cache file
+  FileUtils.mkdir_p '/tmp/alexa'
+  file = File.new("/tmp/alexa/#{site}.txt", 'w')
+  file.puts(rank)
+  file.close
 end
 
-SERVICE_HOST = "awis.amazonaws.com"
-SERVICE_ENDPOINT = "awis.us-west-1.amazonaws.com"
-SERVICE_PORT = 443
-SERVICE_URI = "/api"
-SERVICE_REGION = "us-west-1"
-SERVICE_NAME = "awis"
+# rubocop:disable Layout/LineLength
+raise("\e[31m#{site} has an Alexa ranking above #{max_ranking_string}. (Currently: #{rank})\e[0m") if max_ranking < rank.to_i
 
-def getSignatureKey(key, dateStamp, regionName, serviceName)
-  kDate    = OpenSSL::HMAC.digest('sha256', "AWS4" + key, dateStamp)
-  kRegion  = OpenSSL::HMAC.digest('sha256', kDate, regionName)
-  kService = OpenSSL::HMAC.digest('sha256', kRegion, serviceName)
-  kSigning = OpenSSL::HMAC.digest('sha256', kService, "aws4_request")
-  kSigning
-end
+raise("\e[31m#{site} doesn't have an Alexa rank. #{max_ranking_string} or less required.\e[0m") if rank.to_i.zero?
 
-# escape str to RFC 3986
-def escapeRFC3986(str)
-  return URI.escape(str,/[^A-Za-z0-9\-_.~]/)
-end
+# rubocop:enable Layout/LineLength
 
-action = "UrlInfo"
-responseGroup = "Rank,LinksInCount"
-timestamp = ( Time::now ).utc.strftime("%Y%m%dT%H%M%SZ")
-datestamp = ( Time::now ).utc.strftime("%Y%m%d")
-
-headers = {
-  "host"        => SERVICE_ENDPOINT,
-  "x-amz-date"  => timestamp
-}
-
-query = {
-  "Action"           => action,
-  "ResponseGroup"    => responseGroup,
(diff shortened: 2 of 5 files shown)


.tests/alexa.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-export LC_NUMERIC=en_US.UTF-8
+export LC_ALL=C.UTF-8
 format_number () {
   numfmt --grouping "$@"
 }
```
phallobst: I found a solution that works in `bash` and does not rely on `numfmt` or locales; it uses a hard-coded `,` and groups of three, which is probably what we want:
```
format_number () {
    [ "${#1}" -gt 3 ] && printf "$(format_number ${1:0:-3}),${1: -3:3}" || printf "$1"
}
```
(The space in `${1: -3:3}` is necessary, just in case you're wondering.)

Carlgo11: Alright, I just added that.

.circleci/config.yml
```diff
@@ -10,22 +10,31 @@ jobs:
       - run:
           name: Check Twitter handles
           command: |
-            gem i twitter --no-post-install-message --no-suggestions --minimal-deps --no-verbose -N -q
-            bash twitter.sh
+            git --no-pager diff origin/master..HEAD ../_data | grep "^+[[:space:]]" | grep twitter | cut -c15-
+            [[ $? -z ]] && { gem i twitter --no-post-install-message --no-suggestions --minimal-deps --no-verbose -N -q; ruby twitter.rb }
```
phallobst: Are you sure about `[[ $? -z ]]`? That doesn't look valid to me.

Carlgo11: Checks the exit code of the last executed command. Turns true if it’s 0.

phallobst: > % true; [[ $? -z ]] && echo "that is true"
> -bash: conditional binary operator expected
> -bash: syntax error near `-z'
> % echo $BASH_VERSION
> 4.2.46(1)-release

Do you mean `[[ $? -eq 0 ]] && ...`?

Carlgo11: Whoops!
Fixed :grin:

Carlgo11: Okay this is weird. CircleCI on my fork keeps failing at this line :thinking:

Carlgo11: My local code is:
```bash
[[ -z "$(git --no-pager diff origin/master..HEAD ../_data | grep "^+[[:space:]]" | grep twitter | cut -c15-)" ]] || { sudo gem i twitter --no-post-install-message --no-suggestions --minimal-deps --no-verbose -N -q; ruby twitter.rb; } && echo "No Twitter handles found. Skipping checks.";
```
Which echos "No Twitter handles found. Skipping checks." (returns 0) on the git diff returning null.
If the git diff returns anything it installs the gem and runs `./tests/Twitter.rb`.

phallobst: That's the short circuit evaluation of boolean expressions in bash:
```
% a="not empty"
% [[ -z "$a" ]] || echo "part 1" && echo "part 2"
part 1
part 2
% [[ -n "$a" ]] && echo "part 2" || echo "part 1"
part 2
```
```
% a=""
% [[ -z "$a" ]] || echo "part 1" && echo "part 2"
part 2
% [[ -n "$a" ]] && echo "part 2" || echo "part 1"
part 1
```

`&&` has a higher precedence than `||`.

Carlgo11: Hmm but it should still exit 0 regardless of which part is run.

.tests/twitter.rb
```diff
@@ -1,44 +1,40 @@
...
-		puts "Disregarding Twitter checks due to too many requests."
-		exit 0 # Soft fail if unable to access twitter api
-  elsif e.class == Twitter::Error::BadRequest
-    puts "Invalid authentication. Check Environment variables."
-    exit 1
-	else
-		puts e.backtrace
-		raise
-	end
+if handles.nil?
```
dsoegijono: `.split` returns an array so it might be an empty array, but not `nil`. If it's an empty array, `.each` should handle it.

.circleci/config.yml
```diff
@@ -10,22 +10,30 @@ jobs:
       - run:
           name: Check Twitter handles
           command: |
-            gem i twitter --no-post-install-message --no-suggestions --minimal-deps --no-verbose -N -q
-            bash twitter.sh
+            [[ -n "$(git --no-pager diff origin/master..HEAD ../_data | grep "^+[[:space:]]" | grep twitter | cut -c15-)" ]] && { gem i twitter --no-post-install-message --no-suggestions --minimal-deps --no-verbose -N -q; ruby twitter.rb; } || echo "No Twitter handles found. Skipping checks.";
```
Carlgo11: @phallobst this version built successfully. Does it look good to you?

.tests/twitter.rb
Carlgo11: Thanks for noticing! :+1:

.circleci/config.yml
phallobst: I'd merge the two ` grep`s into one:
```
... | grep "^+[[:space:]]*twitter: " | cut ...
```
That eliminates the risk that "twitter" appears somewhere else in the line without marking the handle itself.

Carlgo11: Alright I did that change as well as move from git diff to git log.
With git log the outdated stuff won't show up which means no more failing builds just because a fork is outdated 🎉

.tests/facebook.sh
```diff
@@ -1,22 +1,22 @@
...
-  handles="$(git --no-pager diff origin/master..HEAD ../_data | grep ^+[[:space:]] | grep facebook | cut -c16-)"
+check_facebook() {
+  handles="$(git log -p origin/master..HEAD ../_data | grep "^+[[:space:]]*facebook: " | cut -c16-)"
 
-  if [ -z "$handles" ]; then
+  [[ -z "$handles" ]] && {
     echo "No Facebook handles found."
     exit 0
-  fi
+  }
```
phallobst: Why, oh why? That if/then/fi was fine, now it's an eye-sore.

Carlgo11: This works the same way. Just less characters.

Carlgo11:
We need this update so that Circle CI builds won't fail on outdated PRs 🙁

phallobst:
The shell scripts look okay to me, but I'm not familiar with ruby, so... :neutral_face:

.tests/alexa.rb
```diff
@@ -1,86 +1,96 @@
...
-  kSigning
+SERVICE_URI = '/api'
+SERVICE_REGION = 'us-west-1'
+SERVICE_NAME = 'awis'
+
+def getsignaturekey(key, datestamp, regionname, servicename)
+  kdate = OpenSSL::HMAC.digest('sha256', 'AWS4' + key, datestamp)
+  kregion = OpenSSL::HMAC.digest('sha256', kdate, regionname)
+  kservice = OpenSSL::HMAC.digest('sha256', kregion, servicename)
+  OpenSSL::HMAC.digest('sha256', kservice, 'aws4_request')
```
dsoegijono: Nitpick: usually ruby uses snake_case (makes things easier to read I think)
```suggestion
def get_signature_key(key, date_stamp, region_name, service_name)
  kdate = OpenSSL::HMAC.digest('sha256', 'AWS4' + key, date_stamp)
  kregion = OpenSSL::HMAC.digest('sha256', kdate, region_name)
  kservice = OpenSSL::HMAC.digest('sha256', kregion, service_name)
  OpenSSL::HMAC.digest('sha256', kservice, 'aws4_request')
```

.tests/alexa.rb
```diff
@@ -1,86 +1,96 @@
...
+def escaperfc3986(str)
+  CGI.escape(str)
 end
 
-action = "UrlInfo"
-responseGroup = "Rank,LinksInCount"
-timestamp = ( Time::now ).utc.strftime("%Y%m%dT%H%M%SZ")
-datestamp = ( Time::now ).utc.strftime("%Y%m%d")
+action = 'UrlInfo'
+responsegroup = 'Rank,LinksInCount'
```
dsoegijono:
```suggestion
response_group = 'Rank,LinksInCount'
```
Suggestion edited by @jamcat22: response_roup -> response_group

.tests/alexa.rb
```diff
@@ -1,86 +1,96 @@
...
+  kdate = OpenSSL::HMAC.digest('sha256', 'AWS4' + key, datestamp)
+  kregion = OpenSSL::HMAC.digest('sha256', kdate, regionname)
+  kservice = OpenSSL::HMAC.digest('sha256', kregion, servicename)
+  OpenSSL::HMAC.digest('sha256', kservice, 'aws4_request')
 end
 
 # escape str to RFC 3986
-def escapeRFC3986(str)
-  return URI.escape(str,/[^A-Za-z0-9\-_.~]/)
+def escaperfc3986(str)
```
dsoegijono:
```suggestion
def escape_rfc3986(str)
```

.tests/alexa.rb
```diff
@@ -1,86 +1,96 @@
...
-url = "https://" + SERVICE_HOST + SERVICE_URI + "?" + query_str
+query_str = query.sort.map { |k, v| "#{k}=#{escaperfc3986(v.to_s)}" }.join('&')
+headers_str = headers.sort.map { |k, v| "#{k}:#{v}" }.join("\n") + "\n"
+headers_lst = headers.sort.map { |k, _v| k }.join(';')
+payload_hash = Digest::SHA256.hexdigest ''
+canonical_request = "GET\n#{SERVICE_URI}\n#{query_str}\n#{headers_str}\n#{headers_lst}\n#{payload_hash}"
+algorithm = 'AWS4-HMAC-SHA256'
+credential_scope = "#{datestamp}/#{SERVICE_REGION}/#{SERVICE_NAME}/aws4_request"
+string_to_sign = "#{algorithm}\n#{timestamp}\n#{credential_scope}\n#{(Digest::SHA256.hexdigest canonical_request)}"
+signing_key = getsignaturekey(secret_access_key, datestamp, SERVICE_REGION, SERVICE_NAME)
```
dsoegijono:
```suggestion
signing_key = get_signature_key(secret_access_key, datestamp, SERVICE_REGION, SERVICE_NAME)
```

.tests/alexa.rb
```diff
@@ -1,86 +1,96 @@
...
+req['Content-Type'] = 'application/xml'
+req['x-amz-date'] = timestamp
+req['Authorization'] = authorization_header
 
 res = Net::HTTP.start(uri.host, uri.port,
-  :use_ssl => uri.scheme == 'https') {|http|
+                      use_ssl: uri.scheme == 'https') do |http|
   http.request(req)
-}
+end
```
dsoegijono: Lines 72-83 is a bit strange cos it passes in the URI twice. I think it can be combined into:
```suggestion
uri = URI("https://#{SERVICE_HOST}#{SERVICE_URI}?#{query_str}")
request_headers = {
  'Accept' => 'application/xml',
  'Content-Type' => 'application/xml',
  'x-amz-date' => timestamp,
  'Authorization' => authorization_header
}

res = Net::HTTP.get_response(uri, request_headers)
```

.tests/alexa.rb
```diff
@@ -1,86 +1,104 @@
...
-req["Authorization"] = authorization_header
+req['Accept'] = 'application/xml'
+req['Content-Type'] = 'application/xml'
+req['x-amz-date'] = timestamp
+req['Authorization'] = authorization_header
 
 res = Net::HTTP.start(uri.host, uri.port,
-  :use_ssl => uri.scheme == 'https') {|http|
+                      use_ssl: uri.scheme == 'https') do |http|
   http.request(req)
```
dsoegijono: I think 72-82 can be removed

Carlgo11: Sorry @dsoegijono but I'm afraid I had to do some major changes as they've changed API 😅

dsoegijono (approved):
👍 on the Ruby code

jamcat22 (approved):

PR #4634, opened 2020-02-22
Add Authy

fainpablo:
Apparently there is no guide on how to setup 2FA on the developer dashboard (where you add and customize your 2FA protected applications). It makes sense since it is an automatic process when you sign up and add your mobile number (you get an SMS message with an invite to download Authy, if already downloaded you'll just see the tokens starts being generating).

https://www.twilio.com/docs/authy

Screen Shot 2020-02-22 at 19 02 36


Carlgo11:
We no longer list 2FA providers.

PR #4639, opened 2020-02-23
Update Figma

phallobst:


_data/other.yml
```diff
@@ -167,7 +167,8 @@ websites:
     tfa:
       - sms
       - totp
-    doc: https://help.figma.com/article/208--
+      - proprietary
```
Carlgo11: Are you sure it's not just TOTP?

phallobst: Duo provides push notifications with a simple tap verification as well, doesn't it?

Carlgo11: Yes but they might just mean that you can scan the RFC-6238 QR code in the Duo app and use it like Google Authenticatior.

Carlgo11 (approved):

jamcat22 (approved):

PR #4646, opened 2020-02-23
okcupid now supports 2FA via SMS

captain-caveman2k:


phallobst:
> Help keep your profile secure and add two factor authentication (2FA) to your account. This means each time your profile is logged into from a new device, we will send you a code via SMS/ text to your phone. Enter the code as well as your password to gain access to your account. Adding a phone number also means you never have to type in a password again. You can log in via SMS password each time.

What is it, then? Either it's 2FA (password + sms code), or it's sms code _instead of_ password (e.g. not 2FA).

captain-caveman2k:
> What is it, then? Either it's 2FA (password + sms code), or it's sms code _instead of_ password (e.g. not 2FA).

I believe it is the usual username/password log in with SMS verification. I don't represent okcupid in any way, I'm just an open source developer who occasionally uses their site ;-)

phallobst:
So if you log in, is it really possible to just use username + sms code, without the password?

Maybe the documentation is simply out-of-date (or was never correct in the first place).

captain-caveman2k:
> So if you log in, is it really possible to just use username + sms code, without the password?

As I mentioned above, I believe it is the usual username and password but with additional SMS verification ;-)

> Maybe the documentation is simply out-of-date (or was never correct in the first place).

I believe the documentation is correct - I've not enabled SMS verification on my account as a) I don't use their site that often and b) I'd rather use a authentication app (as I do for sites like Paypal, Hotmail, github, etc...).

_data/social.yml
```diff
@@ -177,9 +177,10 @@ websites:
 
   - name: OkCupid
     url: https://www.okcupid.com/
-    twitter: okcupid
-    facebook: okcupid
     img: okcupid.png
+    tfa:
+      - sms
+    doc: https://help.okcupid.com/article/202-sms-verification
```
Carlgo11: Another Help Scout page... 🙄 
```suggestion
    doc: https://help.okcupid.com/article/202--
```

Carlgo11 (changes requested):

_data/social.yml
captain-caveman2k: Thanks, I've updated as suggested...

Note: The link on [`https://help.okcupid.com`](https://help.okcupid.com) is as per my original commit.

Carlgo11 (approved):

phallobst (approved):

PR #4652, opened 2020-02-24
Update Cloudflare

phallobst:


jamcat22 (approved):

Carlgo11 (approved):
The old one redirects to the new one but since you already made the PR I don't see why not 🤷‍♂

PR #4671, opened 2020-03-05
Add Bancolombia (Colombia)

julian-alarcon:
Adding Bancolombia (Bank), Alexa Rank: [5.654](https://www.alexa.com/siteinfo/grupobancolombia.com)


phallobst:
According to the [FAQ](https://www.grupobancolombia.com/personas/productos-servicios/canales-servicio/sucursal-web/clave-dinamica/preguntas-frecuentes), this seems to be _authorization_ rather than _authentication_: (translated by deepl.com)
> Once the registration is done, you will be able to use the Dynamic Key immediately to make transactions such as transfers to non-registered accounts, payments of invoices, not own credit cards, among others.

Can you confirm that? If this is the case, we could of course add the site as _not_ supporting 2FA anyway.

julian-alarcon:
Got it, you are right, this is for _Authorization_, currently you don't need an OTP code to access your account, almost any transaction is locked and needs the code to do approve the transaction, but as you mention, it's different from _Authentication_. We should set it then as it needs to implement it. Let me do the change.

phallobst:
Could you also compress the icon by https://tinypng.com?

julian-alarcon:
Done @phallobst

phallobst (approved):

jamcat22 (approved):

PR #4672, opened 2020-03-06
Added GWR (UK train company)

jamiehowarth0:


phallobst:
Should this be named "Great Western Railway" instead? People searching for "GWR" would still find this entry due to the URL being a search keyword. (Renaming would also require moving it above "Greyhound".)

jamiehowarth0:
Updated & pushed!

phallobst:
Please also replace the icon with this one which is smaller in file size:
gwr

Carlgo11 (approved):

jamcat22 (approved):

PR #4584, opened 2020-02-11
Include tfa methods in search keywords

phallobst:
See #521, #3307, #3876

PR #4673, opened 2020-03-06
Adds filters for searching by TFA support

captn3m0:
Totally missed https://github.com/2factorauth/twofactorauth/pull/4584/files, but in case this looks like a better approach - filing this anyway.

My usecase was to filter by just one type of tfa (Hardware).

### DIFF
--- a/_includes/desktop-table.html
+++ b/_includes/desktop-table.html
@@ -30,7 +30,7 @@
         {% assign workonit_tweet = site.data.languages[website.lang].work_tweet %}
       {% endif %}
     {% endif %}
-    <tr class="desktop-tr">
+    <tr class="desktop-tr" data-tfa="{{website.tfa | join: " "}}">
       {% if website.status %}
         <td class="main progress">
           {% include row-title.html section=section_id website=website type='desktop' %}
--- a/index.html
+++ b/index.html
@@ -56,6 +56,13 @@ hash: SupportTwoFactorAuth
     <input type="search" id="jets-search" placeholder="Search websites" autocomplete="off" spellcheck="false"
            tabindex="0">
   </div>
+  <div class="sub header">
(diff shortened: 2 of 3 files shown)


phallobst:
These filters still use "software" and "hardware" as 2FA methods, but meanwhile these have been split into "u2f", "hardware", "totp" and "proprietary". (To be fair, the table headers are still "Hardware Token" and "Software Token"...)

Also, the only way I could make them work properly is to select the methods I wanted and then open a category. Flipping the checkboxes while the category is open does not have any effect for me.

Combining the checkboxes with a filter string yielded strange results as well.

(Tested with Opera, Chrome, Firefox, which all behaved the same.)

captn3m0:
This is WIP for now. I noticed #4584 after I started working on this, and thought filing earlier would be better than finishing but not getting merged. If you think this is worth having, I'll clean it up and fix the edge cases (flipping after checkbox is open / searching etc).

On headers: I considered keeping them tech specific, but I think matching the table headers makes more sense.

I also planned for a "Show All" button next to the 5 checkboxes that opens all categories, but maybe that can be a separate PR.

phallobst:
I see. GitHub allows you to create "Draft Pull Requests" for these cases; unfortunately, it is currently not possible to switch from a "normal" PR back to "draft". I will add "[DRAFT]" to the PR title now to reflect the current state. Once you are confident that your changes shall be reviewed for merging, you can just change it back.

You could also close this PR and open a new "draft" one if you want to.

phallobst:
I took the liberty to convert this PR to a draft, since GitHub now allows this.

captn3m0:
#4584 covers this much cleanly, and fulfills the usecase I had (find all sites across all categories with U2F supported)

Closing this.

PR #4682, opened 2020-03-09
Update Skrill

fainpablo:
Today I just noticed that Skrill has removed the ability to use both software and hardware tokens. When you login to your account, you'll be asked to setup and verify a mobile phone number and/or an email address. The 2FA section is no longer available and was replaced by the "Authentication methods" section, which has those two options.

Screen Shot 2020-03-09 at 11 30 04

As soon as you setup any of those methods, the previous 2FA method is completely disabled. The new login page looks as follows:

Screen Shot 2020-03-09 at 11 30 29

They confirmed this a couple weeks ago through their [Twitter account](https://twitter.com/skrill/status/1228706850868211715): "Hello! Unfortunately, the 2FA is no longer a valid additional authentication option in Skrill. You can learn more about it from here: http://www.skrill.com/en/support/#/path/1362150272"

Not a good move if you ask me.

I've added an exception because email authentication requires an additional PIN, not sure this is considered an exception or not.

### DIFF
--- a/_data/payments.yml
+++ b/_data/payments.yml
@@ -188,7 +188,8 @@ websites:
     tfa:
       - email
       - sms
-    doc: http://www.skrill.com/en/support/#/path/905761532
+    doc: https://www.skrill.com/en/support#/path/SECURITY/1362150272/
+    exception: "Email method requires an additional PIN"
 
   - name: Square
     url: https://squareup.com
--- a/_data/payments.yml
+++ b/_data/payments.yml
@@ -186,9 +186,9 @@ websites:
     url: https://skrill.com/
     img: skrill.png
     tfa:
-      - totp
-      - hardware
-    doc: https://www.skrill.com/en/support#/path/ACCOUNT/Two-Factor-Authentication/905761532
+      - email
+      - sms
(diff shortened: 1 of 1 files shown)


_data/payments.yml
```diff
@@ -186,9 +186,10 @@ websites:
     url: https://skrill.com/
     img: skrill.png
     tfa:
-      - totp
-      - hardware
-    doc: https://www.skrill.com/en/support#/path/ACCOUNT/Two-Factor-Authentication/905761532
+      - email
+      - sms
+    doc: https://www.skrill.com/en/support#/path/SECURITY/1362150272/
+    exception: "Email method requires an additional PIN"
```
Carlgo11, line 192: What does additional PIN mean?

fainpablo: When you setup 2FA, you must create a PIN that then will be used when trying to authenticate using Email. TOTP sent by Email + PIN is one of the options, TOTP sent by SMS (no PIN required) is the other.

Carlgo11 (approved):

jamcat22 (approved):

PR #4687, opened 2020-03-09
Update Name.com

fainpablo:
While not specified in the documentation, SMS is enabled as a backup method by default. You can opt-out anytime.
Screen Shot 2020-03-09 at 12 36 23

Screen Shot 2020-03-09 at 12 37 44


phallobst:
There is actually a documentation page for this: https://www.name.com/support/articles/115013791328

Maybe we should change the doc entry to the "My Account" knowledge base overview page: https://www.name.com/support/categories/200296798

fainpablo:
> https://www.name.com/support/categories/200296798

The first link you shared has a mention to Google Authenticator, maybe we can use this one?

"Note: This process is dependant upon you having a device already setup. If you have not setup Google Authenticator, please see this guide".

phallobst:
I'm confused: You mentioned the first link, but quoted the second...

I'd prefer https://www.name.com/support/categories/200296798 since it's probably more stable (being the overview page) and provides links to all relevant info at one go.

fainpablo:
It wasn't me. I quoted the first link, I swear 🙄

phallobst (approved):

Carlgo11:
@jamcat22 what's our stance on backup methods? Do we list those?

jamcat22 (changes requested):
@Carlgo11 we typically do not list backup methods, unless they're able to be used as primary methods. So in this case, we would not list SMS as a method, as software is the main method and is required for 2FA, while SMS is only used as a backup method. 2FA can't be turned on using only SMS.

The exception to this is when a service requires a backup method to be added. In that case, we add an exception text stating what other method is required, but unless the method can be used as a main method, it's not selected.

Here are a few examples:
**Name.com**
- Site supports software for 2FA.
- Site allows SMS as an optional backup, but not as the primary method.
- Since SMS can't be used as the main method, and SMS is not required as a backup method, SMS is **NOT** listed as a method and exception text is **NOT** added.

**Yandex**
- Site supports software for 2FA.
- Site allows SMS as a required backup, but not as the primary method.
- Since SMS can't be used as the main method, and SMS is required as a backup method, SMS is **NOT** listed as a method and exception text **IS** added stating that SMS is required to enable 2FA.

**WordPress.com**
- Site supports software for 2FA.
- Site allows SMS as a required backup, or as the primary method.
- Since SMS can be used as the main method, and SMS is required as a backup method, SMS **IS** listed as a method and exception text **IS** added stating that SMS is required to enable 2FA.

**Uber**
- Site supports software for 2FA.
- Site allows SMS as an optional backup, or as the primary method.
- Since SMS can be used as the main method, and SMS is not required as a backup method, SMS **IS** listed as a method and exception text **NOT** added.

_data/domains.yml
```diff
@@ -326,8 +326,9 @@ websites:
     img: namecom.png
     tfa:
       - totp
-    doc: https://www.name.com/services/two-step-verification
-
+      - sms
+    doc: https://www.name.com/support/categories/200296798
```
jamcat22:
```suggestion
    doc: https://www.name.com/support/categories/200296798
```

jamcat22 (approved):

PR #4705, opened 2020-03-28
Update Entertainment icons

Carlgo11:
* Resized Floatplane logo to fit inside a 50% border radius
* Resized Google Play logo to fit inside a 50% border radius
* Updated Hulu logo
* Updated IMDb logo
* Fixed transparency on smashcast logo
* Fix colors on Spotify logo
* Resized Twitch logo to fit inside a 50% border radius
* Resized YouTube logo to fit inside a 50% border radius


phallobst:
Instead of shrinking all the icons to fit inside the circle, wouldn't it be easier to enlarge the circle itself so that it is big enough to contain 32x32 pixel icons - or get rid of it entirely?

Carlgo11:
Adding a border around the image would look weird. Really weird.

Making them square would also look weird (albeit not as weird as the suggestion above).
It's also a web design standard to display these types of images in circles.

Carlgo11:
Actually, I might be able to implement some tag for specifying that an image shouldn't have a border.
In the meantime however, I think the best option is to merge these proposed images.

phallobst (approved):

jamcat22 (approved):

PR #4709, opened 2020-03-31
Update Cloudflare

fainpablo:
This just in! Cloudflare has added support for security keys.

https://support.cloudflare.com/hc/en-us/articles/200167906#6Gqe6f3nZtXSTpwyS2PBZ1

Screen Shot 2020-03-31 at 20 22 05

### DIFF
--- a/_data/security.yml
+++ b/_data/security.yml
@@ -60,7 +60,7 @@ websites:
     img: cloudflare.png
     tfa:
       - totp
-      - hardware
+      - u2f
     doc: https://support.cloudflare.com/hc/en-us/articles/200167906
 
   - name: Cobalt
--- a/_data/security.yml
+++ b/_data/security.yml
@@ -60,6 +60,7 @@ websites:
     img: cloudflare.png
     tfa:
       - totp
+      - hardware
     doc: https://support.cloudflare.com/hc/en-us/articles/200167906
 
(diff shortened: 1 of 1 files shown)


_data/security.yml
```diff
@@ -60,6 +60,7 @@ websites:
     img: cloudflare.png
     tfa:
       - totp
+      - hardware
```
Carlgo11:
```suggestion
      - u2f
```

Carlgo11 (changes requested):

_data/security.yml
fainpablo: @Carlgo11 Is there any rule for using u2f vs hardware? Let's take GitHub as example, where 'hardware' is listed instead of 'u2f'.

Carlgo11: If they support U2F keys, add the u2f tag. If they support something else, add the hardware tag.

We rely on user contributions so some sites might not be up to date.

fainpablo: Got it, thanks. I'll take a look at the sites with the "hardware" tag later to see if any of them needs to be updated.

jamcat22 (approved):

Carlgo11 (approved):

Carlgo11:
#4631 🙄

PR #4711, opened 2020-04-01
Add Airbnb exception

leggomyinfo:


phallobst:
Do they use a "trusted device" approach, e.g. by storing a cookie? If so, this probably does not justify an exception text, since this is a widely used policy. Also, the user may be able to enforce 2FA login by disallowing/deleting the cookie.

Carlgo11:
* Made a test account through a VPN connection in chrome private mode.
* Logged out and switched VPN to a different country.
* Opened new private window and signed in. No 2FA prompt.

Looks more like occasional 2SV.

Carlgo11 (approved):

kmpoppe (approved):

PR #4712, opened 2020-04-06
Add Zoom

jimmyvanau:
Add Zoom as a site + link to documentation (https://support.zoom.us/hc/en-us/articles/360038247071-Setting-up-and-using-two-factor-authentication)


_data/communication.yml
```diff
@@ -355,3 +355,10 @@ websites:
     twitter: Yammer
     facebook: Yammer
     img: yammer.png
+
+  - name: Zoom
+    url: https://www.zoom.us
+    img: zoom.png
+    tfa:
+      - totp
+    doc: https://support.zoom.us/hc/en-us/articles/360038247071-Setting-up-and-using-two-factor-authentication
```
jamcat22:
```suggestion
    doc: https://support.zoom.us/hc/en-us/articles/360038247071
    exception: "2FA is only available on paid plans and doesn't protect the desktop or mobile apps."
```

jamcat22 (changes requested):

jamcat22 (approved):
Thanks @jimmyvanau!

Carlgo11:
Is 2FA completely ignored when signing in through a desktop or phone app?

jimmyvanau:
Looks that way, 2FA only for web settings

Carlgo11:
@jamcat22 is that even 2FA then?

prkls:
> @jamcat22 is that even 2FA then?

Arguably, yes. The 2FA is to the web console where changes to the user's defaults are done. Also, if you're an admin or owner of the Zoom account, any changes to the environment, adding and deleting users, etc. can be 2FA protected.

jamcat22:
I'd say adding it with the exception noted is good. It at least adds some protection to admin settings and recordings. The app now integrating chat features makes it a larger target than when it just was used for joining meetings, but that doesn't mean 2FA should be disabled altogether.

Carlgo11 (approved):

PR #4713, opened 2020-04-06
Add itslearning

Carlgo11:
Yes it's all lowercase.
Yes the apostrophe is missing.

The creators of this educational software clearly don't understand basic grammar.

Nonetheless itslearning is a software suit meant for educators and students to communicate with each other. Ranked ~2K on Alexa and mainly used in Norway.

Fun fact: The ISO-3166 2 letter code for Norway is also false in YAML. I doubt that would ever cause problems 🤔 🇳🇴


phallobst:
For admin accounts on the German site, `totp` is actually available, according to https://itslearning.com/de/kunden/2-faktor-authentifizierung/
Do we want a separate entry for `itslearning (DE)` then?

Carlgo11:
@phallobst Finally got a response. Apparently it is available in every country that itslearning operates but only for sys admins. No documentation except for the German one that you found so I think it's better just to skip any doc link.

phallobst (approved):

jamcat22 (approved):

PR #4715, opened 2020-04-07
ecobee has now enabled 2FA (SMS)

LeeBrotherston:
ecobee has now enabled 2FA for all customers.


_data/iot.yml
```diff
@@ -14,6 +14,9 @@ websites:
     twitter: ecobee
     facebook: ecobee
```
phallobst:
```suggestion
```

_data/iot.yml
```diff
@@ -14,6 +14,9 @@ websites:
...
     img: ecobee.png
+    tfa:
+      - sms
+    doc: https://support.ecobee.com/hc/en-us/articles/360040957212-Two-Factor-Authentication-2FA-
```
phallobst:
```suggestion
    doc: https://support.ecobee.com/hc/en-us/articles/360040957212
```

phallobst (changes requested):

_data/iot.yml
LeeBrotherston: Those are both the correct accounts.

phallobst: Yes, but they are not used/needed if some tfa method is present (see the failed Jekyll check).

LeeBrotherston: Aaaah gotcha.... That makes sense.  Sorry... lack of coffee fail!

phallobst (approved):

jamcat22 (approved):

PR #4718, opened 2020-04-08
Add Bendigo Bank

jimmyvanau:


phallobst:
> In order to access some services in Bendigo e-banking, like updating your Contact Details or sending money using Pay Anyone, a Security Token is required.

This sounds more like _authorization_, not _authentication_. Do you need an "e-token" when logging into your account?

jimmyvanau:
This is the logon screen
image

And I've always used the token. Testing again, it seems to ignore that input for my account, as in I can leave it empty. That is interesting 🤔 Seems like it's not proper 2FA then.

phallobst:
Could it be that entering the security token enables a "super user" mode that allows you to change account details or use "Pay Anyone", whereas leaving it blank just gives you basic functions, like checking your balance?
Or do you need to enter a token every time that you want to authorize something?

phallobst:
@jimmyvanau Can you provide more information regarding my [comment](https://github.com/2factorauth/twofactorauth/pull/4718#issuecomment-611477101)?

jimmyvanau:
Hi @phallobst thanks for following up - forgot about this.

Just tested it. It looks like the token is required for extra/secure actions. I can login, check balance and transfer to known users with username + pw. If I attempt to pay a new person after logging in with username + pw only, then it prompts for the token.

Carlgo11:
So that would be closer to two step [authorization](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions) than authentication.

_data/banking.yml
```diff
@@ -222,6 +222,15 @@ websites:
     facebook: bbvacompass
     twitter: BBVACompass
 
+  - name: Bendigo Bank
+    url: https://www.bendigobank.com.au
+    img: bendigobank.png
+    tfa:
+      - proprietary
+    doc: https://www.bendigobank.com.au/help/security/online-security/e-tokens/
```
phallobst: Let's list it as non-supporting then.
```suggestion
    facebook: BendigoBank 
    twitter: bendigobank
```

phallobst (changes requested):

_data/banking.yml
```diff
@@ -234,6 +234,14 @@ websites:
     facebook: bbvacompass
     twitter: BBVACompass
 
+  - name: Bendigo Bank
+    url: https://www.bendigobank.com.au
+    img: bendigobank.png
+    facebook: BendigoBank
```
Carlgo11: Whoops, seems to be an extra space here 😅 
```suggestion
    facebook: BendigoBank
```

Carlgo11 (changes requested):

phallobst (approved):

Carlgo11 (approved):

PR #4732, opened 2020-04-23
Add Virgin Atlantic

fainpablo:


Carlgo11 (changes requested):
virgin

fainpablo:
I though that was the old logo, but looks like it's still being used. Thanks @Carlgo11 !

Carlgo11 (approved):

phallobst (approved):

PR #4744, opened 2020-04-30
Add Quizlet.com to Education category

vmm20:
Quizlet.com does not currently support 2 factor authentication.


phallobst (changes requested):
This is their favicon which is a bit smaller in file size:
quizlet

vmm20:
> This is their favicon which is a bit smaller in file size:
> quizlet

I've replaced the original image with your suggested image.

phallobst (approved):

Carlgo11 (approved):

PR #4752, opened 2020-05-03
OVH: Add documentation and U2F support

rbeuque74:
Adding the documentation link for setting up 2FA on your account.
Also indicating U2F support, as listed in the documentation (I have been testing it myself, works with a Yubikey blue security key)

### DIFF
--- a/_data/hosting.yml
+++ b/_data/hosting.yml
@@ -392,7 +392,8 @@ websites:
     tfa:
       - sms
       - totp
-      - hardware
+      - u2f
+    doc: https://docs.ovh.com/gb/en/customer/secure-account-with-2FA/
 
   - name: PlanetHoster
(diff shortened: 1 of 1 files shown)


_data/hosting.yml
```diff
@@ -393,6 +393,8 @@ websites:
       - sms
       - totp
       - hardware
```
Carlgo11:
```suggestion

```

Carlgo11 (changes requested):
Looks like you don’t support any other hardware implementation than U2F so the hardware tag can be removed.

_data/hosting.yml
rbeuque74: Applied

Carlgo11 (approved):

jamcat22 (approved):

PR #4769, opened 2020-05-16
Add Evergreentech VPS Hosting

NathanFeeser:


phallobst:

@Carlgo11 The _workflow_ check got the rank, but did not fail. Could you check this?

Carlgo11:
Interesting :thinking:
Raising an exception doesn't actually return an exit code of 1

Carlgo11:
@phallobst it should be fixed now. Thanks for noticing it! :+1:

PR #4770, opened 2020-05-17
Add Basic Attention Token

jonade:


phallobst (changes requested):
The "Cryptocurrencies" category is intended to list trading platforms. This website does not fit in there. Please move it to "Other".

phallobst (commented):
Should this be called "Brave Rewards" instead? I'm not quite sure.

jamcat22:
https://creators.brave.com/ appears to go to the same page.

If we wanted to rename the entry to Brave Rewards we'd probably want to change the link as well.

jonade:
Yeah, there are two URLs which can be used as an entry point, the second is mentioned by @jamcat22 . Both domains support TOTP/U2F, but ultimately they end up at the domain in the PR.

phallobst:
Okay, let's leave it as it is then.

phallobst (approved):

Carlgo11 (approved):

PR #4779, opened 2020-05-21
Add Plaid

nagey:
https://www.alexa.com/siteinfo/plaid.com

Rank is 33,177

image


phallobst (approved):
Not quite sure about the "Cloud Computing" category, but I have no better suggestion either.

nagey:
> Not quite sure about the "Cloud Computing" category, but I have no better suggestion either.

I chose cloud since finance seemed to be personal finance stuff. Cloud seemed to make sense since plaid is an api as a service, and I figure anything that’s “as a service” if you squint hard enough, you can call cloud. :shrug:

dsoegijono:
I agree that Cloud doesn't seem to be the right section. I'd say Finance makes more sense. Or maybe even Developer since Plaid is mostly for app integration?

ApCoder123:
Please rebase your branch to the current master.

PR #4800, opened 2020-06-10
Add SimpleBills

jmeridth:
- [x] add logo image file 32x32
- [x] add listing


phallobst:

PR #4801, opened 2020-06-12
Add region tag test

Carlgo11:
This PR uses the ISO 3166-2 list from [datahub](https://datahub.io/core/country-list) to validate all `regions` tags in our site listings as well as the regions in `regions.yml`.

To use less traffic to datahub I made a cache file (`/tmp/iso-3166.txt`) which the script will write to if non exists already. As country codes chagne very unfreqently I decided to to implement any automatic expiry on the cache.

Ran some tests for with/without cache and the results are:
```
no cache:
791ms

cache:
210ms
```

Not much of a difference but once again the main purpose of the cache is to generate the least amount of web traffic to datahub.


phallobst:
`.../data_json/data/.../data_json.json`!? Oh dear...

dsoegijono (approved):

jamcat22 (approved):

PR #4746, opened 2020-04-30
Use "sticky" position on <th> instead of <thead> to increase browser compatibility

phallobst:
Closes #4745

PR #4802, opened 2020-06-12
Minor styling fixes

dsoegijono:
### Update README
Need to `serve` before it's accessible

### Desktop table width
The column widths were set on the `<td>` for icons only, so during search if a section has only non-tfa listed, the width wasn't set. This ensures the table headers width is fixed.

**Before**
Screen Shot 2020-06-12 at 11 14 52 pm

**After**
Screen Shot 2020-06-12 at 11 13 59 pm

### Site name overlaps exception icon

**Before**
Screen Shot 2020-06-12 at 10 46 36 pm

**After**
Screen Shot 2020-06-12 at 10 46 44 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,9 @@ ruby ./_deployment/regions.rb
 
 # Building the site
 bundle exec jekyll build
+
+# Running the site locally
+bundle exec jekyll serve --watch
 ```
 
 The TwoFactorAuth website should now be accessible from `http://localhost:4000`.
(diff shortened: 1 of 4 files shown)


phallobst:
I just noticed that the sticky table headings only work in Firefox again, but not in Chrome (+Chromium-based browsers) - see also #4746.

Can you fix that as well? I was unsuccessful.

Carlgo11 (approved):
Love it! Is the package-lock.json needed?

dsoegijono:
@Carlgo11 Not necessary exactly. Do you manually remove it after `npm i babel-minify`? But since this is a requirement for the site to run, I think we should add `package.json` and `package-lock.json`

@phallobst I can try looking into it but maybe in a separate PR?

phallobst:
The column width are fixed by this. I'm okay with another PR for the "sticky" bug.

phallobst (approved):

PR #4805, opened 2020-06-13
Add missing contact information

phallobst:
"Tell them to support 2FA!" is not so easy without any means of communication...


Carlgo11 (approved):
Could always write a letter or start a petition 🤷‍♂️

dsoegijono (approved):

PR #4809, opened 2020-06-17
add nextdns

rajiv:


rajiv:
nextdns.io has an Alexa ranking above 200,000. (Currently: 239,588)

PR #4810, opened 2020-06-17
Update Heroku

sildur:
Change Heroku URL to id.heroku.com, as that is the URL used when logging in.


Carlgo11:
The URL tag is supposed to direct the user to the main page of a site/service and not a login page.
If you use our data for your own program to compare with other URLs you should truncate any subdomains and instead just use the domain name of the URL tag.

PR #4813, opened 2020-06-20
Add EN World

Zegnat:
Almost missed that EN World had the option for 2FA. Having it on the 2FA list here would have made 1Password tell me, so that is why I am adding it.

Wondering if this may have to go under Social instead of Gaming. It is gaming-related, kind of like Enjin is gaming related while not being a game, but it is also very much a place that exists to engage its “users to interact with each other”. (Quote from #4286 about moving Wikipedia to Social, also used to move ~~Wikia~~ Fandom into Social.)

---

Site: [EN World](https://www.enworld.org/)
Alexa Rank: [45,907](https://www.alexa.com/siteinfo/enworld.org)
Engine: XenForo


phallobst:
I think "Gaming" is okay for this site.

Is there public information about the TFA setup available, or can you supply screenshot of the setup process?

Zegnat:
> Is there public information about the TFA setup available, or can you supply screenshot of the setup process?

Not that I have been able to dig up. Another reason why I almost missed it. Couldn’t find it documented for the XenForo website engine either; other than people asking questions about it on their forums. The only thing they have documented is [how to disable 2FA as a site admin](https://xenforo.com/xf2-docs/manual/config/#enabletfa), which is useless for users.

As for the flow on EN Worlds, it is pretty straight forward:

1. In your account settings there is a _Password and security_ section that lets you enable or change 2FA, called “Two-step verification”:
   
   Screenshot: the password and security section on EN World showing a “Change” button for two-step verification.
   
2. When setting up the two-step verification you will be given the choice between using an app or email. I think it is possible to set-up both, I have not tried.
   
   Screenshot: settings presented for two-step verification after pressing the “Change” button.

Setting up with “an app” gives you the standard QR code to scan, or secret code to manually enter, as well as a form to fill in to validate you adding it to your preferred authenticator. All bog-standard. If you want I can provide screenshots of that as well, have left them out now as it would require me to re-setup my authenticator from scratch to walk through the process.

Cheers!

phallobst:
These screenshots are totally sufficient.

The icon could be replaced with a slightly smaller one (in regards of file size):
enworld

Zegnat:
> The icon could be replaced with a slightly smaller one (in regards of file size)

Hmm, I had already run mine through ImageOptim, but I guess there are always more wins to be had. Just swapped out the image with yours and repushed my commit.

phallobst (approved):

Carlgo11:
We don't list forums and that seems to be the only use that requires login, thus why I'm closing this PR.

Zegnat:
Gotcha @Carlgo11!

I was unable to find a good definitive answer on whether it belonged on the list or not, and added it just because it would’ve been helpful for me to have 1Password recognise it.

Would a PR be accepted to add Forums as an item to EXCLUSION? When I spotted [Scripting Helpers](https://scriptinghelpers.org/) on the list, I just kind of assumed community discussion boards would be acceptd.

PR #4814, opened 2020-06-21
Add Shadow

ddwag1:
I added Shadow.tech to the gaming section. They are a competitor to Geforce Now (cloud gaming service) and GFN is already in this category so I felt it was appropriate to put them here. They do not currently support 2FA. I've requested it from them recently and received a 'thanks for your feedback' email.


phallobst (changes requested):
Please replace the icon by this one which is much smaller in file size:
shadow

Carlgo11 (approved):
Looks good!

phallobst (approved):

PR #4816, opened 2020-06-24
Add hotjar

drubin:
Add hotjar. 

They aren't an Alexa top 200k but I assume this is because its a developer tool.


Carlgo11 (approved):
It is in the top 200K. It's even in the top 1K 😁

drubin:
The search function didn't show it up for some. Thanks for confirming it.

phallobst (approved):

PR #4822, opened 2020-06-28
Add Region codes to banking

Carlgo11:
@jamcat22 you said you wanted to review the PR, right? 😜

PR #4823, opened 2020-06-29
Add Canada, France & Italy

### LINKED ISSUE #4815: Italian version
Hi,
I would be useful to have an Italian version.

Carlgo11:
After #4822 the top regions will be: 
| Region | # of listings |
|--------|---------|
| US | 179|
|DE|39|
|GB|38|
|CA|23|
|SE|22|
|AR|21|
|AU|17|
|FR|12|
|IT|12|

I thought that if there's more than 10 specific listings for a region it would be worth adding said region, so that's what I'm doing right now 😛 

Also closes #4815


jamcat22 (approved):

_data/regions.yml
```diff
@@ -20,4 +29,4 @@
   name: United Kingdom
 
 - id: us
-  name: United States
+  name: United States
```
phallobst: Any reason for the newline being removed?

Carlgo11: IDE Magic

phallobst (approved):

PR #4834, opened 2020-07-03
Add 33mail

clarkeandrew:


phallobst:

clarkeandrew:
@phallobst really, Alexa rank? How is that a reasonable way to determine the popularity of a privacy tool? I guess it's your repo though. Thanks anyway for reviewing.

phallobst:
The [contributing guidelines](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md) specify the basic requirements a website has to meet in order to be included. If there was no restriction on the popularity, we would end up with millions of sites (which would be a maintenance nightmare). The Alexa rank is the metric we currently use.

PR #4835, opened 2020-07-05
Create api.rb

Carlgo11:
By using a ruby script to generate the API files we make the process much more understandable (see the mess that are the current `api/` files).
That in turn means that future changes are easier made and bugs can be found/fixed quicker.


Carlgo11:
Just need to add a trigger in the gh-pages workflow and this will be ready to go.

dsoegijono (approved):

jamcat22 (approved):

PR #4837, opened 2020-07-05
Rename Wikia to Fandom

rail01:
Wikia is not Wikia for a reeeeally long time now. They still don't support 2FA but we shouldn't use their old name. This PR renames it in `_data/other.yml`, adds new image `fandom.png` and removes old `wikia.png` logo.

Let's hope I did everything correctly :)


phallobst:
The name change was also adressed in PR #4290 - that one is still waiting for approval, probably because nobody agrees with the category move...

rail01:
Oh, I didn't know about thay. It's ok to close this PR then ^^ at least I practiced using git with cli :P

PR #4850, opened 2020-07-15
Add LocalCoinSwap P2P exchange to supported exchanges list

CodeForcer:
LocalCoinSwap proudly supports 2FA. Not only that, we actively enforce it for all our users to protect them when using the site. 2FA should be part of every cryptocurrency site.

Therefore, I submit this PR for inclusion into your excellent website. Thank you!


phallobst:

PR #4858, opened 2020-07-23
Use triple-dot notation for "git diff"

phallobst:
This should fix the problem with the check in #4285.


Carlgo11 (approved):
Finally got around to reviewing this locally. Sorry for the long wait! 😅

dsoegijono (approved):

PR #4859, opened 2020-07-23
Calculate correct number of dummy columns

phallobst:
The number depends on the number of categories. This change removes the need to manually adjust that file when adding/deleting a category.


phallobst:
Also, the category icons in the last row keep their position when expanded now.

dsoegijono (approved):

Carlgo11 (approved):

Carlgo11:
Awesome work Michael! 👏

PR #4865, opened 2020-07-30
add ACTIVE 24 hosting

TomasHala:


_data/hosting.yml
```diff
@@ -23,6 +23,20 @@ websites:
...
+      - totp
+    doc: https://faq.active24.com/cz/183996-Ov%C4%9B%C5%99en%C3%AD-p%C5%99ihl%C3%A1%C5%A1en%C3%AD---dvoufaktorov%C3%A1-autorizace-2FA?l=cs
+
+  - name: ACTIVE 24 CZ
+    url: https://www.active24.cz/
+    img: active24.png
+    tfa:
+      - totp
+    doc: https://faq.active24.com/cz/183996-Ov%C4%9B%C5%99en%C3%AD-p%C5%99ihl%C3%A1%C5%A1en%C3%AD---dvoufaktorov%C3%A1-autorizace-2FA?l=cs
+
```
phallobst: I don't think there should be multiple entries, but region specifiers instead.

The fact that there is no doc page in English yet is unfortunate. However, we can at least shorten the URL.

```suggestion
    doc: https://faq.active24.com/cz/183996-
    regions:
      - cz
      - gb
      - nl
      - sk
      - de
      - es

```

phallobst (changes requested):

phallobst (approved):
Once the doc is available in English, we can adjust the link. For now, this should be sufficient.

Carlgo11 (approved):

PR #4866, opened 2020-07-31
Update T-Mobile

andbenn:
Soft token for T-Mobile added

### DIFF
--- a/_data/utilities.yml
+++ b/_data/utilities.yml
@@ -434,6 +434,7 @@ websites:
     tfa:
       - sms
       - totp
+      - hardware
     doc: https://support.t-mobile.com/docs/DOC-2727
     regions:
       - us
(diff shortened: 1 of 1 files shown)


kmpoppe (approved):
Confirmed as per https://www.t-mobile.com/support/account/set-up-and-manage-your-t-mobile-id#heading4

_data/utilities.yml
```diff
@@ -433,6 +433,7 @@ websites:
     img: tmobile.png
     tfa:
       - sms
+      - totp
```
phallobst, line 436:
```suggestion
      - totp
      - hardware
```
Since "Fingerprint or FaceID" is available as well (i.e. biometric features), the `hardware` method should be included as well (which is what we currently map biometric 2FA to).

phallobst (changes requested):

phallobst (approved):

kmpoppe (approved):

PR #4870, opened 2020-08-06
Spain and portugal added.

havocesp:
Added Spain and Portugal id and name.


phallobst:
Currently, there are no websites attributed to Portugal, so there's no benefit in adding it. For Spain, there are only three - not much better either.

PR #4378, opened 2019-12-02
Add BeerAdvocate

phallobst:
Closes #4377

No proper documentation found yet, I sacrified myself for an account :smile:

PR #4871, opened 2020-08-06
Add beeradvocate.com

vihangm:
Ranks ~40k globally https://www.alexa.com/siteinfo/beeradvocate.com
They recently announced a data breach, so would be nice to include them
so folks can enable 2fa.

Wasn't sure what category to pick and if there's issues with the fact
that this site is alcohol related. Given that it's mostly focused around
the social aspect, sharing reviews, forum discussions etc, went with
social. Happy to move if need be.


phallobst:
Duplicate of #4378

PR #4877, opened 2020-08-10
Modify Intesa San Paolo

icovada:
Intesa San Paolo supports SMS and proprietary 2fa auth inside its own app (questionable security...)

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -735,7 +735,8 @@ websites:
     url: https://www.intesasanpaolo.com
     img: intesasanpaolo.png
     tfa:
-      - hardware
+      - sms
+      - proprietary
     doc: https://www.intesasanpaolo.com/it/common/footer/sicurezza.html
 
   - name: Jovia Financial


phallobst:
Are you sure the hardware token is not supported any more?

icovada:
Yes, to activate 2fa on the app it clearly says "insert your hardware token pin for the last time", hardware tokens are no longer supported (even if I wish they were. I mean, what's more secure, a hardware token that's not connected to anything and is inside my house, or a phone... the same one you're doing the transaction with?)

phallobst:
So you can't use the app and the hardware token at the same time - but maybe you can keep using the hardware token without switching to the app? Or is that switch enforced somehow? That would be rather surprising to me.

icovada:
Hardware token has been retired last year. Only app is available right now. I told them it's less secure but they just told me this is the way to go now, or SMS (but if you don't want/can't use the app every SMS you receive costs money)

App is also locked to one bank account, the "Exit" button does not actually log out, only closes the session. Luckily if you have 2fa enabled in the app and for some reason it does not work, it falls back to SMS for free

It's ridiculous and would be legitimate reason to switch banks, unfortunately in the end their service isn't _that_ bad

icovada:
Also unless you have a fingerprint scanner in your phone their "2fa" is just writing your PIN, which you already had to do before AND type in the hardware token code

So basically they actually removed the 2fa thing in their version of it.

So would this still count as 2fa? Now that I think about it... I think it doesn't

icovada:
Third update: just moved some money to my card and purposefully failed the fingerprint reading, it went through just putting in my PIN

Carlgo11:
@icovada sorry for the delay.
If you think their new system doesn't actually provide 2FA then I'd recommend you to change this PR to reflect that 😉 
Their Twitter handle seems to be `IntesaSP_Help` and Facebook page `intesasanpaologroup`.
Only found `info@intesasanpaolo.com` as a support email address but I might be wrong there.

Carlgo11:
Closing due to inactivity.

PR #4912, opened 2020-08-26
Add static timezone suffix

Carlgo11:
The build date is always displayed in UTC/ZULU and so adding the timezone as a suffix would clear up any date confusion.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -84,7 +84,7 @@
       </div>
       <div id="buildNumber" class="desktop-only" title="{{ site.github.build_revision }}">
         Build #: {{ site.github.build_revision | slice: 0,7 }}<br>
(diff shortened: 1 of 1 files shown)


phallobst:
I don't think "ZULU" is widely used in the _real_ world (i.e. outside the military). I'd prefer "UTC", if any.

jamcat22:
What if, similar to the build number, we display the unsliced build date on rollover? It would include the full date and time, as well as the time zone.

Carlgo11:
@phallobst It's the standard as ISO 8601 uses Z.

Carlgo11:
Actually, you're right that it's much easier to read for normal users. Let's make it UTC regardless of standard

phallobst (approved):

jamcat22 (approved):
Still think we should have an unsliced version on roll-over. 😉

PR #4930, opened 2020-09-06
Update Nuget

jamiehowarth0:


Carlgo11:
@jamiehowarth0 there's no need to create an issue if you're going to create a pull request anyway.

phallobst:
Should we add an exception text here since you need an external account to log in?

jamiehowarth0:
Maybe there should be a way to specify that a site relies on another site's ID scheme for authentication? (e.g. Papercall relies on multiple auth providers without having it's own)

Carlgo11 (approved):

Carlgo11:
Microsoft owned site uses Microsoft account to authenticate. Since it's the only way to authenticate and it isn't a third party I don't see any need to specify the auth system further.

phallobst (approved):

PR #4936, opened 2020-09-06
Add Cookiebot

hassankhan:
Add [Cookiebot](https://www.cookiebot.com/)


phallobst (changes requested):
This icon is much more suitable, I guess. It's straight from their favicon:
cookiebot

hassankhan:
Absolutely, I didn't even think to check their favicon. Thanks for pointing that out, will update the PR shortly.

phallobst (approved):

Carlgo11 (approved):

PR #4943, opened 2020-09-08
Performance boost for webp.sh

phallobst:
- read loop instead of "-exec bash ..."
- parameter substitution instead of "sed"

### DIFF
--- a/_deployment/webp.sh
+++ b/_deployment/webp.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 # converting PNG images
-find $1 -type f -and -iname "*.png" \
--exec bash -c '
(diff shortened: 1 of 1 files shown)


Carlgo11:
Are you sure there's a performance boost?
It looks to me like the loop would take a longer time to complete.

phallobst:
The performance boost is quite substantial, actually. The main load was the repeated start of "bash" and "sed" _per PNG file_ - and there are quite a lot.

On my system, the script went from 12s to less than 1s (when "cwebp" was _not_ called at all).

Carlgo11:
wow that's a huge difference! Great work Michael! 👏

Carlgo11 (approved):

jamcat22 (approved):

PR #4947, opened 2020-09-10
Add CuriosityStream

Carlgo11:


phallobst (changes requested):
They are using a round icon on their website, so this should be better.
curiositystream

Carlgo11:
@phallobst check their Apple icon 😝

All icons are rounded on the live site. The square icon just saves some bytes.

dsoegijono (approved):

PR #4941, opened 2020-09-07
Add AWS MFA key exception

hassankhan:
Fixes #4923

PR #4949, opened 2020-09-13
Add exception to AWS

axeleroy:
AWS currently does not support more than one 2FA method at once: you cannot set multiple virtual 2FA devices or U2F keys, or even set a virtual 2FA alongside a U2F key.

Source: https://forums.aws.amazon.com/thread.jspa?threadID=137055


phallobst:
Duplicate of #4941

PR #4959, opened 2020-09-16
Change Centrify "doc" URL

thomas-j-moffett:
Recently, we changed our URL for the page you are linking to. When we changed the URL, we put in a proper 301 redirect, but we'd like it if the URL on your site can be updated so it can go directly to the new URL. Thanks!

### DIFF
--- a/_data/identity.yml
+++ b/_data/identity.yml
@@ -36,7 +36,7 @@ websites:
       - proprietary
       - hardware
       - u2f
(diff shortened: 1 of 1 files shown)


phallobst (approved):

Carlgo11 (approved):

Carlgo11:
Thanks for taking the time to help us keep our documentation up to date! 😄

PR #4961, opened 2020-09-18
Refresh icon.webp if icon.png is newer

phallobst:


phallobst:
Since "*.webp" is present in ".gitignore", this patch is mainly useful for local installations.

dsoegijono (approved):

kmpoppe (approved):

PR #4967, opened 2020-09-22
Add Instagram exception

draco-malfoy:
Activating 2FA is only possible through mobile app

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -133,6 +133,7 @@ websites:
       - sms
       - totp
     doc: https://help.instagram.com/566810106808145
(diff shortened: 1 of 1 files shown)


jamcat22:
You can enable 2FA for Instagram in their [website settings](https://www.instagram.com/accounts/two_factor_authentication/) after logging in.
Instagram web view - 2FA

draco-malfoy:
> You can enable 2FA for Instagram in their [website settings](https://www.instagram.com/accounts/two_factor_authentication/) after logging in.
> Instagram web view - 2FA

My bad! I overlooked "Use Text Message"

PR #4968, opened 2020-09-22
Add WhatsApp exception

draco-malfoy:
activating 2FA only possible through mobile app


phallobst:
I wonder if this worth an exception text, since you need your phone to log in anyway, even for the web frontend.

Carlgo11:
Closing due to inactivity

PR #4971, opened 2020-09-23
Added EasyDMARC website

avag89:


avag89:
@Carlgo11 @phallobst EasyDMARC is in ~206,000 by Alexa rank, and it's only ~6000 above of 200,000 threshold. Is there a chance for merge? Thanks

phallobst:
If the site had a major social media following, we could probably ignore this little gap - but I don't quite see this coming :wink:

I suggest we wait a couple of days to see if the Alexa rank moves in the right direction. If you wish, you can convert this PR to "Draft" in the meantime.

avag89:
Thanks @phallobst for your reply. EasyDMARC is a security product, which is used by more than 10,000 organizations. As you mentioned, the gap of Alexa's rank is really small, and it's matter of just couple of weeks, that the rank will hit 200000. I respect your rules, but you know that the "200000" is just a random number, that can be easily adjusted a little bit. Why not just merge PR now and avoid back and forth movements and waste of your time :)

Carlgo11:
@avag89 the 200K rule is a rule that we've created to keep the lists more manageable.

If your site was closer to 200K then we would probably be able to merge if it's clear that the customer base is large.

At the moment the site is too far from hitting the 200K rank for us to merge it.

Feel free to reopen this pull request if/when your site reaches an Alexa ranking that's less than 200K.

PR #4972, opened 2020-09-25
Move 20i to Hosting

ghost:
Add 20i to hosting category.


phallobst (changes requested):
An icon with a transparent background is preferred; please replace it with this one:
20i

phallobst:
Oh, I didn't even notice that "20i" is already in the _Domains_ category. We don't currently include a website in more than one category, although there are plans to support that properly.

Maybe in this case it would be better to move the entry to _Hosting_ completely.

phallobst (approved):

Carlgo11 (approved):

PR #4976, opened 2020-09-27
Fix various minor errors (spelling, grammar, punctuation, formatting)

phallobst:


CONTRIBUTING.md
```diff
@@ -139,7 +139,7 @@ The fields `name:`, `url:`, `img:`, `tfa:` are required for all entries.
 If a site does provide TFA, it is strongly recommended that you add the `doc:`
 field where public documentation is available. Other fields should be included
 if the website supports them. Any services that are not supported can be excluded.
-Sites supporting TFA should not have a `twitter`, `facebook` or `email_address` field.
+Sites supporting TFA should not have a `twitter:`, `facebook:` or `email_address:` field.
```
kmpoppe:
```suggestion
Sites supporting TFA must not have a `twitter:`, `facebook:` or `email_address:` field.
```
As the checks fail completely (no warning, error) when 2fa is enabled and one these is present I think the language should be a bit more direct.

kmpoppe (changes requested):
See inline comment

kmpoppe (approved):

Carlgo11:
I'm unsure if adding a ":" to each YAML key listing is the right thing to do. Technically we're referring to the key values and not the `key:` part of the string.
Just like when talking about Java variables you might refer to a variable named "a" and when doing that you wouldn't say "Public String a;" but instead just "a" or "string a".

phallobst:
The YAML field were not referenced consistently; sometimes the colon was used, sometimes not. I just wanted to clean that up. If the colon should be dropped, so be it.

Carlgo11:
Yeah we've kind of added it at random 😅
I think the best thing would be to drop it completely and instead just refer to the key value.

Carlgo11 (approved):
Thanks 😄

PR #4979, opened 2020-09-30
Improved Twitter/Facebook/Alexa checks

phallobst:
- ignore (_edit_: leading or trailing) whitespace in yaml values
- no early exit
- use sed to extract Facebook handle in one go

### DIFF
--- a/_deployment/tests/alexa.sh
+++ b/_deployment/tests/alexa.sh
@@ -9,7 +9,7 @@ check_rank () {
     exit 0
   fi
 
-  res=0
+  status=0
   # Loop through all URLs
   for url in $urls; do
 
@@ -21,12 +21,12 @@ check_rank () {
     $cmd
 
     if [ $? -ne 0 ]; then
-      res=1
+      status=1
     fi
 
   done
 
-  return $res
+  return $status
 }
 
 check_rank
--- a/_deployment/tests/alexa.sh
+++ b/_deployment/tests/alexa.sh
@@ -9,8 +9,9 @@ check_rank () {
     exit 0
   fi
 
+  res=0
   # Loop through all URLs
-  echo "${urls}" | while IFS= read -r url; do
+  for url in $urls; do
 
     # Get the domain from the URL
     domain="$(echo ${url} | cut -d'/' -f3)"
@@ -19,12 +20,13 @@ check_rank () {
     cmd="ruby alexa.rb ${domain}"
     $cmd
 
-    # Get exit code from the ruby script
-    status=$?
+    if [ $? -ne 0 ]; then
+      res=1
+    fi
 
-    # Echo ruby script output & exit with its status code
-    return $status
   done
+
+  return $res
 }
 
 check_rank
--- a/_deployment/tests/alexa.sh
+++ b/_deployment/tests/alexa.sh
@@ -17,12 +17,7 @@ check_rank () {
     domain="$(echo ${url} | cut -d'/' -f3)"
 
     # Get Alexa rank for the domain
-    cmd="ruby alexa.rb ${domain}"
-    $cmd
-
-    if [ $? -ne 0 ]; then
-      status=1
-    fi
+    ruby alexa.rb ${domain} || status=1
 
   done
 
--- a/_deployment/tests/twitter.sh
+++ b/_deployment/tests/twitter.sh
@@ -4,26 +4,29 @@
 check_twitter () {
   handles="$(git diff origin/master...HEAD ../../_data | grep "^+[[:space:]]*twitter:" | cut -c15-)"
 
-  if [ -z "$handles" ]; then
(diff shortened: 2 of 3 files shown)


Carlgo11:
Uhm I see the 'error-on-whitespace' as a feature rather than a bug

phallobst:
Why should the check fail on whitespace? YAML allows whitespace of any size between (or after) tokens. Since Facebook and Twitter handles can never contain whitespace, we don't lose any functionality by ignoring that, I suppose.

kmpoppe:
I'd lean to letting the check in as a safety measure when someone entered the handles wrongly (i.e. with a space rather than an underscore) when doing the changes in the yaml.

phallobst:
This change only concerns whitespace _before_ or _after_ the handles. If the handle itself is incorrect (e.g. space instead of underscore), the check will still fail.

kmpoppe:
Okay, that I'm ok with.

_deployment/tests/alexa.sh
```diff
@@ -19,12 +20,13 @@ check_rank () {
     cmd="ruby alexa.rb ${domain}"
     $cmd
 
-    # Get exit code from the ruby script
-    status=$?
+    if [ $? -ne 0 ]; then
+      res=1
+    fi
```
Carlgo11: Shouldn't this work? (And removing `$cmd` on L21)
```suggestion
   $cmd || { res=1}
```

_deployment/tests/twitter.sh
```diff
@@ -4,26 +4,29 @@
...
     echo "No Twitter handles found."
     exit 0
   fi
 
   gem i twitter --no-post-install-message --no-suggestions --minimal-deps --no-verbose -N -q --silent
 
+  res=0
   # Loop through all Twitter handles
-  echo "${handles}" | while IFS= read -r handle; do
+  for handle in $handles; do
```
Carlgo11, line 16: Is this line change needed?

_deployment/tests/twitter.sh
```diff
@@ -4,26 +4,29 @@
 check_twitter () {
   handles="$(git diff origin/master...HEAD ../../_data | grep "^+[[:space:]]*twitter:" | cut -c15-)"
 
-  if [ -z "$handles" ]; then
+  if [[ ! "$handles" =~ [^[:space:]] ]]; then
```
Carlgo11, line 7: While having a space at the end of the value of the `twitter` tag doesn't break anything, it also doesn't improve anything.
As I tried to weigh up the pros and cons I can't really find any pros for it, and a con would be that it adds unnecessary data to the file (yes, it's a bit extreme but I'm just laying everything on the table here).

_deployment/tests/alexa.sh
```diff
@@ -9,8 +9,9 @@ check_rank () {
     exit 0
   fi
 
+  res=0
```
Carlgo11: I think something like `status`, `code` or `status_code` would better describe what the variable is for.
```suggestion
  status=0
```
```suggestion
  code=0
```
```suggestion
  status_code=0
```

_deployment/tests/twitter.sh
phallobst: Yes, it is - the pipe symbol "|" starts a new sub-shell, and any assignments to variables in there do _not_ propagate to the outer shell.

_deployment/tests/alexa.sh
phallobst: Adding a number to a variable would be something like "res=$(($res+1))". That would also mean that the return code is the number of errors here, which is rather unusual for shell scripts. Maybe the worst result should be returned instead.

_deployment/tests/alexa.sh
phallobst: "status" it shall be!

_deployment/tests/alexa.sh
Carlgo11: Updated

phallobst: And the `cmd` variable is not needed either...

Carlgo11: Are you sure? $domain might be the latest run command otherwise 🤔

phallobst: How would `$domain` be executed? It's the argument for `alexa.rb`.

_deployment/tests/twitter.sh
phallobst: Currently, an extra space before or after the Twitter/Facebook handle _does_ make the check fail, which is one of the reasons for this PR. You can't run the Twitter check beforehand in a local installation, since the API key is not public. Also, the extra spaces are not visible most of the time (some editors display them differently), and they are perfectly valid YAML syntax. So even though additional spaces are rather useless, they should not lead to a failed check.

Carlgo11:
The move to JSON makes this issue obsolete.

PR #4980, opened 2020-10-01
Add G2A

rikogercke:


phallobst (changes requested):
Since this is not a gaming platform, it fits better to the _Retail_ category, I suppose.

The doc URL is quite a mouthful, but I can't see how to shorten it either. :man_shrugging:

rikogercke:
I was not sure about it because i spotted CDKeys in the gaming category. Should I move G2A anyway?

Carlgo11:
Duplicate of #3894

PR #4998, opened 2020-10-07
Update Tesla

bricesanchez:


Carlgo11:
The documentation talks about security keys. Is it still totp or do they also support U2F?

phallobst:
The "security key" mentioned there is most likely the _secret_ for the TOTP application - it's just worded in an unfortunate way.

Carlgo11:
ah, yeah switching to another language made it more obvious what they meant. Thanks @phallobst :smile:

Carlgo11 (approved):

phallobst (approved):

PR #5004, opened 2020-10-08
Add Conrad

jokay:
Adds [Conrad](https://www.conrad.ch/).

rank-conrad


phallobst:
Since Conrad is predominantly a German company, I guess we should use https://conrad.de instead, maybe with the `regions` as well - although there are quite a number of them according to https://www.conrad.de/de/ueber-conrad/unternehmen.html.

phallobst:
@Carlgo11 The Facebook check is currently broken, since the header does not include `location` anymore. :frowning_face:

Carlgo11:
@phallobst Nooooo! 😭
I'll try and fix it tonight

_data/retail.yml
```diff
@@ -110,6 +110,31 @@ websites:
...
+      - at
+      - pl
+      - ro
+      - se
+      - ch
+      - sk
+      - si
+      - cz
+      - hu
+      - uk
```
phallobst:
```suggestion
      - gb
```

phallobst (changes requested):

phallobst:
Facebook is messing around currently, I guess. Right now the headers include `Location` (capital L), but a misspelled handle is not corrected.
_Edit_: And now the check is working again?! I have no idea what is happening...

jokay:
> Facebook is messing around currently, I guess. Right now the headers include `Location` (capital L), but a misspelled handle is not corrected.
> _Edit_: And now the check is working again?! I have no idea what is happening...

Conrad Switzerland has `conrad.schweiz` (with `.`) and Conrad Germany `conradelectronic` as Facebook account, this is why it may work again :wink:

Carlgo11 (approved):

phallobst (approved):

PR #5026, opened 2020-10-11
Allow SVG icons

Carlgo11:
And I might have cleaned up a few methods as well :sweat_smile:

### DIFF
--- a/_deployment/tests/verify.rb
+++ b/_deployment/tests/verify.rb
@@ -14,7 +14,8 @@ require 'diffy'
 }.freeze
 
 # Image max size (in bytes)
-@img_max_size = 2500
+@img_max_size_png = 2500
+@img_max_size_svg = 10_240
 
 # Image dimensions
 @img_dimensions = [32, 32]
@@ -40,7 +41,7 @@ def test_img(img, name, imgs)
 
   # Check image dimensions
   error("#{name}: #{img} is not #{@img_dimensions.join('x')} pixels.", img)\
-    unless FastImage.size(img) == @img_dimensions
+    unless img.match?(/svg$/) || FastImage.size(img) == @img_dimensions
 
   test_img_file(img, name)
 end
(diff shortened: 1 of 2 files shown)


Carlgo11:
All done on my end. Surprisingly it looks like the mobile view doesn't use webms at all :thinking: 
Probably something to look further into

phallobst:
The `CONTRIBUTING.md` file should be updated accordingly as well.

Edit: Maybe point to https://vecta.io/nano or a similar service.

phallobst:
Also, is it still useful to have SVG icons cropped by the circle? For bitmap icons, it was easily possible to shrink them a bit (i.e. have a small transparent border), so no vital parts were cut off. This might not be so easy with SVG. (To be honest, I was never a fan of that cropping anyway...)

Carlgo11:
> Also, is it still useful to have SVG icons cropped by the circle? For bitmap icons, it was easily possible to shrink them a bit (i.e. have a small transparent border), so no vital parts were cut off. This might not be so easy with SVG. (To be honest, I was never a fan of that cropping anyway...)

Carlgo11:
I think both the collaborator.md and possible allowance of square icons might be better fit for a new separate issue & PR.

phallobst:
I just realized that you can set `no_img_border: true` for an entry to get rid of the cropping. You live and learn.

Carlgo11:
@phallobst is that something I implemented? 👀 
Because it sounds familiar.

phallobst:
Yep, it was part of https://github.com/2factorauth/twofactorauth/commit/84549c8b

phallobst (dismissed):

phallobst:
The image dimension check (32x32) does not apply to vector images, so it should be skipped for these.

Also, the file size limit of 2.5K will not be met for many of the more detailed icons. For example, the left part of the "Simple" bank logo alone is almost 10K even after passing it through https://vecta.io/nano...
[Official logo](https://www.simple.com/assets/images/simple-logo.svg), [Cropped and optimized](https://raw.githubusercontent.com/phallobst/resterampe/4d8ef4544d77ad1e0a42e9cde2ca53b91c197fba/simple.svg)

phallobst:
@Carlgo11 The latest commit adds more SVG handling in the verify script. This is what I used in #5057 as well.

phallobst (approved):

dsoegijono (approved):

PR #4134, opened 2019-10-17
Update CrashPlan

AnalogCyan:
Fixes #4064

PR #5040, opened 2020-10-17
Update CrashPlan

christantoan:


phallobst:
Could you have a look at PR #4134? That one was closed due to inactivity, but there still was an [unanswered question](https://github.com/2factorauth/twofactorauth/pull/4134#discussion_r341888195). I would imagine this could still be relevant.

christantoan:
I'm really sorry. I currently don't have time for that. I just noticed that the entry doesn't match with my experience and thought a quick fix is all it needed.

PR #5050, opened 2020-10-20
Update WebP & SVG implementation

Carlgo11:
__Requires #5026__

This PR implements the following:
* WebP is finally shown on mobile devices
* Page doesn't link to non-existing WebP images when the source is in the SVG-format
* SVGs are square by default


phallobst:
Given that SVG (hopefully) will be the default format in the future, and very few icons actually _need_ the border, should the "no-border" style be the default one perhaps?

Carlgo11:
@phallobst I think that's a question for a different ticket.

phallobst (approved):

phallobst:
I'd like to have this approved in order to "Release the SVG Kraken"... :wink:

@2factorauth/collaborators Anyone?

PR #5055, opened 2020-10-22
Add University of Nebraska

cbohn4:
This change adds the University of Nebraska and the subsequent universities such as UNL, UNO, UNK, and UNMC.


phallobst (dismissed):
Please replace the icons with these versions:
* uneb.png: uneb

* unk.png: unk
* unl.png: unl
* unmc.png: unmc
* unomaha.png: unomaha

Carlgo11:
@cbohn4 how many of the sites share the same login system(s)?

cbohn4:
That is a good question and it is a little weird in this case.  There are shared credentials through the Nebraska University system using a NU ID. There are also more campus-specific logins that use usernames that are not shared across the campuses for resources like Canvas and Email.  I can see an argument for limiting it to just "University of Nebraska" and an argument for it being individual universities. 
The shared campus identity and 2FA point mixed with campus-specific logins do make it interesting. DUO is shared across campuses and using the system-wide ID number works which is certainly the side for keeping it to only "University of Nebraska." 
I believe either solution would work. 
Another solution that may be worth looking at would be "University of Nebraska" with the docs page going to a note page, similar to Chase Banking, listing the different campuses as under the "University of Nebraska".

Carlgo11 (changes requested):
@cbohn4 I think it would be best to just have one entry then.
I suggest leaving the `University of Nebraska` entry and remove all the other entries (along with the linked images)

Carlgo11:
Closing due to inactivity

PR #5051, opened 2020-10-20
Merge BB&T and SunTrust to Truist

phallobst:

PR #5057, opened 2020-10-22
SVG icons for Banking

phallobst:
The first (and largest) batch of SVG icons has arrived! :partying_face: 

Reviewing this is probably easier if you just clone this branch and scroll through "Banking" in a local installation.

This is still a draft due to the following:

1. Some SVG icons were far too large. I set an arbitrary limit of 10K for the check, so these PNGs are still in place:
  - swedbank.png
  (_Edit_: Only one left!)
  If someone has small SVG versions, message me please.
2. Two additional PNGs are there, but these entries are bound to be removed:
  - BoFI: #4822
  - BB&T: #5051
3. Changes from #5050 have been used already. Once that one is merged, this PR will be updated accordingly.

Try it out if you wish, preferably with a large zoom. It's _very_ satisfying to look at, I must say.

(see #5061)


Carlgo11:
@phallobst I've tried to find a good SVG icon of Swedbank before but the only one I could find was around 30Kib :|
I think it's better to just use a good PNG for it.

PR #5058, opened 2020-10-22
add Bitrix24

trockenasche:


phallobst (changes requested):
The icon does not seem to be in use on the website itself. Maybe the 24 or 24+clock would be better.

trockenasche:
It's the mobile app icon.
But I can have a look for a difference icon, if you want to.

trockenasche:
I added a probably better logo.

phallobst (approved):

Carlgo11 (approved):

PR #5060, opened 2020-10-23
add Bitwala

trockenasche:
Information about the regions: https://www.bitwala.com/supported-nationalities/

> Bitwala is available for residents of the European Economic Area (incl. Switzerland) and accepts identification documents from 70+ nationalities


trockenasche:
I don't understand the error `Bitwala: 'false': not a string.` ?

Never mind, YAML identifies a 'no' as a 'false'.

phallobst (approved):
That's quite a list of regions... As soon as there are other entries referencing all countries in a union, we should think about region aliases "eea", "eu", "efta" or something. But for now, listing all countries will do, I guess.

dsoegijono (approved):

PR #5068, opened 2020-10-31
CSS fixes

phallobst:
* Logo scales properly (to a max size)
* Search bar is not overlapping the text anymore

### DIFF
--- a/css/main.scss
+++ b/css/main.scss
@@ -79,8 +79,7 @@ thead {
   padding: 0;
 
   #main {
-    margin: 5vh auto;
-    height: 40vh;
+    margin: 1em auto 1em auto;
     text-align: center;
 
     h2 {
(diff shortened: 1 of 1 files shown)


dsoegijono (approved):
Before & after screenshots would be great for UI changes I think.

phallobst:
Here are the screenshots for comparison, with 200% and 25% zoom respectively.

Before, zoomed in: Logo is not scaled, search bar overlaps the text
before-zoom-in
Before, zoomed out: Logo is not scaled, large gaps around search bar
before-zoom-out

After, zoomed in:
after-zoom-in
After, zoomed out:
after-zoom-out

Carlgo11 (approved):

Carlgo11:
uhm...

phallobst:
:astonished: Was the CSS rebuilt from an old state somehow? It _did_ work before. On a local installation, I don't see the problem.

The `#logo` should _not_ have `height: 30vh` (but `20em`).

PR #5075, opened 2020-11-09
Convert APIv1 to Ruby source

Carlgo11:
API v1 is a mess and recent changes in the YAML scheme have added tags that shouldn't be present in the JSON file.
This PR makes sure that only the right elements are present in API v1. I also rewrote it in Ruby is it's much easier to modify and read.

As I did this I found the `api/` directory (along with subdirectories) to be unused and thus have removed them. When the `apiv1.rb` and `apiv2.rb` scripts are run they will create any necessary directories and files as needed.


phallobst:
I tested this and compared two pretty-printed `api/v1/data.json` files of the old and new kind. There was a difference in the representation of exception texts:
```diff
-      "exceptions": {
-        "text": "foo bar"
-      }
+      "exceptions": [
+        {
+          "text": "foo bar"
+        }
+      ],
```
Which is the correct one? Given that the key is `exceptions` (plural), I'd guess that an array would be more appropriate, but that would be a change in the API file format, if I understand this correctly.

Carlgo11:
My bad :sweat_smile: 
Thought I had checked that everything looks like before but apparently I missed this. Nice catch! :+1: 

It's called exception**s** because the old YAML schema was designed to support several different types of exceptions and not just in the text format.

phallobst (approved):

dsoegijono (approved):

PR #5078, opened 2020-11-10
Duplicates check

Carlgo11:
Runs through all domains in _data and raises an exception if the same is listed twice.
This would help us act quicker when someone creates a duplicate PR.

Depends on #5077

### DIFF
--- a/_deployment/tests/duplicates.rb
+++ b/_deployment/tests/duplicates.rb
@@ -1,4 +1,4 @@
-#!/bin/ruby
+#!/usr/bin/env ruby
 # frozen_string_literal: true
 
 require 'yaml'
--- a/_deployment/tests/duplicates.rb
+++ b/_deployment/tests/duplicates.rb
@@ -1,0 +1,17 @@
+#!/bin/ruby
+# frozen_string_literal: true
+
+require 'yaml'
+require 'json'
+require 'uri/http'
+require 'addressable/uri'
+
+urls = []
+YAML.load_file('_data/sections.yml').each do |section|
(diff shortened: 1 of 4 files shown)


_deployment/tests/duplicates.rb
```diff
@@ -0,0 +1,17 @@
...
+
+urls = []
+YAML.load_file('_data/sections.yml').each do |section|
+  YAML.load_file("_data/#{section['id']}.yml")['websites'].each do |website|
+    domain = Addressable::URI.parse(website['url']).host
+    raise("::error:: Duplicate entries for #{domain}") if urls.include?(domain)
+
+    urls << domain
+  end
+end
```
phallobst: Collecting all errors is better than failing at the first one:
```suggestion
status = 0
urls = []
YAML.load_file('_data/sections.yml').each do |section|
  YAML.load_file("_data/#{section['id']}.yml")['websites'].each do |website|
    domain = Addressable::URI.parse(website['url']).host
    if urls.include?(domain)
      STDERR.puts "::error:: Duplicate entries for #{domain}"
      status = 1
    end

    urls << domain
  end
end
exit status
```

This yields a number of duplicates:
```
::error:: Duplicate entries for www.discover.com
::error:: Duplicate entries for www.nearlyfreespeech.net
::error:: Duplicate entries for www.rackspace.com
::error:: Duplicate entries for www.upcloud.com
::error:: Duplicate entries for ventraip.com.au
::error:: Duplicate entries for aws.amazon.com
::error:: Duplicate entries for www.virginmedia.com
```
:astonished:

PR #5079, opened 2020-11-10
Remove duplicates

Carlgo11:
Removed duplicate entries.


phallobst (commented):
I think "UpCloud" should stay in _Cloud Computing_ and be removed from _Hosting/VPS_, not the other way around.

phallobst (approved):

PR #5080, opened 2020-11-10
Remove AWS WorkSpaces

Carlgo11:
Amazon AWS WorkSpaces was added to twofactorauth.org back in 2014. At a time when Amazon and AWS didn't have a system-wide 2FA setup.
Now that they do, I don't see much of a reason to keep this entry.

Furthermore, the blog post (doc link) explains how they integrate with external RADIUS services in order to offer 2FA. That would make it a non-native 2FA implementation (like how some sites offer 2FA but only by signing up via Facebook or Google).


jamcat22 (approved):
My [previous comment](https://github.com/2factorauth/twofactorauth/pull/4605#pullrequestreview-380874649) on AWS still applies. I'm ok with removing AWS WorkSpaces as a separate entry from the main AWS entry since most people will search for that anyway. I believe AWS should stay as a separate entry in Cloud though.

phallobst (approved):

PR #5084, opened 2020-11-14
Add API documentation

Carlgo11:
A more comprahensive documentation than the one that #5075 removes.

Unsure about the grammar. @jamcat22 might want to look at that 🙈

### DIFF
--- a/API.md
+++ b/API.md
@@ -21,7 +21,7 @@
 |software|Boolean||Software token support (including RFC-6238)|
 |hardware|Boolean||Hardware token support (including U2F-tokens)|
 |doc|URL||URL to documentation page|
-|exceptions|Mapping\<"text": String>||Array containing the key `text` describing any discrepancies in the 2FA implementation|
+|exceptions|Object\<"text": String>||Object containing the key `text` describing any discrepancies in the 2FA implementation|
 |twitter|String||Twitter handle|
 |facebook|String||Facebook page name|
 |email_address|String||Email address to support|
@@ -90,7 +90,7 @@ If a website only supports some 2FA methods, the unsupported 2FA methods won't b
 |---|-----|---------------|-----------|
 |url|URL|:heavy_check_mark:|URL to the main page of the site/service|
 |img|String|:heavy_check_mark:|Image name used|
-|tfa|Sequence\<String>||Array containing all supported 2FA methods|
+|tfa|Array\<String>||Array containing all supported 2FA methods|
 |doc|URL||URL to documentation page|
 |exception|String||Text describing any discrepancies in the 2FA implementation|
 |twitter|String||Twitter handle|
--- a/API.md
+++ b/API.md
@@ -28,7 +28,7 @@
 
 ### Example website with 2FA disabled
 
-```YAML
+```JSON
 {
   "Category name": {
     "Website name": {
@@ -43,7 +43,7 @@
 
 ### Example website with 2FA enabled
 
-```YAML
+```JSON
 {
   "Category name": {
     "Website name": {
--- a/API.md
+++ b/API.md
@@ -19,7 +19,7 @@
 |phone|Boolean||Phone call support|
 |email|Boolean||Email token support|
 |software|Boolean||Software token support (including RFC-6238)|
-|hardware|Boolean||Hardware token support (including U2F-tokens)|
+|hardware|Boolean||Hardware token support (including U2F tokens)|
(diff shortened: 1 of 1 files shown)


phallobst:
Maybe it would better to put the v2 paragraph _before_ v1 to indicate the preference.

API.md
```diff
@@ -0,0 +1,151 @@
...
+|---|--------|
+|https://twofactorauth.org/api/v2/all.json|All sites|
+|https://twofactorauth.org/api/v2/tfa.json|All 2FA-supporting sites|
+|https://twofactorauth.org/api/v2/sms.json|SMS|
+|https://twofactorauth.org/api/v2/phone.json|Phone calls|
+|https://twofactorauth.org/api/v2/email.json|Email 2FA|
+|https://twofactorauth.org/api/v2/hardware.json|non-U2F hardware 2FA tokens|
+|https://twofactorauth.org/api/v2/u2f.json|U2F hardware tokens|
+|https://twofactorauth.org/api/v2/totp.json|RFC-6238|
+|https://twofactorauth.org/api/v2/proprietary.json|non-RFC-6238 software 2FA
```
phallobst:
```suggestion
|https://twofactorauth.org/api/v2/proprietary.json|non-RFC-6238 software 2FA|
```

API.md
```diff
@@ -0,0 +1,151 @@
...
+|Key|Value|Always defined|Description|
+|---|-----|---------------|-----------|
+|url|URL|:heavy_check_mark:|URL to the main page of the site/service|
+|img|String|:heavy_check_mark:|Image name used|
+|tfa|Boolean|:heavy_check_mark:|2FA support|
+|sms|Boolean||SMS token support|
+|phone|Boolean||Phone call support|
+|email|Boolean||Email token support|
+|software|Boolean||Software token support (including RFC-6238)|
+|hardware|Boolean||Hardware token support (including U2F-tokens)|
```
phallobst:
```suggestion
|hardware|Boolean||Hardware token support (including U2F tokens)|
```

API.md
```diff
@@ -0,0 +1,151 @@
...
+    }  
+  }
+}
+```
+
+## Notes and remarks
+
+### URL & Domain matching
+
+If you're using our API to match client URLs with our dataset make sure you're only use the domain of the `url`-element as a website commonly uses subdomains and subdirectories.
```
phallobst:
```suggestion
If you're using our API to match client URLs with our dataset make sure you only use the domain of the `url`-element, as a website commonly uses subdomains and subdirectories.
```

phallobst (changes requested):

Carlgo11:
I put v1 at the top to preserve chronological order but I guess we could use reverse chronological order. Should the remarks still be at the bottom then?

phallobst:
Having an introduction (along with some development notes/remarks) at the top seems to be better, I think.

Carlgo11:
@phallobst I've done a couple of drafts with your suggestions but they all suffer in the readability aspect.
As the remarks rely on Version 2 having been read it can only really be below Version 2. That means that the order would be Version 2, Remarks, Version 1 which looks really weird.

So basically, how strong do you feel about the changes? Because making the changes would require rewriting the entire document in order for it to be coherent.

phallobst:
What about his one? I didn't really change that much, and it's still quite readable (although lacking a strong story arc).

----

# API usage

## Introduction

The data collected for the twofactorauth.org website is also available as JSON files in order to enable developers to use it in their own programs. It is recommended to use the API with the highest version number, since older versions might not include all available information.

### URL & Domain matching

If you're using our API to match client URLs with our dataset make sure you only use the domain of the `url`-element, as a website commonly uses subdomains and subdirectories. Please note that there are exceptions like `.co.uk`, `.com.au` or `.co.nz` where the actual domain is found at a lower level.

### Caching

If you intent to query our JSON files often and with a lot of traffic, you may be blocked by Cloudflare, our reverse proxy provider. We therefore recommend that you cache the files locally for any large traffic cases.

### Avoid downloading unnecessary data

If you only intent on using a specific dataset, like all sites supporting RFC-6238, we recommend that you use the URI which lists just that. See [URIs](#uris-1) for a list of available paths. The smaller the better.

## Version 2
[...]

## Version 1 (Deprecated) :warning:
[...]

phallobst (approved):

dsoegijono (approved):

PR #5108, opened 2020-11-28
Add GA

jdavis:
There's a lot of improvement we can do to make the site more usable as well as more discoverable. I never wanted to add GA or any other trackers but I want to see where users are coming from. Then use that data as the basis for improving usability.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -24,6 +24,14 @@
     <meta property="twitter:image" content="{{ site.social_image }}">
     <meta property="twitter:description" content="{{ site.description }}">
     <title>{{ site.title }}</title>
+    <script async src="https://www.googletagmanager.com/gtag/js?id=G-SMXGN4C672"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag() { dataLayer.push(arguments); }
+      gtag('js', new Date());
+
+      gtag('config', 'G-SMXGN4C672');
+    </script>
     {% include icons.html %}
     <link href="/manifest.json" rel="manifest">
     <link rel="stylesheet" href="/css/bootstrap.css"/>
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -24,19 +24,12 @@
     <meta property="twitter:image" content="{{ site.social_image }}">
     <meta property="twitter:description" content="{{ site.description }}">
     <title>{{ site.title }}</title>
-    <script async src="https://www.googletagmanager.com/gtag/js?id=G-SMXGN4C672"></script>
-    <script>
-      window.dataLayer = window.dataLayer || [];
-      function gtag() { dataLayer.push(arguments); }
-      gtag('js', new Date());
-
-      gtag('config', 'G-SMXGN4C672');
-    </script>
     {% include icons.html %}
     <link href="/manifest.json" rel="manifest">
     <link rel="stylesheet" href="/css/bootstrap.css"/>
     <link rel="stylesheet" crossorigin="anonymous" integrity="sha512-9BwLAVqqt6oFdXohPLuNHxhx36BVj5uGSGmizkmGkgl3uMSgNalKc/smum+GJU/TTP0jy0+ruwC3xNAk3F759A==" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.12.1/css/all.min.css"/>
     <link rel="stylesheet" crossorigin="anonymous" integrity="sha512-xbKj38YamwOa1yyC9gTFLExhbbV6r8WZwaEYYYaedfU0gosQuncv6p//9zcjK1NO2nkV/Nv5X5TkM/zjOnlc+w==" href="https://cdnjs.cloudflare.com/ajax/libs/flag-icon-css/3.4.6/css/flag-icon.min.css"/>
+    <link rel="stylesheet" crossorigin="anonymous" integrity="sha512-LQ97camar/lOliT/MqjcQs5kWgy6Qz/cCRzzRzUCfv0fotsCTC9ZHXaPQmJV8Xu/PVALfJZ7BDezl5lW3/qBxg==" href="https://cdn.jsdelivr.net/npm/cookieconsent@3.1.1/build/cookieconsent.min.css">
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:400,700|Roboto:300|Spartan:500&display=swap">
     <link href="/css/main.css" rel="stylesheet">
     <link href="/css/navbar.css" rel="stylesheet">
@@ -60,14 +53,6 @@
         <path class="octo-body" d="M115.0,115.0 C114.9,115.1 118.7,116.5 119.8,115.4 L133.7,101.6 C136.9,99.2 139.9,98.4 142.2,98.6 C133.8,88.0 127.5,74.4 143.8,58.0 C148.5,53.4 154.0,51.2 159.7,51.0 C160.3,49.4 163.2,43.6 171.4,40.1 C171.4,40.1 176.1,42.5 178.8,56.2 C183.1,58.6 187.2,61.8 190.9,65.4 C194.5,69.0 197.7,73.2 200.1,77.6 C213.8,80.2 216.3,84.9 216.3,84.9 C212.7,93.1 206.9,96.0 205.4,96.6 C205.1,102.4 203.0,107.8 198.3,112.5 C181.9,128.9 168.3,122.5 157.7,114.1 C157.9,116.9 156.7,120.9 152.7,124.9 L141.0,136.5 C139.8,137.7 141.6,141.9 141.8,141.8 Z"></path>
       </svg>
     </a>
-
-    {% unless page.name == "index.html" %}
-      <div class="back-arrow">
-        <a onclick="history.back()">
-          <i class="fas fa-arrow-left"></i>
-        </a>
-      </div>
-    {% endunless %}
 
     <div class="container">
       <noscript>
@@ -96,14 +81,18 @@
       </div>
     </div>
 
+    <div id="consentCookie"></div>
     <script crossorigin="anonymous" integrity="sha512-C5x7QiS49O4ubITbqnAX93oOLk9/cASa+m8GawJMki1i4z8AjFDeklRCqiBBjj7niQULNW6nDhqf7l06zrZ9mA==" src="https://cdnjs.cloudflare.com/ajax/libs/jets/0.14.1/jets.min.js"></script>
     <script crossorigin="anonymous" integrity="sha512-bLT0Qm9VnAYZDflyKcBaQ2gg0hSYNQrJ8RilYldYQ1FxQYoCLtUjuuRuZo+fjqhx/qtq/1itJ0C2ejDxltZVFg==" src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
     <script crossorigin="anonymous" integrity="sha512-nkVN47hpd3LOCD7BXeONKhZVrx2EE69VNvOwH28U6QTfjZ/LBfUN6EkmsinvuGVD5UWhyQeJudkVkkT+B4ffrA==" src="https://cdn.jsdelivr.net/npm/vanilla-lazyload@15.1.1/dist/lazyload.min.js"></script>
+    <script crossorigin="anonymous" integrity="sha512-yXXqOFjdjHNH1GND+1EO0jbvvebABpzGKD66djnUfiKlYME5HGMUJHoCaeE4D5PTG2YsSJf6dwqyUUvQvS0vaA==" src="https://cdn.jsdelivr.net/npm/cookieconsent@3.1.1/build/cookieconsent.min.js"></script>
+    <script async src="https://www.googletagmanager.com/gtag/js?id=G-SMXGN4C672&aip=1"></script>
     <script src="/js/bootstrap.min.js"></script>
     <script src="/js/popup.js"></script>
     <script src="/js/main.js"></script>
     <script src="/js/social-media.js"></script>
     <script src="/js/search.js"></script>
+    <script src="/js/g.js"></script>
   </body>
 
 </html>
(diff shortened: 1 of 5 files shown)


Carlgo11:
Adding Google Analytics would be bad for our GDPR compliance. As of right now we're in a grey area but adding tracking would definitely be against GDPR.

jdavis:
Yeah good point. I'll add the consent dialog.

Carlgo11:
I'm afraid that's not enough. Due to privacy shield falling through, PII can't leave the EU.

jdavis:
We don't deal with PII directly other than IP addresses. I'll turn on IP address anonymization.

phallobst:
Also, given the nature of 2fa.org, our users are more likely to be privacy/security concious, i.e. they probably have an ad-blocker running, which may render this useless (e.g. by blocking googletagmanager.com).

What kind of information are you actually hoping to gather?

jdavis:
> Also, given the nature of 2fa.org, our users are more likely to be privacy/security concious,

That's fine. A sample of data is what I'm looking for. In particular, what pages people are coming from, what are they searching for to end up on the site.

We just plop people on the main page when it's very possible they are searching for "banks with two factor auth". We aren't making it easy for those users.

jdavis:
Also, just a personal gripe: the site is hard to use (especially on mobile, it doesn't look good). There are a ton of sites, which is great, but at a certain point, there's too much for many of the sections. Usability just really could use some improving. I think with more data, it'll be clearer how people are using it.

Carlgo11:
It technically is doable but you'd have to make it opt-in, excluding any EU-citizen (along with users abroad) or make arrangements with each EU country on any data transfers outside the EU, provide a privacy policy detailing how you and any third party (Google) intend to use the data plus a public address to send any legal letters to.

And the fines for misuse would be enormous.

Carlgo11:
I instead suggest sticking with the data that Cloudflare provides. It includes regions, volumes etc all while doing it in a way that complies with EU regulations.

jdavis:
Cool, updated it and it should be good to go.

jdavis:
When the site loads, it'll try to determine the location of the user. If it can't, it'll show the prompt. If it can, it'll look at regional rules to determine if it needs to show a cookie consent. If and only if the user has allowed or is in a country that doesn't care, it'll load GA.

Also added a bare bones privacy policy.

Here's what it looks like:
Screenshot 2020-11-29 002305

Carlgo11:
Things left to do:
- [x] Only show notification box when the IP is outside of the EU (Unlawful to collect even when the user gives consent) and DNT is disabled. ([DNT example](https://dev.to/corbindavenport/how-to-correctly-check-for-do-not-track-with-javascript-135d))
- [x] Initialize Google Analytics on accept button press instead of cookie? (Cookie laws differ from region to region so it's always better not to use them when an alternative is available)
- [x] Use session instead of cookie storage. ([example](https://developers.google.com/analytics/devguides/collection/analyticsjs/cookies-user-id#disabling_cookies))
- [x] List all parties that will have access to the data. (Josh Davis, Alphabet Inc.)
- [x] Provide legal name and address for all parties. (Alphabet Inc have the primary address: 1600 Amphitheater Parkway, Mountain View, CA 94043, USA )
- [x] Provide way of requesting data deletion.
- [x] Make the agree text clearer. (example text: `I'm not a citizen or resident of any EU/EEA country, Switzerland or the United Kingdom and I volunteer to give pseudo-anonymized telemetry data to the legal entities listed in <privacy policy section>`) (GDPR still applies to people in the EU that travel outside the EU so extra care needs to be put to make sure they're not from the EU)
- [x] List in privacy policy: what data is collected, how it is stored, what measures are taken to prevent unlawful usage of the data, how long the data is stored.

And personally I'd skip the part about usage by minors and parental monitoring since it doesn't really apply to the service we provide or the data that you want to collect 😉

jdavis:
1. Added DNT support. We want to show the notification box when they are in the EU or a country where it is relevant.
2. It only initializes GA when the allow button is clicked unless it determines it doesn't need to show the box based on IP.
3. I don't see why we should use localStorage over a cookie.
4. Done.
5. Done
6. Done.
7. I'm not putting that verbose of text. That's what the privacy policy is for. Also, it only applies to citizens outside of a GDPR country if the data processing company is based in a country covered by GDPR. In this case, that's not true so it only applies to those citizens when their data is collected within a GDPR country.
8. Done

Carlgo11:
1. I don't see why a notification box is needed then. It's illegal to move the data even with the users consent so if the user is inside the EU/EEA/EFTA/UK then they shouldn't be able to press the consent button.

\7. Actually there have been cases of GDPR violation cases due to EU tourists so extra verification needs to be done.

Regarding the privacy policy, it isn't a contract and even if it were, that would most likely be nullified in an EU court as many courts take in to account the odds that the user would read and comprehend such a contract.
Therefore the privacy policy is to be more perceived as a "in case you want to read more" document rather than a contract.
I understand that this is opposite to the US system.

jdavis:
As far as I can tell, that's only for PII. This is based off of looking at some recent digests of Google Analytics + GDPR: [source one](https://www.cookieyes.com/google-analytics-and-gdpr/) and [source two](https://www.cookiebot.com/en/google-analytics-gdpr/).

Again, this is not PII. I'm not seeing these claims about tourists anywhere. https://gdpr.eu/companies-outside-of-europe

Also, I'm not claiming the privacy policy is a contract. It links off to it and details what is all being done with the GA data. That's literally why I said they can click the link if they want more info. The cookie consent message is very similar to most other sites.

privacy.html
```diff
@@ -0,0 +1,91 @@
...
+    <h1>Privacy Policy for Two Factor Auth</h1>
+    <p>At Two Factor Auth, accessible from https://twofactorauth.org/, one of our main priorities is the privacy of our
+      visitors. This Privacy Policy document contains types of information that is collected and recorded by Two Factor
+      Auth and how we use it.</p>
+    <h2>General Data Protection Regulation (GDPR)</h2>
+    <p>We are a Data Controller of your information.</p>
+    <p>Two Factor Auth legal basis for collecting and using the personal information described in this Privacy Policy
+      depends on the Personal Information we collect and the specific context in which we collect the information:</p>
+    <ul>
+      <li>Two Factor Auth needs to perform a contract with you</li>
```
Carlgo11, line 16: ?

privacy.html
```diff
@@ -0,0 +1,91 @@
...
+    <ul>
+      <li>The right to access, update or to delete the information we have on you.</li>
+      <li>The right of rectification.</li>
+      <li>The right to object.</li>
+      <li>The right of restriction.</li>
+      <li>The right to data portability</li>
+      <li>The right to withdraw consent</li>
+    </ul>
+    <h2>Data Access</h2>
+    <p>The only parties that have access to the data include Josh Davis (Seattle, WA) as well
```
Carlgo11, line 39: Address needs to be more precise so that legal notices can be recieved.

privacy.html
```diff
@@ -0,0 +1,91 @@
...
+      reports on user activity and shares it with Google's services. The data collected includes but not limited to: referrer, page visit data
+      such as time of visit, duration of visit, web browser and OS details. Data is anonymized.</p>
+    <p>View more information about Google Analytics here: <a href="https://analytics.google.com/analytics/web/">Google Analystics Info</a></p>
+    <p>More details:</p>
+    <ul>
+      <li><strong>Data process location:</strong> United States</li>
+      <li><strong>Personal Data processed:</strong> Tracker, usage data</li>
+    </ul>
+    <h2>Cookies and Web Beacons</h2>
+    <p>Like any other website, Two Factor Auth uses 'cookies'. These cookies are used to store information including
```
Carlgo11:
```suggestion
    <p>Like other websites, Two Factor Auth uses 'cookies'. These cookies are used to store information including
```

privacy.html
```diff
@@ -0,0 +1,91 @@
...
+    <h2>Privacy Policies</h2>
+    <p>You may consult this list to find the Privacy Policy for each of the advertising partners of Two Factor Auth.</p>
+    <p>Third-party ad servers or ad networks uses technologies like cookies, JavaScript, or Web Beacons that are used in
+      their respective advertisements and links that appear on Two Factor Auth, which are sent directly to users'
+      browser.
+      They automatically receive your IP address when this occurs. These technologies are used to measure the
+      effectiveness of their advertising campaigns and/or to personalize the advertising content that you see on
+      websites that you visit.</p>
+    <p>Note that Two Factor Auth has no access to or control over these cookies that are used by third-party
+      advertisers.
```
Carlgo11:
```suggestion
```
The site doesn't serve ads.

privacy.html
```diff
@@ -0,0 +1,91 @@
...
+    </p>
+    <h2>Third Party Privacy Policies</h2>
+    <p>Two Factor Auth's Privacy Policy does not apply to other advertisers or websites. Thus, we are advising you to
+      consult the respective Privacy Policies of these third-party ad servers for more detailed information. It may
+      include their practices and instructions about how to opt-out of certain options. </p>
```
Carlgo11:
```suggestion
```

privacy.html
```diff
@@ -0,0 +1,91 @@
...
+    <h2>Children's Information</h2>
+    <p>Two Factor Auth does not knowingly collect any Personal Identifiable Information from children under the age of
+      13.</p>
+    <h2>Online Privacy Policy Only</h2>
+    <p>Our Privacy Policy applies only to our online activities and is valid for visitors to our website with regards to
+      the information that they shared and/or collect in Two Factor Auth. This policy is not applicable to any
+      information
+      collected offline or via channels other than this website.</p>
+    <h2>Consent</h2>
+    <p>By using our website, you hereby consent to our Privacy Policy and agree to its terms.</p>
```
Carlgo11:
```suggestion
```
Consent has to be opt-in and can't be required for using the website.

privacy.html
```diff
@@ -0,0 +1,91 @@
...
+    <ul>
+      <li><strong>Data process location:</strong> United States</li>
+      <li><strong>Personal Data processed:</strong> Tracker, usage data</li>
+    </ul>
+    <h2>Cookies and Web Beacons</h2>
+    <p>Like any other website, Two Factor Auth uses 'cookies'. These cookies are used to store information including
+      visitors' preferences, and the pages on the website that the visitor accessed or visited. The information is used
+      to
+      optimize the users' experience by customizing our web page content based on visitors' browser type and/or other
+      information.</p>
```
Carlgo11, line 57: Website is static. Do you mean that the collected information will be used for future website improvements?
```suggestion
```

privacy.html
```diff
@@ -0,0 +1,91 @@
...
+      <li>The right to withdraw consent</li>
+    </ul>
+    <h2>Data Access</h2>
+    <p>The only parties that have access to the data include Josh Davis (Seattle, WA) as well
+      as Google LLC (1600 Amphitheater Parkway, Mountain View, CA 94043, USA)</p>
+    <h2>Google Analytics</h2>
+    <p>The services contained in this section enable the Owner to monitor and analyze web traffic and can be used to keep track of User behavior. </p>
+    <p>Google Analytics is a service that analyzes web traffic and is provided by Google LLC (“Google”). The data is collected and used to prepare
+      reports on user activity and shares it with Google's services. The data collected includes but not limited to: referrer, page visit data
+      such as time of visit, duration of visit, web browser and OS details. Data is anonymized.</p>
```
Carlgo11:
```suggestion
      such as time of visit, duration of visit, web browser and OS details. Data is pseudo-anonymized.</p>
```
Anonomized data would mean it can't be correlated to a sender.

_layouts/default.html
```diff
@@ -88,14 +81,19 @@ <h2>It looks like you have JavaScript disabled!</h2>
       </div>
     </div>
 
+    <div id="consentCookie"></div>
     <script crossorigin="anonymous" integrity="sha512-C5x7QiS49O4ubITbqnAX93oOLk9/cASa+m8GawJMki1i4z8AjFDeklRCqiBBjj7niQULNW6nDhqf7l06zrZ9mA==" src="https://cdnjs.cloudflare.com/ajax/libs/jets/0.14.1/jets.min.js"></script>
     <script crossorigin="anonymous" integrity="sha512-bLT0Qm9VnAYZDflyKcBaQ2gg0hSYNQrJ8RilYldYQ1FxQYoCLtUjuuRuZo+fjqhx/qtq/1itJ0C2ejDxltZVFg==" src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
     <script crossorigin="anonymous" integrity="sha512-nkVN47hpd3LOCD7BXeONKhZVrx2EE69VNvOwH28U6QTfjZ/LBfUN6EkmsinvuGVD5UWhyQeJudkVkkT+B4ffrA==" src="https://cdn.jsdelivr.net/npm/vanilla-lazyload@15.1.1/dist/lazyload.min.js"></script>
+    <script crossorigin="anonymous" integrity="sha512-yXXqOFjdjHNH1GND+1EO0jbvvebABpzGKD66djnUfiKlYME5HGMUJHoCaeE4D5PTG2YsSJf6dwqyUUvQvS0vaA==" src="https://cdn.jsdelivr.net/npm/cookieconsent@3.1.1/build/cookieconsent.min.js"></script>
+    <script async src="https://www.googletagmanager.com/gtag/js?id=G-SMXGN4C672&aip=1"></script>
```
Carlgo11, line 89: Script should be retrived when consent is given.

Carlgo11 (changes requested):
* IP, user-agent, window size etc are PII so 1. still applies.
* Off the top of my head one insident that comes to mind is an incident in south africa where a speed fine-tool got breached and EU citizens had data compromised. It is of course not possible to fine a company that doesn't operate in the EU but it's still illegal.
* I was referring to https://github.com/2factorauth/twofactorauth/pull/5108/files#diff-2bf02a33e2f1187fb2aeac94746b4508724cbcaf55dea6630162731b7c8f726cR89

privacy.html
jdavis: This was just from a privacy policy template I found.

privacy.html
jdavis: Sure, but I think that's irrelevant from a technical/legal standpoint.

privacy.html
jdavis: Oh thanks. Like I said, I used a template for Google services. I missed this section on ads.

_layouts/default.html
jdavis: There's no reason to do this other than just a preference. When the script loads, it's started in `consent mode` and the IPs are anonymized, so other than the initial HTTP request to load the script, nothing else happens.

Carlgo11: Except that we don't get that nice green sheild in Firefox and Safari 🛡️ 😁 
The less is collected by default, the better.

Carlgo11 (changes requested):
Things left to do:
- [ ] Only show notification box when the IP is outside of the EU
- [ ] Make the agree text clearer.
  * Make sure user isn't from the EU/EEA/Switzerland or the UK. ([this](https://www.osano.com/cookieconsent/documentation/location-services-cookie-law/) might help with that?)
  * Point out it's for telemetry purposes
  * Point out that it's completly volonteery

jdavis:
Thanks for the additional suggestions, Carl but I disagree with them. I think things are fine at the moment.

dsoegijono (approved):

PR #5115, opened 2020-11-30
Update namecheap

mmomjian:
Namecheap now uses standard TOTP 2FA, not proprietary. See: [article](https://www.namecheap.com/support/knowledgebase/article.aspx/10073/45/how-can-i-use-the-totp-method-for-twofactor-authentication/)


mmomjian:
I didn't realize that some websites have *both* TOTP and proprietary app listed, my mistake.

PR #5116, opened 2020-11-30
Add US Postal Service

mmomjian:
No public-facing documentation present

Edit: documentation page added to commit

image


phallobst:
I'm not sure about the _Government_ category here. There are some postal services under _Other_ already; maybe the entry should be moved there for consistency.

mmomjian:
I wasn't sure about this, however all the services I saw in other were private companies (DHL, UPS, etc) - I wasn't sure if this was different because it is the official US mailing service, and I didn't see any other governmental postal carriers on the website yet.

_data/government.yml
```diff
@@ -201,6 +201,15 @@ websites:
     regions:
       - us
 
+  - name: US Postal Service
+    url: https://usps.com
+    img: usps.png
+    tfa:
+      - sms
+    doc: https://reg.usps.com/entreg/AccountRecoveryHelpAction_input
```
phallobst:
```suggestion
    doc: https://faq.usps.com/s/article/USPS-com-Account#Two-Factor_Authentication
```

phallobst (changes requested):

phallobst:
Although the USPS is a governmental agency, I do not perceive it as such (as a non-American), but rather as a postal service. In my opinion, it's better to move it to _Other_. Maybe it's even time to create a new category _Postal Service_ or something.

mmomjian:
If we made a new category such as "Mailing", "Shipping", or "Postal service" we could include the following. Is that enough to justify a new category? If not, I will move USPS to Other with the ones below.

- Australia Post
- DHL Deutschland
- DPD
- Hermes DE
- UK Postbox
- UPS
- US Postal Service

https://www.poste.it/ (Poste italiane) is listed under Banking, they could possibly be moved as well. I assume that the post office provides banking services (like many of the ones in Other) but the website linked seems to be the main site of the Italian Post Office.

phallobst:
There's _GLS-ONE_ as well. Regarding _Poste Italiane_, I agree that it rather belongs to the _Postal Services_.

We do not have a fixed number of entries required to start a new category, but 5+ seems reasonable to me.

However, I'd suggest to tackle the new category with a new PR, and add USPS to _Other_ with this one. That way the site is added quickly to the data set, and we can take our time with the new category.

mmomjian:
I will work on a new category soon.

PR #5118, opened 2020-11-30
Add MyChart by Epic

mmomjian:
Documentation is in commit. Low ranking on Alexa for this site but the platform / backend is used by many hospitals and healthcare organizations in the US.


_data/health.yml
```diff
@@ -153,6 +153,16 @@ websites:
     regions:
       - us
 
+  - name: MyChart by Epic
```
phallobst:
```suggestion
  - name: MyChart
```

phallobst (changes requested):

_data/health.yml
mmomjian: Epic is the company providing MyChart backend product, you can see it is included in the logo on the website. For people searching for setup / clarification I think that the full name may be more helpful to find it easily

phallobst: We usually do not include the company name if the product name is distinctive enough. People will probably search for "MyChart" rather than "Epic", I suppose.

phallobst:
The text on the logo is very hard to read when it's scaled down like that. Maybe this one (which is also their favicon) is better:
mychart

mmomjian:
Good idea

phallobst (approved):

dsoegijono (approved):

PR #5126, opened 2020-12-01
Remove Sprint

mmomjian:
Since acquisition, sprint.com now redirects to t-mobile.com, and shares a login portal as well


phallobst:
Duplicate of #5073

PR #5127, opened 2020-12-01
Update T-Mobile

mmomjian:
image

### DIFF
--- a/_data/utilities.yml
+++ b/_data/utilities.yml
@@ -436,7 +436,7 @@ websites:
     tfa:
       - sms
       - totp
-      - proprietary
+      - hardware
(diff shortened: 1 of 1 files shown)


_data/utilities.yml
```diff
@@ -436,8 +436,9 @@ websites:
     tfa:
       - sms
       - totp
-      - hardware
+      - proprietary
```
phallobst: The proprietary T-Mobile app is not the authentication factor here, but rather biometric features. These are currently classified as _hardware_.
```suggestion
      - hardware
```

phallobst (changes requested):

phallobst (approved):

dsoegijono (approved):

PR #5133, opened 2020-12-03
Update Cloudbet

40504:
I didn't find any other article about how to enable 2FA on cloudbet.com

### DIFF
--- a/_data/betting.yml
+++ b/_data/betting.yml
@@ -19,7 +19,7 @@ websites:
     img: cloudbet.png
     tfa:
       - totp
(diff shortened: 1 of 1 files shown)


40504:
Seems like this article is a little out of date :), the right path is: "My account -> Security - > 2-Factor Authentification"

40504:
Screenshot 2020-12-03

phallobst (approved):
It's better than nothing.

dsoegijono (approved):

PR #5140, opened 2020-12-04
Add Postal and Shipping category

mmomjian:
Name was selected because this category includes both governmental mail services and private shippers. I attempted to run a local test which seems to have worked fine (see below)

image


Carlgo11:
Would `Post & Logistics` work? The word "post" is common in many more languages.

mmomjian:
Sounds good, but I'm unsure what "logistics" means as an American. If I was searching for this category I would probably look for "mail, delivery, shipment/shipping, postal service", etc. How about "Post and shipping" or "post and delivery"?

Carlgo11:
Logistics is the handling of the object from sender to reciever; Including packaging, transporting, storing and delivery.
Shipping is the transporting part of the chain.

Companies like DHL, UPS, FedEx, USPS etc. are logistics companies while companies like AeroLogic, Atlast Air, American Airlines etc. are shipping companies.

Carlgo11:
Actually, yeah "Post & Shipping" could be better. Let's do that :+1:

mmomjian:
Done!

Carlgo11 (approved):

phallobst (approved):

PR #5141, opened 2020-12-04
Clean up HTML entities in Utilities

mmomjian:
`& -> &amp;`


dsoegijono:
Is this necessary? It looks like it's currently rendered properly on the site. If anything, I'd prefer having `&` in the data and ensure the HTML renders it properly.

mmomjian:
I made a commit a few days ago including the & character and I was requested to change `& to &amp;`, so when I saw these I thought it was in error. Currently it seems that it's about 50/50 which one is used. Maybe there should be a discussion to settle on one style?

phallobst:
I think it's good practice to use entities for those characters that have a specific meaning for HTML:
- &amp; (`&amp;`)
- &lt; (`&lt;`)
- &gt; (`&gt;`)
- &quot; (`&quot;`)

dsoegijono (approved):

ApCoder123 (approved):

PR #5152, opened 2020-12-10
Add Countly

40504:


phallobst:
This seems to be a self-hosted software package - at least I can see no login page. We don't list those [see EXCLUSION.md].

40504:
https://try.count.ly/login it's possible to use both

phallobst:
Is there an official link to that login page? I can't see it... The "try" part might indicate that it's just a demo installation on their site.

40504:
on the premium, https://count.ly/pricing, is available on both, self and cloud

phallobst:
I'm still not quite sure if it meets our requirements, to be honest.

Carlgo11:
As our site is focused on aiding normal consumers in choosing services and not focused on enterprise customers it would mean the enterprise edition of Countly is out of our scope. Since the non-enterprise edition is self-hosted it's out of scope as well.

Sorry for the inconvenience!

PR #5153, opened 2020-12-10
Add ThriveCart

40504:


phallobst (approved):

Carlgo11 (changes requested):
Hola!
There seems to be a weird white border around the image you uploaded. How about this one instead:

Carlgo11 (approved):

PR #5160, opened 2020-12-15
Update Skrill

40504:
now support TOTP

### DIFF
--- a/_data/payments.yml
+++ b/_data/payments.yml
@@ -235,12 +235,13 @@ websites:
       - us
 
   - name: Skrill
-    url: https://skrill.com/
+    url: https://skrill.com
     img: skrill.png
     tfa:
       - email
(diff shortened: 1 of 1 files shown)


jdavis (approved):

jdavis:
Does it still support SMS/email? I don't see that documented on that page.

40504:
I created an account on skrill, and it's really confusing if you use a mobile app(in my case iOS app) and website. 

Artboard

Sometimes, across these methods, you have to prove and the PIN also. https://www.skrill.com/en-us/support#/path/SECURITY/905782642/Skrill-PIN-Authentication.htm

jdavis:
Interesting. Yeah, that UI isn't great.

Well thanks for checking!

jdavis (approved):

PR #5166, opened 2020-12-23
Breakout some of the 2FA details of major US banks into landing pages

danbirken:
The site has a lot of great information but it isn't very accessible via
google search because of the site's formatting.

As a small, hacky-ish, test, try breaking out the banking pages into
individual landing pages and see if this ends up being more user
friendly in terms of google search and site experience.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -45,6 +45,7 @@ websites:
       - sms
       - email
     doc: https://www.ally.com/messages/sms/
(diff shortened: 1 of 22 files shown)


jdavis:
Just tested it. It's pretty hacky but let's roll with it for the time being. I'm going to improve the build process which should make it easier to customize Jekyll with plugins so these weird yaml files aren't needed

jdavis (approved):

PR #5171, opened 2020-12-28
LastPass does not support U2F.

multifactor-manifesto:
Hello!

LastPass does not support U2F. It only supports the OTP function of YubiKeys, which is significantly weaker than U2F.

From https://www.lastpass.com/yubico:
> The YubiKey offers flexibility for LastPass users looking to deploy Yubico One Time Password to access their password vault [...]

There's a good [demonstration of phishing LastPass](https://pberba.github.io/security/2020/05/28/lastpass-phishing/) which shows the difference.

Thanks for twofactorauth.org! It's an awesome reference!

### DIFF
--- a/_data/identity.yml
+++ b/_data/identity.yml
@@ -87,7 +87,6 @@ websites:
       - totp
       - hardware
       - proprietary
-      - u2f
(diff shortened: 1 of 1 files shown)


jdavis:
Thank you!

jdavis (approved):

PR #5212, opened 2021-01-27
Update Mint

40504:
"Mint", "Turbotax"(already listed), and "QuickBooks" already listed) and are part of the intuit.com family.


40504:
Prior to activating TOTP, you have to verify the phone number 

intuit

PR #5248, opened 2021-02-28
Change twofactorauth.org references to 2fa.directory

phallobst:
The mail address contact@2fa.directory currently forwards to me, as long as no proper mail handling is set up.
I can change the forward if there is a better alternative.


Carlgo11 (approved):
Great work! 👍

kmpoppe (approved):

PR #5257, opened 2021-03-05
Update email.yml

sneakernuts:
Added Gmx.com

Now supports totp just like gmx.net but requires sms verification to start the setup.


sneakernuts (commented):
Added Gmx.com

Now supports totp just like gmx.net but requires sms verification to start the setup.

PhamThiBexx:
[]()

PR #3378, opened 2018-08-14
Add Laravel Forge

jakebathman:
Forge is a server provisioning, deployment, and management service. I've added the site to `cloud.yml` and included a logo.

Please let me know if there are any questions or issues. Thanks!

PR #5271, opened 2021-03-08
Add Laravel applications

jbrooksuk:


Carlgo11 (changes requested):
* The category `utilities` doesn't fit these services. Utilities are meant to include things like water, electricty, broadband.
* If the services all use the same authentication process I think it's better to include just one entry and calling it Laravel.
* Do they really just serve the United States?
* SVG files are prefered over PNGs. If you have any SVG versions of the logos that would be appriciated 🙏

ApCoder123:
Closing, as all of these sites have now been listed (#3378, #5590, #5993)

PR #5276, opened 2021-03-09
Update with Target current 2FA status

bencooper222:
Target

### DIFF
--- a/_data/retail.yml
+++ b/_data/retail.yml
@@ -501,7 +501,6 @@ websites:
 
   - name: Target
     url: https://www.target.com/
-    twitter: Target
     img: target.png
     tfa:
       - sms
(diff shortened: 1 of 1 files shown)


_data/retail.yml
```diff
@@ -503,6 +503,10 @@ websites:
     url: https://www.target.com/
     twitter: Target
     img: target.png
+    tfa:
+      - sms
+      - email
+    doc: https://security.target.com/FAQs
```
Carlgo11, line 511: Let's add a region tag as well while you're at it 😉 
```suggestion
    regions:
      - us

```

Carlgo11 (changes requested):

Carlgo11 (approved):

phallobst (approved):

PR #5277, opened 2021-03-10
Add PGP signed API files at deployment

Carlgo11:
The script goes through all JSON files in the /api folder at deployment and creates seperate .sig files that can be used to verify the integrity of the API files.

The PGP key is already stored as a "secret" environment variable in the repository.

### DIFF
--- a/_deployment/sign.sh
+++ b/_deployment/sign.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+for file in api/v*/*.json; do
+echo "$file.sig"
+echo $1 | gpg --yes --passphrase --local-user $2 --output "$file.sig" --sign "$file"
+done
--- a/_deployment/sign.sh
+++ b/_deployment/sign.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/bin/sh
 
 for file in api/v*/*.json; do
 echo "$file.sig"
--- a/_deployment/sign.sh
+++ b/_deployment/sign.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
 for file in api/v*/*.json; do
 echo "$file.sig"
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -5,7 +5,7 @@ on:
     branches: master
 
 jobs:
-  build:
+  deploy:
     runs-on: ubuntu-latest
     if: github.repository == '2factorauth/twofactorauth'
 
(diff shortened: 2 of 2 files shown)


_deployment/sign.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
phallobst:
```suggestion
#!/bin/sh
```

phallobst (changes requested):

_deployment/sign.sh
Carlgo11: @phallobst the code depends on Bash to work. Many sh versions don't support wildcards.

phallobst: :confused: In the last 25 years of hacking shell scripts, I have never seen one that didn't support wildcards. I mean, we are not talking about recursive globbing like `**` in `zsh` or something...

Carlgo11: Busybox ash is one example of a system shell not supporting it.

phallobst:
```
BusyBox v1.30.1 (Ubuntu 1:1.30.1-4ubuntu6.3) built-in shell (ash)
Enter 'help' for a list of built-in commands.

/tmp $ touch a1 a2
/tmp $ for i in a*; do echo $i; done
a1
a2
/tmp $ mkdir b
/tmp $ mkdir b/c
/tmp $ echo ?/?
b/c
```
Are you running `busybox` on a toaster or something? :laughing:

Carlgo11:
```sh
#!/bin/sh

for file in api/v*/*.json; do
  echo "$file"
done
```
outputs
```
api/v*/*.json
```

phallobst: That's because there is no matching file, not because the shell wouldn't support wildcards. The `bash` shows the same behaviour, unless you set the `nullglob` option there.

Carlgo11: 🤦‍♂️ my bad.
How about `#!/usr/bin/env sh`?

phallobst: I don't really see the benefit, actually.

Carlgo11: Okay but can you see any downsides to having it? 🤔

phallobst: Other than "nobody in their right mind would do it like this"? No... :stuck_out_tongue_winking_eye:

_deployment/sign.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
```
Carlgo11:
```suggestion
#!/bin/sh
```

phallobst (approved):

kmpoppe (approved):

PR #5285, opened 2021-03-10
Add justhost

ApCoder123:


phallobst (changes requested):
Their favicon scaled to size is more readable, I suppose:
justhost

phallobst (approved):

Carlgo11 (approved):

PR #5288, opened 2021-03-13
Update twofactorauth.org references

ApCoder123:
There were a couple of twofactorauth.org references that weren't fixed by #5248 so I have replaced them with 2fa.directory


phallobst (approved):
That only happened because _someone_ wasn't searching case-insensitive the last time... :man_facepalming:

Carlgo11 (approved):

Issue #5061, opened 2020-10-25
Switch to SVG icons


phallobst:
PR #5026 will allow for SVG icons. This issue is used to track the switch to SVG for each category to ensure no work is done twice.

_Before_ you want to work on a certain category, tick the corresponding checkbox (or edit this issue and set it to `[x]`). In your PR you can refer to this issue with `see #5061` - do _not_ use the regular keyword that would close an issue.
- [x] Backup and Sync
- [x] Banking
- [x] Betting
- [x] Cloud Computing
- [x] Communication
- [x] Cryptocurrencies
- [x] Developer
- [x] Domains
- [x] Education
- [x] Email
- [x] Entertainment
- [x] Finance
- [x] Food
- [x] Gaming
- [x] Government
- [x] Health
- [x] Hosting/VPS
- [x] Hotels and Accommodations
- [x] Identity Management
- [x] Investing
- [x] IoT
- [x] Legal
- [x] Other
- [x] Payments
- [x] Post and Shipping
- [x] Remote Access
- [x] Retail
- [x] Security
- [x] Social
- [x] Task Management
- [x] Transport
- [x] Utilities
- [x] VPN Providers

40504:
How can I contribute to this issue? I would like to add svg files

phallobst:
1. Use a new git branch per category, including the changes from #5050 (these are not currently in the _master_ branch).
2. Tick off the category you want to work on in the list above _before_ you start, and be prepared to do the whole category.
3. For each website in your category `foo`...
    1. Find a suitable SVG logo, preferably a (nearly) square one. This could be on the website itself, in a _Press Kit_, an icon font or even embedded in a PDF file offered on that site.
    2. Remove any whitespace around it - we already care about that.
    3. Place it centered in a square viewbox/page, i.e. it should cover the whole width or height (or both if it's actually square).
    4. Save it as "optimized SVG" (e.g. in _inkscape_), or optimize it with something like [svgo](https://github.com/svg/svgo) or [nano](https://vecta.io/nano). This is needed to keep it to a reasonable file size, e.g. no more than 10 KB.
    5. Place the icon in the corresponding `img/foo` directory, remove the old `png` file and update the `_data/foo.yml` file.

Totally simple and not tedious at all... :wink:

If you can't find a usable SVG icon, just leave the entry as it is. We can still work it out later.

Once #5050 is merged, you can rebase your branch to _master_ and create a PR as usual.

PR #5306, opened 2021-03-16
Update Rebel

ApCoder123:
Updated image to svg in process


phallobst (changes requested):
The icon is not centered in its viewbox, so it looks a bit strange.

FYI: You may want to save the effort to update all the icons to SVG. We were already planning to do just that (see #5061) and have completed most of the work by now.

ApCoder123:
Ok, I have fixed this icon
If that is the case with updating svgs, I will probably revert the commits to change to svg on some of my unreviewed PRs to save both of our time. Thanks for letting me know :+1:

Carlgo11 (approved):

phallobst (approved):

PR #5317, opened 2021-03-18
Display Other category at the end

rsapkf:
Since "Other" is not a very specific category, it makes sense to put this at the end (like any other list) instead of somewhere in the middle.


phallobst:
Please see PR #5449 for a solution that leaves the section list as-is and just changes the HTML generator.

PR #5326, opened 2021-03-20
Remove Iniz

ApCoder123:


phallobst:
The Alexa rank is way over 200k nowadays - I reckon this entry should be removed instead.

phallobst (approved):

Carlgo11 (approved):

PR #5330, opened 2021-03-22
Update Paypal

40504:


Carlgo11 (changes requested):
While your PNG is smaller, SVGs are perfered as they allow for perfect resolution in all sizes.
I suggest this one instead:

40504:
> While your PNG is smaller, SVGs are perfered as they allow for perfect resolution in all sizes.
> I suggest this one instead:
> 

Done

Carlgo11 (approved):

phallobst (approved):

PR #5345, opened 2021-03-22
Update Action Network

ApCoder123:
According to the doc, action network supports sms

### DIFF
--- a/_data/communication.yml
+++ b/_data/communication.yml
@@ -4,6 +4,7 @@ websites:
     img: actionnetwork.png
     tfa:
       - totp
+      - sms
     doc: https://help.actionnetwork.org/hc/en-us/articles/217206826
 
   - name: Airship


Carlgo11 (approved):

_data/communication.yml
```diff
@@ -4,6 +4,7 @@ websites:
     img: actionnetwork.png
     tfa:
       - totp
```
phallobst, line 6:
```suggestion
      - proprietary
```
The sentence
> Turn off multi-device in the Authy app for the highest level of security.

suggests to me that it's not actually `totp`, but rather `proprietary`, correct? I can't think of a way for this to work with regular `totp`.

phallobst (dismissed):

_data/communication.yml
ApCoder123: My understanding of [Authy multi-device](https://authy.com/features/multiple-devices/) was that regular TOTP tokens would be stored in the cloud and could then be accessed from anywhere you were logged in, similar to how [1Password supports TOTP](https://support.1password.com/one-time-passwords/). Presumably for this site, you could use another TOTP authenticator, although it is recommending Authy and giving advice on securing Authy account? That's what I understood from it anyways.

phallobst: Okay, let's leave it at is is then.

phallobst (approved):

PR #5357, opened 2021-03-22
Update Aha

ApCoder123:
Pointed doc url to more specific place on page

### DIFF
--- a/_data/developer.yml
+++ b/_data/developer.yml
@@ -14,7 +14,7 @@ websites:
       - phone
       - proprietary
       - hardware
-    doc: https://support.aha.io/hc/en-us/articles/360031319232-Account-settings#SSO
+    doc: https://support.aha.io/hc/en-us/articles/360031319232#enable-2fa-for-your-account
 
   - name: Airbrake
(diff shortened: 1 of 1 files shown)


_data/developer.yml
```diff
@@ -14,7 +14,7 @@ websites:
       - phone
       - proprietary
       - hardware
```
Carlgo11:
```suggestion
      - u2f
```

Carlgo11 (changes requested):
The hardware 2fa seems to be U2F keys

Carlgo11 (approved):

_data/developer.yml
phallobst: I see no mention of hardware-based 2FA (U2F or otherwise) - am I missing something?

ApCoder123: I believe that since all 2FA for Aha is done via Duo, users can use security keys through that proprietary method as well as phone, sms, push notifications. See https://guide.duo.com/security-keys

phallobst (approved):

PR #5365, opened 2021-03-22
Update infobip

ApCoder123:
The current doc link doesn't work and this blog post was the best I could find

### DIFF
--- a/_data/developer.yml
+++ b/_data/developer.yml
@@ -307,7 +307,7 @@ websites:
     img: infobip.png
     tfa:
       - sms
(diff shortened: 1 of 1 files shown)


_data/developer.yml
```diff
@@ -307,7 +307,7 @@ websites:
     img: infobip.png
     tfa:
       - sms
```
Carlgo11, line 309: Sounds like they support 2FA using both SMS and Phone calls.
```suggestion
      - sms
      - phone
```

Carlgo11 (changes requested):

Carlgo11 (approved):

phallobst (approved):

PR #5388, opened 2021-03-23
Add NordLocker

40504:
Screen Shot 2021-03-23 at 01 19 11


40504:
Alexa Rank
Estimate
This site ranks:
86.5 K

# 212,579

In global internet traffic and engagement over the past 90 days

Carlgo11:
The important number is the actual rank (213K) and not the dy/dx of 86.5K.
As the rank is above 200K and thus the build is failing I'm afraid the PR can't be merged at the moment.

Hopefully the rank improves so that the PR can be re-opened but by looking at the graph I don't think that will happen in the near months 😔

PR #5396, opened 2021-03-23
Add Hack The Box

ApCoder123:


Carlgo11 (changes requested):
There are already similar services listed under `education`. Whilst not being a perfect fit I still think it's the better category for now.
If enough similar services gets added we could potentially make it a seperate category in the future.

Carlgo11 (approved):

phallobst:
If only we had keywords instead of categories... :wink:

phallobst (approved):

PR #5413, opened 2021-03-23
Added Glassdoor (UK)

ohheyrj:
Added glassdoor.co.uk which also supports 2FA, docs are the same as .com


Carlgo11:
If it's the same service but with another TLD I think this PR is better off being converted to an issue for when we add support for multi-domain listings.

ohheyrj:
Yup, agree with that :) I'll open an issue.

PR #5439, opened 2021-03-27
Update Compose

ApCoder123:
According to this [blog post](https://www.compose.com/articles/introducing-fido-universal-2nd-factor-authentication/), u2f is supported in addition to sms and totp

### DIFF
--- a/_data/developer.yml
+++ b/_data/developer.yml
@@ -188,6 +188,7 @@ websites:
     tfa:
       - sms
       - totp
+      - u2f
     doc: https://help.compose.com/docs/two-factor-authentication
 
   - name: Cookiebot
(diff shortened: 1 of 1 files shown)


_data/developer.yml
```diff
@@ -188,6 +188,7 @@ websites:
     tfa:
       - sms
       - totp
+      - u2f
     doc: https://help.compose.com/docs/two-factor-authentication
```
phallobst: Since the old page is not up to date, why not use the blog entry as the new (interim) doc link?
```suggestion
    doc: https://www.compose.com/articles/introducing-fido-universal-2nd-factor-authentication/
```

phallobst (changes requested):

phallobst (approved):

Carlgo11 (approved):

PR #5444, opened 2021-03-27
Add American Kennel Club

ApCoder123:


_data/other.yml
```diff
@@ -17,6 +17,13 @@ websites:
       - proprietary
     doc: https://helpx.adobe.com/manage-account/using/secure-your-adobe-account.html
 
+  - name: American Kennel Club (AKC)
```
phallobst: Since the URL already contains "akc", this entry will be found when searching for that, so there's no need to include it in the name.
```suggestion
  - name: American Kennel Club
```

phallobst (changes requested):

phallobst (approved):

Carlgo11 (approved):

PR #5455, opened 2021-03-28
Update Luno

40504:
Phone number is required for activating 2fa:

and when sign in from a new location:

P.S. temp mail address was used

### DIFF
--- a/_data/cryptocurrencies.yml
+++ b/_data/cryptocurrencies.yml
@@ -480,8 +480,12 @@ websites:
     url: https://www.luno.com/
     img: luno.png
     tfa:
+      - sms
+      - email
       - totp
     doc: https://www.luno.com/help/en/articles/1000203420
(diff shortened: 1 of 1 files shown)


Carlgo11 (approved):

_data/cryptocurrencies.yml
```diff
@@ -480,8 +480,12 @@ websites:
     url: https://www.luno.com/
     img: luno.png
     tfa:
+      - sms
+      - email
```
phallobst:
```suggestion
```
The doc page does not mention email or SMS. Can you elaborate on that?

Regarding your screenshot: Does this happen only when you use a new device? It also seems to use a 4 digit code there, instead of the usual 6 digits as seen in the Google Authenticator example video.

phallobst (approved):

PR #5458, opened 2021-03-29
Vultr uses Yubikey U2F now

bluinch:


Carlgo11:
The [doc page](https://www.vultr.com/docs/using-two-factor-authentication-to-login-to-vultr-control-panel) still only mentions Yubikey OTP.

bluinch:
The instructions are a bit unclear. 

It says the when prompted to enter authentication code, all you have to do is insert the yubikey and "Press the Y button, or touch the edge of the device if using a Yubikey Nano". There's no mention of adding the tokens to the Yubico Authenticator app, nor any mention of inputting the six digit code. 

Might be wrong, will have to follow up with someone who has an account.

Carlgo11:
@7redeagle Yubico OTP isn't the same as TOTP. [Yubico OTP](https://developers.yubico.com/OTP/OTPs_Explained.html) codes can be generated from a Yubikey by pressing (or holding) the button and wait for the Yubikey to write a string of 44 characters containing the ID of the key and an encrypted one time key. It's on all Yubikeys (except the NFC key) by default and doesn't require another app to set up.

The documentation page describes instructions on how to use Yubico OTP (as U2F doesn't require a text field, see 7. in the "2FA via Yubikey" instructions).

If you have any other information confirming that they do support U2F, please share that here. If not, I'll close this PR as it doesn't seem like Vultr's 2FA system have changed.

bluinch:
So that's how it works. It seems I was wrong. The system hasn't changed.

PR #5485, opened 2021-04-02
Add GoPro

ApCoder123:
Decided to use the favicon for the icon, as the logo is too rectangular.


phallobst (changes requested):
I suggest to go with the normal "too rectangular" logo instead of just the blue square.

phallobst (changes requested):
That logo doesn't look incredibly appealing either... How about this one (it has a white box at the right, don't worry):

phallobst (approved):

Carlgo11 (approved):

PR #5487, opened 2021-04-02
Add Virgin Pulse

ApCoder123:
Alexa rank 31,007


phallobst (approved):

_data/health.yml
```diff
@@ -230,6 +230,15 @@ websites:
...
 
+  - name: Virgin Pulse
+    url: https://www.virginpulse.com/
+    img: virginpulse.png
+    tfa:
+      - sms
+      - email
+    doc: https://virginpulse.zendesk.com/hc/en-us/articles/360033807991
+    exception: "SMS 2FA is only available for US and Canada phone numbers."
+
```
Carlgo11: Seems like it should be limited to U.S. use only.
virginpulse

```suggestion
    regions:
      - us

```

Carlgo11 (changes requested):

_data/health.yml
ApCoder123: That's strange - when I visit the site, I get redirected to a UK version. There also seems to be regional sites for various other countries. e.g. https://www.virginpulse.com/de/

And they do seem to have various global contacts (https://www.virginpulse.com/en-gb/contact/): 
image

The T&C's might not be fully accurate?

Carlgo11: I emailed them and asked. With no GDPR part in their privacy policy and no mention of any SCC I get the feeling they may not yet be cleared to operate in the EU, EEA, UK :thinking:

phallobst: Is this supposed to be our problem? We can hardly monitor for changes in privacy policies or the like and update our data accordingly.That would create a maintenance nightmare.

ApCoder123: I've had a look, and they do seem to mention GDPR and SCC in their [privacy policy](https://www.virginpulse.com/en-gb/privacy-notice/) now:
image

They also mention that they are available in 190+ countries on their [About Us](https://www.virginpulse.com/en-gb/about-us/) page:
image

It seems that they do operate globally, so no need for a regions array.

phallobst:
There is an SVG logo available:

ApCoder123:
Decided to just use the 'P', like in the favicon; as even optimised, the full logo was over 10KB in size.

Carlgo11 (approved):
Couldn't manage to get in touch with them. Merging now...

PR #5489, opened 2021-04-02
Update Wirex

ApCoder123:
According to the docs

> Сlick Activate.
> You will receive an email with the confirmation link. Open the email and confirm the 2FA for your account. The link will be active for 15 minutes.

so have added an exception

### DIFF
--- a/_data/cryptocurrencies.yml
+++ b/_data/cryptocurrencies.yml
@@ -575,6 +575,7 @@ websites:
     tfa:
       - totp
     doc: https://wirexapp.com/hc/articles/207160259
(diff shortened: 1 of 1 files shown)


phallobst:
Since opening the account itself already requires an email confirmation, this exception is probably not needed.

ApCoder123:
What I understood from the site was that upon enabling 2FA, you would get a separate confirmation email to enable 2FA (even if your email is already confirmed after opening the account).

phallobst:
That might be true, but is no reason for us to list this as an exception text here.

ApCoder123:
Ok, I understand now.

PR #5497, opened 2021-04-02
Add Current RMS

ApCoder123:
Although the docs make it seem like only Authy can be used, according to the enable page

> If you are already using another authenticator app, we will provide a QR code to scan on the next screen.

image


phallobst (changes requested):
This SVG version of the icon seems to be better:

phallobst (approved):

Carlgo11 (approved):

Carlgo11:
@ApCoder123 can you fix the merge conflict please?

PR #5501, opened 2021-04-03
Add Quidco

ApCoder123:


phallobst (changes requested):
I don't think this fits into "Retail". It's similar to "AwardWallet" in the "Other" category.

ApCoder123:
Ok, category updated.

phallobst (approved):

Carlgo11 (approved):

PR #5504, opened 2021-04-03
Add Perimeter81

ApCoder123:
Although not explicitly mentioned, phone and u2f should be available as Duo Security is supported as a 2FA method (https://guide.duo.com/security-keys).


phallobst (changes requested):
This icon has a transparent background:
perimeter81

phallobst (approved):

Carlgo11 (approved):

PR #5517, opened 2021-04-03
Add Pepper Communities

ApCoder123:
Shopping communities all run by pepper and all use the same auth system. All sites are in top 200k.
See https://www.pepper.com/#our-communities for a list


Carlgo11 (approved):

phallobst (approved):
May this entry could use the "additional-domains" feature once the switch to JSON is complete.

PR #5546, opened 2021-04-05
Add docker files

Carlgo11:
Adds an option to test the site locally with the same scripts and environment that the GH action uses.

### DIFF
--- a/_deployment/minify-js.sh
+++ b/_deployment/minify-js.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 for file in _site/js/*.js; do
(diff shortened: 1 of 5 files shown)


_deployment/minify-js.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
phallobst, line 1:
```suggestion
#!/bin/sh
PATH="./node_modules/.bin:$PATH"
```
Now this should also work for people that have installed the node modules locally.

phallobst (changes requested):

phallobst (approved):

Carlgo11:
Closing in favor of #5695

alphachart (commented):
.

PR #5549, opened 2021-04-06
Fix Bug: Broken Link in noscript tag

ApCoder123:
At the moment, the link to the GH repo in the noscript tag doesn't do anything and when the page is built, this is the HTML produced:
`<a href="" style="color:#f5f5f5"> GitHub repository</a>` 

as `site.repo` doesn't seem to be defined. Changing it to `site.github.repository_url` like in the GitHub icon in the corner should fix this.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -68,7 +68,7 @@
           <p>Without JavaScript, this page probably won't work very well.
                   For more information, or if you would like to talk to us about this,
                   please take a look at our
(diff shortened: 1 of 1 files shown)


phallobst (approved):

Carlgo11 (approved):
It probably was meant to say `site.repository` as it is in the [config](https://github.com/2factorauth/twofactorauth/blob/master/_config.yml#L9).
`site.github.repository_url` works as well 👍

PR #5572, opened 2021-04-07
Add Demio

ApCoder123:
Decided to use PNG as the only SVG logo I could find was quite large (in file size) and messy with curves.
SVG Logo


Carlgo11 (approved):
That SVG is pure nightmare fuel 😨

40504:
try this one svg

phallobst (approved):

PR #5577, opened 2021-04-07
Remove border radius on logos

Carlgo11:
As @phallobst has pointed out, the logos look fine without the border radius and SVGs are often better without it.


phallobst (approved):
Some logos have a white square background without the border. I guess it's not really worth it to update them all, since the new repo already has SVG icons for most of them.

ApCoder123 (approved):

PR #5579, opened 2021-04-08
Adding exception info to kickstarter in others.yml

jgfam69qy2:
Info received as response from kickstarter's support team in a ticket.
image

### DIFF
--- a/_data/other.yml
+++ b/_data/other.yml
@@ -373,7 +373,7 @@ websites:
       - phone
       - totp
     doc: https://help.kickstarter.com/hc/en-us/articles/115005127094
-    exception: "Currently requires a verified SMS Number as fallback for Software 2FA."
+    exception: "Software 2FA requires setting up SMS 2FA as a fallback."
 
   - name: Liftopia
     url: https://www.liftopia.com
--- a/_data/other.yml
+++ b/_data/other.yml
@@ -373,6 +373,7 @@ websites:
       - phone
       - totp
     doc: https://help.kickstarter.com/hc/en-us/articles/115005127094
+    exception: "Currently requires a verified SMS Number as fallback for Software 2FA."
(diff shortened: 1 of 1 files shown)


_data/other.yml
```diff
@@ -373,6 +373,7 @@ websites:
       - phone
       - totp
     doc: https://help.kickstarter.com/hc/en-us/articles/115005127094
+    exception: "Currently requires a verified SMS Number as fallback for Software 2FA."
```
ApCoder123:
```suggestion
    exception: "Software 2FA requires setting up SMS 2FA as a fallback."
```

I think that this reads slightly better?

It sounds more similar to existing exception text as well: E.g. for Dropbox "Hardware 2FA requires enabling Software or SMS 2FA as a back-up."

ApCoder123 (changes requested):

ApCoder123 (approved):

_data/other.yml
jgfam69qy2: Thank you. 
It is much clearer that way.

phallobst (commented):
Do they require to actually enable SMS 2FA, or just to supply a phone number? Or is SMS 2FA available anyway as soon as the phone number is given?

jgfam69qy2:
> Do they require to actually enable SMS 2FA, or just to supply a phone number? Or is SMS 2FA available anyway as soon as the phone number is given?

I went through the process
There seems to be no option to use ONLY SMS or TOTP as it verifies both.

Effectively both 2nd factors end up enabled (see "Send code to my Phone" at the bottom of the snapshot).

image

Furthermore, SMS cannot be disabled, only updated, and disabling OTP disables 2FA altoghether

image

phallobst (approved):

PR #5580, opened 2021-04-08
Add Charles Schwab under Banking

hon1nbo:
Added Charles Schwab under Banking.

I did *not* include `sms` as it has in investing since I couldn't find SMS as an actually valid option in the current state of things when I tried referencing it in my account. It may exist, but it isn't the default they provide information to current users about and would need validation.

Cheers,
~H


hon1nbo:
I'll have build environment setup in the future to avoid the rapid-fire commit; new machine doesn't have the dev tools templated out yet for one-offs like this

ApCoder123:
Hi @hon1nbo 
Thanks for your PR.

Currently, we don't allow entries of the same site across multiple categories. There have already been some PRs to add Charles Schwab bank (see the discussion in #3244). If the login systems for investing and banking are separate, we may be able to add this entry (although they seem to use the same based on the discussion in the referenced PR). Otherwise, feel free to submit a PR updating the methods/doc page for the current Charles Schwab entry.

Thanks!

hon1nbo:
Interesting; I know people don't bother checking the other tabs when they're looking for a provider. What prompted this PR was just that earlier today when someone was trying to find a primary bank that supported hardware tokens / TOTP methods.

From a user perspective it's extremely misleading when they're looking for providers with known 2FA options. What's the rationale for not cross-listing when they fit in multiple categories?
For example in the USA there are fairly few banks that support TOTP/hardware/soft tokens, and there are definitely users trying to find answers to services that can provide that and using sites such as this as reference.

ApCoder123:
There was a discussion about cross-listing in #3707 and it could potentially be available in a future release, but I'm probably not the right person to ask about this. Maybe @Carlgo11 might be able to give you a better answer

hon1nbo:
Self-Closing since it requires an overall improvement for multi-category listing.

PR #5593, opened 2021-04-11
Remove Pozible

40504:


phallobst:
:thinking: Its Alexa rank is at 215k right now, with a general upward trend. Let's give it a week or two and see where we are.

phallobst:
It's below 200K now, still with an upward trend - we can close this PR.

PR #5609, opened 2021-04-14
Update Twitter

svensven:
Now supports multiple hardware tokens:
https://twitter.com/TwitterSupport/status/1371510300973875201

### DIFF
--- a/_data/social.yml
+++ b/_data/social.yml
@@ -299,7 +299,7 @@ websites:
       - totp
       - u2f
     doc: https://support.twitter.com/articles/20170388
-    exception: "Mobile apps do not support Hardware 2FA. Hardware 2FA requires enabling back-up method. Cannot add multiple security keys. SMS only available on select providers."
+    exception: "Mobile apps do not support Hardware 2FA. Hardware 2FA requires enabling back-up method. SMS only available on select providers."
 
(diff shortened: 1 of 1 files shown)


_data/social.yml
```diff
@@ -299,7 +299,7 @@ websites:
       - totp
       - u2f
     doc: https://support.twitter.com/articles/20170388
-    exception: "Mobile apps do not support Hardware 2FA. Hardware 2FA requires enabling back-up method. Cannot add multiple security keys. SMS only available on select providers."
+    exception: "Mobile apps do not support Hardware 2FA. Hardware 2FA requires enabling back-up method. SMS only available on select providers."
```
Carlgo11: Thanks for the heads up! Let's also remove this bit as it's also outdated 😉 
```suggestion
    exception: "Hardware 2FA requires enabling back-up method. SMS only available on select providers."
```

Carlgo11 (approved):

_data/social.yml
svensven: Ah, I wasn't paying attention to that bit, and I don't use the app anyway, so couldn't confirm the current state.

ApCoder123 (approved):

PR #5618, opened 2021-04-17
Add Alexa

ApCoder123:


phallobst (changes requested):
Icon in SVG format:

phallobst (approved):

Carlgo11 (approved):
Would be funny if Alexa didn't pass the Alexa rank test

PR #5626, opened 2021-04-20
Add Deel

40504:


ApCoder123 (changes requested):
This might be better suited to the "Finance" category. It seems quite similar to sites like Gusto or JustWorks which are already listed there.

ApCoder123 (approved):

phallobst (approved):

PR #5629, opened 2021-04-23
Added link to NECU online security describing MFA procedure.

vgarmash:
NECU supports MFA procedure based on watermark picture and challenge questions. They described the procedure at https://necu.org/business/online-services/online/online-security/.

Since their online banking based on [Episys](https://www.symitar.com/core-solutions/pages/episys.aspx) you can see more detailed manual about it in this document https://seacomm.org/assets/NetTeller-Online-End-User-Guide.pdf which is published by other credit union who uses the same Episys platform.

By the way - this MFA procedure is used by big number of financial institutes in the USA. I think you can consider adding the option to your "tfa" section for such.

### DIFF
--- a/_data/banking.yml
+++ b/_data/banking.yml
@@ -1066,8 +1066,9 @@ websites:
   - name: Northeast Credit Union
     url: https://necu.org
     img: necu.png
-    twitter: NortheastCU
-    facebook: northeastcu
+    tfa:
+      - proprietary
+    doc: https://necu.org/business/online-services/online/online-security/
 
   - name: Nubank
     url: https://www.nubank.com.br


Carlgo11:
Hello @vgarmash,
since you mention MFA I assume you know the different authentication factors. Which two (or more) factors are used here?
I see that passwords are mentioned which would be 'something you know' but I don't see any of the other factors, 'something you have' and 'something you are'.

Please describe in more detail what possesion or biometric based systems are used.

vgarmash:
> Hello @vgarmash,
> since you mention MFA I assume you know the different authentication factors. Which two (or more) factors are used here?
> I see that passwords are mentioned which would be 'something you know' but I don't see any of the other factors, 'something you have' and 'something you are'.
> 
> Please describe in more detail what possession or biometric based systems are used.

They do not store any hashed biometric information.

NECU Online Banking uses a challenge-response technology called multi-factor authentication (MFA). There are two pieces to MFA; a personal identification image (sometimes called a watermark) and "challenge questions". The first time you log into Online Banking you will be required to choose an image (picture). This image will display on the password screen when you enter your correct Login ID and on each page within Online Banking to let you know that you are on the genuine Online Banking website. If you do not see your Personal Identification Image, you should not enter your password. You will also be prompted to select three "challenge" questions and answers. Challenge questions are an additional way we confirm your identity.

This procedure is implemented by many credit unions and banks in USA. Here I found couple examples:

1. https://mycitizens.bank/security-center/prevention
2. https://www.firstcapitalfcu.com/education/member-security/security-faqs.html
3. https://www.edfed.org/resources/faqs/account-security/
4. https://www.sugarriverbank.com/e-banking/security
5. https://www.communityalliancecu.org/anywhere-banking/online-banking

Carlgo11:
That's not MFA.

vgarmash:
> That's not MFA.

If you say so then it doesn't. How would you educate those who numerous banks and credit unions who call it MFA?

About this PR though: should I close it or you would do it?

Carlgo11:
As the password is a knowledge [factor](https://en.wikipedia.org/wiki/Multi-factor_authentication#Factors) and the security question (in this case a picked image among other images) is also a knowledge factor, the process doesn't require authentication through multiple different factors, thus it isn't multi-factor authentication.
For it to be multi-factor authentication at least two different factors would have to be required.

We currently don't have any particular way of telling a user that a company is lying about offering multi-factor authentication; We just don't mark them as having it.

I'll close the PR now :wink:

PR #5635, opened 2021-04-24
Add Talkdesk

ApCoder123:


phallobst (approved):
Although it's their favicon, the logo does not seem to be up-to-date anymore. However, I don't see an alternative that's obviously better right now.

Carlgo11 (approved):

PR #5637, opened 2021-04-24
Add Seneca

ApCoder123:


_data/education.yml
```diff
@@ -513,6 +513,18 @@ websites:
     regions:
       - se
 
+  - name: Seneca Learning
+    url: https://senecalearning.com
+    img: seneca.svg
+    twitter: SenecaLearning
+    facebook: senecalearning
```
Carlgo11, line 520: `@senecalearning` seems to be only for India while `@SenecaLearn` is global
```suggestion
    facebook: SenecaLearn
```

Carlgo11 (changes requested):

_data/education.yml
ApCoder123: I changed the Twitter handle, but not the Facebook one, as `@SenecaLearn` doesn't seem to exist for Facebook

Carlgo11 (approved):

phallobst (approved):

PR #5648, opened 2021-04-26
Per current ancestry.com documentation, added Email 2FA flag

betsythefc:


_data/other.yml
```diff
@@ -22,6 +22,7 @@ websites:
     img: ancestry.png
     tfa:
       - sms
+      - email
     doc: https://support.ancestry.com/s/article/MFA
     exception: "2FA is currently only available for customers with a U.S. based phone number."
```
phallobst:
```suggestion
    exception: "SMS 2FA is currently only available for customers with a U.S. based phone number."
```

phallobst (changes requested):

_data/other.yml
betsythefc: I re-read the documentation, and it doesn't look like you need to contact customer support to you email:

The steps are under the heading "What to do if you don’t receive the code", and I was able to enable it without contacting customer support.

betsythefc: I agree with putting SMS in front of that exception.

phallobst: At the top of the doc page, it says
> If you don't have a mobile phone, you can enable it for email by giving us a call.

Is this information outdated?

betsythefc: That's interesting. It must require the mobile phone first to enable email delivery.

betsythefc: It looks like this information is out of date. Without a phone number, I was able to enable 2FA, and it gave me the option to select a phone number or an email.

phallobst: Okay, I updated my suggestion accordingly.

phallobst (approved):

ApCoder123 (approved):

PR #5649, opened 2021-04-27
Add Termly

ApCoder123:


_data/legal.yml
```diff
@@ -76,3 +76,13 @@ websites:
...
+
+  - name: Termly
+    url: https://termly.io
+    img: termly.svg
+    email_address: support@termly.io
+    twitter: Termly_io
+    facebook: termly.io
+    regions:
+      - gb
+      - us
```
phallobst: Why are the regions restricted to `gb` and `us`?

ApCoder123: I noticed a US/GB country switcher in the top right corner and assumed it meant that the site was only available in those regions, however it seems that the service is actually available globally, so I have removed the regions array.

phallobst (approved):

Carlgo11 (approved):

Carlgo11:
Yikes. Looks like Twitter have changed their access to their API :disappointed: 
The PR looks fine though :+1:

PR #5654, opened 2021-04-27
Add cPanel

40504:


phallobst:
Does 2FA apply to the actual account at https://cpanel.net, or is this just a feature of the installation at the hosting service?

40504:
> Does 2FA apply to the actual account at https://cpanel.net, or is this just a feature of the installation at the hosting service?

I don't know

PR #5662, opened 2021-04-28
Update Flexera

40504:


_data/cloud.yml
```diff
@@ -64,9 +64,10 @@ websites:
 
   - name: Flexera
     url: https://www.flexera.com/
-    img: flexera.png
-    twitter: flexera
-    facebook: flexera
+    img: flexera.svg
+    tfa:
+      - totp
```
phallobst:
```suggestion
      - totp
      - sms
```

phallobst (changes requested):

40504:
If services support multiple 2FA options, like in Flexera case, TOTP, and SMS, it's not better `doc` to link to https://docs.flexera.com/svr/ug/Content/helplibrary/Configure_Two_Factor_Authentication__2FA_.htm instead of https://docs.flexera.com/svr/ug/Content/helplibrary/Token_Based_Two_Factor_Authentication.htm

phallobst:
The link to "Configure Two-Factor Authentication" seems to be the better choice here.

40504:
> The link to "Configure Two-Factor Authentication" seems to be the better choice here.

done

ApCoder123:
From the doc link provided, it appears that 2FA only applies to their [Software Vulnerability Research](https://www.flexera.com/products/software-vulnerability-research.html) product. I can't find any reference to 2FA for the main Flexera account. If this is the case, then it may be better to leave the current entry how it is and create a separate entry in the security category? It seems to use a separate login system as well - this is the [main login page](https://app.flexera.com/login) while it seems you can only login to the Vulnerability research product [here](https://csi7.secunia.com/csi/). Please could you confirm whether 2FA only applies to this product?

40504:
I can't confirm because don't have an account there. From this page, https://app.flexera.com/login seems like they have multiple services with separate login details

ApCoder123:
Okay, since that is the case, I think that a separate entry should be made for the Vulnerability Research product in the `Security` category, and the current Flexera entry should be kept as it is.

40504:
doc for 2FA is related to Flexera’s Software Vulnerability Research

Carlgo11:
@40504 do you intend on updating the PR with the [suggestions](https://github.com/2factorauth/twofactorauth/pull/5662#issuecomment-830607457) from @ApCoder123?

40504:
At the moment I can't. I will do it as soon as possible.

PR #5668, opened 2021-04-28
Add CleverTap

40504:


phallobst (changes requested):
I think the logo should either be on a red circle instead of a red square, or just the red arc with no backdrop.

40504:
> I think the logo should either be on a red circle instead of a red square, or just the red arc with no backdrop.

done

phallobst (approved):

ApCoder123 (approved):

PR #5674, opened 2021-05-02
Add UC Berkeley

dotlambda:
I'm afraid of infringing on copyright if I include the logo. Is there a way to link to a file?

Should I include `hotp` in the list since you don't need the proprietary DUO app if you use e.g. https://github.com/scintill/DuOTP-android?

### DIFF
--- a/_data/education.yml
+++ b/_data/education.yml
@@ -130,7 +130,7 @@ websites:
     img: cod.png
     regions:
       - us
-      
+
   - name: Colorado School of Mines
     url: https://www.mines.edu/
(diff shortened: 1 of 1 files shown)


Carlgo11 (changes requested):
We don't support external links for the `img` tag. The SVG needs to be placed in the corresponding `img/` subfolder.
While not qualified to give legal advise, most countries have fair use laws which would allow the use of logos here.

Carlgo11:
Closing due to inactivity

PR #5676, opened 2021-05-03
Update signNow

40504:


ApCoder123 (commented):
The doc page describes how to protect a specific document with "_dual factor authentication_". This seems to force the signer to receive a text to the number that the sender has specified before they can sign the document. I don't believe that this meets our definitions of 2FA, as it does not affect the user's sign in process & is not protecting the account with a redundant layer of authentication. Please confirm whether the account has 2FA or just specific documents.

40504:
Just created an account and can confirm:

_data/other.yml
```diff
@@ -606,11 +606,12 @@ websites:
...
-  - name: SignNow
+  - name: signNow
     url: https://www.signnow.com/
-    img: signnow.png
-    twitter: signnow
-    facebook: esignaturesignnow
+    img: signnow.svg
+    tfa:
+      - email
+    doc: https://www.signnow.com/features/setup-dual-factor-authentication
```
ApCoder123:
```suggestion

```
I think we can remove the doc link - it doesn't have any useful info about the 2FA features that you showed in the screenshot.

ApCoder123 (changes requested):

40504:
done

ApCoder123 (approved):

phallobst (approved):

PR #5685, opened 2021-05-06
Add POST

40504:


Carlgo11 (approved):

phallobst:
Does this apply to mypost.lu as well? Maybe it would be good to create two entries (one each in their respective category) or just one that covers all.

40504:
> Does this apply to mypost.lu as well? Maybe it would be good to create two entries (one each in their respective category) or just one that covers all.

I successfully created an account on mypost.lu and there is no 2FA option, and login details that I used for mypost.lu doesn't work for https://webmail.pt.lu

phallobst (approved):

PR #5686, opened 2021-05-06
Add BEE Plugin

40504:


Carlgo11 (commented):
Looks like a better fit for the communications category along with all the other email creation and customer outreach sites.

40504:
> Looks like a better fit for the communications category along with all the other email creation and customer outreach sites.

they use different login details for this https://beefree.io/ and this https://developers.beefree.io/
https://beefree.io/ is suitable for communications category and there is no 2FA option, at least for free users, but for https://developers.beefree.io/ I successfully enabled two-factor authentication.

Carlgo11:
So what does the developer sub domain offer?

Carlgo11:
I'm not certain it should be listed then. We don't list things like [developer.twitter.com](https://developer.twitter.com) and [developers.facebook.com/](https://developers.facebook.com/).
If the main business is to design emails it doesn't make much sense to list their side-business. It will probably confuse more than it would help.

PR #5689, opened 2021-05-07
Add CoinSmart

40504:


_data/cryptocurrencies.yml
```diff
@@ -323,6 +323,15 @@ websites:
       - totp
     doc: https://support.coins.ph/hc/en-us/articles/202604924
 
+  - name: CoinSmart
+    url: https://www.coinsmart.com/
+    img: coinsmart.svg
+    tfa:
+      - totp
+    doc: https://coinspot.zendesk.com/hc/en-us/sections/115000049254
```
ApCoder123:
```suggestion
    doc: https://coinsmart.zendesk.com/hc/en-us/articles/360014948893
```
The doc link provided is for a different website called "CoinSpot". Here is the doc link I think you meant to include.

_data/cryptocurrencies.yml
```diff
@@ -323,6 +323,15 @@ websites:
...
+    regions:
```
ApCoder123: I don't think a regions array is required here - the site seems to allow users from different countries to register on the sign up page and the terms of service mention that users outside of Canada can sign up.

ApCoder123 (changes requested):

ApCoder123 (approved):

phallobst (approved):

PR #5107, opened 2020-11-28
Rename and update Dome9

phallobst:

PR #5714, opened 2021-05-22
Rename Dome9 to Check Point

phallobst:
Supersedes #5107


ApCoder123 (commented):
You forgot to remove the Dome9 entry and image 😉.

_data/security.yml
```diff
@@ -61,6 +61,13 @@ websites:
       - hardware
     exception: "Requires X.509-capable cryptotoken or smartcard."
 
+  - name: Check Point
```
ApCoder123, line 64:
```suggestion
  - name: Check Point CloudGuard Dome9
```

2FA only seems to apply to this specific product (I believe it has a separate auth system as well https://secure.dome9.com/v2/login rather than the main Check Point account at https://accounts.checkpoint.com/)

ApCoder123 (changes requested):

PR #5715, opened 2021-05-22
Add HUK-Coburg

phallobst:
Supersedes #4260


Carlgo11:
Is their Facebook page maybe only available for some people?

phallobst:
It's probably a country restriction. :frowning_face: Maybe there's a chance to detect this and not fail in those cases - I have no idea.

Carlgo11:
There's no way to validate pages with restrictions I'm afraid :(

ApCoder123 (approved):

Carlgo11 (approved):

PR #5722, opened 2021-05-29
Fix category display

phallobst:
Some categories were missing due to incorrect modulo/offset handling.


Carlgo11 (approved):
Thanks for the fix!
The offset system is really complicated :sweat:

ApCoder123 (approved):

PR #5727, opened 2021-05-30
Fix tests

Carlgo11:
Some tests didn't fail because the status variable didn't change as expected. This PR fixes that along with entries causing subsequent test failures


phallobst (approved):

tests/validate-json.rb
```diff
@@ -11,22 +11,22 @@
...
 
   unless schema.valid?(document)
-    puts '================================='
-    puts 'Document not valid:'
+    puts ''
     schema.validate(document).each do |v|
-      puts "- file: #{file}"
-      puts "  error : #{v['type']}"
+      puts "::error file=#{file}:: #{v['type']} error in in #{file}"
+      puts "- tag: #{v['data_pointer'].split('/').last}"
```
ApCoder123: Looking at the [test output from 076f6ae](https://github.com/2factorauth/twofactorauth/runs/2706348385) from this PR,

```
Error:  pattern error in in entries/l/logicmonitor.com.json
- tag: 0
  data: phone
  expected: ^(sms|call|email|u2f|totp|custom-software|custom-hardware)$
Error: Process completed with exit code 1.
```

The `tag: 0` isn't particularly useful - I know that the way the slashes are being split will work for non-array types in the schema, but for items in arrays, wouldn't it be more helpful to display the tag in the format `parent/index`. 
In this case, `-tag: tfa/0`. 

Should a check for array types (or just if `v['data_pointer']` contains one specific strings of the properties e.g. `tfa`, `custom-.`) be put in place before outputting the tag to help those looking at test outputs, or is the current output obvious enough on what is causing the test to fail?

Carlgo11: @ApCoder123 would you prefer the entire v['data_pointer'] value?
It would look like this then:
```
Error: Pattern error in in entries/l/logicmonitor.com.json
- tag: /LogicMonitor/tfa/0
  data: phone
  expected: ^(sms|call|email|u2f|totp|custom-software|custom-hardware)$
```

ApCoder123: The changes in 70acdc6 definitely make it clearer
```
Pattern error in in entries/l/logicmonitor.com.json
- tag: tfa
  data: phone
  expected: ^(sms|call|email|u2f|totp|custom-software|custom-hardware)$
```
When running now, it is more obvious that the problem is within the `tfa` tag rather than just an index number like before.  

I reckon that the current output is clear enough now, but I am also fine with displaying the entire path, if you think that is a better option?

Carlgo11: I'm happy with how it works now.

ApCoder123 (approved):

PR #5736, opened 2021-06-02
Add Image Validation Test

ApCoder123:
Commit 51b20ff in #5733 passed tests even though it referenced an image that does not exist, so added a test to check whether images referenced exist for an entry.

### DIFF
--- a/.github/workflows/repository.yml
+++ b/.github/workflows/repository.yml
@@ -26,6 +26,8 @@ jobs:
         run: bundle exec ruby ./tests/entry-tags.rb
       - name: Validate file extensions
         run:  tests/file-extensions.sh
+      - name: Validate Images
+        run: bundle exec ruby ./tests/validate-images.rb
 
   build:
     name: Minimal build
--- a/tests/validate-images.rb
+++ b/tests/validate-images.rb
@@ -4,8 +4,6 @@
 require 'json'
 status = 0
 
-accepted_extensions = %w[.png .svg]
-
 seen_sites = []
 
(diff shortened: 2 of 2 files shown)


Carlgo11:
If the path variable is stored in an array, a check for orphaned images could also be done at the end. As long as the `img/icons` folder is ignored that is 😬

ApCoder123:
The tests now:
- Check an image exists for each entry (uses `img` tag or finds relevant svg)
- Check images folder to ensure there are no unused images
- Checks file extensions of all images to ensure they are valid, as the current extension test does not do this for images
https://github.com/2factorauth/twofactorauth/blob/8193e81a1fde10b5a47591329f202650e523383c/tests/file-extensions.sh#L22
- Allows more than 1 error to be outputted at a time

Carlgo11:
Thanks Arun!
I did some minor changes to optimize the code using Rubocop. Ruby is a weird language and some best practices are unlike most other languages.

Regarding the file extensions, won't the JSON schema catch that before this script?

ApCoder123:
Yeah, the JSON schema will catch bad file extensions if specified in `img` in the entry. The only case where it won't if is there is a file that isn't specified, but then the unused image file check will catch that,
```
::error file=img/e/example.gif:: Unused image at img/e/example.gif
::error file=img/e/example.gif:: Invalid file extension for img/e/example.gif. Only [".png", ".svg"] are allowed
```
so the file extension checking isn't necessary. I've removed it from the checks.

Carlgo11 (approved):

phallobst (approved):

PR #5737, opened 2021-06-03
Update Pepper

ApCoder123:

PR #5246, opened 2021-02-25
Update netcup

JensSpanier:
Three changes: 

- Update casing (it's **n**etcup, not **N**etcup)
- Update logo ([changed in 2019](https://www.netcup-news.de/2019/07/09/redesign-aller-websites-und-controlpanel/))
- Update domain (user interface is now only availabe at [customercontrolpanel.de](https://www.customercontrolpanel.de/), [also changed in 2019](https://www.netcup-news.de/2019/02/05/netcup-stellt-neues-customer-control-panel-vor/))

PR #5738, opened 2021-06-03
Allow searching for additional domains

ApCoder123:


Carlgo11 (approved):

phallobst:
Is it really useful to search for additional domains as long as they are not displayed? It might be confusing if you had a random or seemingly unrelated search result.

ApCoder123:
I see what you mean. For example in #5246, it mentions an alternate domain of `customercontrolpanel.de` and I see how it would be confusing if you searched "customer" (e.g. looking for customer.io) and a seemingly unrelated hosting provider appeared. 

However, in some cases (e.g. #5737) where you would be more likely to search for an additional domain than the main domain, having the ability to search for additional domains in the results would be useful. 
Additional domains searching could also be useful for when a domain name is changed. For example, in #5739, some users would be more likely to search for the old domain "teespring" (which would come up in results if additional domains are searched for), rather than the new name "spring".


I'm not quite sure what the best option is here 🤔

phallobst:
Yeah, the benefits probably outweigh the drawbacks.

phallobst (approved):

PR #5170, opened 2020-12-26
Add Teespring

alexander-dejeu:

PR #5739, opened 2021-06-04
Add Teespring

40504:
reopen #5170


40504:
sorry for the question but when to use "domain" and "url" together?

phallobst:
You only have to specify a "url" value if "https://**domain**" does not work, or if you want to address a certain sub-domain.

phallobst (approved):

ApCoder123 (commented):
At the top of the page there is a notice that

> Teespring is now SPRING. Learn more
image

It seems that the main name for the site is now Spring, with main domain `spri.ng`, but the login page still appears to be at `teespring.com/login`, but uses Spring branding. 

[This documentation on the Spring site](https://www.spri.ng/blog/two-factor-authentication-guide) also seems much more up to date for the current login page (teespring.com/login).

I think it would be better to have the name and domain for Spring, using the more recent documentation as that seems to be the most up to dateand then have an additional domains array with teespring
```JSON
"additional-domains": [
  "teespring.com"
]
```

ApCoder123 (approved):

PR #5745, opened 2021-06-05
Move "Other" category to the end

phallobst:
Having the categories as an array made the handling much simpler.
The sorting order is now determined by the title instead of the name.
Currently it's no difference, but there might be in the future.

Supersedes #5449

### DIFF
--- a/_data/categories.json
+++ b/_data/categories.json
@@ -1,134 +1,167 @@
-{
-  "backup": {
+[
+  {
+    "name": "backup",
     "title": "Backup and Sync",
     "icon": "fas fa-cloud-upload-alt"
   },
-  "banking": {
+  {
+    "name": "banking",
     "title": "Banking",
     "icon": "fas fa-dollar-sign"
   },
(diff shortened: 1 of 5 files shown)


Carlgo11:
I like the `other` category at the bottom but I'm not sure about the categories.json changes.
I think it's more important to have an easy to read categories.json over the messy Jekyll code. Having the category names as keys makes it easier to work with in my opinion. :thinking:

phallobst:
The problem is that Liquid does not provide a way to dynamically create a collection (as far as I know), but only properly supports arrays. As soon as you need a different sorting order, you are basically screwed.

I could revert this to use key/value pairs as before, but then `categories.json` would need to be in display order, i.e. `other` at the end, and not alphabetical. Is that okay?

Carlgo11:
Yeah nah this is fine then.

Carlgo11 (approved):

ApCoder123 (approved):

PR #5760, opened 2021-06-10
Fix padding on dark theme

ApCoder123:
There is padding that is applied to the category buttons (only on the dark theme), which causes search results to be pushed far down the page. This padding is not applied to the light theme, so removing it doesn't really affect the spacing of the category buttons.

image
image

This is particularly problematic on the mobile table
image
image

When the padding is removed, the search results look much better
image

### DIFF
--- a/_includes/html/category-button.html
+++ b/_includes/html/category-button.html
@@ -2,7 +2,7 @@
 {% assign title = include.category-param[1].title %}
 {% assign icon = include.category-param[1].icon %}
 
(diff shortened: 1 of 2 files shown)


Carlgo11 (approved):

phallobst:
The class id `category-icon-outer` can then be dropped from `_includes/html/category-button.html` as well, I suppose.

phallobst (approved):

PR #5763, opened 2021-06-10
Update loopia

Carlgo11:
Updated Sweden's most expensive domain registrar's listing.
Specified the custom-software solutions offered.
Not sure if the note is still necessary if the custom-software tag is there :thinking:

### DIFF
--- a/entries/l/loopia.se.json
+++ b/entries/l/loopia.se.json
@@ -5,12 +5,14 @@
       "custom-software"
     ],
     "documentation": "https://blogg.loopia.se/2015/03/16/nu-kan-du-logga-in-med-tvafaktorsautentiering-hos-loopia-genom-bankid/",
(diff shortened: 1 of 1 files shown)


phallobst:
If _BankID_ is only available for Swedish residents, the `notes` can be dropped, I guess.

Carlgo11:
Since Loopia only offers 2FA through BankID, and BankID is issued through Swedish banks, and Swedish banks require people to have an ID number, and ID numbers are given to Swedish residents (or in some cases people needing one for work at Swedish companies);
In practice only Swedish residents can use 2FA. It is however not a requirement by Loopia but a result of the third party authentication system they rely on.

phallobst (approved):

ApCoder123 (approved):

PR #5794, opened 2021-06-15
Update First Direct

ApCoder123:
Changes from 12c3bb2. They have updated the logo since that commit was made, so I have updated accordingly.


phallobst (approved):

entries/f/firstdirect.com.json
```diff
@@ -1,14 +1,16 @@
 {
   "First Direct": {
-    "domain": "www1.firstdirect.com",
-    "img": "www1.firstdirect.com.png",
+    "domain": "firstdirect.com",
```
Carlgo11, line 3: Looks like `https://firstdirect.com` isn't set up correctly. It will redirect to either the `www1` or `www2` subdomain but it does this using HTTP.

I suggest using `www2.firstdirect.com` as a `url` tag.
```suggestion
    "domain": "firstdirect.com",
    "url": "www2.firstdirect.com",
```

Carlgo11 (changes requested):

Carlgo11 (approved):

PR #5807, opened 2021-06-15
Add Additional domains to Amazon

### LINKED ISSUE #5072: International Domains for Amazon
My password manager **Bitwarden** reports me entries, where 2FA is available but not set up.
For my Amazon account I get no warning that 2FA is missing.
The reason is that I live in Germany and here the local Amazon Domain is https://www.amazon.de/ (country domain **.de**)
In the Tow Factor Auth List the domain for Amazon is https://www.amazon.com/ (country domain **.com**)

So I suggest that you add all local Amazon URLs to the 2FA list.
Because it is essential to protect the Amazon account with 2FA and I know hat your service is used by many other password managers.

ApCoder123:
Closes #5072
Moves ae1d770, 82312b3

### DIFF
--- a/entries/a/amazon.com.json
+++ b/entries/a/amazon.com.json
@@ -1,7 +1,26 @@
 {
   "Amazon": {
     "domain": "amazon.com",
-    "url": "https://www.amazon.com",
+    "additional-domains": [
+      "amazon.ae",
+      "amazon.ca",
+      "amazon.cn",
+      "amazon.co.jp",
+      "amazon.co.uk",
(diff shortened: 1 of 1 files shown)


Carlgo11 (approved):
Because using one domain for all countries, like the international .com TLD, is too hard :roll_eyes:

person1234-ux:
if we are going to do this shouldn't we do it for a bounch of website that do the same thing

person1234-ux:
ok

phallobst (approved):

PR #5820, opened 2021-06-19
Fix CircleCI Failures

ApCoder123:
Alexa tests are currently failing with 403 error (see #5819). The name of the environment variable in the CircleCI settings is `alexa_access_key` rather than `ALEXA_API_KEY`

Twitter tests are failing with error `Twitter API keys not found or invalid.` (see #5823), as the names of the environment variables are wrong and the changes from #5697 haven't been included.


Carlgo11:
Fixed in 453834d

PR #5829, opened 2021-06-19
Remove Vimla

40504:
Alexa Rank

#2,153,760


Carlgo11:
From looking at the Alexa stat graph it looks to be some sort of bug on Alexa's end. From 600K to immediate no activity doesn't look real.

phallobst:
A rank of 600K+ is enough to remove it anyway, isn't it?

Carlgo11:
Ah right. Sorry I read it as 60K 🤪

Carlgo11 (approved):

phallobst (approved):

PR #5831, opened 2021-06-19
Update Talkdesk

40504:


Carlgo11 (approved):
I rounded the edges as they seem to use that on their website

phallobst (approved):

PR #5833, opened 2021-06-19
Update Pixieset

40504:
Cleaning

### DIFF
--- a/entries/p/pixieset.com.json
+++ b/entries/p/pixieset.com.json
@@ -1,12 +1,12 @@
 {
-    "Pixieset": {
-        "domain": "pixieset.com",
-        "tfa": [
-            "totp"
-        ],
-        "documentation": "https://help.pixieset.com/hc/en-us/articles/360046803431",
-        "keywords": [
-            "other"
-        ]
-    }
+	"Pixieset": {
+		"domain": "pixieset.com",
+		"tfa": [
+			"totp"
+		],
+		"documentation": "https://help.pixieset.com/hc/en-us/articles/360046803431",
+		"keywords": [
+			"other"
+		]
+	}
 }
--- a/img/p/pixieset.com.svg
+++ b/img/p/pixieset.com.svg
@@ -1,1 +1,1 @@
-<svg xmlns="http://www.w3.org/2000/svg" height="10" width="10" style="background:#fff"><text font-family="Times New Roman" font-size="8.1" ><tspan x="2.52" y="7.96">T</tspan></text></svg>
+<svg xmlns="http://www.w3.org/2000/svg" height="10" width="10" style="background:#fff"><text font-family="Times New Roman" font-size="8.1" ><tspan x="2.52" y="7.96">P</tspan></text></svg>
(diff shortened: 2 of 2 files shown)


img/p/pixieset.com.svg
```diff
@@ -1 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 11.388 14.949"><text x="-74.459" y="-97.876" font-size="22.578"><tspan x="-.375" y="14.949"><tspan font-family="'Times New Roman'">P</tspan></tspan></text></svg>
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 90 90"><g fill="none" fill-rule="evenodd"><rect width="90" height="90" fill="#FFF"/><text fill="#000" font-family="TimesNewRomanPSMT, Times New Roman" font-size="68"><tspan x="28.091" y="69">P</tspan></text></g></svg>
```
Carlgo11: What's the reason for this change?
It's both larger and inaccurate as the font-family used by Pixieset for their icon is `'Times New Roman',Times,serif`. They also don't use any padding on their icon themselves so I don't see any reason for doing that here.

entries/p/pixieset.com.json
```diff
@@ -1,12 +1,12 @@
...
+		"domain": "pixieset.com",
+		"tfa": [
+			"totp"
+		],
+		"documentation": "https://help.pixieset.com/hc/en-us/articles/360046803431",
+		"keywords": [
+			"other"
+		]
+	}
+}
```
Carlgo11, line 12: Like I said in another PR, this isn't something worth fixing IMO. As long as it is valid JSON and complies with the JSON schema it's fine as it is.

When you create a new JSON file or do major reworks of an existing JSON file you can do the correct space indentation. That way we get more time for things like PRs with site additions :wink:

Carlgo11 (changes requested):

entries/p/pixieset.com.json
40504: > Like I said in another PR, this isn't something worth fixing IMO. As long as it is valid JSON and complies with the JSON schema it's fine as it is.
> 
> When you create a new JSON file or do major reworks of an existing JSON file you can do the correct space indentation. That way we get more time for things like PRs with site additions 😉

OK

img/p/pixieset.com.svg
40504: I think adding a background for the dark logo is enhances visibility. Moreover, when the logos from 2fa.directory are used in the password managers, specifically on iOS devices with dark mode on, the dark logo can't be seen.

This icon it's exactly as their twitter picture profile. 
https://twitter.com/pixiesetapp

Carlgo11: Okay let's do this then:
```suggestion
<svg xmlns="http://www.w3.org/2000/svg" height="10" width="10" style="background:#fff"><text font-family="Times New Roman" font-size="8.1" ><tspan x="2.52" y="7.96">T</tspan></text></svg>
```

Carlgo11 (approved):

phallobst (approved):

PR #5839, opened 2021-06-20
Remove Campfire

40504:


Carlgo11 (changes requested):
If they've retired Campfile I don't think there's much use to still have it be listed here.
Probably better to just remove the listing completely.

40504:
https://basecamp.com/retired/campfire

phallobst (approved):

ApCoder123 (approved):

Carlgo11 (approved):

PR #5842, opened 2021-06-20
Update Internet.bs

40504:

PR #5826, opened 2021-06-19
Add Localize

ApCoder123:

PR #5846, opened 2021-06-20
Add File Name Checking to JSON Validation Tests

ApCoder123:
Adds a short test to make sure that the file name matches the `domain` field in thatfile. 
For example, this test would fail in cases such as #5842 and #5826 where the file name does not match the domain.


tests/validate-json.rb
```diff
@@ -30,6 +30,12 @@
     end
     status = 1
   end
+  
+  unless file.split('/')[2].eql? document.values[0]['domain'] + '.json'
+    puts "::error file=#{file}:: File name should be the same as the domain name.
+    Recieved: #{file.split('/')[2]}. Expected: #{document.values[0]['domain'] + '.json'}"
```
phallobst:
```suggestion
    Received: #{file.split('/')[2]}. Expected: #{document.values[0]['domain'] + '.json'}"
```

phallobst (changes requested):

Carlgo11 (approved):
Should we use this instead of the domain tag then?

phallobst (approved):

PR #5847, opened 2021-06-20
Update RBC Royal Bank - Add RBC Caribbean

40504:
I'm not sure what type of 2fa is used, please check the documentation page :
https://www.rbcroyalbank.com/caribbean/digital-hub/security-two-factor.html


ApCoder123 (changes requested):
From the links provided, it looks like 2FA for RBC is only available in the Caribbean. 

Given that this is the case, I think it makes sense to break the RBC entries into one without 2FA and RBC Caribbean with 2FA, similar to this:
image

- Please create a new entry under `entries/c/caribbean.rbcroyalbank.com`
- Make sure to add the region codes for the countries listed below.
- image
```JSON
"regions": [
  "ag",
  "aw",
  "bs",
  "tc",
  "bq",
  "ky",
  "cw",
  "dm",
  "gd",
  "ms",
  "kn",
  "lc",
  "sx",
  "vc",
  "tt"
]
```

- The 2FA type would be `custom-software` based on the description of what is needed to create a Soft Token:

> The Advice Centre representative will provide you with the information needed to setup your Soft Token including:
> Name: RBC User ID
> Serial Number
> Activation Code

```JSON
"custom-software": [
  "Entrust Identity Guard"
]
```

- It looks like `custom-hardware` is also available, but only for business customers (There should be a `notes` field to mention this)

Thanks.

entries/c/caribbean.rbcroyalbank.com.json
```diff
@@ -0,0 +1,26 @@
+{
+  "RBC Caribbean": {
+    "domain": "caribbean.rbcroyalbank.com",
+    "img": "rbcroyalbank.com.svg",
+    "tfa": [
+      "custom-software"
+    ],
+    "keywords": [
```
ApCoder123:
```suggestion
    "regions": [
```

entries/c/caribbean.rbcroyalbank.com.json
```diff
@@ -0,0 +1,26 @@
...
+      "cw",
+      "dm",
+      "gd",
+      "ms",
+      "kn",
+      "lc",
+      "sx",
+      "vc",
+      "tt"
+    ]
```
ApCoder123, line 33:
```suggestion
    ],
    "keywords": [
      "banking"
    ]
```

entries/c/caribbean.rbcroyalbank.com.json
```diff
@@ -0,0 +1,26 @@
+{
+  "RBC Caribbean": {
+    "domain": "caribbean.rbcroyalbank.com",
+    "img": "rbcroyalbank.com.svg",
+    "tfa": [
+      "custom-software"
```
ApCoder123:
```suggestion
      "custom-software",
      "custom-hardware"
```

entries/c/caribbean.rbcroyalbank.com.json
```diff
@@ -0,0 +1,26 @@
...
+    ],
```
ApCoder123, line 8:
```suggestion
    ],
    "custom-software": [
      "Entrust Identity Guard"
    ],
    "notes": "Hardware 2FA methods are only available for business customers",
    "documentation": "https://www.rbcroyalbank.com/caribbean/digital-hub/security-two-factor.html",
```

ApCoder123 (changes requested):

ApCoder123 (approved):

phallobst (approved):

PR #5854, opened 2021-06-21
Fix git diff bug

Carlgo11:
This PR removes the white space on git diff lines. This in turn fixes recent build failures that where due to `cut` being incorrectly used to remove white space.

I also reformatted  twitter.rb.

### DIFF
--- a/tests/alexa.rb
+++ b/tests/alexa.rb
@@ -42,15 +42,13 @@ end
 status = 0
 # Fetch changes
 # rubocop:disable Layout/LineLength
-diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"domain\\":" | sed -e 's/ //g;s/"//g;s/+domain://g'`
+diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"domain\\":" | sed -n 's/.*"domain"[^"]*"\\(.*\\)".*/\\1/p'`
 # rubocop:enable Layout/LineLength
 # Strip and loop through diff
 diff.gsub("\n", '').split('",').each do |site|
-  begin
-    fetch_from_api(site) unless fetch_from_cache(site)
-  rescue StandardError => e
-    puts "\e[31m#{e.message}\e[39m"
-    status = 1
-  end
+  fetch_from_api(site) unless fetch_from_cache(site)
+rescue StandardError => e
+  puts "\e[31m#{e.message}\e[39m"
+  status = 1
 end
 exit(status)
--- a/tests/alexa.rb
+++ b/tests/alexa.rb
@@ -41,7 +41,9 @@ end
 
 status = 0
 # Fetch changes
-diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"domain\\":" | cut -c17-`
+# rubocop:disable Layout/LineLength
+diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"domain\\":" | sed -e 's/ //g;s/"//g;s/+domain://g'`
+# rubocop:enable Layout/LineLength
 # Strip and loop through diff
 diff.gsub("\n", '').split('",').each do |site|
   begin
--- a/tests/alexa.rb
+++ b/tests/alexa.rb
@@ -41,14 +41,14 @@ end
 
 status = 0
 # Fetch changes
-# rubocop:disable Layout/LineLength
-diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"domain\\":" | sed -n 's/.*"domain"[^"]*"\\(.*\\)".*/\\1/p'`
-# rubocop:enable Layout/LineLength
+diff = `git diff origin/master...HEAD entries/ | sed -n 's/^+.*"domain"[^"]*"\\(.*\\)".*/\\1/p'`
 # Strip and loop through diff
-diff.gsub("\n", '').split('",').each do |site|
-  fetch_from_api(site) unless fetch_from_cache(site)
-rescue StandardError => e
-  puts "\e[31m#{e.message}\e[39m"
-  status = 1
+diff.split('\n').each do |site|
+  begin
+    fetch_from_api(site) unless fetch_from_cache(site)
+  rescue StandardError => e
+    puts "\e[31m#{e.message}\e[39m"
+    status = 1
+  end
 end
 exit(status)
--- a/tests/alexa.rb
+++ b/tests/alexa.rb
@@ -43,7 +43,7 @@ status = 0
 # Fetch changes
 diff = `git diff origin/master...HEAD entries/ | sed -n 's/^+.*"domain"[^"]*"\\(.*\\)".*/\\1/p'`
 # Strip and loop through diff
-diff.split('\n').each do |site|
+diff.split("\n").each do |site|
   begin
     fetch_from_api(site) unless fetch_from_cache(site)
   rescue StandardError => e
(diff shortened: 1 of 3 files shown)


phallobst:
Something like
```rb
domains = `diff ... | sed -n 's/.*"domain"[^"]*"\\(.*\\)".*/\\1/p'`
domains.split("\n").each do |domain|
  ...
end
```
is shorter and faster, and you need the key pattern only once. (same for `facebook` and `twitter`)

tests/alexa.rb
```diff
@@ -41,14 +41,14 @@ def store_cache(site, rank)
 
 status = 0
 # Fetch changes
-diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"domain\\":" | cut -c17-`
+# rubocop:disable Layout/LineLength
+diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"domain\\":" | sed -n 's/.*"domain"[^"]*"\\(.*\\)".*/\\1/p'`
+# rubocop:enable Layout/LineLength
 # Strip and loop through diff
 diff.gsub("\n", '').split('",').each do |site|
```
phallobst:
```suggestion
diff.split('\n').each do |site|
```

tests/alexa.rb
```diff
@@ -41,14 +41,14 @@ def store_cache(site, rank)
 
 status = 0
 # Fetch changes
-diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"domain\\":" | cut -c17-`
+# rubocop:disable Layout/LineLength
+diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"domain\\":" | sed -n 's/.*"domain"[^"]*"\\(.*\\)".*/\\1/p'`
```
phallobst:
```suggestion
diff = `git diff origin/master...HEAD entries/ | sed -n 's/^+.*"domain"[^"]*"\\(.*\\)".*/\\1/p'`
```

tests/facebook.rb
```diff
@@ -5,7 +5,9 @@
 require 'uri'
 
 status = 0
-diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"facebook\\":" | cut -c21-`
+# rubocop:disable Layout/LineLength
+diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"facebook\\":" | sed -n 's/.*"facebook"[^"]*"\\(.*\\)".*/\\1/p'`
```
phallobst:
```suggestion
diff = `git diff origin/master...HEAD entries/ | sed -n 's/^+.*"facebook"[^"]*"\\(.*\\)".*/\\1/p'`
```

tests/facebook.rb
```diff
@@ -5,7 +5,9 @@
...
+# rubocop:enable Layout/LineLength
 diff.gsub("\n", '').gsub(',', '').split('"').each do |page|
```
phallobst:
```suggestion
diff.split('\n').each do |page|
```

tests/twitter.rb
```diff
@@ -11,20 +11,20 @@
 end
 
 status = 0
-diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"twitter\\":" | cut -c20-`
+# rubocop:disable Layout/LineLength
+diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"twitter\\":" | sed -n 's/.*"twitter"[^"]*"\\(.*\\)".*/\\1/p'`
```
phallobst:
```suggestion
diff = `git diff origin/master...HEAD entries/ | sed -n 's/^+.*"twitter"[^"]*"\\(.*\\)".*/\\1/p'`
```

tests/twitter.rb
```diff
@@ -11,20 +11,20 @@
...
+# rubocop:enable Layout/LineLength
 diff.gsub("\n", '').split('"').each do |handle|
```
phallobst:
```suggestion
diff.split('\n').each do |handle|
```

phallobst (changes requested):
I guess I wasn't clear enough with my suggestion.

phallobst (approved):

tests/facebook.rb
```diff
@@ -5,8 +5,8 @@
 require 'uri'
 
 status = 0
-diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"facebook\\":" | cut -c21-`
-diff.gsub("\n", '').gsub(',', '').split('"').each do |page|
+diff = `git diff origin/master...HEAD entries/ | sed -n 's/^+.*"facebook"[^"]*"\\(.*\\)".*/\\1/p'`
+diff.split('\n').each do |page|
```
ApCoder123:
```suggestion
diff.split(' ').each do |page|
```
I might be missing something here, but when I run this, I get the error:
```
bad URI(is not URI?): "https://www.facebook.com/pg/myinvalidfacebookpage\\n" (URI::InvalidURIError)
```
There seems to be a `\n` that isn't getting removed. The test works when splitting at new lines is changed to splitting at spaces or removing the `\n` from the string.

Carlgo11: Nice catch.

I have no clue why that would be the case. The variable `diff` is literally `rbc\nvygruppen\n`. No spaces, just linebreakers. Splitting with `' '` does indeed fix it though.
Maybe @phallobst knows what is happening?

phallobst:
It seems that `foo.split('\n')` is not the same as `foo.split("\n")` - the double quotes are needed for the line break processing (or `foo.split(/\n/)` as an alternative).

Also, I've learned that a single space for the pattern means "any whitespace character", so that works with line breaks as well.

Who designed this? :vomiting_face:

tests/alexa.rb
```diff
@@ -41,9 +41,9 @@ def store_cache(site, rank)
 
 status = 0
 # Fetch changes
-diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"domain\\":" | cut -c17-`
+diff = `git diff origin/master...HEAD entries/ | sed -n 's/^+.*"domain"[^"]*"\\(.*\\)".*/\\1/p'`
 # Strip and loop through diff
-diff.gsub("\n", '').split('",').each do |site|
+diff.split('\n').each do |site|
```
phallobst:
```suggestion
diff.split("\n").each do |site|
```

tests/facebook.rb
```diff
@@ -5,8 +5,8 @@
 require 'uri'
 
 status = 0
-diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"facebook\\":" | cut -c21-`
-diff.gsub("\n", '').gsub(',', '').split('"').each do |page|
+diff = `git diff origin/master...HEAD entries/ | sed -n 's/^+.*"facebook"[^"]*"\\(.*\\)".*/\\1/p'`
+diff.split(' ').each do |page|
```
phallobst:
```suggestion
diff.split("\n").each do |page|
```

tests/twitter.rb
```diff
@@ -11,12 +11,14 @@
 end
 
 status = 0
-diff = `git diff origin/master...HEAD entries/ | grep "^+[[:space:]]*\\"twitter\\":" | cut -c20-`
-diff.gsub("\n", '').split('"').each do |handle|
+diff = `git diff origin/master...HEAD entries/ | sed -n 's/^+.*"twitter"[^"]*"\\(.*\\)".*/\\1/p'`
+diff.split('\n').each do |handle|
```
phallobst:
```suggestion
diff.split("\n").each do |handle|
```

phallobst (changes requested):

phallobst (approved):

ApCoder123 (approved):

PR #5878, opened 2021-06-26
Update Slack

ApCoder123:


entries/s/slack.com.json
```diff
@@ -6,6 +6,7 @@
       "totp"
     ],
     "documentation": "https://get.slack.help/hc/en-us/articles/204509068",
+    "recovery": "https://slack.com/intl/en-gb/help/articles/204509068#troubleshoot-2fa",
```
phallobst, line 9: Having the `recovery` link point to the _same_ page as `documentation` is a bit excessive, isn't it?

Carlgo11: I agree with @phallobst

PR #5887, opened 2021-06-28
Skip empty categories

### LINKED ISSUE #5886: Skip empty categories when building the site
If a regional site doesn't have any entries in a certain category, the category probably shouldn't have a button or table on that regional site.

If anyone is interested in fixing this then be my guest :wink:
I tried fixing it with an `entries`-array size check and a for loop `{% continue %}` but that doesn't apply to the category icon in desktop view.

phallobst:
Fixes #5886

### DIFF
--- a/scripts/regions.rb
+++ b/scripts/regions.rb
@@ -23,13 +23,24 @@ regions.each do |region|
   end
 
   all = {}
+  used_categories = {}
 
   # Website loop
   websites.each do |name, website|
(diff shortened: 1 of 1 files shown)


phallobst:
The only test case that I can see right now is "Switzerland / Government".

Carlgo11 (approved):
Ah I didn't think of doing it through the regions.rb script. Good work! :+1:

ApCoder123 (approved):

PR #5893, opened 2021-06-30
Add recovery links

### LINKED ISSUE #5892: Add "Recovery" column on 2fa.directory website
for example this one https://github.com/2factorauth/twofactorauth/blob/master/entries/t/technicpack.net.json
```
{
  "Technic": {
    "domain": "technicpack.net",
    "tfa": [
      "totp"
    ],
    "recovery": "https://www.technicpack.net/tfa/remove",
    "keywords": [
      "gaming"
    ]
  }
}
```

has "recovery" option but never used on [2fa.directory](https://2fa.directory)

I propose to display this one as it is useful information, like "Docs" right now.

Carlgo11:
Resolves #1348 and the new issue #5892.
This also applies to the desktop view. I'm uncertain if adding recovery links to the mobile view would help more than it would clutter the limited space available.

The updated view looks like this:
screenshot

### DIFF
--- a/_includes/html/desktop-table.html
+++ b/_includes/html/desktop-table.html
@@ -46,11 +46,18 @@
             {% assign tfa = website.tfa %}
 
             <td class="icon-cell">
+              <div class="row">
               {% if website.documentation %}
-                <a href="{{ website.documentation }}">
+                <a href="{{ website.documentation }}" class="col">
                   <i class="fas fa-book"></i>
                 </a>
               {% endif %}
+              {% if website.recovery %}
+                <a href="{{ website.recovery }}" class="col">
+                  <i class="fas fa-redo"></i>
+                </a>
+              {% endif %}
+              </div>
             </td>
 
             <td class="icon-cell">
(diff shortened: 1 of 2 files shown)


person1234-ux:
I am a little bit worried about how it would look on mobile maybe an option to choose if you want to display the recovery column would work well

Carlgo11:
@person1234-ux it's a static site so there's no way to have any such settings available.

phallobst:
I think the icons are too different looking, given that they provide similar information.

Is it possible to create the new icon by stacking the book (in blue) and the "redo" arrow (in black) on top of each other, with the arrow in the bottom right corner? (I'm hopeless regarding CSS, unfortunately.)

Maybe the two icons should be closer to each other as well.

ApCoder123:
Is something like below what you were picturing, @phallobst?
image

phallobst:
Basically, yes - but now that I see it, I guess it does look rather ugly... Maybe there is something else we could do.

Carlgo11:
Maybe we could merge this for now and then figure out a if we should replace the recovery icon, documentation icon or both? :shrug:

ApCoder123 (approved):

phallobst (approved):

PR #5902, opened 2021-07-01
Update Network Solutions

phallobst:


ApCoder123 (approved):

Carlgo11 (approved):
The viewBox was a bit off so I changed that and aligned the object afterwards.

PR #5908, opened 2021-07-02
Update twitter.com.json

person1234-ux:
hardware keys no longer require another backup 2fa method https://twitter.com/TwitterSupport/status/1410343864431374340?s=20


person1234-ux:
didn't realize there is already a pull request opened to update this

PR #5909, opened 2021-07-02
Update Pepipost/Add Netcore Cloud

40504:


ApCoder123 (commented):
This seems to be a rebrand of `pepipost.com`, which as already listed. However, the doc link for pepipost redirects to the doc link given here for Netcore Cloud (Email services). I think that it is probably better to remove the Pepipost entry/rename it to Netcore Cloud, rather than include both.

40504:
As is now netcorecloud.com

entries/n/netcorecloud.com.json
```diff
@@ -0,0 +1,13 @@
+{
+  "Netcore Cloud": {
+    "domain": "netcorecloud.com",
```
ApCoder123, line 3:
```suggestion
    "domain": "netcorecloud.com",
    "additional-domains": [
      "pepipost.com"
    ],
```

The `pepipost.com` domain still seems to be in use despite the renaming.

entries/n/netcorecloud.com.json
```diff
@@ -0,0 +1,13 @@
+{
+  "Netcore Cloud": {
```
ApCoder123:
```suggestion
  "Netcore Cloud Email API": {
```
2FA only applies to this specific product.

ApCoder123 (changes requested):

entries/n/netcorecloud.com.json
40504: why "Netcore Cloud Email API": when everywhere is "Netcore Cloud"?

ApCoder123: The doc page provided is for their email API:
image

whereas other netcore cloud products (see https://netcorecloud.com/login) don't appear to support 2FA.

ApCoder123 (approved):

entries/n/netcorecloud.com.json
```diff
@@ -0,0 +1,16 @@
...
+    "additional-domains": [
+      "pepipost.com"
+    ],
+    "tfa": [
+      "email",
+      "totp"
+    ],
+    "documentation": "https://emaildocs.netcorecloud.com/docs/two-factor-authentication-2fa",
+    "keywords": [
+      "communication"
```
Carlgo11, line 13: Just like in #6017, marketing would be the proper category.
```suggestion
      "marketing"
```

Carlgo11 (changes requested):

entries/n/netcorecloud.com.json
Carlgo11: I'll change it in that PR instead.

Carlgo11 (approved):

PR #5910, opened 2021-07-03
Remove Skatteverket

Carlgo11:
Skatteverket, along with all other Swedish government sites, support all [european eID systems](https://connector.eidas.swedenconnect.se/idp/extauth/start). As every government site supports all eID systems, which makes them all require 2FA but with no central documentation or setup process, I don't think there's much use in listing them here.

As #5887 now makes it possible to skip the government category completely for the Swedish regional site, I think that would be the better option.


ApCoder123 (approved):

phallobst:
I don't quite understand your reasoning here - do you want to remove the site because 2FA is mandatory?

Carlgo11:
@phallobst as it is mandatory on all sites in the category I think Skatteverket should be removed from the list. This would result in the government category being hidden over at https://2fa.directory/se.

Since the goal of this project is aid users in choosing what services they use based on the 2FA support of services, and publicly shame the services that don't, I don't see much use in a category where all sites have an effective monopoly (ruling out the first use case) and have a mandatory 2FA requirement (ruling out the second use case).

kmpoppe (approved):

PR #5912, opened 2021-07-04
Add Philips Hue

ApCoder123:
image
image
image


phallobst (changes requested):
Please use the main site "philips-hue.com" for the domain instead of the login page.

phallobst:
How about this icon:

phallobst:
BTW: This is a good example why the SVG viewbox is useful sometimes. My example was _visually_ centered, but in your version the viewbox is set tightly around the logo, so it will be _geometrically_ centered, and therefore slightly below the visual middle.

ApCoder123:
Thanks! I didn't realise that geometrically centering the icon would move it below the visual middle. Noted for next time.

phallobst (approved):

Carlgo11 (approved):

PR #5929, opened 2021-07-07
Update Microsoft Services

ApCoder123:
Security keys can only be used as a paswordless login method, not as a second factor. This esentially makes the first factor "something you have", rather than "something you know", but does not constitute 2FA, as there is no second factor involved.


Carlgo11 (approved):
That really sucks :roll_eyes:

phallobst (approved):

PR #5932, opened 2021-07-07
Add Maropost

40504:


entries/m/maropost.com.json
```diff
@@ -0,0 +1,12 @@
+{
+  "Maropost": {
```
ApCoder123:
```suggestion
  "Maropost Marketing Cloud": {
```
They seem to have 2 different products, each with separate auth systems. 
image
and email 2FA seems to only apply to the marketing cloud. The commerce cloud is hosted at a different domain and supports different 2FA methods (https://www.netohq.com/support/s/article/multi-factor-authentication)

ApCoder123 (approved):

phallobst (approved):

PR #5937, opened 2021-07-08
Add PowerReviews

40504:


entries/p/powerreviews.com.json
```diff
@@ -0,0 +1,12 @@
+{
+  "PowerReviews": {
+    "domain": "powerreviews.com",
```
Carlgo11, line 3: Their `https://powerreviews.com` certificate isn't valid.
```suggestion
    "domain": "powerreviews.com",
    "url": "https://www.powerreviews.com"
```

Carlgo11 (changes requested):

entries/p/powerreviews.com.json
Carlgo11: do
```BASH
curl -I https://powerreviews.com/
```
Does it fail?

phallobst (approved):

entries/p/powerreviews.com.json
40504: You are right, the certificate isn't valid

Carlgo11 (approved):

PR #5943, opened 2021-07-08
Add Redis Cloud

ApCoder123:


entries/r/redislabs.com.json
```diff
@@ -0,0 +1,15 @@
+{
+  "Redis Enterprise Cloud": {
+    "domain": "redislabs.com",
+    "tfa": [
+      "totp",
+      "sms"
+    ],
+    "documentation": "https://docs.redislabs.com/latest/rc/security/admin-console-security/multi-factor-authentication/",
+    "notes": "SMS 2FA must be enabled before TOTP 2FA can be enabled.",
```
phallobst: Can SMS 2FA be disabled after that while keeping TOTP active? If not, the text should reflect that.

ApCoder123: SMS 2FA cannot be disabled when TOTP is active. I've updated the exception text

phallobst (approved):

Carlgo11 (approved):

PR #5954, opened 2021-07-09
Remove Medium

ApCoder123:
- Medium only supports login via email link (no passwords), but this is still just one factor, so does not constitute 2FA.
- Given the way in which users are authenticated, it doesn't make sense to replace the entry with contact info, as it doesn't really make sense to get a TOTP code just to have an email sent to you, so I doubt Medium would ever consider implementing 2FA. 
- Therefore, the site should be removed.


ApCoder123:
Closing in favour of #6026

PR #5955, opened 2021-07-10
Update YouTube

40504:
added additional domains


phallobst:
These domains just seem to forward to `youtube.com`. I'm not sure it's useful to include these then.

ApCoder123:
I agree with @phallobst. Due to the redirect, I can't really see a benefit to password managers or to searching for sites on 2fa.directory.

PR #5962, opened 2021-07-13
Update BigCommerce

40504:


entries/b/bigcommerce.com.json
```diff
@@ -1,17 +1,26 @@
 {
-    "BigCommerce": {
-      "domain": "bigcommerce.com",
-      "url": "https://www.bigcommerce.com",
```
ApCoder123, line 4: The `url` field is needed in this case, as the `url.rb` tests are failing

ApCoder123 (changes requested):

ApCoder123 (approved):

phallobst:
It's not really `retail`, is it?

40504:
> It's not really `retail`, is it?

I don't know which category to include 🤷🏻‍♂️

ApCoder123:
> It's not really `retail`, is it?

@phallobst Do you think that other similar services should be moved out of the `retail` category? E.g. Shift4Shop and Shopify

phallobst:
Maybe a `commerce` category would be useful. This could include some more entries:
2checkout, clover, refersion, revenuecat, spring, thrivecart, tito

phallobst (approved):

PR #5963, opened 2021-07-13
Add BlueSnap

40504:


entries/h/home.bluesnap.com.json
```diff
@@ -0,0 +1,12 @@
+{
+  "BlueSnap": {
+    "domain": "home.bluesnap.com",
+    "tfa": [
+      "totp"
+    ],
+    "documentation": "https://support.bluesnap.com/docs/two-factor_authentication",
```
ApCoder123:
```suggestion
    "documentation": "https://support.bluesnap.com/docs/two-factor_authentication",
    "notes": "2FA must be enabled by an admin before use.", 
```

ApCoder123 (approved):

phallobst:
I reckon it would be better to use "bluesnap.com" for the domain and file names - without "home". For entries where "example.com" redirects to "www.example.com", we wouldn't include the "www" either.

40504:
> I reckon it would be better to use "bluesnap.com" for the domain and file names - without "home". For entries where "example.com" redirects to "www.example.com", we wouldn't include the "www" either.

done ✅

phallobst (approved):

PR #5982, opened 2021-07-19
Improve Entry Tags Test

ApCoder123:
Checks now test for the following cases:

If `tfa` is not present:
- `notes` is not present (included in current tests)
- `documentation` is not present
- `recovery` is not present
- `custom-property` is not present

If `tfa` is present:
- `contact` is not present (included in current tests)

If `custom-software`/`custom-hardware` is present:
- `custom-property` must be included in `tfa`
<br>

One of `contact` or `tfa` must always be present


phallobst:
I wonder if these dependencies and checks rather belong in the `tests/schema.json` file.

ApCoder123:
Ok, I've modified the schema so that `validate-json.rb` checks for all of those cases now and I have removed the entry-tags test.

phallobst (approved):

Carlgo11 (approved):

PR #5460, opened 2021-03-30
Add Studio Ninja

ApCoder123:
Alexa Rank 145,636

PR #5986, opened 2021-07-20
Add Studio Ninja

ApCoder123:
Supersedes #5460


Carlgo11 (approved):

phallobst (approved):
I prefer a good PNG over a bad SVG, but there we go...

PR #5991, opened 2021-07-24
Update Geforce

ApCoder123:
Although the default is authorisation when adding a 2FA method, settings can be changed to force 2 factor authentication rather than authorisation.
image

### DIFF
--- a/entries/g/geforce.com.json
+++ b/entries/g/geforce.com.json
@@ -1,11 +1,16 @@
 {
   "GeForce (Nvidia)": {
     "domain": "geforce.com",
-    "url": "https://www.geforce.com/",
-    "contact": {
-      "facebook": "NVIDIAGeForce",
-      "twitter": "NVIDIAGeForce"
-    },
+    "additional-domains": [
+      "nvidia.com",
+      "login.nvgs.nvidia.com"
+    ],
+    "tfa": [
+      "email",
+      "totp",
+      "u2f"
+    ],
+    "documentation": "https://login.nvgs.nvidia.com/v1/help/nfactor",
     "keywords": [
       "gaming"
     ]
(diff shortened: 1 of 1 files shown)


entries/g/geforce.com.json
```diff
@@ -1,11 +1,16 @@
...
+    "additional-domains": [
+      "nvidia.com",
+      "login.nvgs.nvidia.com"
+    ],
+    "tfa": [
+      "email",
+      "totp",
+      "u2f"
+    ],
+    "documentation": "https://login.nvgs.nvidia.com/v1/help/nfactor",
```
Carlgo11, line 13:
```suggestion
    "documentation": "https://login.nvgs.nvidia.com/v1/help/nfactor",
    "notes": "Email 2FA is always enabled."
```

Carlgo11 (changes requested):
Really annoying but the option to auth with email can't be disabled :roll_eyes:

Carlgo11 (approved):

entries/g/geforce.com.json
```diff
@@ -1,13 +1,19 @@
...
   "GeForce (Nvidia)": {
     "domain": "geforce.com",
-    "url": "https://www.geforce.com/",
-    "contact": {
-      "facebook": "NVIDIAGeForce",
-      "twitter": "NVIDIAGeForce"
-    },
+    "additional-domains": [
+      "nvidia.com",
+      "login.nvgs.nvidia.com"
```
phallobst, line 6: Is this still useful when `nvidia.com` is already listed?
```suggestion
```

ApCoder123: I thought that it could potentially be useful for password manager support. I know that bitwarden doesn't check subdomains for 2FA support and 1Password only does it 1 subdomain deep: 
image
image

phallobst (approved):

PR #6017, opened 2021-07-30
Add CoinPayU

40504:
Group 12
Group 13


phallobst (approved):

Carlgo11 (commented):
Marketing & Analytics from #6051 might be a better category as CoinPayU describes themselves as an Advertising Network (which is related to Marketing). Alternatively using both :shrug:

Carlgo11 (approved):

PR #6022, opened 2021-08-01
Fix sorting order in APIv3.rb

ApCoder123:
The same bug from #5724 is present in the APIv3 script.


phallobst (approved):
Not sure if the order is relevant in the API files, but it doesn't hurt either.

Carlgo11 (approved):

PR #6043, opened 2021-08-13
Update WoltLab

TimWolla:
see https://community.woltlab.com/thread/291216-woltlab-suite-5-4-released/

----------------

Disclosure: I'm a WoltLab employee, as also indicated by the source repository being within the [WoltLab organization on GitHub](https://github.com/WoltLab).

PS: I was not able to get the build running locally based off the `Dockerfile`, because the `Dockerfile` does not include all the necessary scripts. Hope I adjusted everything correctly.

```
Converting scripts to Unix format:
dos2unix: scripts/*: No such file or directory
dos2unix: Skipping scripts/*, not a regular file.
dos2unix: tests/*: No such file or directory
dos2unix: Skipping tests/*, not a regular file.
Creating API directories:
Generating API files:
Traceback (most recent call last):
ruby: No such file or directory -- ./scripts/APIv*.rb (LoadError)
Building site:
Configuration file: none
            Source: /twofactorauth/site
       Destination: /twofactorauth/site/_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
                    done in 0.021 seconds.
 Auto-regeneration: disabled. Use --watch to enable.
Generating regions:
Traceback (most recent call last):
ruby: No such file or directory -- ./scripts/regions.rb (LoadError)
Minifying JavaScript files:
/twofactorauth/entrypoint.sh: line 31: ./scripts/minify-js.sh: not found
```


entries/w/woltlab.com.json
```diff
@@ -1,15 +1,16 @@
 {
   "WoltLab": {
     "domain": "woltlab.com",
+    "tfa": [
+      "totp",
+      "u2f",
+      "email"
+    ],
+    "documentation": "https://www.woltlab.com/account-security/",
```
Carlgo11: The documentation page can't 403 I'm afraid. Could you post a screenshot of the page here?

entries/w/woltlab.com.json
```diff
@@ -1,15 +1,16 @@
...
     "url": "https://www.woltlab.com/",
     "img": "woltlab.com.png",
```
Carlgo11: Since you're working for WoltLab, you wouldn't happen to have the logo in an SVG format? We try to use SVG icons for all entries but we haven't been able to find one for WoltLab thus the PNG at `img/w/woltlab.com.png`.

entries/w/woltlab.com.json
```diff
@@ -1,15 +1,16 @@
 {
   "WoltLab": {
     "domain": "woltlab.com",
+    "tfa": [
+      "totp",
+      "u2f",
+      "email"
+    ],
+    "documentation": "https://www.woltlab.com/account-security/",
     "url": "https://www.woltlab.com/",
```
Carlgo11: We direct to `https://{domain}/` by default and your website seem to handle that fine :wink: 
```suggestion
```

Carlgo11 (changes requested):
Hi Tim! :wave:
Thanks for the bug report. Looks like @ApCoder123 is working on that now in #6044.

Regarding this PR there are just some minor tweaks I'd like to see

entries/w/woltlab.com.json
TimWolla: We have an SVG available, but due to the large number of colors in our logo it is much larger than the current PNG:

- The raw SVG is 7472 Bytes.
- The optimized SVG coming out of svgo is 6076 Bytes.
- The current PNG is exactly 1024 Bytes.

I can add the SVG if you like, but I'm not sure this is the intended outcome?

entries/w/woltlab.com.json
TimWolla: Yes, it does. I'll make the necessary changes in bulk once the other two discussions are resolved.

entries/w/woltlab.com.json
TimWolla: The linked "documentation" is the page where the user is able to set-up multi-factor authentication and thus only visible to logged in users. We do not have a dedicated manual for this, because we try to make the process as self-explanatory as possible.

Find screenshots of the page of my local development environment below. The first screenshot is what the user will see when following the link.

Screen Shot 2021-08-14 at 12 31 01
Screen Shot 2021-08-14 at 12 31 13
Screen Shot 2021-08-14 at 12 31 27
Screen Shot 2021-08-14 at 12 32 10
Screen Shot 2021-08-14 at 12 34 11

If this is not what you expect to see for a documentation link, just let me know. I'll remove the link then.

Carlgo11: Thanks for the screenshots. As it is the actual setup page and not a documentation page it still needs to be removed though.
```suggestion
```

Carlgo11 (changes requested):

entries/w/woltlab.com.json
TimWolla: Understood. Removed now.

entries/w/woltlab.com.json
TimWolla: Removed now.

TimWolla:
@Carlgo11 Thank you for the timely review. I have applied the requested changes and I believe that the JSON is good now.

Carlgo11 (approved):

entries/w/woltlab.com.json
phallobst: I wouldn't mind a 6KB SVG icon. There are several ones that are bigger already, and it will automatically be zipped when serving the site as well (I hope), so it's not a big difference to an already compressed PNG of size 1KB.

TimWolla: @phallobst Okay, I've just updated the image in a separate commit. `gzip -1` for the SVG results in 2280 Bytes for me. I'm happy to drop the commit again if you don't feel like the SVG is an improvement. Thanks!

phallobst (approved):

TimWolla:
Thanks!

PR #6044, opened 2021-08-13
Fix Docker Issues

ApCoder123:
Currently when running the site in a docker container, an error is encountered where `./entrypoint.sh` cannot be found, as when mounting a volume to `/twofactorauth`, the existing content (e.g. entrypoint) is overwritten. Additionally, there were missing steps in the entrypoint.sh file, such as not joining entries to `_data/all.json`. 

This PR modifies the Dockerfile and entrypoint to (attempt to) fix all of these issues and allow the site to be run.

```bash
docker build -t twofactorauth .
docker run -p 4000:4000 -v $(pwd):/twofactorauth twofactorauth
```

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -1,6 +1,5 @@
 #!/bin/sh
 cd /twofactorauth
-bundle config set path '/app/vendor/cache'
 
 if [ -z "${SKIP_DOS2UNIX}" ]; then
   echo "Converting scripts to Unix format:"
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
-cd /twofactorauth/site
-bundle config set path '/twofactorauth/vendor/cache'
+cd /twofactorauth
+bundle config set path '/app/vendor/cache'
 
 if [ -z "${SKIP_DOS2UNIX}" ]; then
   echo "Converting scripts to Unix format:"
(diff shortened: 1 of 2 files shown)


scripts/entrypoint.sh
```diff
@@ -14,11 +14,14 @@ if [ -z "${SKIP_API}" ]; then
   mkdir -p api/v2
   mkdir -p api/v3
   echo "Generating API files:"
-  ruby ./scripts/APIv*.rb
```
ApCoder123, line 17: Had to modify this, as the wildcard wasn't working and was only running the v1 script

scripts/entrypoint.sh
```diff
@@ -14,11 +14,14 @@ if [ -z "${SKIP_API}" ]; then
...
+  ruby ./scripts/APIv1.rb
+  ruby ./scripts/APIv2.rb
+  ruby ./scripts/APIv3.rb
```
phallobst: Maybe run them in a loop instead?
```suggestion
  for script in ./scripts/APIv*.rb; do
    ruby "$script"
  done
```

scripts/entrypoint.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
-cd /twofactorauth/site
-bundle config set path '/twofactorauth/vendor/cache'
+cd /twofactorauth
+bundle config set path '/app/vendor/cache'
```
Carlgo11, line 3: I building the regional sites doesn't work for me unless this is added:
```suggestion
bundle install --gemfile=/app/Gemfile

```

ApCoder123: I've made some modifications in 50f01f3 that should resolve this issue without having to re-install any gems

Carlgo11: It does indeed work now. Nice work! :clap:

Carlgo11 (approved):

phallobst (approved):

PR #6048, opened 2021-08-16
Add mxtoolbox.com

ghost:


entries/m/mxtoolbox.com.json
```diff
@@ -0,0 +1,14 @@
+{
+  "MxToolbox": {
+    "domain": "mxtoolbox.com",
+    "url": "https://www.mxtoolbox.com/",
```
phallobst: This line is not necessary in this case.
```suggestion
```

entries/m/mxtoolbox.com.json
```diff
@@ -0,0 +1,14 @@
...
+    "tfa": [
+      "sms",
+      "totp"
+    ],
+    "documentation": "https://mxtoolbox.com/user/twostepverification",
```
phallobst: This documentation page requires the user to log in. Is there a publicly accessible page available?

phallobst (changes requested):

entries/m/mxtoolbox.com.json
ghost: unfortunately no public documentation is available. Only documentation is on internal activation page. 

`
The basic idea is that a single password for your important accounts simply isn’t enough. If your password is guessed, or hackers steal a database with your login information in plain text, your account is a sitting duck. Two-factor authentication tries to address that flaw by requiring a secondary code called a one-time password (OTP)—usually six characters in length and generated by a smartphone app—before you can gain access to your account. That way even if a hacker has your password they’ll still need to crack a secondary code, which makes getting in that much harder.
`

not sure whether it makes any sense at all for it to be listed then.

Carlgo11:
```suggestion
```

phallobst (approved):

Carlgo11 (approved):

PR #6054, opened 2021-08-18
Optimise gem installs

ApCoder123:
I've removed some unused dependencies from the `Gemfile` that used to be used in the old Rakefile and `verify.rb` test. I've also modified some of the installs in the GitHub Actions workflow to avoid installing unnecessary dependencies (i.e. ignore `jekyll_plugins` group for Tests, which should speed up execution time.

### DIFF
--- a/.github/workflows/repository.yml
+++ b/.github/workflows/repository.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Bundle install
         run: |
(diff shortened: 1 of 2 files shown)


Carlgo11 (approved):
Major time difference for the `Tests` job! :alarm_clock:
I remember the Travis-CI tests taking +5 minutes once upon a time :sweat_smile:

phallobst (approved):

PR #6064, opened 2021-08-23
Rework join-entries.rb

ApCoder123:
I've reworked the join-entries.rb script to allow for entries with the same name. 

The only test case I can see is Emma (emma.ca/myemma.com).

Before:
image

After:
image


phallobst:
This may help with the website, but I'm not sure that anyone using the API files is expecting to have two entries with the same name. Renaming one (or both) is probably better, maybe even add a check to prevent this situation.

PR #6066, opened 2021-08-24
Check for duplicate site names

ApCoder123:
Supersedes #6064

### DIFF
--- a/tests/validate-json.rb
+++ b/tests/validate-json.rb
@@ -36,6 +36,7 @@ Dir.glob('entries/*/*.json') do |file|
       puts "  expected: 'tfa' to contain '#{v['schema']['contains']['const']}'" if v['type'].eql?('contains')
     end
     status = 1
+    exit(status)
   end
 
   domain = document.values[0]['domain']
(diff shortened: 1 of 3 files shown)


phallobst (commented):
I think it's better to do this check after the schema validation. Also, having a schema error should probably result in skipping all other checks (for that file) as well, since we can't rely on the JSON elements being present anyway.

tests/validate-json.rb
```diff
@@ -34,6 +36,7 @@
       puts "  expected: 'tfa' to contain '#{v['schema']['contains']['const']}'" if v['type'].eql?('contains')
     end
     status = 1
+    exit(status)
```
phallobst: Better go on with the next file instead of exiting early:
```suggestion
    next
```

phallobst (changes requested):

Carlgo11:
I found some style warnings and code style inconsistencies (partly made by me) that I fixed.

phallobst (approved):

Carlgo11 (approved):

PR #6068, opened 2021-08-25
Adding toshl.com

germanleonz:


phallobst (approved):

ApCoder123 (approved):
I modified your SVG slightly, as it was looking a bit stretched for me when testing on the site:
image

PR #6073, opened 2021-08-25
Add MyFRITZ!

phallobst:
Supersedes #5703 

`myfritz.net` has no Alexa rank, but `sso.myfritz.net` has... :man_shrugging: 

Although the doc page refers to FRITZ!Box 7590, the MyFRITZ! login is device-independent.


ApCoder123 (approved):

Carlgo11 (approved):

Carlgo11:
Oh, what about regions?

phallobst:
Regions should not matter here.

Carlgo11:
But according to Wikipedia Ftitz! is available in Germany, Austria and Switzerland so shouldn't that be the regions?

phallobst:
https://en.avm.de/service/distributors/europe/germany/

Those are just the countries with distribution partners - using a FRITZ!Box elsewhere is possible as well.

Carlgo11:
Thanks for clearing that up :+1:

PR #6075, opened 2021-08-25
Add TU Berlin

peremen:
My university supports 2FA, so adding one.

Question: If there are documents available in multiple languages, then which one should be used? (In this case: German/English)


Carlgo11 (approved):
Thanks for the pull request @peremen!

In general English versions are preferred as it is much easier for us to validate those links. Many maintainers here can read German though :wink: 

I removed the `url` tag as it isn't needed when the default link `https://{domain}/` automatically redirects to the desired URL.
I also minimized the SVG a bit.

phallobst:
> Bei bestimmten sicherheitsrelevanten Anwendungen im TU-Portal werden sogenannte Transaktionsnummern – kurz TAN – zur Bestätigung von Aktionen verwendet

This sounds more like _authorisation_ than _authentication_. Do you need 2FA for logging into your account?

peremen:
> This sounds more like _authorisation_ than _authentication_. Do you need 2FA for logging into your account?

Currently only _selected_ websites of TU Berlin (at least mail service) requires 2FA, but not for whole of them.

entries/t/tu.berlin.json
```diff
@@ -0,0 +1,19 @@
...
+  "Technische Universität Berlin": {
+    "domain": "tu.berlin",
+    "additional-domains": [
+      "tu-berlin.de"
+    ],
+    "tfa": [
+      "totp",
+      "sms"
+    ],
+    "documentation": "https://www.campusmanagement.tu-berlin.de/menue/dienste/konto_karte/tan_verfahren/",
```
phallobst, line 11: Maybe we should note this exception:
```suggestion
    "documentation": "https://www.campusmanagement.tu-berlin.de/menue/dienste/konto_karte/tan_verfahren/",
    "notes": "2FA is only available for selected services",
```

phallobst (changes requested):

entries/t/tu.berlin.json
peremen: Looks good to me.

phallobst (approved):

PR #6076, opened 2021-08-25
Add translation of some elements in Korean

peremen:
While it is a bit sad that the website itself is not available in other languages, I want to add Korean translation of e-mail titles and tweets. Also some websites indicate "kr" as available region, so I added that entity too.

### DIFF
--- a/_data/languages.yml
+++ b/_data/languages.yml
@@ -19,6 +19,9 @@ he:
 it:
   email_subject: Suportare l'autenticazione a due fattori
   tweet: La sicurezza è importante, @TWITTERHANDLE. Ci piacerebbe che supportaste l'autenticazione a due fattori (2FA).
+ko:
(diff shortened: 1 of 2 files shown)


Carlgo11:
Thanks for the translation @peremen! For some reason, a script way down the supply chain pipeline failed because of the special characters (non-latin). I added quotes around your values and at least now the build doesn't fail.
We'll have to see if it parses correctly once your PR is merged. :crossed_fingers:

Carlgo11 (approved):
The translation strings look good! I've removed South Korea from the `regions.yml` as according to the [api](https://2fa.directory/api/v3/regions.json) there are only 6 entries listing South Korea as a supported region. We typically don't list regions with less than 10 entries for that region.

phallobst (approved):

PR #6078, opened 2021-08-26
Rename Radius Bank

ApCoder123:


phallobst:
Two entries "LendingClub" in the _Banking_ category seems weird. Should the original one be moved to `finance` maybe?

entries/b/bank.lendingclub.com.json
```diff
@@ -0,0 +1,23 @@
...
+    "additional-domains": [
+      "banking.lendingclub.com",
+      "radiusbank.com"
+    ],
+    "tfa": [
+      "sms",
+      "call",
+      "email",
+      "totp"
+    ],
```
phallobst: The doc page doesn't list the methods unfortunately, and on https://radiusbank.custhelp.com/app/answers/detail/a_id/70 they mention "email or text" only. Do you have further information?

ApCoder123: I got in touch with support and they confirmed that SMS, call and email are supported

> Yes, we offer two factor authentication to be received via emails and for SMS/phone call.

phallobst (approved):

Carlgo11 (approved):

PR #6083, opened 2021-08-29
Rewrite cURL test

Carlgo11:
I remade the cURL test and added a few things.

It now tests the documentation and recovery URLs as well as the `additional-domains` array if one exists.
I found it easier to fetch changed file names from `git diff` and parse the JSON, instead of using `sed` on the `git diff`. After I reworked that, I realized the script didn't need to be run via CircleCI, so I also moved the test to the GitHub Workflow.


Carlgo11:
I have no idea what happened but every commit from ada147c to d933c0e failed because either `curl` couldn't be found or it tried to curl null. Now that I added `next if path.empty?` it works and but suddenly all the broken builds show up as successful.
No idea what's up but I'd recommend just leaving the `next if path.empty?` as it is. This script is haunted :ghost: :fearful:

tests/validate-urls.rb
```diff
@@ -0,0 +1,27 @@
...
+  next if path.empty?
+
+  entry = JSON.parse(File.read(path)).values[0]
+
+  if entry.key? 'url'
+    curl(entry['url'], path)
+  else
+    curl("https://#{entry['domain']}/", path)
+    entry['additional-domains'].each { |domain| curl("https://#{domain}/", path) }
+  end
```
phallobst: `additional-domains` should be checked regardless of `url` being present:
```suggestion
  end
  entry['additional-domains'].each { |domain| curl("https://#{domain}/", path) } if entry['additional-domains']
```

tests/validate-urls.rb
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'English'
+require 'json'
+diff = `git diff --name-only --diff-filter=AM origin/master...HEAD entries/`.split("\n")
```
phallobst, line 9:
```suggestion

@status = 0
diff = `git diff --name-only --diff-filter=AM origin/master...HEAD entries/`.split("\n")
```

tests/validate-urls.rb
```diff
@@ -0,0 +1,27 @@
...
+
+def curl(url, path)
+  puts `curl --fail -sSI #{url}`
+  puts "::error file=#{path}:: Unable to reach #{url}" unless $CHILD_STATUS.success?
```
phallobst:
```suggestion
  unless $CHILD_STATUS.success?
    puts "::error file=#{path}:: Unable to reach #{url}"
    @status = 1
  end
```

tests/validate-urls.rb
```diff
@@ -0,0 +1,27 @@
...
+  if entry.key? 'url'
+    curl(entry['url'], path)
+  else
+    curl("https://#{entry['domain']}/", path)
+    entry['additional-domains'].each { |domain| curl("https://#{domain}/", path) }
+  end
+
+  curl(entry['documentation'], path) if entry.key? !entry['documentation'].start_with?('/notes/')
+  curl(entry['recovery'], path) if entry.key? 'recovery'
+end
```
phallobst, line 34:
```suggestion
end

exit(@status)
```

phallobst (changes requested):
I can't see why `next if path.empty?` would be needed, although it's probably harmless.

tests/validate-urls.rb
Carlgo11: The build will fail regardless of the exit code since `::error::` is outputted.
Do you still want the exit code to be set?

phallobst: I think it's good style to use the return code of the script to indicate success or failure, and not rely on the caller to parse the output, especially since the output goes to `stdout` and not `stderr` here.

tests/validate-urls.rb
Carlgo11: My thought was that it if the `url` tag is set there's probably a high likelihood of `https://#{domain}/` not working as expected on the the `additional-domains` values as well.

tests/validate-urls.rb
```diff
@@ -0,0 +1,35 @@
...
+
+diff.each do |path|
+  # Make path global for curl()
+  @path = path
+  entry = JSON.parse(File.read(@path)).values[0]
+
+  # Process the url,domain & additional-domains
+  curl((entry.key?('url') ? entry['url'] : "https://#{entry['domain']}/"))
+  entry['additional-domains'].each { |domain| curl("https://#{domain}/") }
```
phallobst:
```suggestion
  entry['additional-domains'].each { |domain| curl("https://#{domain}/") } if entry.key? 'additional-domains'
```
Otherwise I get ```undefined method `each' for nil:NilClass``` for entries without `additional-domains`. Maybe that is related to the `nil` problem you had before?

phallobst (changes requested):

phallobst (approved):

tests/validate-urls.rb
Carlgo11: What a weird language Ruby is. I would have bet on it just skipping the iteration if the the array is nil.

Apparently `&` does the same thing as `if entry.key? 'additional-domains'`.

tests/validate-urls.rb
```diff
@@ -0,0 +1,35 @@
...
+  # Make path global for curl()
+  @path = path
+  entry = JSON.parse(File.read(@path)).values[0]
+
+  # Process the url,domain & additional-domains
+  curl((entry.key?('url') ? entry['url'] : "https://#{entry['domain']}/"))
+  entry['additional-domains']&.each { |domain| curl("https://#{domain}/") }
+
+  # Process documentation and recovery URLs
+  curl(entry['documentation']) if entry.key? !entry['documentation'].start_with?('/notes/')
```
ApCoder123:
```suggestion
  if entry.key? 'documentation'
    unless entry['documentation'].start_with?('/notes/')
      curl(entry['documentation'])
    end
  end
```

ApCoder123 (changes requested):

tests/validate-urls.rb
Carlgo11: @ApCoder123 what's the reasoning? From what I can tell the suggested change does the same as the existing line but with more lines.

ApCoder123: When testing this locally, the documentation link wasn't actually being tested because the current line is missing the `'documentation'` parameter with the `entry.key?` method call. 

When adding this in with the line `if entry.key? 'documentation' && !entry['documentation'].start_with?('/notes/')`, the link still wasn't being tested. After changing to the current code, the documentation link was being tested - I did try shorter versions, but rubocop wasn't satisfied.

Carlgo11: @ApCoder123 it should work now.
I also set an old user-agent because some sites would 403 on the curl user-agent.

ApCoder123 (approved):

PR #6084, opened 2021-08-29
add note that fastmail requires SMS as a backup

elindsey:
Documented here: https://www.fastmail.com/help/account/2fa.html

> To help make sure that you are not locked out of your own account, before you can enable two-step verification, you must add a recovery phone to your account. If you ever lose access to your primary form of two-step verification, your recovery phone can be used to prevent you from being locked out of your account. You get a code sent to your phone instead to complete your second step when you log in.

### DIFF
--- a/entries/f/fastmail.com.json
+++ b/entries/f/fastmail.com.json
@@ -9,6 +9,7 @@
       "u2f"
     ],
     "documentation": "https://www.fastmail.com/help/account/2fa.html",
+    "notes": "SMS must be set up as a backup method.",
     "keywords": [
       "email"
     ]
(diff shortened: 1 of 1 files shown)


entries/f/fastmail.com.json
```diff
@@ -9,8 +9,9 @@
       "u2f"
     ],
     "documentation": "https://www.fastmail.com/help/account/2fa.html",
+    "notes": "SMS must be set up as a backup method.",
```
phallobst: Further down the doc page it says:
> Users who accept the risk of being locked out of their account may remove the recovery phone number from their account after two-step verification has been enabled.

Maybe this text (as used for other entries) is clearer:
```suggestion
    "notes": "Setup requires SMS verification.",
```

phallobst (changes requested):

phallobst (approved):

Carlgo11 (approved):

PR #6096, opened 2021-08-31
Modify docker workflow

ApCoder123:
Docker workflow is still not running - this might work 🤷‍♂️

It looks a bit closer to the [examples in the docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths)


Carlgo11 (approved):

phallobst (approved):
Not sure what this would change, to be honest...

PR #6104, opened 2021-09-02
Replace custom-hardware references with U2F

ApCoder123:
There were a couple entries that support U2F/WebAuthn, but were labelled as `custom-hardware`.

### DIFF
--- a/entries/s/stripe.com.json
+++ b/entries/s/stripe.com.json
@@ -5,7 +5,7 @@
     "tfa": [
       "sms",
       "totp",
-      "custom-hardware"
+      "u2f"
     ],
     "documentation": "https://support.stripe.com/questions/how-do-i-enable-two-step-verification",
     "keywords": [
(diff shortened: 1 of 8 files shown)


Carlgo11 (approved):

entries/s/stripe.com.json
```diff
@@ -5,7 +5,7 @@
     "tfa": [
       "sms",
       "totp",
-      "custom-hardware"
```
phallobst, line 8: They support _Windows Hello_ and _Touch ID_ as well - maybe we need both `custom-hardware` and ` u2f` here?

entries/v/vanguard.com.json
```diff
@@ -6,10 +6,10 @@
     "tfa": [
       "sms",
       "call",
-      "custom-hardware"
```
phallobst, line 9: Same here with _Voice Verification_ I guess.

phallobst (approved):

PR #894, opened 2014-12-04
Update Vanguard details

jdavis:
This PR:
- This continues #892.
- Uses relative link for notes to test locally.
- Included a note for Vanguard as it doesn't appear Vanguard has external documentation (yet).
- Also moved some things to an `_include/` for less copying and pasting.
- A bit of tidying.

PR #6107, opened 2021-09-03
Use absolute paths for notes

ApCoder123:
Currently, any entries with a `/notes` page for their documentation use the relative path. From what I can tell (#894), this was to help test locally. However, currently, these relative paths are appearing in the API files, which is probably not the desired behaviour.

### DIFF
--- a/entries/c/chase.com.json
+++ b/entries/c/chase.com.json
@@ -7,7 +7,7 @@
       "call",
       "email"
     ],
-    "documentation": "/notes/chase/",
+    "documentation": "https://2fa.directory/notes/chase/",
     "keywords": [
       "banking"
     ],
--- a/entries/i/internetbs.net.json
+++ b/entries/i/internetbs.net.json
@@ -4,7 +4,7 @@
     "tfa": [
       "totp"
     ],
-    "documentation": "/notes/internetbs/",
+    "documentation": "https://2fa.directory/notes/internetbs/",
     "keywords": [
       "domains"
     ]
--- a/entries/i/internetbs.net.json
+++ b/entries/i/internetbs.net.json
@@ -1,6 +1,7 @@
 {
   "Internet.bs": {
     "domain": "internetbs.net",
+    "url": "http://internetbs.net",
     "tfa": [
       "totp"
     ],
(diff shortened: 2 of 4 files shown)


ApCoder123:
Had to add an http `url` field to `internetbs.net.json`, as the curl test kept failing to reach `https://internetbs.net` with a 504 Gateway Timeout error.

Carlgo11:
@ApCoder123 Did you figure out what happened?
I only get 504 when the -I flag is set.

ApCoder123:
@Carlgo11 
I think their server was rejecting the `HEAD` requests that are sent with the `-I` flag. 

An alternative command to print headers without using the `HEAD` HTTP verb, but still supress the body could be:
`curl -sS -D - -o /dev/null https://url.com`

Edit: The original command in this comment was incorrect.

Carlgo11:
The curl output may not even be needed. As long as it can connect to the site and doesn't get a 4xx or 5xx status code.

phallobst:
Why would the relative paths not be the desired behaviour? By definition they would use the URL of the API file as the base, i.e. `https://2fa.directory/api/v3/all.json` + `/notes/chase/` = `https://2fa.directory/notes/chase/`. The same applies to `http://localhost:4000` accordingly.

Prefixing the URLs with the main domain definitely breaks testing, with no apparent advantage (at least for me).

ApCoder123:
My thoughts were that when receiving the API data e.g.
```JSON
["Internet.bs",{"domain":"internetbs.net","tfa":["totp"],"documentation":"/notes/internetbs/","keywords":["domains"]}]
```
users may not be expecting to receive a relative path, so probably won't consider appending the domain to beginning. 

I noticed this, as the "Missing 2FA" notification for internet.bs doesn't show up in 1Password and [Bitwarden's implementation](https://github.com/bitwarden/web/blob/master/src/app/tools/inactive-two-factor-report.component.ts) doesn't append 2fa.directory to the beginning either.

I'm open to keeping the relative paths how they are, although it may be useful to add a note about this to the api.md documentation.

Carlgo11:
I remember someone created an issue a couple of years ago complaining that the doc tags had relative paths instead of absolute paths and it messing up his program. We fixed it for the time being but it seems like the latest revision of the project removed that fix.

Carlgo11 (approved):

phallobst:
`tests/validate-urls.rb` could/should be adjusted as well.

phallobst (approved):

PR #6108, opened 2021-09-03
Multiple tweet messages

### LINKED ISSUE #1813: Make "Tell them to support 2FA" tweets more constructive

ApCoder123:
Partially fixes #1813

This PR allows for multiple tweet messages to be added and picks a random tweet message when the twitter button is clicked. Although this doesn't completely fix the issues mentioned in that issue, it does start to make the 2FA tweets a bit less spammy.

### DIFF
--- a/_data/languages.yml
+++ b/_data/languages.yml
@@ -38,3 +38,4 @@ pt:
 sv:
   tweets:
     - Säkerhet är viktigt, @TWITTERHANDLE. Vänligen lägg till two factor authentication på er hemsida.
+    - @TWITTERHANDLE snälla lägg till tvåfaktorsautentisering för användare.
--- a/_data/languages.yml
+++ b/_data/languages.yml
@@ -38,4 +38,4 @@ pt:
 sv:
   tweets:
     - Säkerhet är viktigt, @TWITTERHANDLE. Vänligen lägg till two factor authentication på er hemsida.
-    - @TWITTERHANDLE snälla lägg till tvåfaktorsautentisering för användare.
+    - "@TWITTERHANDLE snälla lägg till tvåfaktorsautentisering för användare."
--- a/_data/languages.yml
+++ b/_data/languages.yml
@@ -7,6 +7,8 @@ de:
   email_subject: Unterstützung von Zwei-Faktor-Authentifizierung
   tweets: 
     - Sicherheit ist wichtig, @TWITTERHANDLE. Bitte aktiviert Zwei-Faktor-Authentifizierung auf eurer Webseite.
+    - Zwei-Faktor-Authentifizierung ist ein wichtiges Sicherheitsmerkmal. Bitte unterstützt dies doch auch, @TWITTERHANDLE.
+    - Wir würden es sehr begrüßen, wenn ihr Zwei-Faktor-Authentifizierung unterstützen würdet, @TWITTERHANDLE. Die Sicherheit der Accounts wird dadurch wesentlich erhöht.
 en:
   email_subject: Support Two Factor Authentication
   tweets: 
--- a/_data/languages.yml
+++ b/_data/languages.yml
@@ -1,28 +1,40 @@
 # Languages should be categorized according to ISO 639-1
 # Language codes (including within data files) should be in lowercase
 cs:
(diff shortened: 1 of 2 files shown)


_data/languages.yml
```diff
@@ -1,28 +1,41 @@
...
 # Language codes (including within data files) should be in lowercase
 cs:
-  tweet: Bezpečnost je důležitá, @TWITTERHANDLE. Byli bychom rádi, kdybyste podporovali dvoufaktorové ověření.
+  tweets: 
+    - Bezpečnost je důležitá, @TWITTERHANDLE. Byli bychom rádi, kdybyste podporovali dvoufaktorové ověření.
 de:
   email_subject: Unterstützung von Zwei-Faktor-Authentifizierung
-  tweet: Sicherheit ist wichtig, @TWITTERHANDLE. Bitte aktiviert Zwei-Faktor-Authentifizierung auf eurer Webseite.
+  tweets: 
+    - Sicherheit ist wichtig, @TWITTERHANDLE. Bitte aktiviert Zwei-Faktor-Authentifizierung auf eurer Webseite.
```
phallobst, line 9:
```suggestion
    - Sicherheit ist wichtig, @TWITTERHANDLE. Bitte aktiviert Zwei-Faktor-Authentifizierung auf eurer Webseite.
    - Zwei-Faktor-Authentifizierung ist ein wichtiges Sicherheitsmerkmal. Bitte unterstützt dies doch auch, @TWITTERHANDLE.
    - Wir würden es sehr begrüßen, wenn ihr Zwei-Faktor-Authentifizierung unterstützen würdet, @TWITTERHANDLE. Die Sicherheit der Accounts wird dadurch wesentlich erhöht.
```

phallobst (changes requested):
Please don't tell me that JS requires adding and subtracting those 0s...

phallobst (approved):

Carlgo11 (approved):

PR #6110, opened 2021-09-03
Automatically redirect to regional sites

Carlgo11:
This PR adds the requirements at GitHub Pages to automatically redirect users to the different regional versions based on their geo-ip thanks to Cloudflare.
Using Cloudflare Workers it is possible to intercept requests to `https://2fa.directory/` and either 302 or 301 redirect the user to `https://2fa.directory/{region}/`.

Proof of concept: [2fa.carlgo11.com](https://2fa.carlgo11.com/).

As timing the merging of this PR and the activation of the Cloudflare Worker is difficult, I decided to leave the `index.html` as is and instead create a new "region" called Global `https://2fa.directory/int/` resulting in both the root dir and `/int/` working simultaneously for the time being.

Here's the CF Worker and keep in mind that I absolutely hate JS :wink: 
```JS
const base = "https://2fa.directory/"
const status = 302

async function gatherResponse(response) {
  const { headers } = response
  const contentType = headers.get("content-type") || ""
  if (contentType.includes("application/json")) {
    return await response.json()
  }
}

async function redirect(request) {
  // Use the cf object to obtain the country of the request.
  // The cf object is only avaialble in production.
  const init = {
    headers: {
      "content-type": "application/json;charset=UTF-8",
    },
  }
  const res = await fetch('https://2fa.directory/api/v3/regions.json', init)
  const countries = await gatherResponse(res) 
  const { pathname, search } = new URL(request.url)

  let country = null

  try {
  country = request.cf.country.toLowerCase()
  } catch(e){
    console.warn("Request not sent via CF")
  }

  if (country != null && country in countries && countries[country]['selection']) { 
    return Response.redirect(base + country + "/" + pathname + search, status)
  } else {
    console.info("Country code \"" + country + "\" not found.")
    return Response.redirect(base + 'int' + pathname + search, status)
  }
  
  
}

addEventListener("fetch", event => {
  event.respondWith(redirect(event.request))
})
```


phallobst:
I tried this with a VPN that was (allegedly) set to _Asia_ and was forwarded to `https://2fa.directory/se/`. Is this the current fallback on `2fa.carlgo11.com`, or is there something wrong?

Carlgo11:
@phallobst That was me testing a thing.
It's back to normal now and should be working as expected. I just tried with VPN and it worked for me.

scripts/APIv3.rb
```diff
@@ -2,14 +2,22 @@
...
+    regions[region]['count'] = 1 + regions[region]['count'].to_i
+  end
 end
-{ 'all' => all }.merge(tfa).each { |k, v| File.open("api/v3/#{k}.json", 'w') { |file| file.write v.sort_by{ |k,v| k.downcase }.to_json } }
-File.open('api/v3/regions.json', 'w') { |file| file.write regions.sort_by(&:last).reverse.to_h.to_json }
+
+avail_regions = YAML.load_file('_data/regions.yml').group_by { |hash| hash['id'] }.keys
+regions.each { |k, v| v['selection'] = avail_regions.include? k }
+{ 'all' => all }.merge(tfa).each { |k, v| File.open("api/v3/#{k}.json", 'w') { |file| file.write v.sort_by { |k, _v| k.downcase }.to_json } }
+
```
ApCoder123, line 28:
```suggestion

regions['int'] = { "count" => all.length, "selection" => true }

```

Adding the `int` region to the `regions.json` file will allow the Cloudflare worker to be simplified by a fair amount:
```js
const base = "https://2fa.directory/";
const status = 302;

async function gatherResponse(response) {
  const { headers } = response;
  const contentType = headers.get("content-type") || "";
  if (contentType.includes("application/json")) {
    return await response.json();
  }
}

async function redirect(request) {
  // Use the cf object to obtain the country of the request.
  // The cf object is only avaialble in production.
  const init = {
    headers: {
      "content-type": "application/json;charset=UTF-8",
    },
  }
  const res = await fetch('https://2fa.directory/api/v3/regions.json', init);
  let countries = await gatherResponse(res);
  const { pathname, search, hash } = new URL(request.url);

  let country = request.cf?.country.toLowerCase() || "int";

  if (!(country in countries) || !countries[country]['selection']){
    country = "int";
  }
  
  return Response.redirect(`${base}${country}${pathname}${search}${hash}`, status);
}

addEventListener("fetch", event => {
  event.respondWith(redirect(event.request));
})

```

Carlgo11: Thanks for the suggestion!
I made the change to https://redirect.twofactorauth.workers.dev/ but obviously it doesn't work as expected until the regions.json changes are live :wink:

PR #6121, opened 2021-09-08
Update afip.gob.ar

fainpablo:
Updated link to documentation. Removed notes as the website removed the limitations around which accounts could enable MFA.

### DIFF
--- a/entries/a/afip.gob.ar.json
+++ b/entries/a/afip.gob.ar.json
@@ -6,8 +6,7 @@
       "custom-hardware",
       "custom-software"
     ],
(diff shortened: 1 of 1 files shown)


Carlgo11 (approved):

Carlgo11:
The build tests is successful now but it looks like the tests also run independently on your fork and those tests try to compare the changes in this PR with your branch `master` which is really outdated. Because of that the tests continue to run and will probably do so for some time now.

@phallobst would it be a good idea to limit the tests to only run on the repository `2factorauth/twofactorauth`?

dsoegijono (approved):

PR #6124, opened 2021-09-11
Append URL to tweet message

ApCoder123:
Currently when using the tweet button on desktop, the site url is automatically appended to the tweet message.
However, this behaviour is not the same on mobile. Adding the `url` parameter to the tweet link fixes this.


ApCoder123:
`site.url` doesn't seem to be getting set with the the [Jekyll GitHub Metadata](https://github.com/jekyll/github-metadata)

Carlgo11:
{{ site.github.url }} ?

ApCoder123:
`{{ site.github.url }}` is equal to https://github.com/pages/2factorauth/twofactorauth

phallobst:
I tried it with Chrome on Android and the `url` parameter was already present. Maybe Twitter adds this automatically by using the `referer` property, and you have disabled that on your phone?

ApCoder123:
I've just had a look and it seems that when the twitter link is opened in the browser, the `url` parameter is automatically added. However, if the twitter app is opened when the twitter button is clicked, the `url` parameter is not present unless explicitly defined.

I think that this is what happened in [this tweet](https://twitter.com/DovahClown/status/1436000107611164677) (I'm assuming it was sent from 2fa.directory as it uses one of the new tweets from #6108).

Carlgo11:
Is the appending of the URL even a desired behavior? I didn't yet work on the project when the tweet button was created but I can imagine the appending of the site URL being something which wasn't explicitly sought after but rather just happened due to the way Twitter's tweet URIs work.

It might be worth debating if the tweet without the URL would be the desired tweet format.

Carlgo11 (approved):

phallobst (approved):

PR #6119, opened 2021-09-08
Remove SMS Requirement from wise.com

rkelly:
wise.com works with Google Voice.

PR #6128, opened 2021-09-14
Switch from cURL validation to httpclient

Carlgo11:
Using [nahi/httpclient](https://github.com/nahi/httpclient.git) allows for more fine tuned handling of failed HTTP requests.
Here I have made 50x responses fail without failing the entire build. This allows us to list sites that are blocking scripts from accessing their site (like sierra.com, alaskanair.com, #6123, #6119)

### DIFF
--- a/tests/validate-urls.rb
+++ b/tests/validate-urls.rb
@@ -13,12 +13,11 @@ def curl(url)
   headers = { 'User-Agent' => 'Mozilla/5.0 (compatible;  MSIE 7.01; Windows NT 5.0)', 'FROM' => '2fa.directory' }
   req = HTTPClient.new
   req.receive_timeout = 8
-  return 0 if (res = req.get(url, nil, headers, follow_redirect: true).status == 200)
-
-  raise(nil) unless res.status.match?('/50\d/')
+  res = req.get(url, nil, headers, follow_redirect: true)
+  return if res.status == 200
+  raise(nil) unless res.status.to_s.match(/50\d|403/)
 
   puts "::warning file=#{@path}:: Unexpected response from #{url} (#{res.status})"
-  0
 rescue StandardError => _e
   puts "::error file=#{@path}:: Unable to reach #{url} #{res.respond_to?('status') ? res.status : nil}"
   1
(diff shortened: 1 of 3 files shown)


tests/validate-urls.rb
```diff
@@ -3,20 +3,25 @@
...
-  unless $CHILD_STATUS.success?
-    # Break build if above cURL exited with non-zero value
-    puts "::error file=#{@path}:: Unable to reach #{url}"
-    @status = 1
-  end
-  # rubocop:enable Style/GuardClause
+  headers = { 'User-Agent' => 'Mozilla/5.0 (compatible;  MSIE 7.01; Windows NT 5.0)', 'FROM' => '2fa.directory' }
+  req = HTTPClient.new
+  req.receive_timeout = 8
+  return 0 if (res = req.get(url, nil, headers, follow_redirect: true).status == 200)
```
ApCoder123:
```suggestion
  res = req.get(url, nil, headers, follow_redirect: true)
  return 0 if res.status==200
```
This line wasn't working as intended, as `res` was being set to a boolean, so the matching for 50x codes below couldn't find the `status` property. I think it's probably easier to read as well if this is split into 2 lines.

tests/validate-urls.rb
```diff
@@ -3,20 +3,25 @@
...
+
+  raise(nil) unless res.status.match?('/50\d/')
```
ApCoder123:
```suggestion
  raise(nil) unless res.status.to_s.match(/50\d|403/)
```

The pattern wasn't being detected due to the quotes and there was an error with trying to match an integer. Also added 403 codes to be warned for, as those were also common when bot detection scripts were blocking (see #6123)

ApCoder123 (changes requested):

tests/validate-urls.rb
Carlgo11: I somewhat applied your suggestions as I had to rewrite some stuff due to Rubocop complaints. Let me know what you think.

ApCoder123 (approved):

phallobst (approved):
Maybe the function should be called `check_url` or something instead of `curl`.

PR #6130, opened 2021-09-15
Update Personal Capital

max-rousseau:
Personal Capital also supports sms & totp now. The documentation link proposed is also more pertinent.

### DIFF
--- a/entries/p/personalcapital.com.json
+++ b/entries/p/personalcapital.com.json
@@ -9,6 +9,9 @@
       "totp",
       "custom-software"
     ],
+    "custom-software": [
+      "Personal Capital App"
+    ],
     "documentation": "https://support.personalcapital.com/hc/en-us/sections/200230560-Security",
     "keywords": [
       "investing"
--- a/entries/p/personalcapital.com.json
+++ b/entries/p/personalcapital.com.json
@@ -12,7 +12,7 @@
     "custom-software": [
       "Personal Capital App"
     ],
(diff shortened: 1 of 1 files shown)


entries/p/personalcapital.com.json
```diff
@@ -4,11 +4,13 @@
     "url": "https://www.personalcapital.com",
     "tfa": [
       "call",
-      "email"
+      "email",
+      "sms",
+      "totp"
```
Carlgo11: According to https://support.personalcapital.com/hc/en-us/articles/360056729413 it only works with Personal Capital app. Can you confirm this?

max-rousseau: It does work with the Personal Capital app using either Push flow or totp code. The generated QR code shown in enrollment however is RFC compliant and can be also registered with any totp app.

Carlgo11: Okay so it sounds like both `totp` and `custom-software` should be set.
See [CONTRIBUTING.md](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#other-properties) for more info.

max-rousseau: Fair point, pushed update.

Carlgo11 (changes requested):
still needs the custom-software tag with `Personal Capital App` value.

Carlgo11 (approved):

phallobst (approved):

PR #6132, opened 2021-09-18
Add quality checks for entries

ApCoder123:
Added some checks to:
- Set a [notice message ](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-notice-message) when no documenation is provided to attach screenshots
- Set a warning message when `custom-property` is listed in `tfa`, but the `custom-property` field is not present.

I've set these to only run on newly created files.


Carlgo11 (commented):
The script fails when entries don't contain a tfa array.

Would you be happy with this layout instead?
```RUBY
#!/usr/bin/env ruby
# frozen_string_literal: true

require 'json'

# Fetch created (but not renamed) files in entries/**
diff = `git diff --name-only --diff-filter=Ar origin/master...HEAD entries/`.split("\n")

diff&.each do |path|
  entry = JSON.parse(File.read(path)).values[0]
  next unless entry.key? 'tfa'

  # rubocop:disable Layout/LineLength
  puts "::notice file=#{path} title=Missing Documentation:: Since there is no documentation available, please could you provide us with screenshots of the setup/login process as evidence of 2FA? Please remember to block out any personal information." unless entry['documentation']
  puts "::warning file=#{path}:: A `custom-software` tag is needed since it has been included in the `tfa` array." if entry['tfa'].include?('custom-software') && !entry['custom-software']
  puts "::warning file=#{path}:: A `custom-hardware` tag is needed since it has been included in the `tfa` array." if entry['tfa'].include?('custom-hardware') && !entry['custom-hardware']
  # rubocop:enable Layout/LineLength
end

```

Carlgo11 (approved):

ApCoder123:
Thanks, Carl. I completely forgot about entries for non 2fa supporting sites 🤦‍♂️.

phallobst (approved):

PR #6135, opened 2021-09-20
Add U2F to Outlook.com

kilasuit:
Outlook.com uses Microsoft Accounts which do support hardware tokens


kilasuit:
Closing this as from #5929 it seems you don't want to show where users can opt to use a HW token when it allows for simplified 2FA (something you know being the login & something you have being the token, which negates the need for a password)

Will reopen a PR adding comments to this regard instead

PR #6139, opened 2021-09-26
Update Dropbox

ApCoder123:


ApCoder123:
Looks like the URL validation test is failing, because dropbox is returning a 409 error due to an outdated user agent

From the response body:
 `<h1>Upgrade your browser to use Dropbox</h1> <div> <p>Your current browser is not supported. We support the latest versions of the following browsers:</p> `.

@Carlgo11, Would it maybe be better to only fail for specific status codes (e.g. 404) and warn for anything else?

Carlgo11:
@ApCoder123 ~it seems to be because of the user-agent.~ I don't really understand their reasoning for completely blocking old versions instead of trying to serve the content and let the client interpret as much as it can. Anyways, anything containing `(compatible; MSIE xx.xx)` gets blocked by Dropbox.

I specified a user-agent because sierra.com would 200 user-agents that are known to not support JS but since 503 responses are now just a soft fail I guess the user-agent can be changed or removed.
Either to a modern user-agent or to something signalling that it is our bot.

phallobst (approved):

Carlgo11 (approved):

PR #6145, opened 2021-09-28
Standardize user agents

Carlgo11:
This PR changes all user-agents used for tests to a more standardized user-agent containing the script name, HTTPClient version and Ruby version. I have tested this user-agent on sites known to block bots and it seems to work fine :thinking:

I also added a note about our bots for any webmasters.

(Also fixes #6139)

Here are screenshots of how bots.md would look like on 2fa.directory
light bg
dark bg

### DIFF
--- a/bot.md
+++ b/bot.md
@@ -15,4 +15,4 @@ As a result you will most likely only receive a couple of requests each year. We
 
 ## robots.txt
 
-Since each script only makes one request per website, the same number as if we would have fetched any robots.txt-file, we have opted to not comply with robots.txt-files.
+Since each script only makes one request per website, the same number as if we would have fetched any robots.txt file, we have opted to not comply with robots.txt files.
--- a/bot.md
+++ b/bot.md
@@ -1,0 +1,18 @@
+# 2FactorAuth Web bots
+
+In order to validate user contributions we use scripts or "bots".  
+These scripts only make requests to your website when someone tries to edit data about your site on 2fa.directory.   
+As a result you will most likely only receive a couple of requests each year. We would be very thankful if you didn't block these HTTP requests.
+
+## User agents:
+
+|User-Agent|Script source|
+|----------|-------------|
+|2FactorAuth/URLValidator|/tests/validate-urls.rb|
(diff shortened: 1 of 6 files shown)


Carlgo11:
4929edd overrides the function [default_redirect_uri_callback](https://github.com/nahi/httpclient/blob/master/lib/httpclient.rb#L717-L734).
The purpose of default_redirect_uri_callback is to fail when encountering [redirects from HTTP to HTTP](https://stackoverflow.com/a/23431136) as happened in #6146, or relative paths which according to https://github.com/nahi/httpclient/issues/385 shouldn't be displayed.

Therefore my conclusion is that overriding the function is completely safe for our usage.

bot.md
```diff
@@ -0,0 +1,18 @@
...
+|User-Agent|Script source|
+|----------|-------------|
+|2FactorAuth/URLValidator|/tests/validate-urls.rb|
+|2FactorAuth/LanguageValidator|/tests/language-codes.rb|
+|2FactorAuth/RegionValidator|/tests/region-codes.rb|
+|2FactorAuth/FacebookValidator|/tests/facebook.rb|
+
+## robots.txt
+
+Since each script only makes one request per website, the same number as if we would have fetched any robots.txt-file, we have opted to not comply with robots.txt-files.
```
ApCoder123:
```suggestion
Since each script only makes one request per website, the same number as if we would have fetched any robots.txt file, we have opted to not comply with robots.txt files.
```

ApCoder123 (commented):
According to this [MDN doc page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/From), the `From` header should be an email address rather than a url. Should we replace the current value with the contact email, or is it probably fine given that the 2fa.directory/bot page can be viewed?

Carlgo11:
@ApCoder123 if anyone would want to complain it would be better if they did that here on GitHub rather than to an email-address of the org.
I'd say we either leave it as is or replace it with our [issue link](https://github.com/2factorauth/twofactorauth/issues/new).

ApCoder123 (approved):
Probably fine to leave it as it is then

phallobst (approved):

PR #6148, opened 2021-09-30
Update jekyll exclusions & add site title

Carlgo11:
* Set a new site title to better reflect our new domain name.
* Removed obsolete file exclusions
* Grouped together similar excluded files


ApCoder123 (approved):
It may also be worth removing `*.webp` from the gitignore if we're tidying up obselete extensions.

phallobst (approved):

PR #6150, opened 2021-10-01
Workaround for old CAs in httpclient

anurag:
See https://github.com/nahi/httpclient/issues/445.

The default implementation uses a list of CAs from 2015, and resetting it to use default paths uses (newer) CA certificates from the underlying OS.


Carlgo11 (approved):
Thank you so much @anurag!
Sorry for the extra work you had to put in :sweat_smile:

ApCoder123 (approved):

PR #6154, opened 2021-10-01
Added U2F to Outlook since it has support

eriktheswede:
Outlook.com and Microsoft accounts have U2F support (and have had for quite some time) and was updated to reflect that.

### DIFF
--- a/entries/o/outlook.com.json
+++ b/entries/o/outlook.com.json
@@ -6,6 +6,7 @@
       "call",
       "email",
       "custom-software",
+	  "u2f",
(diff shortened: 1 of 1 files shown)


phallobst:
I think security keys are not used as a second factor there, but rather as a replacement for passwords ("Go passwordless").

eriktheswede:
I checked and you are correct. They aren't used as a second factor but as a FIDO2 passwordless authenticator, so it's used as both factors instead of in addition to a password.

Issue #3501, opened 2018-10-16
Capital One only has MFA for mobile


chubirka:
SwiftID isn't really two-factor auth. It's only called for certain activities known only to Capital One, that can't be set by the user.

Carlgo11:
Seems like a duplicate of #2994, #2891 and #3460.

Issue #2994, opened 2018-02-08
Capital One WEBUI 2FA is flawed


schwit61:
On the first login Capital One uses 2FA and identifies the source device via the browser’s user-agent. All subsequent logins using the same browser and device fail to use 2FA. I get the same results on multiple systems with multiple browsers. Clearing browser cookies does not help. The only thing that reenables 2FA is changing the browser’s user-agent and again only the first time you log in with that user-agent.

mxxcon:
Unfortunately, many sites work like that where they don't prompt you for 2fa on every login. :(

RichJeanes:
Please see my comment on #2898 and please do not open essentially duplicate issues for the same topic.

PR #6170, opened 2021-10-05
Update Capital One

### LINKED ISSUE #2891: Capital One 360 software 2FA is weak

Carlgo11:
Resolves #2891, #3501, #2994, #3460.

### DIFF
--- a/entries/c/capitalone.com.json
+++ b/entries/c/capitalone.com.json
@@ -1,16 +1,19 @@
 {
   "Capital One": {
     "domain": "capitalone.com",
-    "url": "https://www.capitalone.com/",
     "tfa": [
       "sms",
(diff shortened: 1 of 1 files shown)


phallobst:
What about these verification methods? They seem to be used for logins as well.
https://www.capitalone.com/support-center/bank/what-is-otp
https://www.capitalone.com/digital/mobile-app-verification/

dsoegijono (approved):

PR #6176, opened 2021-10-09
Update Sites - Backup

ApCoder123:
- Added recovery documentation
- Specified custom-* values
- Updated a few doc links


ApCoder123:
@Carlgo11 I think something might have gone wrong with overriding `redirect_uri_callback`?

Carlgo11:
@ApCoder123 😩
I'll have a look at it when I get home in an hour.

ApCoder123:
I've done some debugging and it looks like one of the redirects is to a relative path, which isn't handled by the new `redirect_uri_callback`. Adding the relevant code from [`default_redirect_uri_callback`](https://github.com/nahi/httpclient/blob/master/lib/httpclient.rb#L727) (which allows relative redirects) should solve the issue.

entries/m/mega.io.json
```diff
@@ -4,7 +4,7 @@
     "tfa": [
       "totp"
     ],
-    "documentation": "https://mega.io/blog_48",
+    "documentation": "https://mega.io/help/client/webclient/security-and-privacy#how-do-i-enable-two-factor-authentication-on-my-account-5bb525f9f1b70989348b45de",
```
phallobst:
```suggestion
    "documentation": "https://mega.io/help/client/webclient/security-and-privacy#5bb525f9f1b70989348b45de",
```

phallobst (changes requested):

entries/d/drive.google.com.json
```diff
@@ -8,7 +8,12 @@
       "totp",
       "u2f"
     ],
+    "custom-software": [
+      "Google Prompts via Android",
+      "Google Prompts via Compatible iPhone App (Smart Lock, Gmail, Google app)"
```
phallobst: That's a bit of a mouthful, isn't it? Wouldn't `Google Prompts` or `Google push notification` be sufficient? Nobody will be updating the list of compatible apps anyway.

phallobst (approved):

Carlgo11 (approved):

PR #6179, opened 2021-10-09
Fix relative location bug

Carlgo11:
Fixes bug found in #6176.

HTTPClient is starting to become rather annoying with all its bugs... (╯°□°)╯︵ ┻━┻

### DIFF
--- a/tests/validate-urls.rb
+++ b/tests/validate-urls.rb
@@ -11,6 +11,7 @@ status = 0
 # Fetch created/modified files in entries/**
 diff = `git diff --name-only --diff-filter=AM origin/master...HEAD entries/`.split("\n")
 
+# rubocop:disable Metrics/AbcSize
 def http_client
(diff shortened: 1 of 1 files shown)


tests/validate-urls.rb
```diff
@@ -4,12 +4,21 @@
...
 require 'json'
 require 'httpclient'
+require 'uri'
 
 # Exit code
 status = 0
 
 # Fetch created/modified files in entries/**
-diff = `git diff --name-only --diff-filter=AM origin/master...HEAD entries/`.split("\n")
+diff = `git diff --name-only --diff-filter=AM entries/`.split("\n")
```
phallobst: Was this intentional?

Carlgo11: Nice catch :facepalm:

phallobst (approved):

dsoegijono (approved):

PR #6180, opened 2021-10-10
Allow for multiple PNG resolutions

Carlgo11:
My intention was to just add support for different PNG dimensions.
Then I found things that could be cleaned up and now most of the script is rewritten. Whoops 😳

### DIFF
--- a/tests/validate-images.rb
+++ b/tests/validate-images.rb
@@ -2,31 +2,27 @@
 # frozen_string_literal: true
 
 require 'json'
-status = 0
 
-PNG_SIZE = [32, 32].freeze
-
+@status = 0
+PNG_SIZE = [[32, 32], [64, 64], [128, 128]].freeze
 seen_sites = []
 
+def error(msg)
+  puts "::error file=#{@file}:: #{msg}"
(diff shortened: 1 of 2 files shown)


tests/validate-images.rb
```diff
@@ -2,31 +2,27 @@
...
-
-PNG_SIZE = [32, 32].freeze
 
+@status = 0
+PNG_SIZE = [[32, 32], [64, 64], [128, 128]].freeze
 seen_sites = []
 
+def error(msg)
+  puts "::error file=#{@file}:: #{msg}"
+  @status = 1 unless @status.eql? 1
```
dsoegijono: So we expect `@status` to be 1 after this right? Do we need the unless check?

Carlgo11: @dsoegijono the idea was that the unless-statement would aid the interpreter in skipping the line quicker on subsequent runs.
From examining it now however, it looks like it would probably have no impact as the interpreter will most likely skip the unless-statement and not the entire line.

dsoegijono (approved):

phallobst (approved):

PR #6182, opened 2021-10-10
Add NiceHash

archit11111:
Nicehash has Alexa Rank - 2835 and supports 2FA.


entries/n/nicehash.com.json
```diff
@@ -0,0 +1,15 @@
+{
+  "NiceHash": {
+    "domain": "nicehash.com",
+    "img": "nicehash.com.png",
```
phallobst: You could use this SVG icon instead:


```suggestion
```

phallobst (changes requested):

entries/n/nicehash.com.json
archit11111: Done

phallobst (approved):

entries/n/nicehash.com.json
```diff
@@ -0,0 +1,14 @@
+{
+  "NiceHash": {
+    "domain": "nicehash.com",
+    "tfa": [
+      "totp",
+      "u2f"
+    ],
```
ApCoder123:
```suggestion
    "tfa": [
      "totp",
      "custom-hardware"
    ],
    "custom-hardware": [
      "Yubico OTP"
    ],
```

The proprietary Yubico OTP system is being used rather than U2F. See https://www.nicehash.com/support/general-help/security/how-to-add-otp-device

ApCoder123 (changes requested):

entries/n/nicehash.com.json
archit11111: Done. I was confused. So, u2f should be used when it supports all forms rather than just specific hardware like yubikey, right?

ApCoder123: `u2f` should be used when a site supports U2F/WebAuthn, however this site only supports a proprietary method that is compatible with YubiKeys, called Yubico OTP, which is why the `custom-hardware` value was needed.

ApCoder123 (approved):

PR #6191, opened 2021-10-12
Remove redundant code

ApCoder123:
This `<a>` tag doesn't seem to do anything


Carlgo11:
It's there so that URI fragments work for categories.

PR #6192, opened 2021-10-12
Clean up custom-hardware values

ApCoder123:
These entries are appearing in search results for "Apple", which is probably not the desired outcome.


phallobst:
We have the same problem with `custom-software` as well. Searching for e.g. "push" brings up unexpected entries as well.

Carlgo11 (approved):

phallobst (approved):

PR #6194, opened 2021-10-14
Remove Braziliex

phallobst:
see https://braziliex.com/blog/comunicados/braziliex-encerramento-atividades/


ApCoder123 (approved):

phallobst:
I'm not quite sure which is the better diff option: `--diff-filter=Ard` or `--diff-filter=A`

Carlgo11 (approved):

PR #6209, opened 2021-10-17
Add Metro Bank

ApCoder123:


entries/c/commercial.metrobankonline.co.uk.json
```diff
@@ -0,0 +1,25 @@
+{
+  "Metro Bank Commercial": {
+    "domain": "commercial.metrobankonline.co.uk",
+    "url": "https://metrobankonline.co.uk",
```
ApCoder123: Could setting the url as an exisiting domain potentially cause issues for users matching sites using API v2?

Carlgo11: Potentially but we haven't made any statements that the value would be unique so any bugs would be due to improper implementation.

entries/c/commercial.metrobankonline.co.uk.json
```diff
@@ -0,0 +1,25 @@
...
+      "custom-hardware",
+      "custom-software"
+    ],
+    "custom-software": [
+      "Metro Bank Authenticator app"
+    ],
+    "custom-hardware": [
+      "Security Device"
+    ],
+    "documentation": "https://www.metrobankonline.co.uk/help-and-support/what-is-sca/",
```
Carlgo11: Sounds like they only prompt for OTP codes on transactions and not login attempts.

ApCoder123: From what I can tell, the `custom-software` methods can be used for authentication, rather than authorisation:

https://www.metrobankonline.co.uk/help-and-support/what-is-sca/#item17066
> It’s an app you can use instead of your security device to prove it’s you when logging in or making changes on your Business Online Plus and Commercial Online Banking. You can use the app to:
> 
> log in using a different device or browser
> make certain payments and transactions including online shopping
> make admin preference/access changes
> The app is only available to those with a UK mobile number.

Carlgo11 (approved):

entries/c/commercial.metrobankonline.co.uk.json
```diff
@@ -0,0 +1,20 @@
+{
+  "Metro Bank Commercial": {
+    "domain": "commercial.metrobankonline.co.uk",
+    "img": "metrobankonline.co.uk.svg",
+    "tfa": [
+      "custom-software"
```
phallobst: What about the "security device" (i.e. `custom-hardware`)?

See https://www.metrobankonline.co.uk/help-and-support/what-is-sca/#item15017
> You’ll need to use your security device or your Metro Bank Authenticator App to log in, ...

phallobst (approved):

PR #6200, opened 2021-10-16
Update win2day

phallobst:

PR #6213, opened 2021-10-17
Add regions to investing

ApCoder123:


entries/p/prudential.com.json
```diff
@@ -1,12 +1,14 @@
 {
   "Prudential Financial": {
     "domain": "prudential.com",
-    "url": "https://www.prudential.com",
```
Carlgo11: This line needs to be here.

phallobst: I'm afraid that's the same problem as in #6200 - neither URL does work.

Carlgo11: https://prudential.com didn't work for me anyway :shrug:

I'm not sure how to debug the failed test. For starters I can't find what DNS servers the GitHub workflow uses :thinking:

entries/i/ig.com.json
```diff
@@ -1,14 +1,16 @@
 {
```
Carlgo11, line 1: Is it worth listing these countries?

ApCoder123: I didn't include those, as it did seem to allow registrations from any country - see "IG International" at the bottom of that page and https://www.ig.com/en/application-form.

entries/p/prudential.com.json
Carlgo11: After much debugging and some reverse shells into the github workflow I've found that the problem is indeed due to Microsoft Azure's DNS servers. Setting the nameservers to `1.1.1.1` in `/etc/resolv.conf` fixes the problem. Oddly enough it only happens when running the command inside docker on Microsoft Azure :thinking:

Carlgo11 (approved):

entries/w/wealthsimple.com.json
```diff
@@ -10,6 +10,10 @@
     "documentation": "https://help.wealthsimple.com/hc/en-ca/articles/360058451253",
     "keywords": [
       "investing"
+    ],
+    "regions": [
+      "gb",
```
phallobst: See https://help.wealthsimple.com/hc/en-gb/articles/4411041188891
```suggestion
```

phallobst (changes requested):

phallobst (approved):

PR #6227, opened 2021-10-23
Add Tellonym

sn00py1310:


Carlgo11 (changes requested):
SVGs are preferred over PNGs. Here's an SVG file with the proper proportions:

sn00py1310 (commented):
Updated to SVG, thanks for the proper file.

Carlgo11 (approved):

phallobst (approved):

Issue #6225, opened 2021-10-23
Add *.live.com


sn00py1310:
### Site name

Outlook - Microsoft

### Site URL

https://outlook.live.com/owa/

### Documentation URL

https://support.microsoft.com/en-us/help/12408/

### Supported 2FA methods

- [X] TOTP (RFC-6238, Google Authenticator)
- [X] Proprietary Software (Authy, Duo, etc.)
- [x] FIDO2/U2F/WebAuthn security keys
- [ ] Proprietary hardware (Yubikey OTP, RSA keys, etc.)
- [X] Phone calls
- [X] SMS tokens
- [X] Email tokens

### Eligibility

- [X] I have checked that the site meets 2fa.directory's [contributing guidelines and site criteria](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md)
- [X] The site is ranked within the top 200,000 sites on [Alexa](https://www.alexa.com/siteinfo/) or has a substantial social media following.
- [X] The site does not contain or promote content that violates 2fa.directory's [excluded categories and websites guidelines](https://github.com/2factorauth/twofactorauth/blob/master/EXCLUSION.md), including pornographic, discriminatory, or unlawful content.
- [X] The issue I'm creating is not a duplicate of an existing [issue](https://github.com/2factorauth/twofactorauth/issues).
- [X] The issue I'm creating is not a duplicate of an existing [pull request](https://github.com/2factorauth/twofactorauth/pulls)

### Additional information

Notes: `2FA phone calls may not be available in all regions. Not all FIDO2 keys work because of custom requirements.`
FIDO2 note: [Blog post](https://www.microsoft.com/en-us/microsoft-365/blog/2018/11/20/sign-in-to-your-microsoft-account-without-a-password-using-windows-hello-or-a-security-key/) and [specs](https://docs.microsoft.com/en-us/windows/security/identity-protection/hello-for-business/microsoft-compatible-security-key)

Not sure if it's a similar case to #3698. The new scheme in #6185 would fit best.

phallobst:
We already list this site as "Outlook.com".

For U2F support, please also see #6154 - this does not constitute 2FA, but allows for a passwordless login.

sn00py1310:
@phallobst maybe `outlook.live.com` or `*.live.com` should be added in as "additional-domains" for "Outlook.com", because it redirects to outlook.live.com.

PR #6235, opened 2021-10-26
Update Outlook.com

phallobst:
See also #6225

### DIFF
--- a/entries/o/outlook.com.json
+++ b/entries/o/outlook.com.json
@@ -2,7 +2,8 @@
   "Outlook.com": {
     "domain": "outlook.com",
     "additional-domains": [
-      "outlook.live.com"
+      "outlook.live.com",
+      "hotmail.com"
     ],
     "tfa": [
       "sms",
--- a/entries/o/outlook.com.json
+++ b/entries/o/outlook.com.json
@@ -1,12 +1,18 @@
 {
   "Outlook.com": {
     "domain": "outlook.com",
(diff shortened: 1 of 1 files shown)


sn00py1310 (approved):

entries/o/outlook.com.json
```diff
@@ -1,17 +1,23 @@
 {
   "Outlook.com": {
     "domain": "outlook.com",
+    "additional-domains": [
+      "outlook.live.com"
```
Carlgo11: Won't hurt adding hotmail as well. There's probably some edge case where they still use that domain :roll_eyes: 
```suggestion
      "outlook.live.com",
      "hotmail.com"
```

Carlgo11 (changes requested):

entries/o/outlook.com.json
phallobst: "hotmail"... That's a name I haven't heard in a while.

Do we really want `live.com` here? This could clash with "OneDrive" (`onedrive.live.com`).

Carlgo11: I didn't realize they used the live.com domain for anything else. The suggestion is updated now.

Hotmail is still very much alive and sadly being used. Their servers on the other hand are very much [**not** alive](https://postmaster.live.com).

ApCoder123 (approved):

Carlgo11 (approved):

PR #6262, opened 2021-11-08
Add more region tags 

Carlgo11:


ApCoder123:
Some of the `investing` entries are duplicated from #6213

entries/a/alaskaair.com.json
```diff
@@ -8,6 +8,14 @@
     },
     "keywords": [
       "transport"
+    ],
+    "regions": [
+      "ca",
+      "cu",
```
ApCoder123:
```suggestion
```
Alaska Airlines no longer fly to Cuba - see https://newsroom.alaskaair.com/2017-11-14-Alaska-Airlines-will-discontinue-flying-to-Havana-Cuba

entries/t/thalia.de.json
```diff
@@ -12,31 +11,10 @@
...
-      "hr",
-      "cz",
-      "lt",
-      "ro",
-      "lv",
-      "mt",
-      "si",
-      "is"
+      "li",
+      "lu"
```
ApCoder123, line 17: Whats the reasoning behind listing only these 4 countries? https://www.thalia.de/shop/hilfe-versand/show/#D2

entries/u/ukpostbox.com.json
```diff
@@ -7,6 +7,9 @@
     ],
     "keywords": [
       "postal"
+    ],
+    "regions": [
+      "gb"
```
ApCoder123: I don't think this entry should have a `gb` region tag based on the website's description:

> Get a personal or business address in the UK regardless of where you are in the world. Read your post online or forward it internationally - a smarter post office.

entries/u/uniteddomains.com.json
```diff
@@ -6,9 +6,12 @@
     "tfa": [
       "totp"
     ],
-    "documentation": "https://help.uniteddomains.com/hc/en-us/articles/115001458813",
+    "documentation": "https://www.uniteddomains.com/twofactorauth/",
     "keywords": [
       "domains"
+    ],
+    "regions": [
+      "us"
```
ApCoder123: This site seems to offer its services internationally

entries/t/thalia.de.json
Carlgo11: While the site ships to most EU countries, the site is only available in German which would mean that's their primary market. That was my reasoning at least

entries/u/ukpostbox.com.json
Carlgo11: Ah my bad :see_no_evil:

phallobst:
It seems to me that the line `uri = URI.parse(uri)` in `validate-urls.rb` causes the checks to fail. The `uri` lambda parameter is not a string, but a URI object and needs no further parsing.

An alternative would be `uri = URI.parse("#{uri}")`, I guess - but I don't know if we have a case that requires this.

Carlgo11:
@phallobst but it's facebook.rb that's failing :confused:

phallobst:
D'oh! I didn't even look and just ran the tests myself, and they failed there...

Carlgo11:
I put a `next if response.header['location'].nil?` in `tests/facebook.rb`. For some reason Facebook isn't returning a Location header for the conradcom page. This seems to only happen on CircleCI and I am thus not able to debug it further.

entries/c/copaair.com.json
```diff
@@ -9,6 +9,15 @@
     },
     "keywords": [
       "transport"
+    ],
+    "regions": [
+      "ar",
+      "br",
+      "ca",
+      "co",
```
phallobst, line 23:
```suggestion
      "cl",
      "co",
      "cr",
```

entries/f/flyfrontier.com.json
```diff
@@ -8,6 +8,13 @@
     },
     "keywords": [
       "transport"
+    ],
+    "regions": [
+      "do",
+      "sv",
+      "gt",
+      "mx",
+      "us"
```
phallobst: Current list according to [Wikipedia](https://en.wikipedia.org/wiki/List_of_Frontier_Airlines_destinations):
```suggestion
      "ag",
      "bs",
      "bz",
      "cr",
      "cu",
      "do",
      "gt",
      "jm",
      "mx",
      "nl",
      "pr",
      "sv",
      "us"
```

phallobst (changes requested):

phallobst (approved):

entries/c/copaair.com.json
```diff
@@ -9,6 +9,17 @@
...
+    "regions": [
+      "ar",
+      "br",
+      "ca",
+      "cl",
+      "co",
+      "cr",
+      "mx",
+      "pa",
+      "us"
```
ApCoder123:
```suggestion
      "ar",
      "aw",
      "br",
      "bs",
      "bb",
      "bz",
      "bo",
      "ca",
      "cl",
      "co",
      "cu",
      "cr",
      "cw",
      "do",
      "ec",
      "gt",
      "gy",
      "ht",
      "hn",
      "jm",
      "mx",
      "ni",
      "pa",
      "py",
      "pe",
      "pr",
      "sx",
      "sr",
      "sv",
      "tt",
      "us",
      "uy",
      "ve"
```
https://destinationsguide.copaair.com/en/

ApCoder123 (approved):

PR #6265, opened 2021-11-10
Update NAB

jimmyvanau:


jimmyvanau:
Let me know if I should remove the URL because the domain entry redirects appropriately

Carlgo11:
From reading the included documentation it sounds like they offer [authorization](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#authorization) rather than authentication.

jimmyvanau:
That would be right, it's authorisation, not authentication. Let me go and annoy the Aussie banks.

PR #6272, opened 2021-11-12
Update Udemy

phallobst:


Carlgo11 (approved):

Carlgo11 (approved):

ApCoder123 (commented):
This reminds me of the issue brought up in #4306. Is it maybe better to create a separate entry for instructors under `teach.udemy.com` as "Udemy (for Instructors)" in order to avoid confusion? It's probably not ideal for API data, but could be slightly better on the webite? What do you think?

PR #6275, opened 2021-11-12
Add Bankera

40504:

PR #6278, opened 2021-11-13
Allow excluding regions

ApCoder123:
Currently, if a site is available everywhere in the world apart from one region, it is necessary to list all region codes in the `regions` array apart from that region to exclude it from that single region's site. This PR modifies the `regions` array to allow excluding regions to prevent sites that aren't available in certain regions from being displayed on their regional sites.

Examples of entries would be:
- Vanguard (shouldn't be displayed on AU or UK site as they have separate entries)
- Bankera (#6275) which is available everywhere [apart from the US](https://bankera.com/faq/is-bankera-available-globally/)


Carlgo11:
I like the idea! What if instead of a new array we used `-` as a suffix to the region?
Like:
```json
"regions" : [
  "-us"
]
```
That would make it look neater if/when we add region groups like europe and north america.

Carlgo11 (approved):

phallobst (approved):

PR #6283, opened 2021-11-13
Change category icons based on regional site

ApCoder123:
Thought it would be cool to display different currency symbols for the `banking` category based on the regional site accessed.

Also changed `finance` icon

### DIFF
--- a/_data/categories.json
+++ b/_data/categories.json
@@ -67,7 +67,7 @@
   {
     "name": "finance",
     "title": "Finance",
-    "icon": "fas fa-dollar-sign"
+    "icon": "fas fa-file-invoice-dollar"
   },
   {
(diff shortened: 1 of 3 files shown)


phallobst:
:+1: for the "Finance" icon.

Using different currencies for "Banking" is a cute gimmick, but unfortunately does not translate to the "Crowdfunding" or the (new) "Finance" icon, so it's not much of a bonus, I guess.

If we **do** want to support that, the icon definition should rather be in the `categories.json`, don't you think?

ApCoder123:
I agree that the icon definition should be in `categories.json` and I guess the "Crowdfunding" and "Finance" icons will look a bit inconsistent if we decide to go ahead with this.

PR #6296, opened 2021-11-17
Update N26

ApCoder123:


entries/n/n26.com.json
```diff
@@ -5,12 +5,22 @@
...
     "regions": [
-      "de"
+      "at",
+      "br",
+      "fr",
+      "de",
+      "it",
+      "es",
+      "ch",
+      "us"
```
phallobst: Given that one of the region choices on there is "Europe", this list is definitely not complete. Should we drop it or add the rest of the countries here?

phallobst: Also: https://n26.com/en-de/press/press-release/n26-to-discontinue-us-operations

```suggestion
```

phallobst (changes requested):

entries/n/n26.com.json
ApCoder123: I've added region codes based on [this page](https://support.n26.com/en-eu/account-and-personal-details/opening-an-account/can-i-open-an-n26-account-in-my-country) in https://github.com/2factorauth/twofactorauth/pull/6296/commits/ff35dc57b4c372ac254bb805bd4bd207ac3e594f

Carlgo11 (approved):

phallobst (approved):

moviuro:
@ApCoder123 I have serious doubts about if N26 *truly* supports 2FA. Can you please confirm what happens when you:

1. login to your account through a private/incognito window
2. close all private/incognito windows
3. login to your account through a second private/incognito window

If you aren't asked for the 2FA code **twice**, then it's not 2FA.

I'm having very serious doubts because my own 2 banking accounts in France (HelloBank! and LCL) *don't* do it properly.

PR #6305, opened 2021-11-19
Fix excluded regions v3

ApCoder123:
Summary:
- #6300 did nothing useful
- #6304 partially solved the problem
- The `site_excluded_regions` still held the "-regioncode" format, so when checking if that array included `region["id"]`, it was failing, meaning excluded regions weren't being detected.
- When there was an array with exclusion codes, `site_regions` was still being initialised, so wasn't `nil`, so the `next` statement was being triggered, resulting in excluded sites not being displayed on any regional site (other than `int`)

Test cases used locally:
- Bankera (not seen on `us`, but seen on all other regional sites)
- Vanguard (not seen on `au`, `gb`, but seen on all other regional sites)
- HSBC (not seen on `ar`, but seen on all other regional sites)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -213,6 +213,28 @@ government site - you can note this with the `regions` field.
 
 The country codes should be lowercase [ISO 3166-1][iso-country-wikipedia] codes.
 
+#### Excluded Regions
+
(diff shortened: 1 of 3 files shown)


phallobst:
Now that it works properly, maybe a sentence or two in the documentation would be helpful.

phallobst (approved):

Carlgo11 (approved):

PR #6308, opened 2021-11-20
Add Cfx.re (FiveM/RedM) Forums

mcastiglia:
There is no documentation site, but here is an example of what the 2FA settings page looks like:
image


Carlgo11:
Forums aren't something we list.

Issue #6312, opened 2021-11-22
Riot Games is forced email 2fa real 2fa?


sn00py1310:
Should Riot Games email 2fa be considered a real 2fa? Because [here](https://www.reddit.com/r/VALORANT/comments/q0ox4v) a Reddit user claims it isn't required on every login.

> Well, they do! But they also don't. Login through a browser to your Riot Games ID from an unverified hardware ID or browser user agent may require you to have a code sent to the account's verified email address as a form of 2FA.

Riot mentions this step [here](https://support-leagueoflegends.riotgames.com/hc/articles/360010367273). But it isn't required for the login to play the game.

Carlgo11:
More information is needed for us to make a decision here.

It could still be regarded as functioning 2FA if a user is required to enter the 2FA codes on all new devices (i.e cleared cookies / new IP-address).

If the 2FA prompt is more sporadic than that then it most likely wouldn't qualify as functioning 2FA.

sn00py1310:
@Carlgo11 I tested a bit with the Riot Games login.
It appears that only for `account.riotgames.com` the email OTP is required. The desktop client and the sign in button on `www.riotgames.com` don't need it.

So account hijacking is 2fa save but impersonating is still possible.

Carlgo11:
Yikes that sounds really bad.
I think it's a clear case of [authorization and not authentication](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#authorization). If you'd like you can create a PR changing the status of their 2FA. Otherwise I'm sure someone else will.

Thanks for bringing this up! :+1:

PR #6315, opened 2021-11-23
Add Riot Games

sn00py1310:
Discussesd in #6312.


Carlgo11:
LoL is already listed so that would need to be removed before this can be merged.

Carlgo11 (dismissed):

phallobst:
Do we really want to remove LoL which has 2FA support (albeit email only) and "hide" it in this entry without 2FA?

Carlgo11:
@phallobst The auth system for LoL is the same as Riot Games so it's not actual 2FA but 2SV.

dsoegijono (commented):
Looks like they've added it now? https://www.riotgames.com/en/news/multi-factor-authentication-has-arrived

sn00py1310:
I used the Valorant documentation link because they have a different support page for every game, but the content seems to be the same.

phallobst (approved):

Carlgo11 (approved):

PR #6324, opened 2021-11-28
Add Typeracer

sn00py1310:


Carlgo11 (changes requested):
While SVGs are certainly preferred in most cases, this time I think a PNG with 1/14th the size would be a preferable option.
Please replace the SVG with this image:

Carlgo11 (approved):

phallobst (approved):

PR #6337, opened 2021-12-06
Add Gitea

6543:
https://gitea.com/


6543:
`gitea.com is ranked above the maximum rank of 200K` - what does that mean?

PS: gitea.com is only one instance of "Gitea" - all Gitea instances do support this - so that's the main intention to get it into the db, same as gitlab is ...

Carlgo11:
Hi @6543, have you read the contribution.md and exceptions.md?
Those should answer your question

6543:
@Carlgo11 hi :wave: 

> Alexa top 200K: A new site that is not already listed has to be within the Alexa top 200,000 ranking. ...

well https://www.alexa.com/siteinfo/gitea.io would make it - but that's only for the blog/docu of gitea

6543:
@Carlgo11 should I use `gitea.io` as domain even if you can only log into gitea.com?

Carlgo11:
There are several issues with this PR and the eligibility of the listed domains: 
1. Self-hosted software is on our [exclusion list](https://github.com/2factorauth/twofactorauth/blob/master/EXCLUSION.md#self-hosted-services).
2. Gitea.com doesn't meet the [Alexa ranking](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#guidelines) requirement meaning that domain can't be listed.
3. The documentation page you linked to doesn't explain how to enable 2FA.

6543:
as you pointed out, there is more that one issue - I probably cant solve that easy.

So unless gitea.com is to low in the ranking - I'll "pause" this request.
thanks @Carlgo11 for looking at it anyway :)

Issue #5516, opened 2021-04-03
First Tech Federal Credit Union FTFC current issues


symbioquine:
- [x] I have searched [open issues and pull requests](https://github.com/2factorauth/twofactorauth/issues?q=is%3Aopen). The issue I'm creating is not a duplicate of an existing open issue or pull request.

### Information about the bug: ###
**Bug description:**

![image](https://user-images.githubusercontent.com/30754460/113484051-38c06600-945b-11eb-936f-06a975fb8ff4.png)

* Does support `sms`, `phone`, `email` "second factors"
* Broken docs link - it's unclear what the replacement URL should be though; https://github.com/2factorauth/twofactorauth/blob/336b423e04ac4bb421aae59233e2b8b909761ba5/_data/banking.yml#L687
* Info icon says "proprietary software based 2FA" but should instead say "proprietary hardware based 2FA" - these can be requested through the online banking portal and look like this;
  ![image](https://user-images.githubusercontent.com/30754460/113484196-e3d11f80-945b-11eb-8510-d3e1cd3b27bb.png)

**Bug reproduction:**

N/A - content issues

**Impacted devices:**

All

**Expected behavior:**

Content should be accurate

**Additional information:**

Thanks for maintaining this resource!

ApCoder123:
> Broken docs link - it's unclear what the replacement URL should be though

Perhaps this page as a doc link: https://www.firsttechfed.com/help/security/prevent#multifunction

symbioquine:
Oh, yeah that looks like a reasonable replacement.

ApCoder123:
> Does support sms, phone, email "second factors"

Since there is not information about this in the documentation available, please could you provide us with screenshots of the setup process? Please remember to block out any personal information.

symbioquine:
I personally only have hard tokens enabled, but I've captured screenshots showing where other "second factors" could be enabled in the settings. When they are enabled it gives the option of using any of the enabled ones at login.

![image](https://user-images.githubusercontent.com/30754460/114047403-84e61e80-983e-11eb-8a56-6ddb946c94c1.png)
![image](https://user-images.githubusercontent.com/30754460/114047479-8f081d00-983e-11eb-9647-de119e5ecb8c.png)
![image](https://user-images.githubusercontent.com/30754460/114047507-94fdfe00-983e-11eb-823a-410dff162176.png)

> Info icon says "proprietary software based 2FA" but should instead say "proprietary hardware based 2FA"

[...]

PR #6345, opened 2021-12-11
Specify firsttechfed.com custom-software implementation

stephenjust:
First Tech Fed uses Entrust Identity Guard as their solution for soft tokens.

### DIFF
--- a/entries/f/firsttechfed.com.json
+++ b/entries/f/firsttechfed.com.json
@@ -9,6 +9,9 @@
       "call",
       "email"
     ],
+    "custom-software": [
+      "Entrust Identity Guard"
+    ],
     "documentation": "https://www.firsttechfed.com/help/security/prevent#multifunction",
     "keywords": [
       "banking"
--- a/entries/f/firsttechfed.com.json
+++ b/entries/f/firsttechfed.com.json
@@ -12,6 +12,9 @@
     "custom-software": [
       "Entrust Identity Guard"
     ],
+    "custom-hardware": [
+      "Entrust Hardware Token"
+    ],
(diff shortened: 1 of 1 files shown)


Carlgo11:
Would you happen to know the name of their hardware keys as well?

The [documentation](https://www.firsttechfed.com/help/security/prevent#browse) page is hilarious to read! Not only do they advice people to look for the deprecated EV cert browser text but they also don't even use EV certs! It's a bit hard to verify the EV text on a Lets Encrypt certificate :laughing:

stephenjust:
Entrust is their vendor for hardware tokens as well - or at least that's what their customer service suggested. I'm much less confident in this though so I didn't include it.

entries/f/firsttechfed.com.json
```diff
@@ -9,6 +9,9 @@
       "call",
       "email"
     ],
+    "custom-software": [
+      "Entrust Identity Guard"
+    ],
```
Carlgo11, line 14: After reviewing the data you've provided and what was previously provided at #5516 I think `Entrust Hardware Token` can be added as custom-hardware.
The previous submitter also mentioned sms, phone calls and email 2FA. Is that still something that is provided?

```suggestion
    ],
    "custom-hardware": [
      "Entrust Hardware Token"
    ],
```

Carlgo11 (changes requested):

entries/f/firsttechfed.com.json
stephenjust: Yes, the other data is still correct.

Carlgo11 (approved):

phallobst (approved):

PR #6348, opened 2021-12-15
Focus and select search input

jareha:
# Changes
* Focus and select search input on page load.
* Copyedit comments.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -1,6 +1,13 @@
 $(document).ready(function () {
-  // Focus and select search input
-  $('#innerSearchBox').focus().select();
+  // Execute when window resized
+  $(window).resize(function() {
+    const width = $(window).innerWidth(); // New width
+
+    if (width >= 992) {
+      // Focus and select search input
+      $('#innerSearchBox').focus().select();
+    }
+  });
   
   // Show popup notice
   $('.exception').popup({ position: 'right center', hoverable: true, title: 'Exceptions & Restrictions' });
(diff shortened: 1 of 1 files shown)


Carlgo11:
This would cause all devices with on-screen keyboards to activate the keyboards on site load. I don't think that's a desirable thing.

jareha:
Fair point. I checked a couple search engines and found, to your point, they don‘t focus on mobile, but do on desktop:

* Bing
* DuckDuckGo
* Google
* Yahoo

How would y‘all feel about limiting this functionality to desktop, then?

(Might be why I‘ve only wanted this change on 2FA Directory, as it‘s one of the only sites I use exclusively on desktop.)

Carlgo11:
If you find some way to determine that reliably then I see no problem with adding that feature.

jareha:
How do you feel about the following?

```JavaScript
$(document).ready(function () {
  // This will execute whenever the window is resized
  $(window).resize(function() {
    const width = $(window).innerWidth(); // New width

    if (width < 993) {
      // Focus and select search input
      $('#innerSearchBox').focus().select();
    }
  });

  // Rest of block
});
```

ApCoder123:
The width check condition should be `width>=992` in order to only focus on desktop (when width is greater than `lg` breakpoint).

jareha:
@ApCoder123, @Carlgo11 Apologies for the delay. I’ve incorporated the discussed [desktop-only configuration](https://github.com/2factorauth/twofactorauth/pull/6348#issuecomment-996243469) along with @ApCoder123’s [suggested change](https://github.com/2factorauth/twofactorauth/pull/6348#issuecomment-997432757).

Carlgo11 (changes requested):
I think the IF statement would make more sense being in `resizeObserver` instead of being inside a new event listener. Preferably also without the width var being declared and only used once.

The code cleanup in other areas of the file feels like a merge conflict waiting to happen so sticking with one change per PR would be appreciated.

Do you want to make the changes @jareha or would you rather want me to just push the changes to your branch?

jareha:
@Carlgo11 Changes pushed.

phallobst (approved):

Carlgo11 (approved):

PR #6355, opened 2021-12-22
Update termius

mcsitter:
Termius supports 2FA using TOTP. This is not stated in the documentation, but it is there. The documentation also links to the correct [url to change the 2FA settings](https://account.termius.com/).

### DIFF
--- a/entries/t/termius.com.json
+++ b/entries/t/termius.com.json
@@ -2,6 +2,7 @@
   "Termius": {
     "domain": "termius.com",
     "tfa": [
+      "totp",
       "custom-software"
     ],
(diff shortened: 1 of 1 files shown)


Carlgo11:
Since the page you're referring to requires authentication to view, please attach a screenshot of it so that we can verify TOTP is available.

mcsitter:
Happy to do so! Here you go. A screenshot of the 2FA setup page. You get there by clicking on "Two-factor Authentication" on the [account details page](https://account.termius.com/).

image

Carlgo11 (approved):
Thanks for the screenshot.
I've added Authy in the custom-software array to fix the build warning.

phallobst (approved):

PR #6368, opened 2021-12-30
Add Linus Tech Tips (LTT) Forum. 

ACSimon33:
I couldn't find any official 2FA documentation, only some 2FA related topics in the forum itself.

I added the abbreviation in the name because it's widely used in the community.


Carlgo11:
Unfortunately, we do not list forums on 2fa.directory.

Sorry for the inconvenience.

PR #6401, opened 2022-01-10
Add Zip

fletcherdodd:
- Documentation only states 2FA is used when logging in from an unrecognised device
- 2FA is enabled by default when you set up your account (a phone number is needed and you must verify it when recreating an account). Didn't know whether or not to include this in the notes.
- See below for when you log in from an unrecognised device

Zip


entries/z/zip.co.json
```diff
@@ -0,0 +1,21 @@
+{
+	"Zip": {
+		"domain": "zip.co",
+		"tfa": [
+			"sms"
+		],
+		"notes": "2FA is only used when an account login is made from an unrecognised device; user cannot enable 2FA for every login.",
```
phallobst: Usually a device is identified by a cookie, so deleting the cookies may reset the state to "unrecognised". In that case, we don't really need this note, I reckon.
```suggestion
```

phallobst (changes requested):

img/z/zip.co.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 768 285"><path fill="#000" d="m19.53 227.27 6.66 54.26h233.17l-7.63-62.14H143.01l-.95-7.7 100.13-69.97-6.69-54.36H2.35l7.63 62.13h108.9l.95 7.77-100.3 70.01Z"/><path fill="#AA8FFF" d="m262.59 87.36 23.85 194.17h233.34L495.93 87.36H262.59Z"/><path fill="#000" d="M764.47 157.26c-5.38-43.64-39.66-70.09-86.26-69.9H523l23.84 194.18h69.83l-4.78-38.84h73.89c58.15-.01 84.76-36.24 78.69-85.44Zm-86.24 31-73 .08-5.72-46.6 73.41.06c17.26.21 26.09 9.93 27.52 23.23.88 8.56-3.04 23.22-22.21 23.22v.01ZM332.303 61.56c12.281-13.184 10.14-35.139-4.781-49.038-14.921-13.9-36.973-14.48-49.253-1.296-12.281 13.184-10.14 35.139 4.781 49.038 14.921 13.9 36.973 14.48 49.253 1.296Z"/></svg>
```
Carlgo11: I think the favicon works better:
```suggestion
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100" style="background:#aa8fff;fill:#1a0826"><path d="m19.68 62.062 1.812 15.284H85l-2.077-17.499H53.312l-.257-2.172 27.27-19.706-1.82-15.316H15l2.078 17.504h29.661l.265 2.187z"/></svg>
```

entries/z/zip.co.json
```diff
@@ -0,0 +1,20 @@
+{
+	"Zip": {
+		"domain": "zip.co",
+		"tfa": [
+			"sms"
+		],
+		"documentation": "https://zip.co/security",
```
Carlgo11: The documentation page was great to verify that the service indeed supports 2FA but I don't think it will tell users anything of value. I recommend you remove the documentation link.
```suggestion
```

Carlgo11 (changes requested):

entries/z/zip.co.json
fletcherdodd: I did think it was a little pointless to add to considering you don't have to do anything to enable it. But wasn't too sure

img/z/zip.co.svg
fletcherdodd: Agreed, I was trying to find that the more simplified one. But could only see the one I added in their brand assets.

Carlgo11 (approved):

phallobst (approved):

PR #6405, opened 2022-01-12
Update Mojang

fletcherdodd:


Carlgo11 (approved):

fletcherdodd:
@Carlgo11 U2F is available in the advanced security options of your Microsoft account while, fyi. I have 2 setup on mine.

image

Carlgo11:
@fletcherdodd that would be FIDO2 "passwordless" authentication. See [this note](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#passwordless-authentication) on how we classify "passwordless" authentication.

phallobst:
Maybe the main domain should be `minecraft.net` instead of `mojang.com`, with "Minecraft" as the entry name.

fletcherdodd:
I did think that, but the only svg that was good enough was the minecraft text. The grass block logo, I couldn't get to work well so it would have to be a png file

Carlgo11:
I replaced the entry with minecraft.net. Used their favicon as the icon. No idea why it is so large tho :unamused:

phallobst (approved):

PR #6411, opened 2022-01-16
Update Square Enix authentication methods and domain.

cj-wong:
TOTP support was added in addition to the existing two formats.

I also updated the note to refer only to the proprietary implementation. It seems now that other account holders (those who do not have games registered) can freely set TOTP 2FA.


phallobst (approved):

cj-wong:
I think the domain should be changed, since na.square-enix.com (along with eu.square-enix.com) now points to square-enix-games.com, while the log-in page is at membership.square-enix.com. Thoughts whether it should be moved, and to which letter in `entries` if so?

Carlgo11:
Maybe `square-enix-games.com`?

cj-wong:
Domain has been changed (old domain placed under "additional-domains").

Carlgo11 (approved):

PR #6417, opened 2022-01-17
Support contact form URL

### LINKED ISSUE #6415: Contact Us Page Link as a Contact Method
- [x] I have searched [open issues and pull requests](https://github.com/2factorauth/twofactorauth/issues?q=is%3Aopen). The issue I'm creating is not a duplicate of an existing open issue or pull request.

Many sites nowadays don't provide an email for people to openly suggest features or provide feedback and instead have a contact us page where someone can fill in their information and a message they would like to end to the company. This may be the only place for some sites to contact them to suggest adding 2FA. I suggest adding a fourth contact method where a link can be placed to allow someone to be directed to their contact/send us page when an email is not available.

phallobst:
Closes #6415


Carlgo11 (commented):
I think the schema should be set to only allow either email or form but not both at the same time 
( And document that in CONTRIBUTING.md)

phallobst:
I don't see why we couldn't support both. Who's to say which is better if both are offered by a website.

Carlgo11:
Since the contacting customer support through a form or sending an email to the support email address will most likely produce the same result (an email from customer support) I'd argue offering two options for the same result makes it redundant.

fletcherdodd:
Any updates on this one?

phallobst:
If the contact form (and the specific issue) allows for a one-way communication and/or the user does not want to disclose their email address, that form might actually be beneficial.

PR #6418, opened 2022-01-18
Update Steam

fletcherdodd:
Adjusting the wording of the note as the email 2FA method doesn't require a mobile number


entries/s/store.steampowered.com.json
```diff
@@ -10,7 +10,7 @@
 		],
 		"documentation": "https://help.steampowered.com/en/faqs/view/06B0-26E6-2CF8-254C",
 		"recovery": "https://help.steampowered.com/en/wizard/HelpWithSteamIssue/?issueid=809",
-		"notes": "A phone number is required as a fallback for 2FA.",
+		"notes": "A phone number is required as a fallback for TOTP 2FA.",
```
phallobst: I think it doesn't make sense to refer to "TOTP 2FA" if the two options listed are `email` and `custom-software`.

fletcherdodd: Updated wording, is that better?

phallobst (approved):

Carlgo11 (approved):

PR #6422, opened 2022-01-22
Create psono.com.json

chickahoona:


phallobst:
This is a self-hosted project, and also its Alexa rank is above 200K; therefore, this website does not qualify for inclusion in our list.

chickahoona:
@phallobst We have a managed solution too with https://psono.pw that people can use free of charge.

phallobst:
Unfortunately, the Alexa rank for both sites is way over 200K.

chickahoona:
@phallobst I understand. Still thanks alot for your consideration. I hope that you don't mind if we reapply once we meet your requirements. As a German company its quite hard to gain some traction ;)

PR #6424, opened 2022-01-22
Remove favicon.ico

Carlgo11:
According to https://caniuse.com/link-icon-png, all browser versions since 2012 have supported PNG as a link image type (as long as they support favicons) and since the site won't even load on such old browsers (due to TLS and JS support mismatch) I don't think there's any reason for us to keep the ICO file.

Oh, and apparently we haven't even been linking to that file for a while :see_no_evil:


phallobst (approved):
Since it was stored in `img/icons` and not the top level, the `favicon.ico` couldn't have been found anyway, I guess.

ApCoder123 (approved):

PR #6425, opened 2022-01-23
Update LogMeIn

sn00py1310:


entries/l/logmein.com.json
```diff
@@ -1,14 +1,16 @@
...
+      "logme.in"
+    ],
     "tfa": [
       "sms",
       "email",
       "totp",
       "custom-software"
     ],
-    "documentation": "https://support.logmeininc.com/pro/help#logmein-t-login-policy-set-twofactor-ownaccount",
+    "documentation": "https://documentation.logmein.com/webhelp/SecDoc/LogMeIn/t_login_policy_Set_TwoFactor_OwnAccount.html",
```
phallobst, line 16:
```suggestion
    "custom-software": [
      "LastPass Authenticator"
    ],
    "documentation": "https://documentation.logmein.com/webhelp/SecDoc/LogMeIn/t_login_policy_Set_TwoFactor_OwnAccount.html",
```

phallobst (changes requested):

entries/l/logmein.com.json
sn00py1310: I would remove custom-software completely because LastPass authenticator isn't a requirement, it's more like a recommendation.

> Let's take a closer look. This procedure describes setup using our own LastPass Authenticator. The procedure is similar for other popular authenticator apps.

phallobst: The set-up might be similar, but the effect is different:
> LastPass Authenticator: Tap the on-screen notification to confirm your identity.

Here you are coupling your account with the app. You are not required to enter the TOTP code, which also means that your phone needs to be connected to the internet. (You may also use LastAuthenticator as a regular TOTP app, I suppose.)

sn00py1310: Ok, thanks for the help. I just tested for regular TOTP.

phallobst (approved):

Carlgo11 (approved):

PR #6430, opened 2022-01-24
Add Tipico

bennetfabian:
- [ ] Check in other regions
- [ ] Find contact email


bennetfabian:
Hmm, test fails saying TFA field is missing. I thought when there is a contact field there should of course not be a TFA field?

Carlgo11:
@bennetfabian that's because you used the `notes` tag which only can be used when a `tfa` tag is present.

bennetfabian:
> @bennetfabian that's because you used the `notes` tag which only can be used when a `tfa` tag is present.

Oops. Well the information in the notes tag might still actually be useful if not necessary.

Carlgo11:
That may be. I'm just telling you why the build fails :shrug: 
The Twitter handle is also incorrectly capitalized and the domain is ranked above 200K on Alexa (See [guidelines](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#guidelines))

40504:
@bennetfabian Try `tipico.de` Alexa ranking is below 200k

PR #6434, opened 2022-01-25
Add Abra

40504:


Carlgo11 (commented):
> Once enabled, your 2FA code will be required each time a cryptocurrency withdrawal is initiated, a fiat withdraw is initiated, when sending money to an Abra user or when viewing your recovery phrase.

Sounds like authorization rather than authentication.

PR #6437, opened 2022-01-26
Add CoinTracker.io

dnivi3:
Adds CoinTracker.io 2FA information


phallobst:
CoinTracker is already listed (see #5142)

dnivi3:
@phallobst Thanks, my bad - not sure how I missed it!

PR #6438, opened 2022-01-27
Add Santa Clara University

walkerselby:


entries/s/scu.edu.json
```diff
@@ -0,0 +1,22 @@
+{
+  "Santa Clara University": {
+    "domain": "scu.edu",
+    "tfa": [
+      "sms",
+      "custom-software",
+      "custom-hardware",
```
Carlgo11: From reading the [FAQ](https://www.scu.edu/technology/get-connected/duo/2fa-faqs/) it sounds like U2F/FIDO 2 keys are the only hardware 2fa methods supported by SCU. If that's correct then the custom-hardware method can be removed:
```suggestion
```

Carlgo11 (changes requested):

walkerselby:
Sounds good, was looking at Cornell, I believe, and basing it off of that; better to drop custom-hardware though. Thanks!

Carlgo11 (approved):

Carlgo11:
Yeah, every org using Duo is using it differently. Some provide special hardware tokens and some don't. Unless the process is well documented by the organization it's impossible for us to know what login methods they use :sweat_smile:

Thanks for the pull request! Once another member gives a second approval it will be merged and go live on our site.

phallobst (approved):

PR #6441, opened 2022-01-27
Update LastPass

walkerselby:
Clarify the availability of U2F, dependent on plan type.

### DIFF
--- a/entries/l/lastpass.com.json
+++ b/entries/l/lastpass.com.json
@@ -21,7 +21,7 @@
     ],
     "documentation": "https://support.logmeininc.com/lastpass/help/enable-multifactor-authentication-lp010002",
     "recovery": "https://support.logmeininc.com/lastpass/help/i-lost-my-phone-ndash-how-do-i-disable-multifactor-authentication-for-lastpass",
-    "notes": "Availability of 2FA options depends on your account type. U2F hardware tokens are only available through Duo on the free plan. On the premium plans, Yubico is supported directly.",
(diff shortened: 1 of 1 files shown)


Carlgo11 (commented):
I'm guessing that you're referring to Yubico OTP?

I think it should be clarified that U2F works on all plans through Duo.

`Availability of 2FA options depends on your account type. U2F is available through Duo integration on all plans. Yubico OTP is available on paid plans.`

Although I don't see any reason why anyone would choose Yubico OTP instead of U2F...

walkerselby:
Yeah, that is what I meant. Not sure why anyone would choose it over U2F either but hey -- just the messengers. Would you like me to commit that revision you wrote in the comments or do you got it? Thanks.

Carlgo11 (approved):

phallobst:
While we're at it: Shouldn't `sms` be listed as well? [SecureAuth](https://support.logmeininc.com/lastpass/help/secureauth-authentication-lp010034) seems to support it.

walkerselby:
> While we're at it: Shouldn't `sms` be listed as well? [SecureAuth](https://support.logmeininc.com/lastpass/help/secureauth-authentication-lp010034) seems to support it.

Sure, will add later. Looking into it, that seems to only be available on LastPass business/enterprise accounts and is not available on the free, premium, or family plans. I will add a note for that as well.

Carlgo11:
I think "Availability of 2FA options depends on your account type" covers that. If we add more to the string I'm not sure it will fit the popup div.

phallobst (approved):

PR #6446, opened 2022-01-29
Similarweb rank check

### LINKED ISSUE #6339: Alexa.com closing down in May 2022 - change criteria to another service
Just saw that Amazon are closing down the Alexa.com service used to validate if a site meets the criteria for addition.
[https://support.alexa.com/hc/en-us/articles/4410503838999](https://support.alexa.com/hc/en-us/articles/4410503838999
)
I think another service would need to be picked to replace it, perhaps something like [https://www.similarweb.com/](https://www.similarweb.com/)

phallobst:
This is a "soft" check currently, i.e. the return code will always be 0. Once we are happy with the results, we could switch from Alexa (and change the documentation accordingly).

The SIMILARWEB_API_KEY is my own test key - we may want to replace it in the future.

Closes #6339

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,24 +7,31 @@ jobs:
     steps:
       - checkout
 
+      - restore_cache:
+          keys:
+            - bundler
+            - awis-cache
+            - similarweb-cache
+
+      - run:
+          name: Bundler
+          command: |
+            bundle config set --local path '/tmp/bundler'
+            bundle config set --local without 'jekyll_plugins'
+            bundle install
+
       - run:
           name: Check Twitter handles
           command: |
-            gem i twitter
-            ruby ./tests/twitter.rb
+            bundle exec ruby ./tests/twitter.rb
 
       - run:
           name: Check Facebook handles
           command: ruby ./tests/facebook.rb
 
-      - restore_cache:
-          keys:
-            - awis-cache
-            - similarweb-cache
-
       - run:
           name: Check Alexa ranking
-          command: ruby ./tests/alexa.rb
+          command: bundle exec ruby ./tests/alexa.rb
 
       - save_cache:
           key: awis-cache
@@ -39,6 +46,11 @@ jobs:
           key: similarweb-cache
           paths:
             - "/tmp/similarweb/*"
+
+      - save_cache:
+          key: bundler
+          paths:
+            - "/tmp/bundler/*"
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
 
       - restore_cache:
           keys:
-            - bundler
+            - bundler-cache
             - awis-cache
             - similarweb-cache
 
@@ -22,8 +22,7 @@ jobs:
 
       - run:
           name: Check Twitter handles
-          command: |
-            bundle exec ruby ./tests/twitter.rb
+          command: bundle exec ruby ./tests/twitter.rb
 
       - run:
           name: Check Facebook handles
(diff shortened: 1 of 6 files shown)


phallobst:
See "Check Similarweb ranking" here:
https://app.circleci.com/pipelines/github/2factorauth/twofactorauth/6322/workflows/7f8b4a28-561e-492f-8fe3-7f8ddd9a876d/jobs/6391
Looks good to me! :partying_face:

dsoegijono (approved):

tests/similarweb.rb
```diff
@@ -0,0 +1,56 @@
...
+diff = `git diff origin/master...HEAD entries/ | sed -n 's/^+.*"domain"[^"]*"\\(.*\\)".*/\\1/p'`
+# Strip and loop through diff
+diff.split("\n").each do |site|
+  domain = Addressable::URI.parse("https://#{site}").domain
+  fetch_from_api(domain) unless fetch_from_cache(domain)
+rescue StandardError => e
+  puts "\e[31m#{e.message}\e[39m"
+  # status = 1
+end
+exit(status)
```
Carlgo11: I reorganized your code a bit and fixed handling of domains without a rank. I also made the exit code return 1 on errors after the first of May this year.

```suggestion
#!/usr/bin/env ruby
# frozen_string_literal: true

require 'addressable'
require 'fileutils'
require 'json'
require 'net/http'
require 'uri'

# Cache handling
module Cache
  @cache_dir = '/tmp/similarweb/'

  def self.fetch(site)
    path = "#{@cache_dir}#{site}"
    File.read(path) if File.exist?(path)
  end

  def self.store(site, rank)
    FileUtils.mkdir_p(@cache_dir)
    File.open("#{@cache_dir}#{site}", 'w') { |file| file.write rank }
  end
end

# Similarweb API handling
module Similarweb
  def self.connect(url)
    http = Net::HTTP.new(url.host, url.port)
    http.use_ssl = true
    http.request(Net::HTTP::Get.new(url, { 'Accept' => 'application/json' }))
  end

  def self.fetch(site)
    apikey = ENV['SIMILARWEB_API_KEY']
    url = URI("https://api.similarweb.com/v1/similar-rank/#{site}/rank?api_key=#{apikey}")
    response = connect(url)
    raise("#{site} doesn't have a Similarweb ranking") if response.code.eql? '404'
    raise("(#{response.code}) Request failed.") unless response.code.eql? '200'

    rank = JSON.parse(response.body)['similar_rank']['rank']
    Cache.store(site, rank)
    rank
  end
end

raise('Similarweb API key not set') if ENV['SIMILARWEB_API_KEY'].nil?

status = 0
# Fetch changes
diff = `git diff entries/ | sed -n 's/^+.*"domain"[^"]*"\\(.*\\)".*/\\1/p'`
# Strip and loop through diff
diff.split("\n").each do |site|
  domain = Addressable::URI.parse("https://#{site}").domain
  rank = Cache.fetch(domain) || Similarweb.fetch(domain)
  raise("#{site} is ranked above the maximum rank of 200K") if rank.to_i > 200_000
rescue StandardError => e
  puts "\e[31m#{e.message}\e[39m"
  status = 1
end
exit(status) if Time.now.to_i > 1_651_363_200

```

.circleci/config.yml
```diff
@@ -30,6 +31,15 @@ jobs:
           paths:
             - "/tmp/alexa/*"
 
+      - run:
+          name: Check Similarweb ranking
+          command: ruby ./tests/similarweb.rb
```
Carlgo11: The CircleCI jobs don't actually run `bundle install` at the moment as all their required gems come preinstalled.
Therefore, this command would need to also include installation of the required gems (mainly 'addressable').

tests/similarweb.rb
Carlgo11: ignoring the exit status until May this year is also a possibility.
```ruby
exit(status) if Time.now.to_i > 1_651_356_000
```

.circleci/config.yml
phallobst: Isn't that what "Gemfile" is for?

Carlgo11: The Gemfile tells Bundler what to install but it doesn't install things without `bundle install` being executed.

Carlgo11 (approved):

PR #6444, opened 2022-01-28
Add Finary

bartocc:
name: Finary
labels: ~banking~, investing
about: La plateforme de l’investisseur moderne

PR #6449, opened 2022-01-29
Subdirectory name test

ApCoder123:
Fails when an entry is in the wrong entries subdirectory.
E.g. `entries/a/example.com` would fail. See https://github.com/2factorauth/twofactorauth/pull/6444#pullrequestreview-866982353 for an example.

### DIFF
--- a/tests/validate-json.rb
+++ b/tests/validate-json.rb
@@ -64,6 +64,15 @@ Dir.glob('entries/*/*.json') do |file|
           "File name should be the same as the domain name. Received: #{file_name}. Expected: #{expected_file_name}")
   end
 
+  folder_name = file.split('/')[1]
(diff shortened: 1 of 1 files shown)


phallobst:
Come on, only an idiot would miss that during a review! :man_facepalming: 

It would be nice to check the image files as well, though.

ApCoder123:
Is it necessary for the image files? Due to the way that `validate-images.rb` works, it will fail if the image is in the wrong folder.

phallobst:
You're right, it does - although the error message is a bit odd:
`::error file=entries/y/youhodler.com.json:: Unused image at img/a/foo.svg`
The first file name seems to be random and out of place.

Carlgo11 (approved):
I think it's a good addition 👍

phallobst (approved):

PR #6452, opened 2022-01-30
Add Stan

fletcherdodd:


phallobst (approved):
I wouldn't mind using a logo like in their [favicon](https://www.stan.com.au/favicon.ico), but this should do too.

Carlgo11 (approved):

PR #6453, opened 2022-01-30
Ignore exclusions for regions.json

phallobst:
A "region" like `-se` should not be included in the API files IMHO.


Carlgo11 (approved):

ApCoder123:
The note on excluded regions in [API.md](https://github.com/2factorauth/twofactorauth/blob/master/api.md#elements) should probably also be removed.

phallobst:
This PR is only about the `regions.json` file (e.g. how many entries there are per region etc.), not about the regions per entry.

ApCoder123:
Whoops, sorry - I hadn't looked at the PR properly, only briefly read the description.

ApCoder123 (approved):

PR #5000, opened 2020-10-08
Add Migros

jokay:
Adds [Migros](https://www.migros.ch).

rank-migros

PR #6458, opened 2022-01-31
Add Trimble

pitdicker:
Trimble offers products for all kinds of Industries, I am only familiar with their construction solutions. Ca. #5000 on Alexa.


phallobst:
The 2FA refers to the "Trimble ID" service, which in turn is used by [sketchup.com](https://sketchup.com) or [novapoint.com](https://novapoint.com), right?

pitdicker:
O, I didn't realize Trimble could be seen as a '2FA provider'. Is that a reason not to include the site? I only know Tekla Structures and SketchUp, which are products of Trimble but which happen to have their own sites.

Carlgo11 (approved):
From what I can see Trimble isn't a 2FA provider but a authentication provider (among other things).
Since services like Okta, OneLogin and Auth0 are listed I don't think Trimble should be a problem.

PR #6463, opened 2022-02-03
Update Dashlane

louis-tk:
Dashlane does not currently support U2F for public use following the transition to using the webapp and removing their desktop app (the desktop app can be installed, but then refuses to run once the user logs in and instead directs them to the webapp). As such, it is no longer possible (despite advertising saying otherwise) to use U2F with dashlane.

### DIFF
--- a/entries/d/dashlane.com.json
+++ b/entries/d/dashlane.com.json
@@ -2,12 +2,7 @@
   "Dashlane": {
     "domain": "dashlane.com",
     "url": "https://www.dashlane.com",
-    "tfa": [
-      "totp",
-      "u2f"
-    ],
     "documentation": "https://support.dashlane.com/hc/en-us/articles/202625042",
-    "notes": "FIDO U2F is a Premium feature only and requires the desktop app which has been deprecated and is no longer available to the public.",
     "keywords": [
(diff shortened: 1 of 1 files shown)


louis-tk:
I have left the `"u2f"` value as it may well still be supported for corporate customers and I cannot be sure it is not supported at least in _some_ way.

Carlgo11 (changes requested):
According to https://support.dashlane.com/hc/en-us/articles/207051919 the U2F 2FA option is deprecated so I think it's better to just remove U2F completely from the listing.

louis-tk:
According to https://support.dashlane.com/hc/en-us/articles/360021374760#h_01F5AZKY3QW2E9SPWZ1M98JR62
it does look like it is possible to still use a hardware token via WebAuthn by enabling the "Biometric unlock" option. Not sure how best to show this?

Carlgo11:
That would be "passwordless" which we don't list as 2FA.

Carlgo11:
Unless there's activity within the next week I'll close the PR.

Carlgo11 (approved):

phallobst (approved):

PR #6482, opened 2022-02-05
Update MongoDB products

ApCoder123:
Although the domain listed for Cloud manager is cloud.mongodb.com, the same domain is used after logging in to Atlas (rather than the currently listed atlas.mongodb.com). Currently, the Cloud Manager doc link will show up in password managers for Atlas accounts. It would be better to list this domain with Atlas, as Atlas is the more recent, active product (Cloud Manager appears to be more of a legacy product), so I have changed the domain on Atlas to be `cloud.mongodb.com` and changed cloud manager to use `cloudmanager.mongodb.com`.


Carlgo11 (dismissed):

phallobst:
Has this changed in the last couple of days? `cloudmanager.mongodb.com` redirects to `mongodb.com` (but strangely not to `mongodb.com/cloud/cloud-manager`). Both logins now use `account.mongodb.com` and should not differ in their 2FA methods. Maybe I'm just confused. :confused:

ApCoder123:
I've had a look and you're right - both do use a single MongoDB account - the cloud manager page was just outdated. Have combined both into one entry "MongoDB" and linked to atlas doc page with up to date 2FA methods.

phallobst (approved):

Carlgo11 (approved):

PR #6484, opened 2022-02-05
New theme proposal

### LINKED ISSUE #1450: Information disclosure concern - "Tell them to support 2FA" option
While it's a great idea in general, if a user were to only use this feature to notify their own sites it could easily translate to an inadvertent disclosure of a twitter user's affiliations.

Generally-speaking it would take some work to find out what banks a user might be registered with, whether they have cryptocurrencies, where their data is stored. Without at least a note about information disclosure here, this opens the door for a user to disclose that information to a potential attacker on their own. Knowing those sites don't support 2FA additionally reduces the difficulty threshold for a potential attacker.

### LINKED ISSUE #6221: Enable region redirect through Cloudflare Workers
Since the merging of #6163 we now have the option to redirect users connecting to `https://2fa.directory/` to their regional version of our website if such a region is available.
This would benefit the users by only seeing the services that they can use, reducing clutter and load times. If we don't serve a site for the region of which the user connects from, they will be directed to the normal "global" site.

An example of this is available at [twofactorauth.net](https://twofactorauth.net/) (one of the domains we have).

I'd need some approvals before enabling this. Please vote and comment about any concerns/issues.

Carlgo11:
## Background

The category buttons have remained the same since version 1 of this repository which was 8 years ago (wow I feel old now )
When redesigning the buttons I first opted for a [Metro UI](https://en.wikipedia.org/wiki/Metro_UI) like look but, according to @jamcat22, that's not cool anymore so I made it more like [Fluent UI](https://en.wikipedia.org/wiki/Fluent_Design_System) with rounded edges and a box shadow. The buttons are square on desktop down to laptop size after which they turn into mobile interface with one category per row.

## Issues

* Resolves #6221

## Screenshots

desktop main bright theme
desktop main dark theme
desktop categories bright theme
desktop categories dark theme
mobile categories bright theme

### DIFF
--- a/css/navbar.scss
+++ b/css/navbar.scss
@@ -62,6 +62,7 @@
   font-size: 1.5em;
   font-weight: 700;
   line-height: 1;
+  vertical-align: middle;
 }
 
 @media screen and (min-width: 444px) {
--- a/css/navbar.scss
+++ b/css/navbar.scss
@@ -46,6 +46,7 @@
   background: var(--secondary-background);
   padding: .5em 0;
   text-align: center;
+  font-size: 14px;
   margin: 0 0 3rem 0;
 
   &:only-child {
@@ -55,13 +56,21 @@
 }
 
 #region-notice-close-btn {
-  margin-left: 2rem;
   color: var(--text-color);
   border: none;
   background: none;
   font-size: 1.5em;
   font-weight: 700;
-  line-height: 1
+  line-height: 1;
+}
+
+@media screen and (min-width: 444px) {
+  #region-notice-close-btn {
+    margin-left: 2rem;
+  }
+  #region-notice {
+    font-size: 1rem;
+  }
 }
 
 @media (prefers-color-scheme: dark) {
--- a/css/navbar.scss
+++ b/css/navbar.scss
@@ -5,7 +5,7 @@
 }
 
 .navbar {
-  background: #eeeeee96;
+  background: var(--secondary-background);
   box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 3px 1px -2px rgba(0, 0, 0, 0.2), 0 1px 5px 0 rgba(0, 0, 0, 0.12);
   font-size: 17px;
   line-height: 18px;
@@ -23,7 +23,7 @@
   .dropdown-menu {
     position: absolute !important; // Bootstrap normally sets position static for dropdowns in mobile views
     background: #eeeeee;
-    z-index: 1100; // higher than ".cat"
+    z-index: 1100; // higher than category button
 
     .dropdown-item {
       .flag-icon, .fas {
@@ -42,10 +42,29 @@
   }
 }
 
-@media (prefers-color-scheme: dark) {
-  .navbar {
-    background: #6f6f6f96;
+#region-notice {
+  background: var(--secondary-background);
+  padding: .5em 0;
+  text-align: center;
+  margin: 0 0 3rem 0;
+
+  &:only-child {
+    margin: 0 auto;
+    width: max-content;
   }
+}
+
+#region-notice-close-btn {
+  margin-left: 2rem;
+  color: var(--text-color);
+  border: none;
+  background: none;
+  font-size: 1.5em;
+  font-weight: 700;
+  line-height: 1
+}
+
+@media (prefers-color-scheme: dark) {
   .nav-link {
     color: #fff;
 
--- a/js/main.js
+++ b/js/main.js
@@ -59,11 +59,6 @@ let resizeObserver = new ResizeObserver(() => {
     $('.footer').css({position: 'static'});
   }
 
-  // Focus and select search input on desktop
-  if ($(window).innerWidth() >= 992) {
-    $('#innerSearchBox').focus().select();
-  }
-
   // Resize square divs
   $('.box').height($('.box').width());
 });
--- a/js/main.js
+++ b/js/main.js
@@ -25,7 +25,7 @@ $('.category-btn').click(function () {
   let query = window.location.hash.substring(1);
 
   // Collapse all other tables
-  $('.collapse').collapse('hide');
+  $('.category-table.collapse').collapse('hide');
   $('.category-btn').removeClass('active');
 
   // Check if category tables are displayed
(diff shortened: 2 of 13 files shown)


phallobst:
The category buttons are rather massive. When I open one in the top line, I can't even see the first entry anymore in my browser, even with a full-screen window. :frowning: Maybe we could seize the opportunity to get rid of the vast empty space at the top, e.g. by shrinking the logo and/or putting it next to the text and search bar.

What about the given issues? I see stuff regarding #6221, but nothing for the other ones.

Carlgo11:
The width of the buttons are actually reduced from the current 1/5th to 1/6th. Making the buttons smaller would mean more buttons per row and 7 buttons per row looks cluttered and weird.
The height of the buttons are proportional to the width.

What proportions is your viewport? I've tried with the standard 1920x1080 along with common sizes like 1280x800 (laptop), 1080x810 (tablet).
Here's 1280x800 size for comparison:
laptop screenshot

>What about the given issues? I see stuff regarding https://github.com/2factorauth/twofactorauth/issues/6221, but nothing for the other ones.

Doh!


The branch I did the testing on contained those changes but I thought I'd split the changes into multiple PRs to ease reviewing and so I didn't bring those changes into this branch while forgetting to reflect that in the description. I've removed those issue numbers now.

Carlgo11:
I'm open to testing different size values if requested.

Here's a side by side comparison of the current and proposed main page.


Here's the logo in 20em (current) and 15em.

phallobst:
I imagined something like this to reduce the space requirements even further:
image

Carlgo11:
That's a rather big adjustment which would require a lot of work and I mostly just focused on the category buttons in this PR.
Maybe it could be put on the list for future changes?

I do agree that it could look though :+1:

Carlgo11:
Here are screenshots of the changes:
|Device simulation|Resolution|Theme|Screenshot|
|------|---|:--:|:--------------:|
|Desktop|1920x1080|Bright|desktop bright|
|Desktop|1920x1080|Dark|desktop dark|
|Laptop|1200x800|Bright|Screenshot 2022-02-06 at 21-00-22 2FA Directory|
|Laptop|1200x800|Dark|Screenshot 2022-02-06 at 21-00-31 2FA Directory|
|iPad Portrait Mode|810x1080|Bright|Screenshot 2022-02-06 at 20-50-40 2FA Directory|
|iPad Portrait Mode|810x1080|Dark|Screenshot 2022-02-06 at 20-49-16 2FA Directory|
|iPad Landscape Mode|1080x810|Bright|Screenshot 2022-02-06 at 20-50-45 2FA Directory|
|iPad Landscape Mode|1080x810|Dark|Screenshot 2022-02-06 at 20-49-07 2FA Directory|
|iPhone SE|375x667|Bright|Screenshot 2022-02-06 at 20-50-29 2FA Directory|
|iPhone SE|375x667|Dark|Screenshot 2022-02-06 at 20-49-41 2FA Directory|

phallobst:
The CSS for the regional pages does not work in my local test installation, e.g. I get a huge logo and no category button borders there - the main page is okay. I tested with three different browsers. :raised_eyebrow:

Carlgo11:
@phallobst Sounds like a cache problem. Delete _site in the repo directory and any regions in /tmp.
We should probably do some cache checks in the future when generating region sites as that tends to mess things up locally 🤔

phallobst:
Yes, it _was_ the region cache in the `/tmp` directory. :+1:

phallobst (approved):

Carlgo11:
Latest 2 commits fix a bug caused by #6348 where, as a user clicks on a category button in Chromium browsers, the browser scrolls to the top of the page instead of to category table. This also happens when pasting in the URL for the category (example: `https://2fa.directory/#banking`). 

As #6348 sets focus to the search field on document.ready it shifts the focus from the category to the search field as shown by @ApCoder123 here:

https://user-images.githubusercontent.com/3535780/152874164-af422535-6035-4673-a800-a030ce3854f2.mp4

Carlgo11:
e776940 fix a bug caused by #6348 where, as a user clicks on a category button in Chromium browsers, the browser scrolls to the top of the page instead of to category table. This also happens when pasting in the URL for the category (example: `https://2fa.directory/#banking`). 

As #6348 sets focus to the search field on document.ready it shifts the focus from the category to the search field.

The changes are live at https://2fa.carlgo11.com/ (/about/ isn't working as the test page is hosted outside of this repository)

css/navbar.scss
```diff
@@ -42,10 +42,29 @@
...
+}
+
+#region-notice-close-btn {
+  margin-left: 2rem;
+  color: var(--text-color);
+  border: none;
+  background: none;
+  font-size: 1.5em;
+  font-weight: 700;
+  line-height: 1
```
ApCoder123:
```suggestion
  line-height: 1;
  vertical-align: middle;
```
I think that it looks a bit nicer if the close button is aligned with the text:
Current:
image

Suggested:
image

ApCoder123 (commented):
The new design looks great. I've noticed a few minor issues with how it appears:

- Mobile Layout: The region notice looks a bit messy. It might look nicer to split the message across 2 lines. The margin on the close button should probably also be removed on mobile layouts
image


- Mobile + Desktop Layout: The padding issue from #5760, #5812 has returned
image
image

js/main.js
```diff
@@ -1,29 +1,35 @@
...
+$('.exception').popup({position: 'right center', hoverable: true, title: 'Exceptions & Restrictions'});
+
 // On category click
-$('.cat').click(function () {
-  let query = window.location.hash;
+$('.category-btn').click(function () {
+  let query = window.location.hash.substring(1);
 
   // Collapse all other tables
   $('.collapse').collapse('hide');
```
ApCoder123: A consequence of this line is that the region notice disappears when a category button is clicked. Do we want to continue showing it until the user has clicked the cross button, or is this behaviour fine (as it will still appear the next time the user visits the site)?

Carlgo11:
* The padding of the div surrounding category-buttons seems to be the problem. Since the padding needs to be there I suppose a fix would be to give all such divs a class and hide them instead of the category-buttons.
* Thanks for the line-height fix!
* I'll change it to `$('.category-table.collapse').collapse('hide');`

Carlgo11:
@ApCoder123 changing the line-height didn't help :disappointed: 
Screenshot 2022-02-08 at 22-38-33 2FA Directory
Screenshot 2022-02-08 at 22-39-11 2FA Directory

ApCoder123:
I got it to work by adding `vertical-align: middle;` to the button.

Carlgo11:
I'm not sure it is aligned :thinking:

ApCoder123:
It looked good and properly aligned when I checked:
image

ApCoder123 (approved):

PR #6485, opened 2022-02-05
Add self-dox warning to social media buttons

### LINKED ISSUE #1450: Information disclosure concern - "Tell them to support 2FA" option

Carlgo11:
This PR adds a warning message when hovering over social media contact buttons in categories with the tag `sensitive`.

My IDE also cleaned up a bunch of white spaces in desktop-table.html so I apologize if the diff is hard to read :sweat_smile: 

* Resolves #1450
* Resolves #3812


phallobst:
Reading this diff on GitHub is easier if you hide whitespace-only changes:
image

phallobst:
I'm not quite sure if only certain categories should be considered "sensitive". Depending on your local laws and regulations, almost any category might contain a website that would qualify.

Similar to the region notice in #6484, a flag in the `localStorage` could indicate whether the user wants to keep this warning in. We wouldn't need cookies to do so.

ApCoder123:
Are we planning to incorporate this into the mobile layout as well? Also, agree that it would be a good idea to allow users to disable the warning.

Carlgo11:
I've addressed the requested changes the following ways:
* Made the tooltip a modal that's displayed on all device types
* Added a local storage variable to only display the warning once
* Switched from a custom Bootstrap v4 to vanilla Bootstrap v5 as the prev version didn't contain modals

js/social-media.js
```diff
@@ -17,27 +14,43 @@ $('.email-button').click(function () {
   window.open('mailto:' + $(this).data('email') + '?subject=' + langs.get(lang));
 })
 
+$('.facebook-button').click(function () {
+  const uri = `https://facebook.com/${(this).data('facebook')}`;
```
phallobst:
```suggestion
  const uri = `https://facebook.com/${$(this).data('facebook')}`;
```

phallobst (changes requested):
I think that after clicking on "I accept the risk", the modal dialog should also be closed.

phallobst (approved):

dsoegijono (approved):

PR #6492, opened 2022-02-08
Generate regions in parallel

Carlgo11:
This PR allows for the generation of regional sites in parallel which allows for sped up builds on multi-core architectures. As GitHub Actions runs in a 2-core environment it should theoretically half the runtime of `regions.rb`.

I also removed the if-statement ignoring generation of sites if the site already exists in `/tmp/{region}` which should improve testing on local systems.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,7 @@ end
 group :test do
   gem 'httpclient'
   gem 'json_schemer'
+  gem 'parallel'
   gem 'rubocop'
   gem 'twitter'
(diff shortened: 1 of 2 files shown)


phallobst (approved):

ApCoder123 (approved):

Carlgo11:
I forgot that the `complete build` job doesn't install gems in the test group. I'll push a commit to move the parallel gem outside of that group.

PR #6520, opened 2022-02-15
SVG Lint test

Carlgo11:
I've developed a SVG lint test to check for syntax errors and minification possibilities but would like feedback on if it's something that would be desirable to use or not.

While keeping in mind the barrier to entry for new contributors I wanted some way of automate testing of SVG files and provide feedback on ways to enhance the files. I've therefore designed the script to fail on errors which would cause the graphic not to be displayed properly on the website whilst providing messages (warnings) on things which could hint at the file not being very optimized. My idea is for the errors to have to be fixed before a PR can be merged but any warnings to be treated as debug/info messages for the PR author and maintainers.


tests/svg-lint.rb
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'json'
+require 'active_support/core_ext/hash'
```
phallobst, line 6:
```suggestion
require 'active_support'
require 'active_support/core_ext/hash'
```

tests/svg-lint.rb
```diff
@@ -0,0 +1,49 @@
...
+def warn(file, msg)
+  puts "::warning file=#{file}:: #{msg}"
+end
+
+def test(svg, css)
+  svg.css(css).length.positive?
+end
+
+@status = 0
+diff = `git diff --name-only --diff-filter=Ard -name '*.svg' origin/master...HEAD img/`
```
phallobst:
```suggestion
diff = `git diff --name-only --diff-filter=Ard origin/master...HEAD -- 'img/***.svg'`
```

phallobst (changes requested):
The script did not run without these changes on my machine.

Afterwards, I got no SVG errors, but warnings for 2/3 of the icons. :fearful:

phallobst (approved):

Carlgo11:
Thanks for the review @phallobst :pray: 
requirements have been fixed and the git diff command fixed.

two-thirds of the SVGs are indeed causing warnings so imagine the potential of this script :wink:

PR #6523, opened 2022-02-16
Excluded Regions: Skrill

ghost:
Due to regulations by the Japanese financial authorities, people living in Japan are not allowed to sign up for Skrill.

### DIFF
--- a/entries/s/skrill.com.json
+++ b/entries/s/skrill.com.json
@@ -10,7 +10,7 @@
     "custom-software": [
       "Skrill App Push Notification"
     ],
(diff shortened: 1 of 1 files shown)


phallobst:
The list of excluded regions is quite large actually: https://www.skrill.com/en/footer/terms-conditions/non-serviced-countries/
But since Japan is the only one currently featured in the region dropdown, this should be sufficient right now.

I also took the liberty to update the doc link, since the old one gave me a 404.

phallobst (approved):

Carlgo11 (approved):

PR #6524, opened 2022-02-16
Excluded Regions: Neteller

ghost:
As of 2022, Netteller has withdrawn from Japan and the US.


phallobst:
Are they related to Skrill? https://www.neteller.com/en/policies/non-serviced-countries _(Edit: They are, actually.)_

I have updated this doc link as well.

ghost:
Skrill and Neteller have the same parent company.
The reason for the exclusion of regions is likewise due to legal restrictions.

phallobst (approved):

Carlgo11 (approved):

PR #6534, opened 2022-02-18
Replace inline styling

Carlgo11:
Removed all inline styling that I could find.
Also fixed a bug where lots of images would be unnecessarily loaded on document load instead of lazy-loading.

### DIFF
--- a/_includes/html/desktop-table.html
+++ b/_includes/html/desktop-table.html
@@ -20,7 +20,7 @@
         <tr class="table-{% if website.tfa %}success{% else %}danger{% endif %}">
 
           <td>
-            <div class="searchWords" style="display: none;">{{ name }} {{ website.domain }}{% for item in website.tfa %} 2fa:{{ item }}{% endfor %}{% for item in website.custom-hardware %} 2fa:{{item}}{% endfor %}{% for item in website.custom-software %} 2fa:{{item}}{% endfor %} {% for domain in website.additional-domains %} {{ domain }} {% endfor %}</div>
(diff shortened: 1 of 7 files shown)


_includes/html/desktop-table.html
```diff
@@ -32,7 +32,7 @@
                 {% assign image = website.domain | append: ".svg" | prepend: letter %}
               {% endif %}
 
-              <img class="logo" loading="lazy" src="/img/{{ image }}" alt="">
+              <img class="logo desktop-only" loading="lazy" src="/img/{{ image }}" alt="">
```
phallobst, line 35: Is this needed? We are in `desktop-table.html` after all.

_includes/html/mobile-table.html
```diff
@@ -22,7 +22,7 @@ <h2 class="search-table-title">{{ include.category.title }}</h2>
             {% assign image = website.domain | append: ".svg" | prepend: letter %}
           {% endif %}
 
-          <img class="logo" loading="lazy" src="/img/{{ image }}" alt="">
+          <img class="logo mobile-only" loading="lazy" src="/img/{{ image }}" alt="">
```
phallobst, line 25: see above

about.html
```diff
@@ -2,7 +2,7 @@
 layout: default
 ---
 
-<h1 class="display-3 desktop-only title">The 2factorauth Group</h1>
+<h1 class="display-3 desktop-only title">The {{ site.github.owner_name }} Group</h1>
 <h1 class="mobile-only title">The 2factorauth Group</h1>
```
phallobst:
```suggestion
<h1 class="mobile-only title">The {{ site.github.owner_name }} Group</h1>
```

phallobst (changes requested):

_includes/html/desktop-table.html
Carlgo11: Yes apparently it is needed, otherwise some images are loaded even when the table is hidden. I don't quite understand why it happens or why this fixes it but it does 🤷‍♂️

phallobst (approved):

ApCoder123 (approved):

PR #6540, opened 2022-02-19
Update Hi-Rez Studios

phallobst:


entries/h/hirezstudios.com.json
```diff
@@ -1,8 +1,6 @@
 {
   "Hi-Rez Studios": {
     "domain": "hirezstudios.com",
-    "url": "https://www.hirezstudios.com/",
```
Carlgo11, line 4: No response on https://hirezstudios.com/. I suggest we leave the url tag

Carlgo11 (changes requested):

entries/h/hirezstudios.com.json
phallobst: It worked for me both in Chrome and Firefox (and redirected to www.hirezstudios.com), but wget and curl gave no reponse. Sigh...

Carlgo11: Your browsers probably tried to connect to the HTTP version when HTTPS failed. I've disabled that in about:config 😬

Carlgo11 (approved):

ApCoder123 (approved):

PR #6549, opened 2022-02-21
Update Rakuten US

ghost:
Rakuten has websites in [Japan](https://www.rakuten.co.jp/), the [US](https://www.rakuten.com/), [Germany](https://rakuten.de/), and other countries, but they all have independent accounts.
Rakuten US is only a US website, not a global website and needs to be modified.


Carlgo11 (changes requested):
If the login systems are the same then I recommend you add the de, jp, us to the regions list and the domains to ` additional-domains`.

ghost:
> If the login systems are the same then I recommend you add the de, jp, us to the regions list and the domains to ` additional-domains`.

The login system is different.

Carlgo11:
The JP authentication system seems to be independent but the other sites appear to use `login.account.rakuten.com` as SSO.
As no sites use 2FA and the trend seems to be that the systems are merging I'd say it's probably better, and a lot easier, to just use one entry for all Rakuten sites until the start to differ in 2FA acceptance.

German signup page:
Screenshot 2022-02-21 at 16-31-54 Rakuten - Registration
Taiwanese signup page:
Screenshot 2022-02-21 at 16-32-10 Rakuten - Registration

ghost:
I have a Rakuten DE account.
When I tried to log in to Rakuten US with that account, I could not log in.
The hostname is the same, but the login system seems to be different.

entries/r/rakuten.com.json
```diff
@@ -1,5 +1,5 @@
 {
-  "Rakuten": {
+  "Rakuten US": {
     "domain": "rakuten.com",
     "url": "https://www.rakuten.com/",
```
Carlgo11, line 4: Right, I'm not saying the databases are the same but the authentication systems are the same. That would mean that if they enabled as an option on one of the sites it would be enabled for the rest (excluding JP).
Therefore, please add the other domains in the `additional-domains` array. If you want to create a separate listing for Rakuten Japan that's also fine.
```suggestion
    "url": "https://www.rakuten.com/",
    "additional-domains": [
      "rakuten.ca",
      "rakuten.de",
      "rakuten.es",
      "rakuten.fr",
      "rakuten.com.tw"
    ],
```

entries/r/rakuten.com.json
```diff
@@ -1,5 +1,5 @@
 {
-  "Rakuten": {
+  "Rakuten US": {
```
Carlgo11:
```suggestion
  "Rakuten": {
```

Carlgo11 (changes requested):

Carlgo11 (approved):

PR #6556, opened 2022-02-23
Add sitemap

Carlgo11:
Found a useful script to generate sitemap.xml. Edited it to fit this project.

### DIFF
--- a/scripts/sitemap.sh
+++ b/scripts/sitemap.sh
@@ -8,7 +8,7 @@
 URL="https://2fa.directory/"
 
 # values: always hourly daily weekly monthly yearly never
-get_freqency () {
+get_frequency () {
   if [ "$1" -lt 7 ]; then FREQ="daily";
   elif [ "$1" -lt 32 ]; then FREQ="weekly";
   elif [ "$1" -lt 63 ]; then FREQ="monthly";
--- a/scripts/sitemap.sh
+++ b/scripts/sitemap.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+# Copyright (c) 2017, Tino Reichardt <https://github.com/mcmilk/sitemap-generator>
+# Copyright (c) 2022, 2factorauth <https://github.com/2factorauth/twofactorauth>
+
+[[ $# -ne 1 ]] && { echo "USAGE: ./sitemap.sh _site" >&2; exit 1; }
+
+# url configuration
+URL="https://2fa.directory/"
+
+# values: always hourly daily weekly monthly yearly never
+get_freqency () {
+  if [ "$1" -lt 7 ]; then FREQ="daily";
+  elif [ "$1" -lt 32 ]; then FREQ="weekly";
+  elif [ "$1" -lt 63 ]; then FREQ="monthly";
+  else FREQ="yearly";
+  fi;
+}
+
+# begin new sitemap
+exec 1> "${1}/sitemap.xml"
+
+# print head
+echo '<?xml version="1.0" encoding="UTF-8"?>'
+echo '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">'
+
+OPTIONS='
+-prune
+! -iname robots.txt
+! -iname manifest.json
+! -iname service-worker.js
+! -iname "sitemap.*"
+! -iname ".*"
+! -path "./css/*"
+! -path "./js/*"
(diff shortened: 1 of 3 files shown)


Carlgo11:
Hmm file-extensions.sh doesn't like the sitemap.sh.options file :thinking: 
@phallobst how do you feel about adding `options` as a valid file extension to `/scripts`? I don't really have a better place to store the file :disappointed:

phallobst:
I'm not sure why we'd nee the options file at all, since these are references only once in the script and could be used in-place.

If you really want to keep the file, why not just name it `sitemap.options.sh` instead? :wink:

phallobst:
Is the `get_freqency` function any useful? The script will probably work on a freshly checked-out `git` workspace most of the time, i.e. the file timestamps will always lead to "daily" there, I reckon.

Carlgo11:
I did a fetch and hard reset of my environment and it appears to work as expected there.
How do you propose we implement the options data into the shell script? I can't come up with a way that also is readable.

phallobst:
I took the liberty to rename the options file, so that's one alternative. :wink:

This works as well:
```bash
OPTIONS='
-prune
! -iname robots.txt
! -iname manifest.json
! -iname service-worker.js
! -iname "sitemap.*"
! -iname ".*"
! -path "./css/*"
! -path "./js/*"
! -path "./img/*"
! -path "./entries/*"
'
cd "${1}" || exit 1
eval find . -type f $OPTIONS -printf '"%TY-%Tm-%Td%p\n"' | \
while read -r line; do
```

Or if you don't want to use eval:
```bash
cd "${1}" || exit 1
find . -type f \
-prune \
! -iname robots.txt \
! -iname manifest.json \
! -iname service-worker.js \
! -iname "sitemap.*" \
! -iname ".*" \
! -path "./css/*" \
! -path "./js/*" \
! -path "./img/*" \
! -path "./entries/*" \
-printf "%TY-%Tm-%Td%p\n" | \
while read -r line; do
```

Carlgo11:
phallobst the Unix wizard :magic_wand: 
I added the eval suggestion :smile:

phallobst (approved):

scripts/sitemap.sh
```diff
@@ -0,0 +1,53 @@
...
+# Copyright (c) 2017, Tino Reichardt <https://github.com/mcmilk/sitemap-generator>
+# Copyright (c) 2022, 2factorauth <https://github.com/2factorauth/twofactorauth>
+
+[[ $# -ne 1 ]] && { echo "USAGE: ./sitemap.sh _site" >&2; exit 1; }
+
+# url configuration
+URL="https://2fa.directory/"
+
+# values: always hourly daily weekly monthly yearly never
+get_freqency () {
```
dsoegijono:
```suggestion
get_frequency () {
```

scripts/sitemap.sh
```diff
@@ -0,0 +1,53 @@
...
+! -path "./img/*"
+! -path "./entries/*"
+'
+
+cd "${1}" || exit 1
+eval find . -type f $OPTIONS -printf '"%TY-%Tm-%Td%p\n"' | \
+while read -r line; do
+  DATE=${line:0:10}
+  FILE=${line:12}
+  get_freqency $((($(date +%s) - $(date +%s -r "${FILE}")) / 86400))
```
dsoegijono:
```suggestion
  get_frequency $((($(date +%s) - $(date +%s -r "${FILE}")) / 86400))
```

dsoegijono (approved):
Typo on "frequency" but otherwise lgtm

Carlgo11:
@phallobst looks like -printf isn't an option in the busybox eval command.

phallobst:
@Carlgo11
`-printf` is an option for _GNU find_, but the _busybox find_ does not support it. A slower workaround would be something like

```bash
eval find . -type f $OPTIONS -print0 | xargs -0 stat -c '%z%n' | \
while read -r line; do
  DATE=${line:0:10}
  FILE=${line:37}
```
or
```bash
eval find . -type f $OPTIONS -print0 | xargs -0 stat -c '%z%n' | \
while read -r line; do
  DATE=${line:0:10}
  FILE=${line:31}
```
depending on whether `stat` is the _busybox_ variant as well. The offsets for the `DATE` and `FILE` variables are correct on my system, but may need some adjustment.

PR #6557, opened 2022-02-23
Add NAGA

40504:
Restricted regions: NAGA Global Ltd does not provide services to residents of certain jurisdictions including the USA, Iran, North Korea, Cuba, Iraq, Syria and Canada.

source: https://support.naga.com/en/articles/5056295


Carlgo11:
NAGA as in Nake America Great Again? :thinking: 

You could exclude those regions with the regions array though I doubt some of them really need to be listed as I don't think we'll ever create a regional site for North Korea or Iran :upside_down_face: 

```json
"regions": [
  "-ca",
  "-us"
]
```

Carlgo11 (approved):

phallobst (approved):

PR #6578, opened 2022-03-04
Update Mintos

ghost:


entries/m/mintos.com.json
```diff
@@ -8,6 +8,12 @@
     "documentation": "https://help.mintos.com/hc/en-us/sections/360000373858",
     "keywords": [
       "investing"
+    ],
+    "regions": [
+      "-fr",
+      "-gb",
+      "-in",
+      "-us"
```
dsoegijono, line 16: Why the `-` for the regions?

ghost: Those regions will not appear in the country choices at sign-up.

PR #6580, opened 2022-03-04
Update Vanguard

ghost:


entries/v/vanguard.com.json
```diff
@@ -18,8 +18,7 @@
       "investing"
     ],
     "regions": [
-      "-gb",
-      "-au"
+      "us"
```
phallobst, line 21: According to https://global.vanguard.com/portal/site/home the services are offered in various regions across the globe.

ghost: Hmmm, each country has a different domain and a separate account.
It is true that we service multiple countries, but not the whole world.

PR #6602, opened 2022-03-11
Update CoinTracking

40504:


phallobst (approved):

entries/c/cointracking.info.json
```diff
@@ -4,6 +4,7 @@
     "tfa": [
       "totp"
     ],
+    "recovery": "https://cointracking.freshdesk.com/en/support/solutions/articles/29000021946",
```
Carlgo11, line 7: Link doesn't work for me :thinking: 
Unauthorized location

Carlgo11 (approved):
Got it working with a German VPN. Seems like they're using geoblocking for the freshdesk site but not their main page :shrug:

PR #6606, opened 2022-03-13
Replace Font-Awesome with Bootstrap Icons

Carlgo11:
Apologies as this turned out to be a rather massive pull request.
As Font Awesome goes further and further away from FOSS I thought it could be a good idea to replace it with something that better aligns with this project. Bootstrap Icons have equivalent icons for almost all Font Awesome icons that we use and using Bootstrap Icons would result in a smaller library size for all users.

I had some issues with the webfont version of the documentation and checkmark icons which is why those icons use SVG instead of webfonts.

As the Bootstrap Icons project is rather new and a few categories don't have equivalent icons I couldn't remove Font Awesome completely but instead only have the Solid icons which still results in a smaller size.

Here's a screenshot of the page with Bootstrap Icons:

### DIFF
--- a/_data/categories.json
+++ b/_data/categories.json
@@ -2,32 +2,32 @@
   {
     "name": "backup",
     "title": "Backup and Sync",
-    "icon": "fas fa-cloud-upload-alt"
+    "icon": "bi bi-cloud-arrow-up-fill"
   },
   {
     "name": "banking",
     "title": "Banking",
-    "icon": "fas fa-dollar-sign"
+    "icon": "bi bi-currency-dollar"
   },
   {
     "name": "betting",
     "title": "Betting",
-    "icon": "fas fa-trophy"
+    "icon": "bi bi-trophy-fill"
   },
   {
     "name": "cloud",
     "title": "Cloud Computing",
-    "icon": "fas fa-cloud"
+    "icon": "bi bi-cloud-fill"
   },
   {
     "name": "communication",
     "title": "Communication",
-    "icon": "fas fa-comments"
+    "icon": "bi bi-chat-left-fill"
   },
   {
     "name": "creativity",
     "title": "Creativity",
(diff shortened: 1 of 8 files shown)


phallobst:
In my opinion, the Bootstrap icons look worse than the FontAwesome ones (and these weren't great to begin with). Any chance that we can switch to something else, e.g. https://uxwing.com/ or some other collection? I'd rather have a good quality SVG icon instead of a smaller, uglier webfont.

Carlgo11:
@phallobst I think Bootstrap Icons provide a similar style as the Font Awesome icons;
A modern two-tone style allowing for universal recognition of the icon's meaning and while the icons of Bootstrap Icons tend to use more rounded lines than Font Awesome I believe it simply reflects the UI design trend we've seen in the last ten years.

As iconography is about conveying a message quickly to everyone, regardless of background, the icons in a pack should follow a similar style. Modern icons also tend to use as little detail as possible whilst still conveying the intended message.
The [uxwing](https://uxwing.com/) icon pack that you linked to doesn't have a unified design style and many icons incorporate text which I'd argue is somewhat of a red flag in iconography as it reflects an underlying issue of icon readability.

I'm open to replacing Bootstrap Icons with something better if presented but I think this resource is the currently best option as it's both interoperable with Font Awesome, allows for consistent iconography across the website and doesn't make the barrier to entry higher for outsiders who want to contribute to the project, something which I think linking to SVG images as icons would.

The font size of the Bootstrap Icons reflects the size of Font Awesome icons so the icon size on the end product should be indistinguishable from the previous version.
I'm unaware of the woff2 version containing less detail than the SVG version.

phallobst:
How about [Line Awesome](https://icons8.com/line-awesome)? This is a drop-in replacement for Font Awesome (i.e. webfont) and uses the [Good Boy License](https://icons8.com/good-boy-license).

PR #6615, opened 2022-03-25
Update OKX

40504:


Carlgo11 (commented):
A region array excluding the US and UK might be useful

40504:
> A region array excluding the US and UK might be useful

Done ✅

Carlgo11 (approved):
Thanks!

phallobst (approved):

PR #6622, opened 2022-03-28
Add Make, formerly Integromat

40504:


Carlgo11 (approved):

phallobst:
The `documentation` should probably not point to a page for Integromat anymore. Can you provide an updated link?

And while we're at it, we could change the keyword to `developer` as well.

40504:
> The `documentation` should probably not point to a page for Integromat anymore. Can you provide an updated link?
> 
> 
> 
> And while we're at it, we could change the keyword to `developer` as well.

I was searching for a new documentation but without success. I successfully enabled 2fa option for make.com, I'll upload confirmation screenshot later.

40504:
Artboard-1

phallobst (approved):

PR #6624, opened 2022-03-28
add support.gitlab.com

tendervittles:
I'd like to differentiate `support.gitlab.com` from `gitlab.com`.

The former does not support 2fa, while the latter does.


Carlgo11:
1. Chances are you're here because your password manager of choice uses our data but according to our spec any potential subdomains should be ignored and thus you'll just end up with two entries for the same domain.
2. We have a site rank criteria which requires a site to be ranked below 200K on SimilarWeb (previously Alexa). support.gitlab.com doesn't even have a ranking: https://www.similarweb.com/website/support.gitlab.com/#overview

PR #6647, opened 2022-04-09
Add additional domains

tendervittles:
Bitwarden has equivalent domains to denote shared login information.

This seems to fit with 2fa's additional domains.

I've added a few in this PR.


Carlgo11 (approved):
Thanks for the pull request @tendervittles!
Do you know if Bitwarden reads additiona-domains attributes? I don't think 1Password does it yet :thinking:

tendervittles:
@Carlgo11 

Bitwarden does indeed check for [additonal-domains](https://github.com/bitwarden/web/blob/master/src/app/reports/inactive-two-factor-report.component.ts#L102).

I just wanted to test the waters with a small PR. I'll make a few more later with additional equivalent domains.

Carlgo11:
Found [inactive-two-factor-report.component.ts](https://github.com/bitwarden/web/blob/master/src/app/reports/inactive-two-factor-report.component.ts) which answered my question. Also looks like the unsigned API files are used by Bitwarden :unamused:

dsoegijono (approved):

PR #6648, opened 2022-04-10
Add Formula 1

Carlgo11:
Formula 1 provides subscription services like F1TV and Live Timing. Like Netflix but more expensive and with only one show :laughing:


dsoegijono (approved):

phallobst (commented):
Their official logo is just red - should we go with that one? It should still be perfectly visible on our pale red background.

Carlgo11:
I could do that but the red on red doesn't meet the minimum WCAG contrast standards for accessibility. Your call :shrug: 
Screenshot

phallobst:
The logos are "decorative", not "informative" (according to https://www.w3.org/WAI/tutorials/images/), since the name of the website is right next to it. I don't think accessibility is a deciding factor there.

phallobst (approved):

PR #6650, opened 2022-04-10
Update VoIP.ms regions

Carlgo11:


entries/v/voip.ms.json
```diff
@@ -9,6 +9,10 @@
     "notes": "Only one of the 2FA methods can be enabled at the same time.",
     "keywords": [
       "utilities"
+    ],
+    "regions": [
+      "ca",
+      "us"
```
phallobst, line 15: https://wiki.voip.ms/article/Choosing_Server shows servers in Europe and Australia as well. Do you have further information why only `us` and `ca` should be included?

Carlgo11: * The title of the site is "US and Canada VoIP Internet phone service"
* The price plans are for either Canada or the U.S.

phallobst: Well, on the front page it says "Trusted by customers all around the globe", and the sign-up dialog has a large number of countries in the dropdown, so... :man_shrugging: 

In the [Terms and Conditions](https://voip.ms/terms-of-service) they explicitly exclude certain countries - that would not be necessary if only US/CA were available in the first place.

Carlgo11: I think a distinction between intended market and availability needs to be made.
In previous PRs we've used the regions array in cases where a site is available for the entire internet but the company mainly serves a select region. Examples of this are _hosting services_ (Glesys, eUKhost, UKFast, United-domains.de) and _post/shipping companies_ (Deutsche Post, Canada Post, Australia Post, PostNord).
As the intended market seems to be North America based on my previous observations plus the fact that they don't comply with GDPR further indicating that international customers aren't intended.

I just create an account on their website where I set my address as North Korea and I didn't have any problems completing the process but I assume that's not an intended customer base.
If I were to speculate I'd say the countries listed in their terms of service are solely because of previous attempts by those regimes to enforce local laws onto VoIP.ms. Saudi Arabia for example requires biometrics linked to all cell phone numbers.

ApCoder123 (approved):

PR #6655, opened 2022-04-13
Update ProtonMail

40504:
https://protonmail.com/blog/protonme-domain/


Carlgo11:
As the auth system will still only be on protonmail.com I don't think these domains are useful to add.

40504:
when is suitable to add them?

phallobst:
While `additional-domains` are certainly useful for password managers, they also act as search keywords. I'd say add them anyway.

40504:
Ok, I get it, thank you for the explanations. You know better what to do with this PR :)

Carlgo11:
@40504 as Michael mentioned, the additional-domains element is primarily meant for third party apps like password managers.
While we also use it for our search field on 2fa.directory I doubt anyone would search for these specific domains as they are only meant to handle SMTP traffic and anyone trying to access the domains through HTTP/HTTPS would be redirected to the primary domain, protonmail.com.

PR #6666, opened 2022-04-15
Fix Bootstrap 5 CSS bugs

Carlgo11:
Some CSS issues that I missed in #6485

The issues fixed are:
* Regions dropdown not clickable
* tables don't include background color on mobile in with light theme

In my defense for not noticing the last bug, no sane person writes code without system dark theme enabled 😉

### DIFF
--- a/_includes/html/navbar.html
+++ b/_includes/html/navbar.html
@@ -13,7 +13,7 @@
       {% assign regions2 = regions | map: 'id' %}
       {% if page.name == "index.html" %}
       <li class="nav-item dropdown">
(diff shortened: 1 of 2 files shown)


phallobst:
> ... no sane person writes code without system dark theme enabled :wink:

TIL I'm insane... (That does explain a lot, actually.)

"About" and "Region" are left-aligned nowadays - was that intentional?

Carlgo11:
The left alignment was unintentional. Want it fixed?

phallobst:
> The left alignment was unintentional. Want it fixed?

 Maybe it's better to leave it on the left, in case we want/need to have a multi-column _Region_ dropdown in the future.

phallobst (approved):

jamcat22 (approved):

PR #6220, opened 2021-10-19
Accessibility improvements

phallobst:
Fixes #6215

Draft PR to collect ideas - feel free to add bullets points and/or commits.

- [x] remove useless "back" button on about page
- [x] remove filler text
- [x] add empty alt text on about page
- [x] higher contrast for link color
- [x] unique title for region sites
- [x] fix h1/h2 header nesting
- [x] add `tabindex=0` to `regionDropdown`
- [ ] `href` links on category icons (`window.onhashchange`)
- [ ] add `aria-label` to the icons in the table
- [ ] turn social media buttons into links
- [ ] add `role` for `<div>` elements

Issue #6215, opened 2021-10-18
Website Accessibility


not-my-profile:
Hey, nice project :) Just FYI you could improve the accessibility of your website with the following simple changes.

**keyboard accessibility** (website should be navigable with just `Tab`/`Shift+Tab` and `Enter`)

* add `tabindex=0` to `regionDropdown`
* turn the category buttons into actual links e.g. `href=#backup` and expand the tables by listening to `window.onhashchange` instead of the `click` event

**screen reader accessibility**

* add `aria-label` to the icons in the table

The social buttons are also inaccessible. You could add `tabindex=0 role=button` but even better would be turning them into regular links.

I unfortunately don't have time to implement these changes; only for reporting^^ But if you have any questions I can try to answer them :)

Carlgo11:
Seems the requested changes have all been added so I'll close this issue

PR #6668, opened 2022-04-16
Fix some accessibility issues

Carlgo11:
Ran the site through Google Lighthouse and found some accessibility issues.
Some of the issues are already mentioned in #6215 whilst some aren't. None of these are addressed in #6220.

* Made `.site-name` text thicker and with a darker color (same color as dark theme bg)
* Set aria-label on documentation and recovery link buttons.
* Removed duplicate IDs (both the table and `.category-btn` used the category name as their ID value)
* Fixed aria-label on `.flag-icon` being ignored on some screen readers due to lack of element role.

And this time I checked that it looks good in both dark theme and light theme :wink:


phallobst:
On desktop view, the table entries now have the same weight as the table header, which looks a bit off.

I also think that a text link should not just be text. Maybe change at least the color on hovering?

Everything else is okay for me.

Carlgo11:
@phallobst The `thead > th` have a font weight of **700** whilst `.site-name` have a font weight of **600**.
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#syntax) **700** is **bold** while **600** is _**"semi-bold"**_. I guess the thickness of the characters also depends on what font you're using :shrug:
I do however agree that it looks better with a smaller font weight. **500** might be the most appropriate weight

700 th & 500 site-name:
example.exe.mov

I opted for a black text color instead of the `a` default color as it's hard to read blue on green (even when it's #005AA0) and the link surrounds not just to the site name but the outer column making it more of a button than a normal web 1.0 link.

phallobst (approved):

PR #6671, opened 2022-04-17
Split Klarna

Carlgo11:
Klarna offers:
* payment services internationally and authenticates through email/SMS/calls (except in NO & SE).
* banking and payment services to Norway and Sweden using software authentication called BankID.

_Note that "BankID" refers to different methods in the two different countries as it's run by completely unrelated companies. It thus can't be merged into the same entry._

To not make duplicate entries I had to use different domain names for all the entries resulting in Klarna SE and Klarna NO not receiving the appropriate SimilarWeb ranking as the domain which they 301 to. Should still be possible to merge though.

### DIFF
--- a/entries/k/klarna.com.json
+++ b/entries/k/klarna.com.json
@@ -1,12 +1,13 @@
 {
   "Klarna": {
     "domain": "klarna.com",
-    "url": "https://www.klarna.com",
     "tfa": [
       "email",
-      "custom-software"
+      "sms",
+      "call"
     ],
-    "notes": "Software 2FA is only available for Swedish citizens with BankID.",
+    "documentation": "https://www.klarna.com/us/customer-service/how-do-i-log-in/",
+    "recovery": "https://www.klarna.com/us/customer-service/i-cant-log-in-what-can-i-do/",
     "keywords": [
       "payments"
     ],
@@ -25,11 +26,9 @@
       "ie",
       "it",
       "nl",
-      "no",
       "nz",
       "pl",
       "pt",
-      "se",
       "us"
     ]
   }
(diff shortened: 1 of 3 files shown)


entries/k/klarna.com.json
```diff
@@ -25,11 +26,9 @@
       "ie",
       "it",
       "nl",
-      "no",
       "nz",
       "pl",
       "pt",
-      "se",
```
phallobst, line 32: Shouldn't these be `"-no"` and `"-se"` instead?

entries/k/klarna.com.json
```diff
@@ -1,36 +1,19 @@
 {
   "Klarna": {
     "domain": "klarna.com",
-    "url": "https://www.klarna.com",
     "tfa": [
       "email",
```
ApCoder123: > You will receive an email with a login link or a text with a 6-digit verification code 

Do we consider this to be 2FA? Sounds like passwordless

Carlgo11: Yeah I guess that's just one factor :thinking:

entries/k/klarna.com.json
```diff
@@ -1,36 +1,16 @@
...
-      "fr",
-      "ie",
-      "it",
-      "nl",
-      "no",
-      "nz",
-      "pl",
-      "pt",
-      "se",
-      "us"
```
ApCoder123: Shouldn't these regions remain here as Klarna is only available in [these countries](https://www.klarna.com/international/)?

ApCoder123 (changes requested):

ApCoder123 (approved):

PR #6689, opened 2022-04-30
more additonal domains

tendervittles:
These domains are equivalent on bitwarden.

### DIFF
--- a/entries/f/facebook.com.json
+++ b/entries/f/facebook.com.json
@@ -1,6 +1,9 @@
 {
   "Facebook": {
     "domain": "facebook.com",
+    "additional-domains": [
+      "messenger.com"
+    ],
     "tfa": [
       "sms",
       "totp",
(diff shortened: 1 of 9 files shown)


Carlgo11:
Do you have any more?

tendervittles:
That is it for me.

Carlgo11 (approved):

entries/f/facebook.com.json
```diff
@@ -1,6 +1,9 @@
 {
   "Facebook": {
     "domain": "facebook.com",
+    "additional-domains": [
+      "messenger.com"
+    ],
```
phallobst, line 6: I'm not quite sure about this one. It seems to deserve a separate entry in `communication`.

Carlgo11: I'm not sure I agree here. As Messanger is a part of the Facebook experience and can't be severed from it I don't think a separate entry is needed for it. It's basically the PM feature of Facebook like the DMs on Twitter or private messages on Discord.

person1234-ux: I believe this should just be give added as an addition domain. As said before it's basically like the pm feature on twitter.

PR #6694, opened 2022-05-04
Update lhv.ee

Avamander:
The current selection of 2FA methods is somewhat incomplete. I chose "custom-hardware" even though mTLS using Smart Cards isn't more proprietary than say Yubikey Webauthn, but there's no option for that. LHV and all other Estonian banks have supported mTLS authentication for 8 or so years now. I also added "custom-software" because by now there is a mobile app as well. 

Both methods fall under the EU-wide eIDAS regulation/specification, which is now being adopted by more and more countries each year. So maybe it should just be "software-eidas" and "hardware-eidas", if there's a wish to differentiate.

I should also mention that all entries concerning EU-member country banks should probably be updated if they currently list no 2FA. This specific bank offered these options before PSD2 was turned into a law in 2017, but PSD2 has made it mandatory EU-wide.

### DIFF
--- a/entries/l/lhv.ee.json
+++ b/entries/l/lhv.ee.json
@@ -1,7 +1,6 @@
 {
   "LHV": {
     "domain": "lhv.ee",
-    "url": "https://www.lhv.ee/",
     "tfa": [
       "custom-software",
       "custom-hardware"
--- a/entries/l/lhv.ee.json
+++ b/entries/l/lhv.ee.json
@@ -2,12 +2,6 @@
   "LHV": {
     "domain": "lhv.ee",
     "url": "https://www.lhv.ee/",
-    "contact": {
-      "email": "info@lhv.ee",
-      "facebook": "LHVPank",
-      "twitter": "ellhaavee",
-      "language": "et"
-    },
     "tfa": [
       "custom-software",
       "custom-hardware"
--- a/entries/l/lhv.ee.json
+++ b/entries/l/lhv.ee.json
@@ -5,6 +5,13 @@
       "custom-software",
       "custom-hardware"
     ],
+    "custom-software": [
+      "eIDAS (SmartID)"
+    ],
+    "custom-hardware": [
+      "eIDAS (ID-card, mobile-ID)",
+      "Security token"
+    ],
     "regions": [
       "ee"
     ],
--- a/entries/l/lhv.ee.json
+++ b/entries/l/lhv.ee.json
@@ -6,10 +6,11 @@
       "custom-hardware"
     ],
     "custom-software": [
-      "eIDAS (SmartID)"
+      "SmartID"
     ],
     "custom-hardware": [
-      "eIDAS (ID-card, mobile-ID)",
+      "ID-card",
+      "mobile-ID",
       "Security token"
     ],
     "regions": [
--- a/entries/l/lhv.ee.json
+++ b/entries/l/lhv.ee.json
@@ -8,6 +8,10 @@
       "twitter": "ellhaavee",
       "language": "et"
     },
+    "tfa": [
+      "custom-software",
+      "custom-hardware"
+    ],
     "regions": [
       "ee"
     ],


entries/l/lhv.ee.json
```diff
@@ -2,12 +2,10 @@
...
     "url": "https://www.lhv.ee/",
-    "contact": {
-      "email": "info@lhv.ee",
-      "facebook": "LHVPank",
-      "twitter": "ellhaavee",
-      "language": "et"
-    },
+    "tfa": [
+      "custom-software",
+      "custom-hardware"
```
Carlgo11, line 6: `custom-software` and `custom-hardware` tags are needed since they have been included in the `tfa` array.

entries/l/lhv.ee.json
```diff
@@ -2,12 +2,10 @@
   "LHV": {
     "domain": "lhv.ee",
     "url": "https://www.lhv.ee/",
```
Carlgo11:
```suggestion
```

Carlgo11 (changes requested):

entries/l/lhv.ee.json
```diff
@@ -1,13 +1,17 @@
...
+    "tfa": [
+      "custom-software",
+      "custom-hardware"
+    ],
+    "custom-software": [
+      "eIDAS (SmartID)"
+    ],
+    "custom-hardware": [
+      "eIDAS (ID-card, mobile-ID)",
+      "Security token"
```
Carlgo11: As only certificates validated by Estonian authorities are accepted it's not really true eIDAS is it? Because from reading https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=celex%3A32015D0296 it looks like interoperability is required for it to be eIDAS.

Avamander: It is true eIDAS.

Interoperability doesn't mean a business has to offer their services to everyone in the EU.

Carlgo11: True but saying they support eIDAS software when only one eIDAS software client is supported is a bit disingenuous. Leave it as SmartID for the software and ID-card, mobile-ID for the hardware instead.

Avamander: I think there might be a misunderstanding actually, there is no "one client", it's mTLS built into browsers and Smart Card support built-in or available on each OS. This functionality is very interoperable. LHV just limits access only to their customers.

It's not disingenuous. In the same way, if someone required a specific manufacturer's attestation with Webauthn, it'd still be Webauthn.

Carlgo11: So the user receives a x509 certificate to add to their browser?

> It's not disingenuous. In the same way, if someone required a specific manufacturer's attestation with Webauthn, it'd still be Webauthn.

If Google only allowed WebAuthn from Yubikey devices we'd label it as custom-hardware with a value of `Yubikey`.
For general consumers the behind the scenes technology is less important than the actual apps/devices that the can use. That's why the phrasing on our site is "Required hardware token: x" and not "Supported hardware standard: x".

Avamander: > So the user receives a x509 certificate to add to their browser?

Pretty much.

> If Google only allowed WebAuthn from Yubikey devices we'd label it as custom-hardware with a value of Yubikey.

And that's what it is currently, it is custom hardware with a value that describes the requirements. This doesn't mean Google doesn't follow Webauthn or that the devices can't be used elsewhere.

entries/l/lhv.ee.json
```diff
@@ -1,13 +1,17 @@
...
-    },
+    "tfa": [
+      "custom-software",
+      "custom-hardware"
+    ],
+    "custom-software": [
+      "eIDAS (SmartID)"
+    ],
+    "custom-hardware": [
+      "eIDAS (ID-card, mobile-ID)",
```
Carlgo11:
```suggestion
      "SmartID"
    ],
    "custom-hardware": [
      "ID-card",
      "mobile-ID",
```

Carlgo11 (changes requested):

Carlgo11 (approved):

phallobst (approved):

PR #6695, opened 2022-05-04
Update PayPal

Avamander:
Symantec VIP (custom-ish TOTP) is being deprecated, the listed alternatives are:

* Mobile phone SMS
* Authenticator apps: Options include Authy, Google Authenticator, Microsoft Authenticator, DUO, YubiKey Auth App, and many others. 
* YubiKey 5 Series: All of its desktop, authenticator app, and hard tokens will work for primary and secondary users via the “Authenticator Option” in the 2-step verification process.

### DIFF
--- a/entries/p/paypal.com.json
+++ b/entries/p/paypal.com.json
@@ -5,10 +5,9 @@
     "tfa": [
       "sms",
       "totp",
-      "u2f",
-      "custom-software"
+      "u2f"
     ],
-    "documentation": "https://www.paypal.com/ee/smarthelp/article/faq4502",
+    "documentation": "https://www.paypal.com/ee/smarthelp/article/faq4057",
     "keywords": [
       "payments"
     ]
--- a/entries/p/paypal.com.json
+++ b/entries/p/paypal.com.json
@@ -4,10 +4,11 @@
     "url": "https://www.paypal.com/",
     "tfa": [
       "sms",
-      "totp"
+      "totp",
+      "u2f",
+      "custom-software"
     ],
-    "documentation": "https://www.paypal.com/us/smarthelp/article/faq4057",
-    "notes": "2FA is only available in a few countries. Contact PayPal on Twitter for an up-to-date list of available countries.",
+    "documentation": "https://www.paypal.com/ee/smarthelp/article/faq4502",
     "keywords": [
       "payments"
     ]
--- a/entries/p/paypal.com.json
+++ b/entries/p/paypal.com.json
@@ -7,7 +7,7 @@
       "totp",
       "u2f"
     ],
-    "documentation": "https://www.paypal.com/ee/smarthelp/article/faq4057",
+    "documentation": "https://www.paypal.com/us/smarthelp/article/faq4057",
     "keywords": [
       "payments"
     ]
--- a/entries/p/paypal.com.json
+++ b/entries/p/paypal.com.json
@@ -7,6 +7,7 @@
       "totp",
       "u2f"
     ],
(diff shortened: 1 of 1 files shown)


entries/p/paypal.com.json
```diff
@@ -4,12 +4,13 @@
     "url": "https://www.paypal.com/",
     "tfa": [
       "sms",
-      "totp"
+      "totp",
+      "u2f",
+      "custom-software"
     ],
-    "documentation": "https://www.paypal.com/us/smarthelp/article/faq4057",
```
Carlgo11, line 9: This documentation page does still explain the 2FA setup process better than [faq4502](https://www.paypal.com/ee/smarthelp/article/faq4502). I think it should stay.

entries/p/paypal.com.json
```diff
@@ -4,12 +4,13 @@
     "url": "https://www.paypal.com/",
     "tfa": [
       "sms",
-      "totp"
+      "totp",
+      "u2f",
+      "custom-software"
```
Carlgo11: Since the only custom/proprietary 2FA software is Symantec VIP which, as you said, is deprecated if not entirely removed by now, I don't think `custom-software` should be listed.

entries/p/paypal.com.json
```diff
@@ -4,12 +4,13 @@
     "url": "https://www.paypal.com/",
     "tfa": [
       "sms",
-      "totp"
+      "totp",
+      "u2f",
```
Carlgo11: The documentation page mentions FIDO2 and U2F but I don't see a setting for it. It's possible this isn't available anymore.

Carlgo11 (changes requested):

entries/p/paypal.com.json
Avamander: It certainly is available. It might be hidden if your browser doesn't support it.

entries/p/paypal.com.json
```diff
@@ -4,12 +4,12 @@
     "url": "https://www.paypal.com/",
     "tfa": [
       "sms",
-      "totp"
+      "totp",
+      "u2f"
     ],
-    "documentation": "https://www.paypal.com/us/smarthelp/article/faq4057",
-    "notes": "2FA is only available in a few countries. Contact PayPal on Twitter for an up-to-date list of available countries.",
+    "documentation": "https://www.paypal.com/ee/smarthelp/article/faq4057",
```
phallobst:
```suggestion
    "documentation": "https://www.paypal.com/us/smarthelp/article/faq4057",
```

phallobst (changes requested):

entries/p/paypal.com.json
phallobst: It's not available for me either (tested with multiple browsers), although i'm pretty sure that I set up a YubiKey before. Maybe this feature has been deprecated.

But while we're at it: Logging in on the Android app allowed for a phone call where I had to enter the displayed code. So `call` might need to be added as well.

Avamander: It has not been deprecated, it's there and I just tested it out.

Avamander: I wouldn't want to drag on the PR for too long, adding "call" should probably be a separate PR.

Carlgo11: I contacted PayPal support and they say it's only available in certain regions. I therefore suggest a note saying `2FA methods vary depending on region. Contact PayPal on Twitter for an up-to-date list of available methods in your region.`

Carlgo11 (approved):

phallobst (approved):

PR #6706, opened 2022-05-10
Add Magnet.me

rogierslag:
Hi there!

[Magnet.me](https://magnet.me/jobseekers) is an online platform primarily active in the Netherlands and the UK matching job seekers with jobs. As a result many people (over 300k+ accounts) have since uploaded information on themselves or their work and education history. Since 2FA has been added, it seems like a wise idea to let as many users as possible use 2FA.

Thanks for your hard work, and I hope this may be a valuable inclusion.


person1234-ux (commented):
I tested and the everything works as expected

phallobst:
Currently this website has a Similarweb rank of over 200,000. This PR will therefore be closed, but you are welcome to re-open it once the rank drops below that limit.

PR #6725, opened 2022-05-15
Support '?q=foobar' for automatic searches

### LINKED ISSUE #18: Pre-populated search field
- [x] I have searched [open issues and pull requests](https://github.com/2factorauth/twofactorauth/issues?q=is%3Aopen). The issue I'm creating is not a duplicate of an existing open issue or pull request.

Sometimes I show some friends this site when they ask me about 2fa for a specific website. A much easier way would be a pre-populated search field, so they find the information faster. It could also be handy for links from another website.

This could be done client side:
```js
// https://2fa.directory/int/?q=github
let search_field = new URLSearchParams(window.location.search).get('q');
```

### LINKED ISSUE #6680: Pre-populated search field
- [x] I have searched [open issues and pull requests](https://github.com/2factorauth/twofactorauth/issues?q=is%3Aopen). The issue I'm creating is not a duplicate of an existing open issue or pull request.

Sometimes I show some friends this site when they ask me about 2fa for a specific website. A much easier way would be a pre-populated search field, so they find the information faster. It could also be handy for links from another website.

This could be done client side:
```js
// https://2fa.directory/int/?q=github
let search_field = new URLSearchParams(window.location.search).get('q');
```

phallobst:
Closes 2factorauth/frontend#18

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -63,6 +63,10 @@ $(document).ready(function () {
   }).on('search', function () {
     jets.search(this.value);
   });
+
+  // https://2fa.directory/?q=foobar
+  const query = new URLSearchParams(window.location.search).get('q');
+  if (query) $('#innerSearchBox').val(query).trigger('search');
 });
 
 // Wrap the jets.search function with a debounced function


Carlgo11:
`https://2fa.directory/?q=foobar` bypasses the redirection script and sends the user to the root (non-int) page. Is this the intended result?

phallobst:
I addressed this concern on the issue as well. One could either use `https://2fa.directory/int/?q=foobar` to make sure the result is displayed in any case, or leave the region out to force the search within the redirected region.

Maybe we could alter the text "_No results found._" to "_No results found for this region._" if applicable.

Carlgo11:
I think the option of redirecting to a local region is fine and the option of redirecting to the int site is fine but I don't think the current option is a good alternative. The global non-int page is an undocumented page which very likely will be removed in the near future.
If this search param is to be used I think the [redirect script](https://github.com/2factorauth/redirect/) should be updated to catch these requests and handle them in one of the ways mentioned above.

sn00py1310:
As mentioned in [2factorauth/redirect#1](https://github.com/2factorauth/redirect/issues/1#issuecomment-1170253774) there aren't any practical solutions with the redirect script.
A possible workaround could be redirecting the user with clientside Javascript always from the root to the /int/ page. So the feature would at least be usable.

PR #6741, opened 2022-05-22
Small svg

person1234-ux:


person1234-ux:
did I just run all the svg files through a script that shrinks them. The answer to that question is yes

phallobst:
Nobody is going to review 1200+ icons. As long as there is no _significant_ improvement in file size for some image, this PR is of not much use, I suppose.

person1234-ux:
lots of insignificant changes bundled together make a significant change

person1234-ux:
> Nobody is going to review 1200+ icons. As long as there is no _significant_ improvement in file size for some image, this PR is of not much use, I suppose.

Ok I will try to see if I can modify the script to only include significant changes.

PR #6753, opened 2022-05-29
Add more regions

Carlgo11:
Adding more regions that have more than 10 entries.

The regions are:
* China (mainland)
* Chile
* Costa Rica
* South Korea
* Hong Kong
* Peru
* Slovakia
* Taiwan

### DIFF
--- a/_data/regions.yml
+++ b/_data/regions.yml
@@ -21,7 +21,7 @@
   square_flag: true
 
 - id: cn
-  name: China Mainland
+  name: China
 
 - id: cl
   name: Chile
--- a/_data/regions.yml
+++ b/_data/regions.yml
@@ -20,6 +20,15 @@
   name: Switzerland
   square_flag: true
 
+- id: cn
+  name: China Mainland
+
+- id: cl
+  name: Chile
+
+- id: cr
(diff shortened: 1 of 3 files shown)


Carlgo11:
Unsure how China should be displayed. I'm thinking either as `China Mainland`, `Mainland China` or `China (Mainland)`

phallobst:
Everybody knows what "China" is supposed to mean in this context, especially since "Hong Kong" and "Taiwan" are displayed as well, so just stick with that. We're not practising international politics here.

Besides, the `.dropdown-menu` should use multiple columns, as it's getting quite large already (`column-count: 2;` in CSS).
```

Carlgo11:
@phallobst my thought was that since HK and TW are separate entries there was a need for clarification on what the CN region contained.

Regarding col-count, is this what you're after?
screenshot

phallobst:
The two-column list look good - however, I tried this on mobile and the list was slightly shifted out of view on the right. Swapping the "About" button with the "Region" dropdown should fix this (on most devices).

Carlgo11:
new layout:
desktop
phone

phallobst (approved):

PR #6762, opened 2022-06-06
Update int page

Carlgo11:
I've reworked the regions.rb script to not display sites if:
* They have a regions element
* All of the regions in the regions-element have their own page on 2fa.directory (i.e. "[selectable](https://2fa.directory/api/v3/regions.json)")

That way, people from regions without a dedicated page on 2fa.directory don't see sites only relevant for another region.
i.e. if you're from Luxembourg the banking category isn't filled with U.S. credit unions.

I also replaced `2fa.directory/index.html` with a simple HTML-page linking to the international page. This page should never be displayed to users but if it is shown, we're now more likely to notice that something is wrong with the [redirect script](https://github.com/2factorauth/redirect).

### DIFF
--- a/scripts/regions.rb
+++ b/scripts/regions.rb
@@ -19,7 +19,7 @@ Parallel.map(regions.keys) do |key|
   region = regions[key]
   dest_dir = "#{tmp_dir}/#{key}"
   Dir.mkdir(dest_dir) unless File.exist?(dest_dir)
-  FileUtils.cp_r(%w[index.html _includes _layouts], dest_dir)
+  FileUtils.cp_r(%w[index.html _includes _layouts _data], dest_dir)
   File.open("#{dest_dir}/_config_region.yml", 'w') do |file|
     file.write("title: 2FA Directory (#{region['name']})") unless key.eql?('int')
   end
@@ -38,7 +38,6 @@ Parallel.map(regions.keys) do |key|
     website['keywords'].each { |kw| used_categories.push kw }
   end
 
-  FileUtils.mkdir_p("#{dest_dir}/_data")
   File.open("#{dest_dir}/_data/all.json", 'w') { |file| file.write JSON.generate(all) }
   File.open("#{dest_dir}/_data/categories.json", 'w') do |file|
     file.write JSON.generate(categories.select { |cat| used_categories.include? cat['name'] })
--- a/scripts/regions.rb
+++ b/scripts/regions.rb
@@ -8,22 +8,19 @@ require 'parallel'
 
 data_dir = './_data'
 categories = JSON.parse(File.read("#{data_dir}/categories.json"))
-websites = JSON.parse(File.read("#{data_dir}/all.json"))
-regions_array = YAML.load_file("#{data_dir}/regions.yml")
-regions = regions_array.map { |r| [r['id'], r] }.to_h
+regions = YAML.load_file("#{data_dir}/regions.yml").map { |r| [r['id'], r] }.to_h
 regions['int'] = { 'name' => 'Global' }
-tmp_dir = '/tmp/2fa'
-git_dir = Dir.glob('.git')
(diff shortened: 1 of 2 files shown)


phallobst:
Unfortunately, this also means that a specific site may not come up in the search results if the user is not on a matching region. At least with `int`, this was guaranteed. I'm afraid this is a step in the wrong direction. (Handling regions like a search term would be way easier IMHO.)

Carlgo11:
Alright I've reverted that change. Now all sites are displayed on the int page regardless of relevance.
I did however do some cleanup of the regions.rb script and the change to `2fa.directory/index.html` is still included.

scripts/regions.rb
```diff
@@ -7,60 +7,46 @@
...
 
 data_dir = './_data'
-websites = JSON.parse(File.read("#{data_dir}/all.json"))
-regions = YAML.load_file("#{data_dir}/regions.yml")
-tmp_dir = '/tmp'
-regions.insert(0, { 'id' => 'int', 'name' => 'global' })
-git_dir = Dir.glob('.git')
-FileUtils.cp_r(git_dir, "#{tmp_dir}/") unless File.exist?("#{tmp_dir}/.git")
+categories = JSON.parse(File.read("#{data_dir}/categories.json"))
+regions = YAML.load_file("#{data_dir}/regions.yml").map { |r| [r['id'], r] }.to_h
```
ApCoder123, line 11:
```suggestion
regions = YAML.load_file("#{data_dir}/regions.yml").map { |r| [r['id'], { 'name' => r['name'] }] }.to_h
```
For consistency with 'int' key.

scripts/regions.rb
```diff
@@ -7,60 +7,46 @@
...
-    unless website['regions'].nil? || site_regions.empty? || site_regions.include?(region['id']) || region['id'].eql?('int')
-      next
-    end
-
-    next if !site_excluded_regions.nil? && site_excluded_regions.include?(region['id'])
+  JSON.parse(File.read("#{data_dir}/all.json")).each do |name, website|
+    # Ignore if current region is excluded
+    next if website['regions']&.select { |r| r.start_with?('-') }&.map! { |rc| rc.tr('-', '') }&.include?(region['id'])
+    # Ignore unless regions element includes current region or current region is int
+    next unless website['regions']&.reject { |r| r.start_with?('-') }&.include?(key) || key.eql?('int')
```
ApCoder123, line 35:
```suggestion
    next if website['regions']&.select { |r| r.start_with?('-') }&.map! { |rc| rc.tr('-', '') }&.include?(key)
    # Ignore unless site includes current region, current region is int, site has no regions or site is not excluded in current region
    regions = website['regions']&.reject { |r| r.start_with?('-') }
    next unless regions&.include?(key) || key.eql?('int') || website['regions'].nil? || regions.empty?
```
Current code causes sites with no regions to not be displayed on regional sites and sites with excluded regions to not be displayed on any sites.

scripts/regions.rb
```diff
@@ -7,60 +7,46 @@
...
-    file.write JSON.generate(categories.select { |cat| used_categories[cat['name']] })
+    file.write JSON.generate(categories.select { |cat| used_categories.include? cat['name'] })
   end
 
-  out_dir = "#{Dir.pwd}/_site/#{region['id']}"
-  puts "Building #{region['id']}..."
+  out_dir = "#{Dir.pwd}/_site/#{key}"
+  puts "Building #{key}..."
+  # rubocop:disable Layout/LineLength
   puts `bundle exec jekyll build -s #{dest_dir} --config _config.yml,#{dest_dir}/_config_region.yml -d #{out_dir} --baseurl #{region['id']}`
```
ApCoder123, line 49:
```suggestion
  puts `bundle exec jekyll build -s #{dest_dir} --config _config.yml,#{dest_dir}/_config_region.yml -d #{out_dir} --baseurl #{key}`
```

ApCoder123 (changes requested):

PR #6765, opened 2022-06-08
Add Woolworths

fletcherdodd:
As per the note in the JSON file, 2FA is enabled by default on all accounts. You must verify a number and email when you set up your account, there is no way to disable 2FA. 
There is no linkable documentation on the 2FA information, see proof below:
Woolworths 2FA


entries/w/woolworths.com.au.json
```diff
@@ -0,0 +1,16 @@
+{
+  "Woolworths": {
+    "domain": "woolworths.com.au",
+    "tfa": [
+      "sms",
+      "email"
+    ],
+    "notes": "A phone number and email are required for account setup. 2FA is turned on by default.",
```
Carlgo11: Like with a similar PR you submitted, the information isn't really relevant as it isn't a restriction nor an exception.
Good on you for being really thorough, though! :+1: 
```suggestion
```

Carlgo11 (changes requested):

Carlgo11 (approved):

PR #6777, opened 2022-06-10
Re-add TOTP support for Paycor

raidancampbell:
This was originally added and discussed in #2878, but looks like it was accidentally removed in #4436

### DIFF
--- a/entries/p/paycor.com.json
+++ b/entries/p/paycor.com.json
@@ -5,7 +5,8 @@
     "tfa": [
       "sms",
       "call",
-      "email"
+      "email",
+      "totp"
     ],
     "documentation": "https://www.paycor.com/security/how-to-keep-your-account-safe",
     "keywords": [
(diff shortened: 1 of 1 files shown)


entries/p/paycor.com.json
```diff
@@ -5,7 +5,8 @@
     "tfa": [
       "sms",
       "call",
-      "email"
+      "email",
+      "totp"
     ],
     "documentation": "https://www.paycor.com/security/how-to-keep-your-account-safe",
```
phallobst: The `totp` method is not mentioned on the old doc page, so that's probably why it was removed. Changing the link to this page should prevent that from happening in the future.
```suggestion
    "documentation": "https://support.paycor.com/s/article/Multi-Factor-Authentication-MFA-Requirements-and-Setup",
```

phallobst (changes requested):

entries/p/paycor.com.json
raidancampbell: ah, nice catch and thanks for digging up the new URL!

phallobst (approved):

PR #6782, opened 2022-06-11
Add multiple Korea websites

alexkoala:


phallobst:
Cramming that many new sites into a single PR does not really help with the review process, I'm afraid. Please split it into separate ones. (Using the same PR for all `naver.com` entries for example is okay, I guess.)

Also, we prefer SVG logos to PNGs. It would be kind of you to provide these.

Carlgo11:
Continuing on what @phallobst said,
when adding new sites please use the `regions` element for sites that that are intended for one certain countries. In this case, since all the sites target the (Republic of) Korean audience, that would look like this:
```
"regions": [
  "kr"
]
```

I'll close this PR now. You can submit new PRs for the individual sites (or for a group as long as they are related to each other).

PR #6801, opened 2022-06-18
Add BNPL sites

fletcherdodd:
Adding BNPL (Buy Now Pay Later) sites; LatitudePay, Humm & Laybuy


entries/l/latitudepay.com.json
```diff
@@ -0,0 +1,12 @@
+{
+  "LatitudePay": {
+    "domain": "latitudepay.com",
+    "contact": {
+      "facebook": "latitudefinancialservices",
```
phallobst:
```suggestion
      "facebook": "LatitudePay",
```

entries/l/latitudepay.com.json
```diff
@@ -0,0 +1,12 @@
...
+      "twitter": "latitude_fs"
+    },
+    "keywords": [
+      "payments"
+    ]
```
phallobst, line 14:
```suggestion
    ],
    "regions": [
      "au"
    ]
```

phallobst (changes requested):

entries/l/latitudepay.com.json
fletcherdodd: I would argue that the original one is the better option as it has more followers and is actually the verified account/promoted by the company

phallobst: On their [main page](https://www.latitudepay.com/), the linked Facebook account is "LatitudePay" - it doesn't get more official than that, I suppopse.

fletcherdodd: Fair, you win this round

phallobst (approved):

PR #6808, opened 2022-06-21
add Masaryk University

melanger:
Add MUNI Unified Login which supports TOTP and WebAuthn.

PR #6811, opened 2022-06-22
docs: update CONTRIBUTING.md

melanger:
based on discussion in #6808

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,6 +103,7 @@ the corresponding [subdirectory][entries] as shown in the following example:
   }
 }
 ```
+- The domain should point to the main page of the service, not the login page (usually the root domain, not a subdomain).
 - Keywords must be selected from the values listed in [`categories.json`][categories].
 - The default value for the icon is `<domain>.svg`, but can be overridden by an `img`
 field.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -103,6 +103,7 @@ the corresponding [subdirectory][entries] as shown in the following example:
   }
 }
 ```
+- Only top level domains are allowed for new entries
```
Carlgo11: Top level domain is the first part of the domain, like `com`, `net`, `cz`. What you're after is called the root domain.


I wrote a simplified version of the rule to you in #6808. The domain element is meant to reflect the entire service of which the listing is for. In your case that was the university and so I therefore requested you to remove the subdomain. The idea is that when third party tools, like password managers, check if the site supports 2FA, it will first check the current FQDN and then work up to the root domain.
Due to the lack of standardization of domain name usage on the Internet, sometimes there are services which sit under a subdomain, unrelated to what the root domain points to.

phallobst: Maybe the text should say something like
`The domain should point to the main page of the service, not the login page.`

melanger: > The idea is that when third party tools, like password managers, check if the site supports 2FA, it will first check the current FQDN and then work up to the root domain.

Unfortunately it seems that some (at least Bitwarden) only check the domain/subdomains for which the password is saved, e.g. when I have a password for sub.example.com and example.com is listed in 2FA directory, the password manager does not tell me that I should set up 2FA. But I get your point and I agree, password managers should work up to the root domain.

I will update my suggestion.

Carlgo11 (approved):

PR #6812, opened 2022-06-22
Incremental changes to gh-pages

Carlgo11:
[keep_history](https://github.com/crazy-max/ghaction-github-pages#inputs) allows for accurate modification dates on published files which could improve caching as well as the sitemap generation.

### DIFF
--- a/.github/workflows/repository.yml
+++ b/.github/workflows/repository.yml
@@ -146,5 +146,6 @@ jobs:
           jekyll: false
           fqdn: ${{ secrets.DOMAIN }}
           commit_message: "Deploy ${{GITHUB.SHA}}"
+          keep_history: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


Carlgo11:
Tested this and it looks like both the file modification timestamp and git commit timestamp are disregarded when the `gh-pages` branch is read by GitHub Pages. The only potential benefit then, is to get accurate modification intervals for sitemap.xml. This can be done with:
```sh
# Fetch gh-pages branch
git clone -b gh-pages git@github.com:2factorauth/twofactorauth.git /tmp/gh-pages

for file in /tmp/gh-pages/*; do
  name=$(basename $file)
  # Check if file from gh-pages exists in new code
  if [ -f "_site/$name" ]; then
    # Only copy if the file hasn't been modified
    diff $file "_site/$name" 1>/dev/null && touch -r "$file" "_site/$name"
  fi
done

# Now generate sitemap.xml
./scripts/sitemap.sh _site

```
Though I'm not sure this is something worth doing. What do you think @phallobst?

phallobst:
First of all, `basename` yields the actual filename without directories, i.e. `basename /tmp/gh-pages/foo/bar` equals `bar` - that's probably not what you want.

In this case, maybe the use of `rsync` is better, e.g. `rsync -r -c $SOURCE/ $TARGET` would recursively (`-r`) copy only those files that differ by their checksum (`-c`). The slash at the end of `$SOURCE/` ensures that the source directory itself is not created within `$TARGET`, but only its contents. Subdirectories are created automatically if needed.

PR #6813, opened 2022-06-23
Add PSNProfiles

fletcherdodd:


Carlgo11:
Apart from the forum aspect of the site, which we [don't list](https://github.com/2factorauth/twofactorauth/blob/master/EXCLUSION.md#forums), I'm not sure it meets any of the [site critera](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria).

fletcherdodd:
> Apart from the forum aspect of the site, which we [don't list](https://github.com/2factorauth/twofactorauth/blob/master/EXCLUSION.md#forums), I'm not sure it meets any of the [site critera](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#site-criteria).

I would argue otherwise. PSNProfiles has a subscription service offered on the main website (not through the forums) which they would need to hold onto your payment information, therefore it falls under that money criteria. Up to you I guess, I don't mind either way.

Secondly, would that mean that sites like Hypixel and Mineplex should be removed from the entries as they are forum-type websites?

PSNProfiles Premium purchase page
PSNProfiles Screenshot

Carlgo11 (approved):
I'm satisfied with the arguments you provided for why PSNProfiles should be listed.

>would that mean that sites like Hypixel and Mineplex should be removed from the entries as they are forum-type websites?

Believe me, if I could remove those pesky Minecraft server-websites from our site I would! Sadly there seems to be some tie in with in-game currency which means it's not just a forum :roll_eyes:

PR #6825, opened 2022-06-26
Add MLP

phallobst:


Carlgo11:
Michael! :disappointed:
ssllabs report

Carlgo11:
What's Sm@rt-TAN and is it a relevant thing to also list?

phallobst:
The method is called "Sm@rt-TAN plus" (so fancy!), but you actually use a TAN generator (together with your bank card) - so that's the `custom-hardware` one.

Carlgo11 (approved):
My Little Pony Credit Union

PR #6827, opened 2022-06-29
Create etax.ch.json

whjvenyl:
Added authenticator setup for: etax.ch


phallobst:
Currently this website has a Similarweb rank of over 200,000. This PR will therefore be closed, but you are welcome to re-open it once the rank drops below that limit.

PR #6834, opened 2022-07-01
Add islogin.cz

ghost:
I am not sure whether using national eID qualifies for your definition of two-factor auth. I listed it as a custom software 2FA.


Carlgo11:
eID systems are allowed as custom-software or custom-hardware depending on how they work. We do not, however, list 2FA provider websites in our entries list. From what I understand islogin.cz is the website for the eID system and as such can't be listed.

You can list the individual websites using the eID system but they should be listed under separate entry files and without any subdomains (such as `is.`). Notes explaining which system is being used is also redundant as the custom-software element already explains that.

ghost:
> eID systems are allowed as custom-software or custom-hardware depending on how they work. We do not, however, list 2FA provider websites in our entries list. From what I understand islogin.cz is the website for the eID system and as such can't be listed.

It is not the eID system, it is an administrative information system for universities (for students and teachers) which uses eID as 2FA, in addition to it's username and password. It is used by several universities but the login page is always `islogin.cz`. The domains which I listed under `additional-domains` are entry pages into the system for each university.

> You can list the individual websites using the eID system but they should be listed under separate entry files and without any subdomains (such as `id.`). Notes explaining which system is being used is also redundant as the custom-software element already explains that.

I will remove the "notes" element then.

Carlgo11:
I've reopened the PR if you wish to update the PR so that it follows the guidelines.
The main issues as of right now are:
* The entry name is Masaryk University, which already exists. Please change this name to something more relevant for all affected domains.
* If the auth system is used for authentication at the entire university, the `is.` subdomain should be removed. If the system is only used for `is.<uni domain>.cz` and the university uses another authentication system (like Microsoft AD) then this point can be disregarded.
* The domain `islogin.cz` doesn't meet the ranking criteria of our contributing guidelines and thus can't be listed.
* The custom-software value of `national eID` seems vague. Is this the name of the system or should it be `eidentita.cz`?

ghost:
>     The entry name is Masaryk University, which already exists. Please change this name to something more relevant for all affected domains.
Fixed
>     If the auth system is used for authentication at the entire university, the `is.` subdomain should be removed. If the system is only used for `is.<uni domain>.cz` and the university uses another authentication system (like Microsoft AD) then this point can be disregarded.
The second point is true, so I am keeping the domain.
>     The domain `islogin.cz` doesn't meet the ranking criteria of our contributing guidelines and thus can't be listed.
I guess I cannot fix that. Country rank is 2k, global rank almost 250k. But the main domains for which it serves rank higher (muni.cz is 6k globally, slu.cz is 65k globally, vstecb.cz is 175k globally etc.)
>     The custom-software value of `national eID` seems vague. Is this the name of the system or should it be `eidentita.cz`?
Updated to [identitaobcana.cz](https://identitaobcana.cz/), which also works as a link to an information page.

phallobst:
This seems to me like a 2FA provider rather than a website we should list on its own. I mean, you want to log into your `muni.cz` account for example, and `islogin.cz` is just the way to do so.

This would probably mean to include `islogin.cz` as an `additional-domain` for each university it supports, in order to facilitate password managers - but since a student will probably have an account for one university only, I don't consider this to be a problem.

Carlgo11:
Closing due to inactivity.

PR #6838, opened 2022-07-03
Update Halifax

ApCoder123:
According to doc page, `custom-software` methods can only be used for approving purchases, not signing in.


phallobst:
According to [this page](https://www.halifax.co.uk/aboutonline/sign-in.html), they don't use 2FA at all (anymore), only "double 1FA". :frowning_face:

ApCoder123:
The documentation seems to be inconsistent. On the sign in page, there is still a reference to SMS/call with a link to the current doc page.
image

Carlgo11 (approved):

PR #6851, opened 2022-07-07
Add BitLaunch 

mrbluehollywood:
BItLaunch IO supports 2FA and was under 200k on Alexa before Alexa was retired. 

Is it possible to get it added?

Thanks!


Carlgo11:
As Alexa was retired we switched to SimilarWeb. Unfortunately BitLaunch doesn't meet the 200K ranking requirement on SimilarWeb and thus can't be listed.

PR #6833, opened 2022-06-30
Replace exception popups and add popovers for custom-* data

ApCoder123:
Replaces current exception popups with Bootstrap popovers. 
Removes `popup.scss` (43 KB) and `popup.js` (30.9 KB) and replaces `bootstrap.min.js` (59.2 KB) with `bootstrap.bundle.min.js` (78.1 KB) leading to an overall saving of 55 KB (file size).

image

PR #6857, opened 2022-07-11
Noscript Site

### LINKED ISSUE #2536: Unusable without javascript enabled
Hi:

Thanks for consolidating all this information and putting pressure on sites/services to use 2fA. I wanted to make one critique, however. The site is impossible to view with without enabling javascript because most links don't work. It would be great if there was a static version.

Thanks!

ApCoder123:
Closes #2536
This PR creates a separate page noscript.html on which the site can be used without JS. Most of the functionality other than the search and social actions have been replicated.

Key changes on noscript page:
- Category buttons are `<a>` elements to allow displaying the table via the `:target` pseudo-class (tables can still be accessed via their ids in the URL)
- Logos on `noscript.html` are `<div>`s with background image set as logo as this prevents the browser from loading images until the category table is displayed. This avoids the problem mentioned in the issue with large amounts of assets being served.
- ~~Exception popovers are displayed with `<details>` element.~~
- Custom-* data displayed in title attributes like they were before #6833
- Social buttons replaced with links to pages

image
image
image

### DIFF
--- a/_includes/html/category-button.html
+++ b/_includes/html/category-button.html
@@ -2,7 +2,11 @@
 {% assign title = include.category-param.title %}
 {% assign icon = include.category-param.icon %}
 <div class="col {{ include.viewport }}-only category-btn-outer">
+  {% if page.noscript %}
+  <a class="row {% unless include.viewport == 'mobile' %}box{% endunless %} category-btn" href="#{{ name }}">
+  {% else %}
   <button class="row {% unless include.viewport == 'mobile' %}box{% endunless %} category-btn" id="{{ name }}" href="#{{ name }}" aria-controls="{{ name }}-table">
+  {% endif %}
     <div class="col"></div>
     <div class="col-12 align-self-center">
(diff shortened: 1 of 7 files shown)


phallobst:
For the exception popup, the `title` attribute is sufficient as well, I reckon. There is no real need to mix two different display variants.

The Facebook/Twitter links should probably open in a new window, i.e. `target="_blank"`.

Other than that, it looks good - perfectly usable for the non-JS folks.

phallobst (approved):

Carlgo11:
Found a bug.

The noscript message is being displayed on the noscript page. It redirects to `{region}/noscript/noscript`.

ApCoder123:
What browser are you using? I'm can't seem to be able to reproduce the bug

Carlgo11:
Firefox but I used uBlock Origin to disable JS. Disabling JS via about:config doesn't show the message :thinking:

PR #6867, opened 2022-07-19
Update servus.ca.json

asclepiadae:


asclepiadae:
I am not sure why the repository build test is failing. Literally copied the examples from documentation and rewrote the entry...

phallobst:
The JSON format itself is pretty strict regarding the syntax, and we have some additional content checks.
- Only one of `tfa` and `contact` may be present.
- The `documentation` tag was not separated by a comma. (I also moved it further up for consistency.)
- The `url` tag was not really needed, so I removed it as well.

phallobst:
I just had a look at https://www.servus.ca/security/protect-yourself/two-step-authentication:
> 4. Do I need to use two-step authentication to sign in to Servus online banking?
> No, you don't have to use two-step authentication to sign in to online banking today. However, it will eventually be required to better protect you when performing certain types of transactions.

Unfortunately, this is not _authentication_, but _authorization_ (regardless of what they call it). That means we will keep this entry as it is - maybe/hopefully they'll offer true two-factor _authentication_ in the future.

PR #2973, opened 2018-01-31
Add National Australia Bank (NAB)

Jawshy:
- Add: [National Australia Bank](https://www.nab.com.au/) (NAB) in [banking](https://github.com/2factorauth/twofactorauth/blob/master/_data/banking.yml). 
  - NAB [appears to have two-step authorisation via SMS](https://www.nab.com.au/help?return#/?m=c&id=23229), but not [2FA by 2factorauth's definition](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#a-note-on-definitions).

PR #6871, opened 2022-07-22
Update Bendigo Bank

willidb:


phallobst:
According to the original PR #4718, the "security token" is only used for _authorization_, but not _authentication_. Has anything changed in this regard?

willidb:
The TOTP provided by the Security Token (Symantec VIP) may be configured by the user to be required at login in addition to User ID and Password. The Security Token is also used for step-up authentication commensurate to sensitivity of transaction. However having reviewed the PR https://github.com/2factorauth/twofactorauth/pull/2973 I now understand that the 2factorauth project defines _step-up authentication_ as a form of _authorisation_. Hopefully this definition will be amended. Irrespective of the _step-up authentication_ definition, does configurable (optional) enforcement of the Security Token at login meet guidelines?

Carlgo11 (changes requested):
>[...] does configurable (optional) enforcement of the Security Token at login meet guidelines?

Yes if it is required at the login stage it would qualify. Just make sure to link to documentation describing how to set it up and add a 'note' briefly explaining what actions need to be taken for 2FA to be used instead of 2SV.

You've added 'totp' as an option. Is it actual RFC-6238 TOTP authentication or does it require certain apps? If only certain apps are supported, change the 'totp' option to 'custom-software' and make a custom-software array containing the name(s) of the supported app(s). Read the 'custom-hardware'/'custom-software section' of [CONTRIBUTING.md#other-properties](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#other-properties) for more information on how to specify custom-software.
A Semantic VIP hardware key is also mentioned on the documentation page. Please add a 'custom-hardware' property accordingly.


Lastly, the build is failing because the contact element still is present. Please remove that section from the entry.

Carlgo11:
Unless there's activity within the next week I'll close the PR.

Carlgo11:
Closing due to inactivity.

PR #6872, opened 2022-07-22
Update nab.com.au.json

willidb:


phallobst:
According to the original PR #2973, the "SMS Security" is only used for _authorization_, but not _authentication_. Has anything changed in this regard?

willidb:
"SMS Security" is used by nab for _step-up authentication_ commensurate to sensitivity of transaction. However having reviewed the PR https://github.com/2factorauth/twofactorauth/pull/2973 I now understand that the 2factorauth project defines _step-up authentication_ as a form of _authorisation_. Hopefully this definition will be amended, but until it is I understand that this pull request should be declined. Apologies.

Carlgo11:
@willidb It's not just _our definition_ but part of the foundation to information security. The Wikipedia article on [Multi Factor Authentication](https://en.wikipedia.org/wiki/Multi-factor_authentication) states:
>Authentication takes place when someone tries to [log into](https://en.wikipedia.org/wiki/Log_into) a computer resource (such as a network, device, or application).

We won't amend our definition to include post-authentication user verification as the definition for authentication can't possibly include actions taken after said process.

PR #6879, opened 2022-07-27
Update ESET

Aldaviva:
Hello,

ESET HOME (previously known as My ESET) is an online management portal for users' antivirus licenses and devices. It used to only support username and password authentication, but at some point they added support for generic TOTP.

They also offer their own TOTP mobile app called ESET Secure Authentication, but it's not mandatory to use that, you can just use Google Authenticator or whatever you like with the provided QR or numeric codes.

- 2FA can be enabled from the [My Account](https://home.eset.com/account) page of ESET HOME.
- [Documentation for 2FA](https://help.eset.com/home_eset/en-US/?two_factor_authentication.html)
- [Documentation for 2FA with ESET Secure Authentication mobile app](https://support.eset.com/en/kb8269-set-up-two-factor-authentication-in-eset-home)

My Account screenshot

### DIFF
--- a/entries/e/eset.com.json
+++ b/entries/e/eset.com.json
@@ -2,10 +2,10 @@
   "ESET": {
     "domain": "eset.com",
     "url": "https://www.eset.com",
-    "contact": {
-      "facebook": "eset",
-      "twitter": "ESET"
-    },
+    "tfa": [
+      "totp"
+    ],
+    "documentation": "https://help.eset.com/home_eset/en-US/two_factor_authentication.html",
     "keywords": [
       "security"
     ]
--- a/entries/e/eset.com.json
+++ b/entries/e/eset.com.json
@@ -1,7 +1,6 @@
 {
-  "ESET": {
+  "ESET HOME": {
     "domain": "eset.com",
-    "url": "https://www.eset.com",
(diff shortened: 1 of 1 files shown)


entries/e/eset.com.json
```diff
@@ -2,12 +2,12 @@
   "ESET": {
     "domain": "eset.com",
     "url": "https://www.eset.com",
```
Carlgo11: If you go to `https://<domain>` and it redirects to the value of the `url`-element then the `url`-element isn't needed.
```suggestion
```

entries/e/eset.com.json
```diff
@@ -2,12 +2,12 @@
   "ESET": {
```
Carlgo11:
```suggestion
  "ESET HOME": {
```

Carlgo11 (changes requested):

entries/e/eset.com.json
Aldaviva: Makes sense! I was wondering why it was like that in the first place.

Carlgo11 (approved):
Thanks for the PR!

Aldaviva:
Thanks!

PR #6882, opened 2022-07-30
Update Zapier

fletcherdodd:
As per logo change announced here: https://zapier.com/blog/zapiers-new-look/

Updating logo to their new favicon as it presented the best.


Carlgo11 (approved):
Thanks for the pull request @fletcherdodd! Your version looked good but I saw a chance of reducing the file size even further by replacing the path with a rectangle object. Sorry for hijacking your PR :sweat_smile:

PR #6889, opened 2022-08-04
OCI domains updated

likhner:
The current domain (`oracle.com`) is used by Oracle accounts (separate from OCI accounts) that do not support 2FA.

Authorization and 2FA setup for Oracle Cloud Infrastructure occurs on separate domains (`cloud.oracle.com`, `oraclecloud.com`).


entries/c/cloud.oracle.com.json
```diff
@@ -1,6 +1,9 @@
 {
   "Oracle Cloud Infrastructure": {
-    "domain": "oracle.com",
+    "domain": "cloud.oracle.com",
+    "additional-domains": [
+      "oraclecloud.com"
+    ],
```
Carlgo11, line 6: Since the entry can't redirect to a login page I think redirecting to this page would be the best fit:
```suggestion
    ],
    "url": "https://www.oracle.com/cloud/",
```

Carlgo11 (changes requested):

Carlgo11 (approved):

PR #6892, opened 2022-08-06
Add University of Cambridge

christopherlam888:


phallobst:
Although we prefer SVG logos, this one is quite chunky (57K). Even with some tweaks I was not able to get it under 20K. I'd suggest to use a PNG logo in this case:
cam ac uk

entries/c/cam.ac.uk.json
```diff
@@ -0,0 +1,27 @@
...
+    "tfa": [
+      "totp",
+      "sms",
+      "custom-software",
+      "custom-hardware",
+      "call"
+    ],
+    "custom-software": [
+      "Microsoft Authenticator",
+      "Authy"
```
Carlgo11: While Authy is mentioned, it's only used in the context of a TOTP RFC-6238 application and no methods specific to Authy are supported.
```suggestion
      "Microsoft Authenticator"
```

Carlgo11 (changes requested):

Carlgo11 (approved):

PR #6900, opened 2022-08-11
Add Prepr.io

DavydeVries:


DavydeVries:
Apparently, the SimilarWeb global ranking is 1,289,569 and the local ranking 86,745. So Prepr is probably too small to add?

I guess this should be clarified in the contribution guide.

Carlgo11:
Indeed it is

DavydeVries:
Reopening this PR with some additional context.

Prepr currently doesn’t meet the 200k SimilarWeb threshold (rank ~1.1M), but we believe it fits the documented Site Criteria. We handle PII and sensitive content for B2B clients. Our CMS powers personalization and content pipelines, making 2FA support critical.

Would this still qualify for inclusion based on the criteria?

Carlgo11:
Oh a blast from the past :sweat_smile: 
Hi again Davy! Unfortunately, the site would have to meet the site criteria AND the ranking limit.
We do have an exception for apps without a major web preference, but in those circumstances we look at social media following (as that's the best publicly available metric we have :shrug:) and I'm afraid it doesn't look like Prepr would qualify there either 🙁

Your product looks great, though! I'll even check with my company if we could switch to Prepr :thinking: 

P.S. it looks like someone forgot to complete the "Everything in..." sentence for your Flex price plan
image

PR #6903, opened 2022-08-11
Add Teleport

benarent:
This PR add Teleport. We've also recently added Passwordless, but happy to add it to the list.


Carlgo11:
What an odd error. Missing origin remote 🤔
I'll set the overriding variable and rerun the build.

Edit: not even that worked

phallobst:
Your branch has an usual name: `benarent:benarent/teleport`. Was that intentional? Normally, it would be `user:foobar` instead of `user:user/foobar`. Maybe gh-pages is confused by this. Could you try to create a branch `benarent:teleport` and open another PR from there?

benarent:
Closing in favor or https://github.com/2factorauth/twofactorauth/pull/6904 .

The branch name thing is just another practice we do at work.. good to know that it doesn't play nice with GH branches.

PR #6907, opened 2022-08-13
Update Twitter Card

Carlgo11:
The old Twitter Card URL no longer exists.


phallobst:
How about adding it to `img/icons/` and linking to the repository file inside the config?

phallobst (approved):

PR #6912, opened 2022-08-17
Exclude entries dir from Jekyll build

Carlgo11:
Excluding `entries/` as there's no reason for the it to be published on the website.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -18,4 +18,5 @@ exclude:
   - vendor/
   - scripts/
   - tests/
+  - entries/
 theme: null


phallobst:
Ignoring the `entries/` directory basically disables the `jekyll serve --watch` feature, which is quite handy for development. I use a local config file anyway, so it doesn't really affect me, but it should ne noted nonetheless.

Carlgo11:
It shouldn't interfere with `jekyll serve --watch` as Jekyll doesn't read directly from `entries/` but from `_data/all.json` which is generated by join-entries.rb.

phallobst:
Well, color me bamboozled... But it does make sense, once you can be bothered to think about it. :smile:

phallobst (approved):

PR #6988, opened 2022-10-02
Add back to top button

### LINKED ISSUE #14: Should there be a back-to-top button in the UI?
- [ ] I have searched [open issues and pull requests](https://github.com/2factorauth/twofactorauth/issues?q=is%3Aopen). The issue I'm creating is not a duplicate of an existing open issue or pull request.

Should there be a back-to-top button functionality at the bottom right corner to prevent too much scrolling? 
Just an idea to enhance user experience.

### LINKED ISSUE #6120: Should there be a back-to-top button in the UI?
- [ ] I have searched [open issues and pull requests](https://github.com/2factorauth/twofactorauth/issues?q=is%3Aopen). The issue I'm creating is not a duplicate of an existing open issue or pull request.

Should there be a back-to-top button functionality at the bottom right corner to prevent too much scrolling? 
Just an idea to enhance user experince.

sn00py1310:
Closes 2factorauth/frontend#14

Moved out of the window by default, only visible with focus and through scrolling.
Only for mobile because of mobile-only.
~~Without JavaScript `visibility: hidden;` used.~~
Preview


js/main.js
```diff
@@ -71,3 +71,12 @@ const customSoftwarePopoverList = [...document.querySelectorAll('.custom-softwar
   title: 'Custom Software 2FA'
 }));
 
+const backToTopElement = $('.back-to-top-link');
+backToTopElement.removeClass('invisible');
+$(window).scroll(() => {
```
Carlgo11, line 75: I'm not very good at JS but instinctively feel like a function like this should be asynchronous. Would there be any reason for it not to?

sn00py1310: I'm also not good at JS, so I tested it with async and it did the same job.

Carlgo11: Great to hear I'm not the only person who isn't a JS genius :sweat_smile:
@ApCoder123 if I remember correctly you know a bit of JS. Mind reviewing this PR?

ApCoder123: I'm, not sure that adding the `async` keyword will do anything to this function. `await` isn't used inside this function, so using `async` will just cause this function to wrap the return value in a promise. I think that even with `async`, this code will still run synchronously. See https://stackoverflow.com/a/51201849

Also, IMHO it's not worth hiding the back to top button for users without JS as the site is unusable without it anyway and we have the separate noscript page for non-JS users. Other than that, looks good 👍!

sn00py1310: @ApCoder123 thanks for helping with the JS. I applied your suggestions.
Little side notice around the `async` topic. Could you please check these lines/functions?

- [main.js#20](https://github.com/2factorauth/twofactorauth/blob/ba52fb5b52450dd0de376b20c6ee3362ce9bbb98/js/main.js#L20)
- [main.js#31](https://github.com/2factorauth/twofactorauth/blob/ba52fb5b52450dd0de376b20c6ee3362ce9bbb98/js/main.js#L31)

ApCoder123: Yeah, I'm pretty sure that those uses of `async` will have no effect on the execution of the function

PR #6991, opened 2022-10-03
Add Smyths Toys

windmueller:
Smyths Toys is running hundreds of shops for toys and entertainment products. They also offer shipping when using their website. Sadly, the order history and personal data can not be secured with 2FA.

Because of legal issues in my country, I am not allowed to add the logo myself. The simplest way in my opinion would be to use the [favicon](https://www.smythstoys.com/favicon.ico).


Carlgo11 (commented):
I'm not sure this meets any site criteria. Is there any sensitive data behind the login portal?

windmueller:
When logged in, I can see

- Personal data like my address
- Stored alternative shipping addresses
- Payment methods
- Order history

In my opinion this is no different than Walmart or Zalando.

fducasse:
hello what's the payment Module of zalando please

Carlgo11:
@fducasse I don't think you're in the right place here. This PR isn't related to Zalando.

Carlgo11 (changes requested):
From reading their shipping & delivery page it looks like their main audience is GB,IE,DE,CH,AT. A regions array would be good here.

Carlgo11 (approved):
Thank you! 🙂

PR #6994, opened 2022-10-05
Adding Service Canada (MSCA) Login

Photonicsguy:
Service Canada (MSCA) supports totp authentication.
MSCA is visually similiar to CRA Login (in canada.ca.json), however, the accounts are different.

### DIFF
--- a/entries/s/servicecanada.gc.ca.json
+++ b/entries/s/servicecanada.gc.ca.json
@@ -1,0 +1,22 @@
+{
+  "Service Canada": {
+    "domain": "servicecanada.gc.ca",
+    "additional-domains": [
+      "canada.ca"
+    ],
+    "url": "https://www.canada.ca/en/employment-social-development.html",
+    "img": "canada.ca.svg",
+    "tfa": [
+      "sms",
+      "call",
+      "totp"
+    ],
+    "documentation": "https://www.canada.ca/en/employment-social-development/services/my-account/multi-factor-authentication.html",
+    "keywords": [
+      "government"
+    ],
+    "regions": [
+      "ca"
+    ]
+  }
+}


entries/s/servicecanada.gc.ca.json
```diff
@@ -0,0 +1,22 @@
+{
+  "Service Canada": {
+    "domain": "servicecanada.gc.ca",
+    "additional-domains": [
+      "canada.ca"
+    ],
+    "url": "https://www.canada.ca/en/employment-social-development.html",
```
Carlgo11, line 7: What's the actual login URL? I suggest you use that domain name instead of `servicecanada.gc.ca` and `canada.ca` as if you want third-party programs (i.e. password managers) to know 2FA is available on the site, the best way to do that is to use the actual login page as the `domain`-element or `additional-domains`-element. The `url`-element can then be used to direct people to the main page of the service, such as https://www.canada.ca/en/employment-social-development.html.

Photonicsguy: The actual login process starts at: https://www.canada.ca/en/employment-social-development/services/my-account.html
Though after logging authenticating via GCKey or a Bank, Service Canada still requires a separate two-factor authentication.
 
There are many different government services with different accounts, and different levels of security all through canada.ca. It's complicated, and CRA for example, (canada.ca.json) doesn't allow GCKey, nor do they support totp yet.

Carlgo11: Canada.ca is a tricky website to list as they don't follow the internet standard of one FQDN per service.
If 2FA is mandatory, the best option may be to remove the interfering `canada.ca` from additional-domains and just point to `servicecanada.gc.ca`.

Carlgo11:
:warning: Unless there's activity within the next 7 days I'll close the PR.

Carlgo11:
Closing due to inactivity.
> **Note**

PR #6998, opened 2022-10-16
Update paysafecard 2FA options

Jackenmen:
Standard TOTP is not an option anymore, instead, SMS and their custom app are now always enabled 2FA options.

### DIFF
--- a/entries/p/paysafecard.com.json
+++ b/entries/p/paysafecard.com.json
@@ -9,7 +9,7 @@
     "custom-software": [
       "paysafecard app"
     ],
-    "documentation": "https://www.paysafecard.com/en/help/2-faktor-authentifizierung/",
+    "recovery": "https://www.paysafecard.com/en/help/2-faktor-authentifizierung/",
     "keywords": [
       "payments"
     ],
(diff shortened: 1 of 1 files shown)


entries/p/paysafecard.com.json
```diff
@@ -3,9 +3,13 @@
...
     "tfa": [
-      "totp"
+      "sms",
+      "custom-software"
     ],
-    "documentation": "https://www.paysafecard.com/en-au/lp-products/2-step-login/",
+    "custom-software": [
+      "paysafecard app"
+    ],
+    "documentation": "https://www.paysafecard.com/en/help/2-faktor-authentifizierung/",
```
Carlgo11: As you mention 2FA is mandatory and the page you linked to only covers recovery I think the `recovery`-element is a better fit.
```suggestion
    "recovery": "https://www.paysafecard.com/en/help/2-faktor-authentifizierung/",
```

Carlgo11 (changes requested):

Carlgo11 (approved):

PR #7007, opened 2022-10-21
Update Simplii

sarahmrrp:
Added support for 2FA

### DIFF
--- a/entries/s/simplii.com.json
+++ b/entries/s/simplii.com.json
@@ -2,15 +2,20 @@
   "Simplii Financial": {
     "domain": "simplii.com",
     "url": "https://www.simplii.com",
-    "contact": {
-      "facebook": "SimpliiFinancial",
-      "twitter": "SimpliiFin"
-    },
-    "regions": [
-      "ca"
+    "tfa": [
+      "sms",
+      "call",
+      "custom-software"
     ],
+    "custom-software": [
+      "Simplii App Push Notification"
+    ],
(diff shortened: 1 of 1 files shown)


Carlgo11 (commented):
From reading [this page](https://simplii.intelliresponse.com/index.jsp?requestType=NormalRequest&source=3&id=2286&sessionId=eff4be9f-516e-11ed-bb3e-73b23b154f85&question=I+was+asked+to+give+a+one-time+verification+code.+What+is+this+and+why+is+it+required) it looks like these codes are only required for certain actions and not for logging in. Is this correct?

sarahmrrp:
> From reading [this page](https://simplii.intelliresponse.com/index.jsp?requestType=NormalRequest&source=3&id=2286&sessionId=eff4be9f-516e-11ed-bb3e-73b23b154f85&question=I+was+asked+to+give+a+one-time+verification+code.+What+is+this+and+why+is+it+required) it looks like these codes are only required for certain actions and not for logging in. Is this correct?

Unless I have my card saved in my browser, I get prompted for a 2FA code every time,

Carlgo11:
Unless you have your card saved? Could you clarify what you mean?

sarahmrrp:
Unless you select "Save bank card" in your browser (like a "remember me" features) it will ask you for 2FA when logging in.

Carlgo11 (approved):
Thanks for the clarification. Your contribution will be public in about 10 minutes.

PR #7016, opened 2022-10-29
Update OneDrive / Xbox / Outlook.com

DandelionSprout:
Microsoft accounts have had a "Use a security key" option for seemingly some time by now, so I figured I should add support for U2F to the Microsoft tools that I was most confident would use Microsoft accounts as-is.

Screenshot of the setting:
<details>

image

</details>

#### Notes:
• I lacked recent experience with Skype, Azure, Minecraft, and Microsoft To-Do, and thus I did not wager a guess that they supported Microsoft accounts with U2F as well. I don't think the listing for Office 365 corporate plans would use typical Microsoft accounts.
• Strangely, it doesn't appear to me that Microsoft accounts have native Windows Hello support (At least not when choosing "Use a security key"), meaning that e.g. the Kensington VeriMark is not supported by Microsoft accounts' security key support.

### DIFF
--- a/entries/o/onedrive.live.com.json
+++ b/entries/o/onedrive.live.com.json
@@ -6,7 +6,8 @@
       "call",
       "email",
       "custom-software",
-      "totp"
+      "totp",
+      "u2f"
     ],
     "custom-software": [
       "Microsoft Authenticator"
--- a/entries/o/outlook.com.json
+++ b/entries/o/outlook.com.json
@@ -10,7 +10,8 @@
       "call",
       "email",
       "custom-software",
-      "totp"
+      "totp",
+      "u2f"
     ],
     "custom-software": [
       "Microsoft Authenticator"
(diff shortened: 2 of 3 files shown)


Carlgo11:
Unless Microsoft have recently updated their authentication system, what they're is using isn't U2F/WebAuthn but FIDO2 "passwordless", a good authentication alternative but something which we don't label as multi-factor authentication. See [Passwordless](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#passwordless-authentication) in CONTRIBUTING.md for more information on the matter.

DandelionSprout:
I admit that I didn't expect any companies out there to only ask for a USB key without also asking for username and password, but it appears you're correct about it: https://support.microsoft.com/en-us/windows/sign-in-to-your-microsoft-account-with-windows-hello-or-a-security-key-800a8c01-6b61-49f5-0660-c2159bea4d84

PR #7017, opened 2022-10-30
Update PayPal note

person1234-ux:
I went to enable a u2f on paypal and they wouldn't let me enable it and then forced me to login using only sms. so I added some not so nice notes

### DIFF
--- a/entries/p/paypal.com.json
+++ b/entries/p/paypal.com.json
@@ -10,7 +10,7 @@
       "totp",
       "u2f"
     ],
-    "notes": "Some methods of 2FA are only available in a few countries. Contact PayPal for more information.",
+    "notes": "Some methods of 2FA are only available in a few countries. Contact PayPal for more information. (2fa isn't usefull for paypal as they default to single factor sms)",
     "documentation": "https://www.paypal.com/us/smarthelp/article/faq4057",
     "keywords": [
       "payments"


entries/p/paypal.com.json
```diff
@@ -10,7 +10,7 @@
       "totp",
       "u2f"
     ],
-    "notes": "Some methods of 2FA are only available in a few countries. Contact PayPal for more information.",
+    "notes": "Some methods of 2FA are only available in a few countries. Contact PayPal for more information. (2fa isn't usefull for paypal as they default to single factor sms)",
```
Carlgo11, line 13: Looks like TOTP is also supported as a backup method.

```suggestion
    "notes": "Some methods of 2FA are only available in a few countries. Contact PayPal for more information. U2F/WebAuthn requires a backup authentication method",
```

Carlgo11 (changes requested):

entries/p/paypal.com.json
person1234-ux: @Carlgo11 I realized  I need to give more context paypal current state doesn't actually meet the requirements of 2fa as they allow you to login via sms as the only factor on any device even if it wasn't previously used. This was my experience yesterday that lead me to closing my account

Carlgo11: @person1234-ux I tried to reproduce what you described but the result differs from yours.

If I enter my correct password I get the following options:
Screenshot 2022-10-31 at 00-00-09 PayPal

If I click on _forgot password_ or _Having trouble logging in?_ I indeed get the option of SMS authentication.
If we're to assume that factor gets compromised (due to SIM-swapping) I still get the following requirement after successfully entering the SMS code:
Screenshot 2022-10-31 at 00-07-17 PayPal

So by that process the authentication factors are the same but the methods are weaker:
|Circumstance|Knowledge method|Possession method|
|-|-|-|
|Normal Authentication|Password|U2F Key|
|Password Reset|Security Questions|SMS code|

Of course SMS + Questions are far from ideal but technically 2FA is still enforced.

Carlgo11:
Closing due to inactivity.
> **Note**

PR #7031, opened 2022-11-02
Add note for proton.me

BurnerWah:
proton.me requires enabling TOTP in order to use a security key, [as stated on their help page](https://proton.me/support/two-factor-authentication-2fa):
> Note that to use a security key, you must first set up 2FA using an authenticator app, as described below.

This is also restated on the [help page for setting up security keys](https://proton.me/support/2fa-security-key).

### DIFF
--- a/entries/p/proton.me.json
+++ b/entries/p/proton.me.json
@@ -11,7 +11,7 @@
       "u2f"
     ],
     "documentation": "https://proton.me/support/two-factor-authentication-2fa",
-    "notes": "An authenticator app must be enabled to use a hardware token.",
+    "notes": "TOTP must be enabled to use a hardware token.",
     "keywords": [
       "email",
       "backup",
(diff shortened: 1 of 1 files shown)


entries/p/proton.me.json
```diff
@@ -11,6 +11,7 @@
       "u2f"
     ],
     "documentation": "https://proton.me/support/two-factor-authentication-2fa",
+    "notes": "An authenticator app must be enabled to use a hardware token.",
```
Carlgo11: `TOTP` is probably better than `An authenticatior app` since there's many different ways to generate TOTP codes.
```suggestion
    "notes": "TOTP must be enabled to use a hardware token.",
```

Carlgo11 (changes requested):

entries/p/proton.me.json
BurnerWah: i was just staying close to their original wording, but that makes more sense.

Carlgo11 (approved):

PR #7034, opened 2022-11-07
Add Khan Academy

hkamran80:


hkamran80:
The Facebook handles check is failing, even though the [handle exists](https://www.facebook.com/khanacademy). I think this may be due to the check visiting `https://www.facebook.com/pg/khanacademy` when the URL is `https://www.facebook.com/khanacademy`.

entries/k/khanacademy.org.json
```diff
@@ -0,0 +1,10 @@
+{
+  "Khan Academy": {
+    "domain": "khanacademy.org",
+    "contact": {
+      "twitter": "khanacademy",
+      "facebook": "khanacademy"
```
Carlgo11: Something weird is going on with that Facebook page as seems behave more like a Facebook profile or fan page rather than a normal page.

Take CIBC as an example:
CIBC
Here the page we can see the handle and category. There's also an option to contact CIBC through Messanger.

Khan Academy
On KA's page the handle and category is replaced with a followers/following count and there's no way to contact the org.

Due to the inconsistencies and since there doesn't seem to be any way to contact KA through the page, I recommend you remove the listing.
```suggestion
```

Carlgo11 (changes requested):

entries/k/khanacademy.org.json
hkamran80: Thanks!

Carlgo11 (approved):

PR #7037, opened 2022-11-10
Add HSBC CA

jsoref:


Carlgo11 (commented):
What's described sounds more like authorization for certain user actions than authentication at the login phase.
We distinguish between the two in the section about authorization in [Contributing.md](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#authorization).
Please confirm if this indeed is the case or if 2FA is also prompted for on login.

jsoref:
Sorry, I'm not really in a position to do so. I'm happy to remove it. The clueless CSR I spoke w/ this evening while auditioning their services seemed to believe it counted, and the page seems to believe it counts.

My naive understanding is that both the App and the custom hardware token generate a response (quite possibly according to TOTP, but also possibly not -- the docs clearly indicate both generators function offline) that one then has to enter into a website in order to authorize certain transactions. In the case of the app, you apparently have to unlock the number generator.

They have a FAQ. Amusingly, it looks like you can turn off the password feature and switch to 1FA using only your token generator:
https://www.hsbc.ca/support/faq/security-device/

I'm unlikely to select them based on these offerings...

---

Btw, the bot has `in in` in its output which feels buggy (I'm more likely to fix this than to be able to understand the HSBC's content):
>  'Oneof' error in in entries/h/hsbc.ca.json

hkamran80:
@jsoref I think it's because you have a `contact` object and a `tfa` object. Try removing the `contact` object.

jsoref:
@hkamran80 I'm calling out the Typo in the error message itself. I'll apply your suggestion for the file.

hkamran80:
@jsoref Ah, gotcha.

Carlgo11 (changes requested):
For us to approve the PR we need some proof that 2FA is required on login and not just for certain actions such as make transactions. The proof can be in form of a link to further documentation or a screenshot of the login process.

If you're unable to provide such proof I'm afraid we can't merge the PR.

jsoref:
Would you take one that went with the contact section instead?

Carlgo11:
Sure

jsoref:
If the requirement is for actual login, I _suspect_ that they in fact do it wrong -- at least based on my conversation w/ the CSR, it felt like you could half log in and get a semi-readonly view.

It wouldn't shock me if some of the other HSBC instances are the same.

Carlgo11 (approved):

PR #7040, opened 2022-11-10
Readme: Fix command to build with Docker on Windows

sydeslyde:
The command from the README.md for building with Docker does not work on Windows.

`$(pwd)` is not valid. Neither in CMD nor in PowerShell.

This PR adds a separate listing and command for PowerShell with `${pwd}` that works (braces instead of parentheses).

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,10 +71,16 @@ npm i babel-minify
 
 ## Building :running:
 
-Docker (Windows/Linux/MacOS):
+Docker (Linux/MacOS):
 
 ```BASH
 docker run -p 4000:4000 -v $(pwd):/twofactorauth 2factorauth/twofactorauth
+```
+
+Docker (Windows PowerShell):
+
+```powershell
+docker run -p 4000:4000 -v ${pwd}:/twofactorauth 2factorauth/twofactorauth
 ```
 
 Snap/Manual:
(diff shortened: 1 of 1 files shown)


Carlgo11 (commented):
`$(pwd)` could just be replaced with `./` for it to work on all platforms.

Carlgo11:
:warning: Unless there's activity within the next 7 days I'll close the PR.

sydeslyde:
> `$(pwd)` could just be replaced with `./` for it to work on all platforms.

`docker run -p 4000:4000 -v ./:/twofactorauth 2factorauth/twofactorauth`

I tried this on Windows Powershell and CMD, both resulting in the following error:

```
docker: Error response from daemon: create .: volume name is too short, names should be at least two alphanumeric characters.
See 'docker run --help'.
```

Carlgo11:
Ah okay.
I guess `${PWD}` works in PS?
If so, replace `$(pwd)` with `${PWD}` as it should be available cross platform.

Although the entire ordeal might be a bit of a waste due to #7091 removing that part of the README completely 🤷‍♂️

sydeslyde:
Yes for Windows PowerShell `${pwd}` works.
I can't test `${pwd}` on Linux and Mac.
I merged the commands to use `${pwd}` for all platforms.

Carlgo11 (approved):

PR #7041, opened 2022-11-10
Scroll to category when it is opened (fixes table outside of viewport)

sydeslyde:
When a category with many websites is open, and you click a different category with less websites the already open one, the table is collapsed and the new table is shown, but outside of the viewport.

Steps to reproduce:

- open category “Developer”
- scroll down and click on category “Food”
- → table with websites is outside of viewport

This PR implements a fix, by listening to the Bootstrap collapse components shown-event, and scrolling to the opened category when fired. (jQuery `one()` to listen to the event only once)

As far as I can tell it works fine, but please check if my idea using the bootstrap `shown.bs.collapse` event is correct.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -37,7 +37,9 @@ $('#region-notice-close-btn').click(async () => {
 async function showCategory(category) {
   $(`.category-table.collapse:not(#${category}-table, #${category}-mobile-table)`).collapse('hide');
   $(`.category-btn:not([id=${category}])`).removeClass('active');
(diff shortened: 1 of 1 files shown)


Carlgo11:
The current design isn't really being maintained as [version 5.0](https://github.com/2factorauth/frontend) is close to done. It would probably make more sense merging a fix like this over at that repo.

Carlgo11 (approved):

PR #7074, opened 2022-11-16
Add Hackster.io

hkamran80:


entries/h/hackster.io.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Hackster.io": {
+    "domain": "hackster.io",
+    "contact": {
+      "twitter": "Hacksterio"
+    },
+    "keywords": [
+        "education"
```
Carlgo11: I don't think the education category is a very good fit but neither is the creativity or social category. Maybe it's better to put it in the _other_ category until a more fitting one is created.
```suggestion
        "other"
```

Carlgo11 (changes requested):

Carlgo11 (approved):

PR #7080, opened 2022-11-17
Update Action Network

hkamran80:

PR #7082, opened 2022-11-17
Update Action Network's 2FA options

hkamran80:


Carlgo11 (approved):
Thanks!
I checked what you said in #7080 about the new logo not showing up. Looks like it's probably just your browser's cached version. We use long cache TTLs for images to speed up loading but clearing your cache should make the correct logo show up 🙂

PR #7086, opened 2022-11-20
Add Mastodon

Carlgo11:
## Eligibility
Mastodon, was removed by me in #3446 as it matched the [self-hosted services](https://github.com/2factorauth/twofactorauth/blob/master/EXCLUSION.md#self-hosted-services) and did not meet the inclusion criteria for self-hosted services:
> The category "self-hosted services/sites" includes all types of software that's designed to be hosted by the end user. These are excluded from listing unless all these points are met:
>
>    1. All [Site Criteria][site_criteria] and [Guidelines][guidelines] are met
>
>    2. Public interest and public accessibility are given  
>
>    3. The site has an independent authentication database
>
>    4. The underlying service supports built-in 2FA or through a first-party (developer) plugin

Since then Mastodon have risen in popularity, most notably due to the recent Twitter migration, to the level that I'd argue listing Mastodon now would be of public interest. As such I think all the points outlined above are met and thus propose Mastodon should be re-listed in the _socials_ category.

## Instances 
Due to the nature of Mastodon being a self-hosted service, listing every single domain hosting a Mastodon instance is unfeasible. I therefore propose only the instances officially hosted by Mastodon gGmbH should be listed.


kmpoppe (approved):

timonus:
Mastodon also supports security keys for 2FA

Carlgo11:
@timonus I don't see an option for that myself but if what you say is indeed true then please create a new issue with some sort of proof 🙂

PR #7096, opened 2022-11-24
Add Payfit

jychp:


entries/p/payfit.com.json
```diff
@@ -0,0 +1,18 @@
...
+    "documentation": "https://support.payfit.com/en/two-factor-authentication-for-administrators",
+    "tfa": [
+      "sms",
+      "email"
+    ],
+    "keywords": [
+      "finance"
+    ],
+    "regions": [
+      "eu"
```
Carlgo11: Looks like only the following countries are supported at the moment:
```suggestion
      "de",
      "es",
      "fr",
      "gb"
```

entries/p/payfit.com.json
```diff
@@ -0,0 +1,18 @@
+{
+  "Payfit": {
+    "domain": "payfit.com",
+    "url": "https://www.payfit.com",
```
Carlgo11: If you go to `https://<domain>` and it redirects to the value of the `url`-element then the `url`-element isn't needed.
```suggestion
```

entries/p/payfit.com.json
```diff
@@ -0,0 +1,18 @@
+{
+  "Payfit": {
```
Carlgo11:
```suggestion
  "PayFit": {
```

Carlgo11 (changes requested):
We prefer SVG icons as they scale better between different devices.
Here's an SVG version of the logo you provided:
payfit logo

Carlgo11 (approved):

Carlgo11:
Thank you! 🙂

PR #7098, opened 2022-11-27
Update UniSuper

costaparas:
Since June 2021, UniSuper supports 2FA via SMS or Email.

### DIFF
--- a/entries/u/unisuper.com.au.json
+++ b/entries/u/unisuper.com.au.json
@@ -2,9 +2,11 @@
   "UniSuper": {
     "domain": "unisuper.com.au",
     "url": "https://www.unisuper.com.au/",
-    "contact": {
-      "twitter": "UniSuperNews"
-    },
+    "tfa": [
+      "sms",
+      "email"
+    ],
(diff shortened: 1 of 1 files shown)


entries/u/unisuper.com.au.json
```diff
@@ -2,14 +2,16 @@
...
     "domain": "unisuper.com.au",
     "url": "https://www.unisuper.com.au/",
-    "contact": {
-      "twitter": "UniSuperNews"
-    },
+    "tfa": [
+      "sms",
+      "email"
+    ],
+    "documentation": "https://www.unisuper.com.au/contact-us/accessing-online-accounts",
```
Carlgo11: Since the document doesn't mention 2FA set up but instead recovering from different types of authentication issues, I suggest you use the recovery attribute instead:
```suggestion
    "recovery": "https://www.unisuper.com.au/contact-us/accessing-online-accounts",
```

Carlgo11 (changes requested):

Carlgo11 (approved):

PR #7099, opened 2022-12-01
Make dark scroll bar in dark theme for Chromium-based browsers

rakleed:
Make dark scroll bar in dark theme for Chromium-based browsers (show scrollbar must be enabled "Always" in system settings).

### DIFF
--- a/_includes/scss/search.scss
+++ b/_includes/scss/search.scss
@@ -20,6 +20,11 @@
     height: 100%;
     float: right;
     outline: none;
+    @media (prefers-color-scheme: dark) {
+      background-color: #fff;
+      color: #000;
+      caret-color: #000;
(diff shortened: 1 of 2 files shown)


Carlgo11:
As we're in the process of moving over to a [new website design](https://github.com/2factorauth/frontend), PRs concerning HTML, CSS or JavaScript are better directed at that repository.
Merging it here will just result in your work being lost in a month or so when the move is complete.

rakleed:
@Carlgo11 done: https://github.com/2factorauth/frontend/pull/1

If everything is in order, it would be nice if you accepted this PR as well, so that there is a dark scroll bar while the old design is used.

Carlgo11 (approved):

PR #7118, opened 2022-12-15
Add WebMoney

rakleed:
https://en.wikipedia.org/wiki/WebMoney


entries/w/webmoney.ru.json
```diff
@@ -0,0 +1,19 @@
...
+  "WebMoney": {
+    "domain": "webmoney.ru",
+    "tfa": [
+      "sms",
+      "custom-software"
+    ],
+	  "custom-software": [
+	    "E-num service"
+	  ],
+    "documentation": "https://wiki.wmtransfer.com/projects/webmoney/wiki/Login_with_additional_confirmation",
```
Carlgo11: Whilst it does say the list is incomplete, it does look more like authorization than authentication to me.

rakleed: I can’t check it myself, but support said that in any case, you will need a second factor to enter the system. As for the list, these are not the main services of the system, and in any case, a second factor will be needed to access the wallet.

rakleed: @Carlgo11 please take a look

Carlgo11: I can't confirm your PR without any proof that 2FA is required on login and not just to perform certain actions.

rakleed: That is, the answer from support is not enough? Do you need personal confirmation from the user? I just can't do it.

Carlgo11: As previously mentioned, I need proof that it's actually two factor _authentication_ and not _authorization_. That can be a quote from a support chat, describing how it is required __on login and enforced on the entire account__, or screenshots of the setup/login process.
If you're unable to provide any of this, I suggest we maybe wait with adding WebMoney until an actual user of the service requests it to be added.

rakleed: Looks like you were right: 
image

Then should I change the status ​​to the fact that the service does not support 2FA?

Carlgo11: Yes please.

Also, it looks like your CircleCI token might be outdated. Could you log in to [CircleCI](https://circleci.com/vcs-authorize/) with your GitHub account so that your PRs are actually processed? :wink:

rakleed: And add social networks for English-speaking or Russian-speaking users?

Carlgo11: If WebMoney is only available for Russia (and its neighboring Russian-speaking countries) use the Russian social media details, otherwise English please.

rakleed: The service has the most users from Russia, but it also works in other countries, so it added social networks for English-speaking users.

And I can't log in to CircleCI. When I sign in, I am automatically redirected to this page, even if I use a VPN: https://support.circleci.com/hc/en-us/articles/360043679453-CircleCI-Terms-of-Service-Violation-Sanctioned-Country

Carlgo11: Oh wow that's annoying :man_facepalming: 
Well at least it shows as 'skipped' and not failing anymore. Thanks for trying.

Carlgo11 (approved):

PR #7119, opened 2022-12-15
Add Qiwi

rakleed:
https://en.wikipedia.org/wiki/Qiwi

The site only has the ability to confirm the transfer of money through the code in SMS, but for the login there is no such possibility.


entries/q/qiwi.com.json
```diff
@@ -0,0 +1,15 @@
+{
+  "Qiwi": {
+    "domain": "qiwi.com",
+    "contact": {
+      "email": "bankinfo@qiwi.com",
+      "twitter": "QiwiRussia"
+    },
+    "keywords": [
+      "payments"
```
Carlgo11, line 9: The payment category is meant for services like Apple Pay, Visa, Mastercard, Stripe etc.
Is Qiwi a payment service or a bank?

rakleed: It is a payment system similar to Yoomoney, Payeer and Skrill.

Carlgo11 (approved):

PR #7120, opened 2022-12-15
Add Russia to regions

rakleed:
After merge PR #7118 and #7119


Carlgo11:
Russia is already a listed region on https://beta.2fa.directory/ and if the migration goes as planned that design will replace the current one within a week.

PR #7121, opened 2022-12-16
Add Meta

hkamran80:
Meta accounts have the same 2FA as Oculus.

2FA on Meta.com


entries/m/meta.com.json
```diff
@@ -0,0 +1,12 @@
+{
+    "Meta": {
+      "domain": "meta.com",
+      "tfa": [
+        "sms",
+        "totp"
+      ],
+      "keywords": [
+        "gaming"
```
Carlgo11: So if I understand Meta's structure correctly, Meta accounts are used for Oculus devices alongside Instagram & Facebook accounts. Is that the sole use of a Meta account?

From the news it also sounds like Zuckerburg is trying to move Meta's Oculus products away from the Gaming category into a social interaction / futuristic work dystopia. Since we don't have a `futuristic work dystopia` category, maybe `other` or `social` are better fits?

hkamran80: I'll change it to `social`

Carlgo11 (approved):

PR #7126, opened 2022-12-17
Add FanFiction.Net

hkamran80:


entries/f/fanfiction.net.json
```diff
@@ -0,0 +1,13 @@
+{
+  "FanFiction.Net": {
+    "domain": "fanfiction.net",
+    "img": "fanfiction.net.png",
+    "contact": {
+      "twitter": "FictionPress",
+      "email": "support@fanfiction.com"
+    },
+    "keywords": [
+      "entertainment"
```
Carlgo11: Not sure if this is an appropriate category. Maybe `other`? :thinking:

hkamran80: I consider it `entertainment` because writers release stories for people to read, which counts as entertainment to me. I can switch it to `other` if you'd like.

Carlgo11: Yeah I still think it should be in `other`. Sorry 😬

hkamran80: No worries. I'll update it in a few.

Carlgo11 (approved):

PR #7150, opened 2022-12-21
Add Forem

hkamran80:


Carlgo11:
https://github.com/2factorauth/twofactorauth/blob/master/EXCLUSION.md#self-hosted-services

hkamran80:
Forem can be self-hosted, but it also has a centralized authentication service, which is available for all Forem sites.

Carlgo11 (approved):

PR #7155, opened 2022-12-23
Update Barmenia

juliushacker:
The customer login of Barmenia is on the domain meine-barmenia.de - the link "Kundenportal" on barmenia.de with the option "Login webversion" afterwards redirects to meine-barmenia.de


Carlgo11 (changes requested):
the domain element is generally meant to cover the main page. I suggest you use the additional-domains element instead.

Carlgo11 (approved):
Thank you! 🙂

PR #7160, opened 2022-12-25
Update API references

ApCoder123:


Carlgo11:
We could also just point to the API.md in 2factorauth/frontend to avoid duplicates

Carlgo11 (approved):

PR #7170, opened 2022-12-28
Update UWCU

mwinters0:


Carlgo11 (approved):
Thanks for the PR @WintersMichael!
Due to very little demand the UWCU entry hadn't been looked at for a very long time. That's also why we incorrectly listed the credit union as supporting proprietary software 2FA.

PR #7175, opened 2022-12-29
Replace Air France with Flying Blue

### LINKED ISSUE #3031: Add KLM or modify Air France
KLM and Air France has two separate website with different functionality, however the login methods are both tied to FlyingBlue accounts.
I would suggest to modify Air France to Air France/KLM ( Flying Blue) or create separate entries to:

- Air France ( already exists )
- KLM
- Flying Blue ( optional )

Carlgo11:
Resolves #3031.

Air France and KLM both advertise the Flying Blue accounts instead of their own local accounts. As no system exists for listing multiple entries for the same domain, I suggest the Air France listing should be replaced with this one.

Several other airlines are members of Flying Blue but only Air France and KLM support authentication with Flying Blue accounts.


entries/f/flyingblue.com.json
```diff
@@ -0,0 +1,16 @@
+{
+  "Flying Blue": {
+    "domain": "flyingblue.com",
+    "additional-domains": [
+      "airfrance.com",
+      "klm.com"
+    ],
```
hkamran80:
```suggestion
    "additional-domains": [
      "airfrance.com",
      "klm.com",
      "flyingblue.am",
      "flyingblue.at",
      "flyingblue.be",
      "flyingblue.bz",
      "flyingblue.ca",
      "flyingblue.cn",
      "flyingblue.co.in",
      "flyingblue.co.nz",
      "flyingblue.co.uk",
      "flyingblue.co.za",
      "flyingblue.com.au",
      "flyingblue.com.co",
      "flyingblue.com.mx",
      "flyingblue.com.ph",
      "flyingblue.com.tw",
      "flyingblue.com.ve",
      "flyingblue.de",
      "flyingblue.es",
      "flyingblue.fr",
      "flyingblue.la",
      "flyingblue.nl",
      "flyingblue.pl",
      "flyingblue.us"
    ],
```

Flying Blue is also available on some regional domains

hkamran80 (changes requested):

hkamran80 (approved):

PR #7179, opened 2022-12-29
Add Harvey Mudd College

hkamran80:


Carlgo11 (commented):
Have you found any mention of Duo D100 Tokens or is that just an assumption?

hkamran80:
I thought I had seen it somewhere, but now I can't find it. I'll replace with it "Duo token".

Carlgo11:
My question was rather if you had seen they supported hardware tokens at all. If they support Duo hardware tokens that would mean it's a Duo D100 Token but I know some sites have hardware 2FA support specifically turned off.

hkamran80:
They do support hardware tokens. From their [questions page](https://docs.google.com/document/d/1UHlq8zQDmmX2m4Tya0SvVszB1-9xqViVITuyWs5lriI/pub#h.mcvq26aawezy):

> Hardware tokens can be provided to people without smartphones or who do not wish to use their smartphones.  The first hardware token will be provided at no cost.  If you lose your token, it will need to be replaced and you will be billed $20.

Carlgo11 (approved):

PR #7195, opened 2022-12-30
Add OpenAI

hkamran80:


Carlgo11 (approved):
Interesting that they didn't use ML for the logo.
Looking at the result when I asked ChatGPT to draw an SVG, I kinda understand them though.
svg

PR #7198, opened 2022-12-31
Update Uber

leggomyinfo:
I was using TOTP for Uber. I disabled it and re-enabled it in order to set up TOTP on a different authenticator app. Once re-enabled, SMS was the only option.


entries/u/uber.com.json
```diff
@@ -3,13 +3,12 @@
     "domain": "uber.com",
     "url": "https://www.uber.com/",
```
hkamran80:
```suggestion
```

The `url` key is only needed if going to the domain (e.g. `https://uber.com`) fails

hkamran80 (changes requested):

hkamran80:
It's interesting that they haven't updated their docs yet

Carlgo11:
@leggomyinfo Have you reached out to Uber Support?
Their [help article](https://help.uber.com/riders/article/turn-on-2-step-verification?nodeId=b8bb9152-8c91-4f49-83c4-35cf2e1dcf72) on the matter still states TOTP is supported.

leggomyinfo:
I just contacted support. I’ll let you know their reply.

kmpoppe:
Pre-Pandemic I wasn't even able to log into my Uber account that I created for a tour in the UK, which nowadays seems to work.
I had set up TOTP in the account in 2019 and it is still working so I wouldn't remove TOTP as of now.

leggomyinfo:
After a lot of back and forth, Uber support was totally unhelpful. They agreed to escalate the ticket a few days ago — still waiting on a reply.

The primary audience for 2fa.directory is people looking up a site/service to see if a 2FA method is available, not people who already have that 2FA method enabled for the site/service. If you accept that premise, and if TOTP is in fact only available to Uber users who already have it enabled, it makes more sense to remove it and explain the weird situation with the tooltip note. At a minimum, if TOTP remains checked off for Uber, there should be a tooltip note – I'm guessing everyone can agree on that.

If there is someone who is willing to risk losing TOTP for Uber or doesn't already have it, please test it by disabling/re-enabling or trying to enable TOTP for the first time.

kmpoppe:
> After a lot of back and forth, Uber support was totally unhelpful.

Which was exactly my experience with them after debating why I wasn't able to access my account from Germany when I created it in the UK...

> At a minimum, if TOTP remains checked off for Uber, there should be a tooltip note – I'm guessing everyone can agree on that.

Indeed. With `notes` in the file, an exclamation mark is shown with could be used for warning users that TOTP seems to be unavailable for new setup as of a right now.

Carlgo11:
I took it upon myself to verify and yes, it is indeed not an option after resetting it.
I think the note is redundant (and misspelled?) but support removing TOTP from the list.

entries/u/uber.com.json
```diff
@@ -3,13 +3,12 @@
...
     "tfa": [
-      "sms",
-      "totp"
+      "sms"
     ],
     "documentation": "https://help.uber.com/h/b8bb9152-8c91-4f49-83c4-35cf2e1dcf72",
-    "notes": "2FA can only be enabled via the app",
+    "notes": "2FA can only be enabled via the app. TOTP is no longer an option but grandathered in if previously set up.",
```
Carlgo11:
```suggestion
    "notes": "2FA can only be enabled via the app",
```

Carlgo11 (changes requested):

Carlgo11 (approved):

hkamran80 (approved):

PR #7216, opened 2023-01-02
Add Vivup

jamesmacwhite:
Add Vivup and employee benefits service in the UK aimed at public sectors.

Below is the two-factor authentication options available

image


hkamran80 (changes requested):
Please change the name of the file to `vivup.co.uk.json`

jamesmacwhite:
Sorry about that JSON filename corrected. Also added proof of 2FA options.

hkamran80 (approved):

PR #7240, opened 2023-01-10
Minimize ProduBanco icon

phallobst:


Carlgo11 (approved):
javascript inside an SVG file :fearful:

PR #7243, opened 2023-01-11
Add OpenStreetMap

slothtown:


Carlgo11 (approved):
Thanks for the pull request @slothtown!
I replaced your SVG file with a PNG file. SVG files are generally preferred but the one you supplied was over 33KiB in size and that's a bit much for a single logo 😅

slothtown:
Wow, I didn't realize it was that big! Thanks for changing it!

PR #7261, opened 2023-01-13
Add Lyyti

slothtown:
Lyyti.com isn't in similarweb top 200k, but additional domain lyyti.fi is.


Carlgo11 (changes requested):
Could this one be used instead?
lyyti

Carlgo11 (approved):

PR #7262, opened 2023-01-13
Replace svgo with nano in CONTRIBUTING.md

slothtown:
Nano is easier to use than svgo for the average user, so I thought it should replace it. Also in the Wiki page, nano is already recommended instead of svgo.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ are stored in folders corresponding to each of those entries in their own
    merged. GH Actions will only check your changes after you submit a pull request.
    If you want to test locally, instructions are listed below. Keep reading!
 2. **Use a Nice Icon**: SVG is the preferred format. If possible, please also run the image 
-   through an optimizing utility such as [svgo][svgo] (advanced) or [nano][nano] (easy) to reduce the file size.
+   through an optimizing utility such as [svgo][svgo] (command line) or [SVGOMG][SVGOMG] (web) to reduce the file size.
    If an SVG icon is not available, the icon should be a PNG with a resolution of 32x32, 64x64 or 128x128.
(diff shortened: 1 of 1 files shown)


phallobst:
There are various SVG optimizers that have different target audiences, e.g.
* [svgo](https://github.com/svg/svgo): command line tool, especially useful for batch processing
* [SVGOMG](https://jakearchibald.github.io/svgomg/), the WYSIWYG web frontend driven by svgo
* [nano](https://vecta.io/nano), another web service, easier to use but fewer bells and whistles

Maybe these could all be listed in the documentation. None of them is the ultimate tool for _all_ SVG files.

slothtown:
I agree. Nano is so easy and straightforward to use that I think that for most (new) contributors it would be the best choice, as they don't have to install a new program and learn how to use it. I think it should be mentioned at least somewhere in the contributing guide.

hkamran80:
Maybe say "... through an optimizing utility such as svgo or nano to reduce the file size"?

slothtown:
@hkamran80 Maybe something like this? I just hope that new contributors don't feel overwhelmed with svgo and then decide to skip the svg optimizers all together.

hkamran80:
@slothtown Most people would probably use [SVGOMG](https://jakearchibald.github.io/svgomg/) if they use svgo. Should there be a link to that as well?

Also, maybe change "easy" to "simple" for nano? I feel like "simple" fits better with "advanced".

Carlgo11:
Honestly, vecta.io/nano produces such low quality SVG optimization that they have to be run through SVGO anyway before being merged. It's probably better to just link to SVGOMG and SVGO.

slothtown:
Would this work?

Carlgo11 (approved):
Works for me :+1:

PR #7276, opened 2023-01-14
Specify net-http instead of httpclient in header

phallobst:


Carlgo11:
`Gem.loaded_specs['net-http']` just returns nil for me

tests/language-codes.rb
```diff
@@ -17,7 +17,7 @@
   headers = {
     'Accept' => 'application/json',
     'User-Agent' => '2FactorAuth/LanguageValidator '\
-    "(HTTPClient/#{Gem.loaded_specs['httpclient'].version} on Ruby/#{RUBY_VERSION}; +https://2fa.directory/bot)",
+    "(Net::HTTP/#{Gem.loaded_specs['net-http'].version} on Ruby/#{RUBY_VERSION}; +https://2fa.directory/bot)",
```
Carlgo11:
```suggestion
    "(Ruby/#{RUBY_VERSION}; +https://2fa.directory/bot)",
```

tests/region-codes.rb
```diff
@@ -16,7 +16,7 @@
   headers = {
     'Accept' => 'application/json',
     'User-Agent' => '2FactorAuth/RegionValidator' \
-    "(HTTPClient/#{Gem.loaded_specs['httpclient'].version} on Ruby/#{RUBY_VERSION}; +https://2fa.directory/bot)",
+    "(Net::HTTP/#{Gem.loaded_specs['net-http'].version} on Ruby/#{RUBY_VERSION}; +https://2fa.directory/bot)",
```
Carlgo11:
```suggestion
    "(Ruby/#{RUBY_VERSION}; +https://2fa.directory/bot)",
```

Carlgo11 (changes requested):

phallobst:
> `Gem.loaded_specs['net-http']` just returns nil for me

That's basically how I stumbled upon this problem - the `httpclient` gem was not installed and the script gave me an error message. Does the test work for you at all, @Carlgo11? (Without the cached file in `/tmp`, that is.)

Carlgo11:
Script works fine without `Gem.loaded_specs['net-http']`.
httpclient is also in the tests group so running `bundle install --with tests` should install it for you.

phallobst:
Since these tests use Net::HTTP instead of HTTPClient, I figured it would be useless to specify the `httpclient` version, but leaving it out altogether is okay for me as well. :man_shrugging:

Carlgo11 (approved):

Carlgo11:
Net is part of the core Ruby methods so it makes sense leaving it out.

PR #7284, opened 2023-01-17
Update eBay

slothtown:
I removed the regional restrictions as eBay is used worldwide. Also removed url as ebay.com works.


Carlgo11 (changes requested):
As with most other U.S. businesses, eBay only serves customers in countries not sanctioned by the U.S. Apart from that there also needs to be an established postal system in the country for them to ship goods there.
I think the countries listed in the Global Shipping Program can be seen as served by eBay.

https://www.ebay.com/help/selling/shipping-items/setting-shipping-options/global-shipping-program?id=4646

slothtown:
Does 2fa directory have regional filters for any of the countries eBay doesn't serve? If yes, should the ebay json be edited to exclude those countries instead of including all the countries they serve? In any case, the json is currently missing at least Finland

entries/e/ebay.com.json
```diff
@@ -38,56 +37,6 @@
...
-      "py",
-      "ru",
-      "sg",
-      "sv",
-      "tr",
-      "tw",
-      "us",
-      "uy",
-      "ve"
     ]
```
Carlgo11: Looked it up and eBay delivers to 89 countries out of 195-197. 
```suggestion
    "regions": [
        "ae",
        "ag",
        "aw",
        "bd",
        "be",
        "bg",
        "bh",
        "bm",
        "bo",
        "bs",
        "bz",
        "ca",
        "ch",
        "cl",
        "cn",
        "co",
        "cr",
        "cy",
        "cz",
        "de",
        "dk",
        "do",
        "ec",
        "ee",
        "eg",
        "es",
        "fi",
        "fr",
        "gb",
        "gb",
        "gf",
        "gg",
        "gi",
        "gr",
        "gt",
        "hk",
        "hn",
        "hu",
        "ie",
        "il",
        "in",
        "is",
        "it",
        "je",
        "jm",
        "jo",
        "jp",
        "kh",
        "kn",
        "kr",
        "kw",
        "ky",
        "lc",
        "lk",
        "lu",
        "lv",
        "mc",
        "mq",
        "ms",
        "mt",
        "mv",
        "mx",
        "my",
        "ni",
        "nl",
        "no",
        "nz",
        "om",
        "pa",
        "ph",
        "pk",
        "pl",
        "pt",
        "qa",
        "re",
        "ro",
        "sa",
        "se",
        "sg",
        "si",
        "sk",
        "th",
        "tt",
        "tw",
        "ua",
        "us",
        "uy",
        "vn",
        "za"
    ]
```

slothtown:
Great, thanks! I'll update the file when I get to my pc (Github doesn't allow applying edits on deleted lines).

On a side note, are the current additional domains up to date? There were a few urls that Github Actions were unable to reach. There were also some urls without the word "ebay", but those might of course be some regional sites with different branding or something similar.

entries/e/ebay.com.json
```diff
@@ -26,7 +26,6 @@
       "gittigidiyor.com",
```
Carlgo11: Looks like this domain is no longer used. (See [Web Archive](https://web.archive.org/web/20220918233739/http://www.gittigidiyor.com/))
The rest look good :+1: 
```suggestion
```

Carlgo11 (changes requested):

Carlgo11:
After further searching it turns out eBay really only exclude the absolute minimum required for them to operate in the U.S. Updated the regions array to reflect that.

Edit: And Turkey for some reason :shrug:

Carlgo11 (approved):

PR #7285, opened 2023-01-18
Add Discovery Plus

slothtown:
The svg is huge but I was unable to find a smaller one


Carlgo11:
I think everything over 10KiB is a bit much :sweat_smile: 
icon

slothtown:
Yeah, I get it 😁 I'll switch the image when I get to my pc

Carlgo11 (approved):

PR #7287, opened 2023-01-19
Clean up validate-urls.rb

Carlgo11:
Removed last usage of httpclient gem.
Cleaned up code in validate-urls

I also removed the exit code changing on certain HTTP errors as test failures are ignored in the [workflow](https://github.com/2factorauth/twofactorauth/blob/master/.github/workflows/repository.yml#L38). (Uncertain if this is desirable?)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,6 @@ source 'https://rubygems.org'
 
 group :tests, optional: true do
   gem 'addressable'
-  gem 'httpclient'
   gem 'json_schemer'
   gem 'rubocop'
   gem 'twitter'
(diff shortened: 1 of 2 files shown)


phallobst:
Does this check never fail then? There are no more exit codes nor `::error::` lines. At least invalid TLDs or something like that should be handled.

Carlgo11:
Neither the current script nor the script with these proposed changes can fail as the workflow ignores non-zero exit codes.
Instead of completely blocking the build from being merged the script instead show warnings in the "Files Changes" tab on each PR.

This is by design as many sites block requests our script makes.

phallobst (approved):

PR #7288, opened 2023-01-19
Add Killstar

slothtown:
Sources for excluded regions:
https://www.killstar.com/pages/contact-us
https://www.killstar.com/pages/faq#reamaze#0#/kb/shipping-501759cf098a591d/do-you-ship-worldwide-de20ffc0e6f42254


Carlgo11 (approved):
Killstar sounds like a military callsign. 
Killstar, Texaco, return pre-contact 🥴

slothtown:
It does haha

slothtown:
@Carlgo11 I tested searching for Killstar on https://2fa.directory/ru and found it, even though Russia is excluded in the json file. Is there a problem in the json or somewhere else?

Carlgo11:
All this logic hurts my brain 😓

I had accidentally written the if statement so that excluded regions were treated the same way as included regions.

PR #7293, opened 2023-01-20
Add Retroachievements

slothtown:


phallobst:
How about this icon in SVG format:
retroachievements

slothtown:
Thanks, it's better! I was only able to find a black and white svg

Carlgo11 (approved):

PR #7305, opened 2023-01-23
Update Adobe ID

slothtown:
Adobe ID doesn't support totp anymore, as you can see in the documentation: https://helpx.adobe.com/manage-account/using/secure-your-adobe-account.html

### DIFF
--- a/entries/a/adobe.com.json
+++ b/entries/a/adobe.com.json
@@ -4,7 +4,6 @@
     "tfa": [
       "sms",
       "email",
-      "totp",
       "custom-software"
     ],
 	  "custom-software": [


phallobst:
The recent PR #7211 says otherwise, although it's not documented. Do you have different information?

slothtown:
I don't use Adobe ID but the option was missing when I tried enabling it on a friend's account. There's also a post from March 2022 about this problem, and a user said that they contacted Adobe and were told that 3rd party authentication apps aren't supported anymore: https://community.adobe.com/t5/account-payment-plan-discussions/does-anyone-know-how-to-setup-2fa-with-totp-without-secure-signin-app/td-p/12806773

Seems like only some people have the totp option. I don't know whether this has something to do with an account type, or when the account was created (or whether totp is already enabled).

Carlgo11 (approved):

Carlgo11:
Just created an account and there's no TOTP option. It could very well be that it's visible for the author of #7211 because they had it enabled when it was an option.

PR #7311, opened 2023-01-24
Add hardware key support for Apple/iCloud

emielwiedijk:
Since macOS 13.2/iOS 16.3 Apple IDs support FIDO tokens for authentication.

Source: https://support.apple.com/en-us/HT213154


Carlgo11 (approved):
:partying_face:

PR #7314, opened 2023-01-25
Update Schneider Electric category

Carlgo11:
Schneider Electric is a company delivering energy solutions to utility companies and IoT solutions to consumers. As such IoT is a more fitting category.

### DIFF
--- a/entries/s/se.com.json
+++ b/entries/s/se.com.json
@@ -10,7 +10,7 @@
     ],
     "documentation": "https://www.se.com/ww/en/faqs/FAQ000220689/",
     "categories": [
-      "utilities"
+      "iot"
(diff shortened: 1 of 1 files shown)


hkamran80:
`iot` is a good category, but I think `utilities` should stick around. Schneider also has some utility products, such as their [smart residential EV charger](https://www.se.com/us/en/product-range/161147017-schneider-charge).

Carlgo11:
>Schneider Charge is part of the Schneider Home energy management ecosystem.

That still sounds more like an IoT platform rather than a utility provider.

hkamran80:
True

hkamran80 (approved):

PR #7337, opened 2023-01-29
Update Glorious

phallobst:


hkamran80 (approved):

entries/g/gloriousgaming.com.json
```diff
@@ -0,0 +1,12 @@
+{
+  "Glorious": {
+    "domain": "gloriousgaming.com",
+    "contact": {
+      "facebook": "GloriousPCG",
```
hkamran80:
```suggestion
```

The check is failing

hkamran80 (changes requested):

entries/g/gloriousgaming.com.json
Carlgo11: Looks like Facebook is restricting access to some pages more than others.
We might need a Facebook API key in the future :unamused: 
In the meantime I second @hkamran80's suggestion ^

johntheholman: The page exists though and you can access it (like some of the `domain` changes I have been working on lately). Is there something that may need to be updated with the tests?

Screen Shot 2023-02-02 at 2 55 22 AM

hkamran80: Like @Carlgo11 said, we might need to add an API key for Facebook. This removal is temporary, at least until we can fix this bug.

hkamran80 (approved):

PR #7352, opened 2023-01-31
Update to namebase.io

johntheholman:
Added documentation for recovering an account with 2FA

### DIFF
--- a/entries/n/namebase.io.json
+++ b/entries/n/namebase.io.json
@@ -4,6 +4,7 @@
     "tfa": [
       "totp"
     ],
+    "recovery": "https://www.namebase.io/faq/#how-do-i-reset-my-2fa",
     "categories": [
       "domains",
       "cryptocurrencies"


phallobst (approved):

johntheholman:
@phallobst what is the preference for these types of links? Should we try to go to the specific answer if on a very tall page? Or does a link to the page that contains the answer more what we are looking for?

phallobst:
If there is a suitable answer with an HTML anchor (like here), we would prefer to use it.

johntheholman:
Sounds good. I will have some more updates coming soon then 😄 . Thanks @phallobst.

johntheholman:
One last question (for now), if `documentation` also contains `recovery` information should this be left to just `documentation` or should we duplicate the URL to the `recovery` key? In this particular scenario there are no anchor tags to differentiate the two links from each other.

phallobst:
If the recovery information is on the same page (even with its own anchor), we only use `documentaton`.

PR #7353, opened 2023-01-31
Update 15five

johntheholman:
- Updated `documentation` link to include anchor tag to the specific spot to setup 2FA
- Added a `note` about not being able to turn on two-step authentication if your organisation uses SSO

### DIFF
--- a/entries/1/15five.com.json
+++ b/entries/1/15five.com.json
@@ -4,7 +4,8 @@
     "tfa": [
       "totp"
     ],
-    "documentation": "https://success.15five.com/hc/en-us/articles/360002698811",
+    "documentation": "https://success.15five.com/hc/en-us/articles/360002698811#h_01G1GXTSKCGMSC3ECKRFSC8NWM",
+    "notes": "You cannot enable two-step authentication if your organization uses SSO.",
     "categories": [
       "communication"
     ]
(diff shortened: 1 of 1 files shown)


entries/1/15five.com.json
```diff
@@ -4,7 +4,8 @@
     "tfa": [
       "totp"
     ],
-    "documentation": "https://success.15five.com/hc/en-us/articles/360002698811",
+    "documentation": "https://success.15five.com/hc/en-us/articles/360002698811#h_01G1GXTSKCGMSC3ECKRFSC8NWM",
+    "notes": "You cannot enable two-step authentication if your organization uses SSO.",
```
Carlgo11: This part isn't something we usually write about since SSO means authentication is handled externally and external authentication providers are out of scope.

johntheholman: From their documentation it sounds like they leave authentication up to the organisations. If the organisation chooses to use SSO then they (15five.com) have no ability to enforce the 2FA because you are taking authentication out of their hands. If I, as a user, wanted to turn on 2FA but my organisation turned on SSO then I am out of luck. I feel like this note is pretty important since the user is forced into a (potentially) less secure option because the organisation wants to handle SSO their own way. How would you like me to proceed?

Carlgo11: As both you and I mentioned, SSO means the organization/SAML-provider handles all forms of authentication.
SSO/SAML means authentication is outside the trust-boundary of the site and instead inside the trust-boundary of the SAML-provider. Thus out of scope for the entry.

Just like we don't label sites offering authentication via Google or Facebook as having 2FA, we don't note exceptions for SSO/SAML.

Carlgo11 (approved):

PR #7356, opened 2023-01-31
Update Acorns

johntheholman:
- Added a `note` about needing a verified contact to turn on 2FA and the caveat about PIN and FaceID
- Update based on `editorconfig`
- Updated to include `email` as a 2FA option

### DIFF
--- a/entries/a/acorns.com.json
+++ b/entries/a/acorns.com.json
@@ -3,10 +3,11 @@
     "domain": "acorns.com",
     "url": "https://www.acorns.com/",
     "tfa": [
-      "sms"
+      "sms",
+      "email"
     ],
     "documentation": "https://www.acorns.com/support/how-do-i-turn-on-two-factor-authentication-on-my-account/",
-    "note": "Using your PIN or Face ID in-app will not activate 2FA",
+    "note": "You will need to verify one piece of contact informaiton and using your PIN or Face ID in-app will not activate 2FA.",
     "categories": [
       "investing"
     ],
--- a/entries/a/acorns.com.json
+++ b/entries/a/acorns.com.json
@@ -7,7 +7,7 @@
       "email"
     ],
     "documentation": "https://www.acorns.com/support/how-do-i-turn-on-two-factor-authentication-on-my-account/",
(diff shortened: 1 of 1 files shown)


entries/a/acorns.com.json
```diff
@@ -3,14 +3,16 @@
     "domain": "acorns.com",
     "url": "https://www.acorns.com/",
     "tfa": [
-      "sms"
+      "sms",
+      "email"
     ],
     "documentation": "https://www.acorns.com/support/how-do-i-turn-on-two-factor-authentication-on-my-account/",
+    "notes": "You will need to verify one piece of contact informaiton and using your PIN or Face ID in-app will not activate 2FA.",
```
hkamran80:
```suggestion
    "notes": "You will need to verify one piece of contact information before activating 2FA",
```

Isn't the second part (the in-app security) unnecessary? I would argue that most know that it doesn't activate 2FA.

entries/a/acorns.com.json
```diff
@@ -3,14 +3,16 @@
     "domain": "acorns.com",
     "url": "https://www.acorns.com/",
```
hkamran80:
```suggestion
```

hkamran80 (changes requested):

entries/a/acorns.com.json
johntheholman: They are saying that if you use those forms of authentication they will not ask you for your second factor.

entries/a/acorns.com.json
johntheholman: Good catch, I will start paying attention to this more as I go through the sites at this time.

entries/a/acorns.com.json
johntheholman: So maybe better to say: `You will need to verify email or phone in order to activate 2FA. If you use biometrics to unlock the app it will not prompt you for your second factor.`

hkamran80: I would remove the part about the app, because I would assume that most know that enabling app-level security doesn't affect the account overall. Otherwise, looks good to me.

hkamran80 (approved):

PR #7362, opened 2023-02-01
`url` Deleted Based on `domain`: The `A`s

johntheholman:
- Removes the `url` key if the `domain` key is the same
- Small updates to some files based on the `editorconfig`
- This is only looking at the `A` files (as seen when hitting the `v3/all.json` endpoint)


hkamran80:
Out of curiosity, how did you do this?

johntheholman:
I pulled the `JSON` and then have just been going down the list but searched through VSCode for all of the ones that have a `"url"` key (since that is the one that would need to be deleted if they are the same). I could script this more but this is pretty quick since the `"url"` key is not used that much throughout the `JSON` file. Then it is all about running the `tests/validate-urls` script to make sure you haven't deleted something that would break everything.

hkamran80:
Ah. Just be careful because some sites, like Bourse Direct, can only be accessed by prefixing `www`.
Speaking of which, Bourse Direct is one of the sites flagged for `url` removal in #7363.

johntheholman:
Good call out. The tests have been passing so thank you for giving me this example so that I can make these PRs better. I will change these to drafts until I have been able to check all of these out individually.

Carlgo11:
https://github.com/2factorauth/twofactorauth/actions/runs/4061713761/jobs/6991999898#step:11:12
Check aetna.com

johntheholman:
All of them load for me. In fact, I will need to update the `AJ Bell Youinvest` site next since it redirects to `ajbell.co.uk` so it would have to change folders.

Could it be that the script does not allow enough time for it to resolve?

Here is `aetna.com`:
https://user-images.githubusercontent.com/45466974/216299378-88ca8fad-a340-44c4-af50-275292fd10c2.mov

Carlgo11:
Your browser loads the HTTP version (see the address bar @ 1 second).
Try it with cURL or disable HTTP requests in your browser.

Carlgo11:
> **Warning**
> Unless there's activity within the next 7 days I'll close the PR.

johntheholman:
@Carlgo11 updates have been made to this branch.

Carlgo11:
Only `aetna.com` and `aliexpress.com` needed to be fixed tho 😬

johntheholman:
@Carlgo11 let me fix that then, for some reason it look like it was showing more.

johntheholman:
@Carlgo11 updated.

Carlgo11 (approved):

PR #7367, opened 2023-02-02
Support relative redirects

Carlgo11:
Apparently relative URLs are allowed during redirects.

Source: RFC 7231

### DIFF
--- a/tests/validate-urls.rb
+++ b/tests/validate-urls.rb
@@ -27,7 +27,7 @@ def check_url(path, url, res = nil)
 
   puts "::warning file=#{path}:: Unexpected response from #{url} (#{res.code})" unless res.is_a? Net::HTTPSuccess
 rescue StandardError => e
-  puts "::error file=#{path}:: Unable to reach #{url}"
+  puts "::warning file=#{path}:: Unable to reach #{url}"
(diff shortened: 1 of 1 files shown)


tests/validate-urls.rb
```diff
@@ -15,19 +15,22 @@
...
     break unless res.is_a? Net::HTTPRedirection
 
-    url = res['location']
+    url = URI(res['location']).is_a?(URI::HTTP) ? res['location'] : "#{url.scheme}://#{url.host}#{res['location']}"
   end
 
   puts "::warning file=#{path}:: Unexpected response from #{url} (#{res.code})" unless res.is_a? Net::HTTPSuccess
 rescue StandardError => e
-  puts "::warning file=#{path}:: Unable to reach #{url}"
+  puts "::error file=#{path}:: Unable to reach #{url}"
```
Carlgo11: I also switched unreachable URLs from outputting warnings to errors.

phallobst: Some sites are only accessible from a certain region (or with a VPN), so a hard error might be a bit much here.

Carlgo11: Good point :thinking:
I'll revert this line then

phallobst (approved):

PR #7368, opened 2023-02-03
Add file permission validation

Carlgo11:


tests/validate-fs.sh
```diff
@@ -0,0 +1,42 @@
...
+# $2, $3, ... = permitted file permissions
+function checkPerm()
+{
+  dir="$1"
+  shift
+  violation=$(find $dir -type f -exec stat -c '%a' '{}' +| grep -vq "$(echo $* | sed 's/ /|/g')")
+  for file in $violation; do
+    echo "Directory '$dir' may only have files with the following permissions: $*"
+    status=1
+  done
```
phallobst:
```suggestion
  unset pattern
  for p; do
    pattern+="${pattern+ -a }! -perm $p"
  done
  if find $dir -type f $pattern | grep -q .; then
    echo "Directory '$dir' may only have files with the following permissions: $*"
    status=1
  fi
```
The `-perm` option for `find` is much faster. Also, the error message should only be displayed once per directory, not once for each offending file.

phallobst (changes requested):

tests/validate-fs.sh
Carlgo11: Thanks! Great to have a shell wizard on the team :smile:

phallobst:
Do we know which versions of `find` and `grep` are in use? I hope it's not some variant of `busybox` or something...

Locally, this script worked fine.

Carlgo11:
Haha yup it's busybox :stuck_out_tongue_winking_eye:

Carlgo11:
The worker should be run on Ubuntu 22.04 LTS so that would make it GNU?

phallobst (approved):

PR #7370, opened 2023-02-03
Update Bitrix24

rakleed:
- Add regional domains.

### DIFF
--- a/entries/b/bitrix24.com.json
+++ b/entries/b/bitrix24.com.json
@@ -1,7 +1,11 @@
 {
   "Bitrix24": {
     "domain": "bitrix24.com",
-    "url": "https://www.bitrix24.com/",
+    "additional-domains": [
+      "bitrix24.ru",
+      "bitrix24.eu",
+      "bitrix24.net"
+    ],
     "tfa": [
       "totp"
     ],


Carlgo11:
With my limited knowledge about Bitrix24, it looks like they operate in Europe and the U.S. as [bitrix24.com](https://bitrix24.com)/[bitrix24.eu](https://bitrix24.eu) (Bitrix24 Ltd. / Bitrix, Inc.) whilst operating in Russia with [bitrix24.ru](https://bitrix24.ru) (1C-Bitrix OOO).

If you want to list the Russian site I suggest you create a different entry for bitrix24.ru and specify `ru` in the regions array. At the same time you can specify `-ru` in the regions array for bitrix24.com.

As a site note, what's the difference between Б and в? 🤨

rakleed:
Done.

Б/б is pronounced like B/b, and В/в is like V/v. Well, they are different letters. Why are you asking?

entries/b/bitrix24.ru.json
```diff
@@ -0,0 +1,18 @@
+{
+  "Битрикс24>": {
+    "domain": "bitrix24.ru",
+    "additional-domains": [
+      "bitrix24.net"
+    ],
```
phallobst: You can just re-use the original icon here and remove `img/b/bitrix24.ru.svg`.
```suggestion
    ],
    "img": "bitrix24.com.svg",
```

phallobst (changes requested):

entries/b/bitrix24.ru.json
phallobst: The `additional-domains` are primarily targeted at password managers. I'm not sure how they are able to properly handle duplicates across different entries.

rakleed: Done.

rakleed: I can remove an additional domain, but usually saving the login and password is attached to the login page.

phallobst:
I'm actually not quite sure why we need a separate entry for the `.ru` domain in the first place. It would probably be better to just list all the `additional-domains` in the original one, especially since they seem to share a common login page.

rakleed:
@Carlgo11 I have added additional regional domains as suggested by @phallobst since I have checked in the media and on [Wikipedia[ru]](https://ru.wikipedia.org/wiki/1С-Битрикс_(компания)) and nowhere is it mentioned that they are different companies. In general, in 2007, Bitrix created a joint venture with 1C, and therefore their company is called 1C-Bitrix. And for the European and American markets, they have registered separate companies, most likely to simplify the interaction with their customers.

phallobst (approved):

PR #7387, opened 2023-02-10
Update Skrill

wnelson03:
image

totp has been removed from Skrill, it's not in the settings as previously documented https://i.imgur.com/Np3HbCN.png, https://authy.com/guides/skrill/

I also updated the excluded countries to match https://www.skrill.com/en-us/footer/terms-conditions/non-serviced-countries/


Carlgo11:
That's a lot of excluded countries o_o

entries/s/skrill.com.json
```diff
@@ -16,7 +15,118 @@
...
+	  "-tm",
+	  "-tc",
+	  "-tv",
+	  "-vu",
+	  "-ve",
+	  "-wf",
+	  "-eh",
+	  "-ws",
+	  "-ye",
+	  "-zw"
```
Carlgo11: two more spaces would be preferred by our editorconfig.
Also, I think you accidentally removed the blank line at the end of the file :wink: 

Depending on what editor you're using, there might be an [editorconfig](https://editorconfig.org/) plugin available.

Carlgo11 (changes requested):

Carlgo11 (approved):

Issue #7391, opened 2023-02-10
Add wemakeit


svoop:
### Site name

wemakeit

### Site URL

https://wemakeit.com

### Documentation URL

https://wemakeit.com/user/edit/settings

### Supported 2FA methods

- [X] TOTP (RFC-6238, Google Authenticator)
- [ ] Proprietary Software (Authy, Duo, etc.)
- [ ] U2F/WebAuthn security keys
- [ ] Proprietary hardware (Yubico OTP, RSA keys, etc.)
- [ ] Phone calls
- [ ] SMS tokens
- [ ] Email tokens

### Eligibility

- [X] I have checked that the site meets 2fa.directory's [contributing guidelines and site criteria](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md)
- [ ] The site is ranked within the top 200,000 global sites on [Similarweb](https://www.similarweb.com/).
- [X] The site does not contain or promote content that violates 2fa.directory's [excluded categories and websites guidelines](https://github.com/2factorauth/twofactorauth/blob/master/EXCLUSION.md).
- [X] The issue I'm creating is not a duplicate of an existing [issue](https://github.com/2factorauth/twofactorauth/issues) or [pull request](https://github.com/2factorauth/twofactorauth/pulls).

### Additional information

wemakeit.com is the top crowdfunding site by a margin in Switzerland and recently has been sold and handed over to its own community. Since the site is dealing with money and also features credit accounts, it's important to protect those with 2FA. Full disclosure: I'm the lead developer and my colleagues have worked hard to integrate the second factor nicely into our workflows. We'd really like to get the word out particularly to the folks using password managers (which we also encouraged in the past).

I'm making this submission despite the fact that wemakeit.com is not yet in the top 200k sites globally (yet), rather near the top 300k. While available globally, our focus is Switzerland. At 8mio people, it's a really small country and making the top 200k with this audience is beyond our SEO.

Then again, the pozible.com crowdfinding site is on the list even though they don't feature 2FA at all and rank well above 700k.

Thanks for considering us anyways!

PR #7392

svoop:
Apparently not, PR has been sunk, so closing...

PR #7392, opened 2023-02-10
Add wemakeit

svoop:
Issue #7391

wemakeit.com is the top crowdfunding site by a margin in Switzerland and recently has been sold and handed over to its own community. Since the site is dealing with money and also features credit accounts, it's important to protect those with 2FA. Full disclosure: I'm the lead developer and my colleagues have worked hard to integrate the second factor nicely into our workflows. We'd really like to get the word out particularly to the folks using password managers (which we also encouraged in the past).

I'm making this submission despite the fact that wemakeit.com is not yet in the top 200k sites globally (yet), rather near the top 300k. While available globally, our focus is Switzerland. At 8mio people, it's a really small country and making the top 200k with this audience is beyond our SEO.

Then again, the pozible.com crowdfinding site is on the list even though they don't feature 2FA at all and rank well above 700k.

Thanks for considering us anyways!


Carlgo11:
Thanks for the pull request @svoop.
It’s great to hear that you’ve implemented 2FA (although FIDO U2F might be a more secure option than pure RRC-6238).
Unfortunately, as you’ve already mentioned, your site doesn’t qualify to be listed on 2fa.directory.

Feel free to reopen this pull request if/when your site meets our requirements and best of luck with your project.

svoop:
@Carlgo11 That's unfortunate... tough criteria to match when you're in a small country, top 200k might never happen I'm afraid. (Even "your" espressohouse.com barely makes the top 200k.) However, we'll soon reach 100mio€ in pledges, dealing with that much money ought to be a criteria in the context of 2FA as well, just sayin' 😉

Carlgo11:
We’re not hindering anyone from adding 2FA to their websites. However, as this project is entirely maintained by volunteers, we’re not able to maintain a list of every website supporting 2FA and so a cutoff point of 200 000 have been made.

Regarding your remarks about Espresso House, it’s a Swedish coffee chain without any internet presence and still able to make it to the top 200 000 websites. As Sweden and Switzerland are similar in population I don’t think your argument about population have much merit.

svoop:
@Carlgo11 Point taken. Still, traffic is a poor metric in our case. And we're in the process of ditching GA in favor of a privacy-friendly alternative, therefore our Similarweb rank will implode (or rather explode) soon. I get the workload argument, I'm spending most of my time contributing to FOSS myself, however, the merge button was only 2cm (or 0.003 calories) right of the close button. I'll shut up now, your rules, not mine.

PR #7400, opened 2023-02-14
remove Udemy notes

fharper:
2fa is available for every account type now

### DIFF
--- a/entries/u/udemy.com.json
+++ b/entries/u/udemy.com.json
@@ -4,7 +4,7 @@
     "tfa": [
       "email"
     ],
-    "documentation": "https://support.udemy.com/hc/en-us/articles/1500009156921",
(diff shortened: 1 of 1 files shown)


entries/u/udemy.com.json
```diff
@@ -5,7 +5,6 @@
       "email"
     ],
     "documentation": "https://support.udemy.com/hc/en-us/articles/1500009156921",
```
phallobst: This page is probably more applicable then.
```suggestion
    "documentation": "https://support.udemy.com/hc/en-us/articles/4420349685015",
```

phallobst (changes requested):

entries/u/udemy.com.json
fharper: Good call, I didn't check it!

phallobst (approved):

PR #7411, opened 2023-02-19
Add Scribd

hkamran80:
They have two-factor authorization, not two-factor authentication. To quote their [help page](https://support.scribd.com/hc/en-us/articles/4406311611156):
> To add additional security, our users will need to verify the email address linked to their Scribd account. Verification is simple and should only need to be completed once. After your email is verified, any changes to your account settings or password will need to be confirmed with a custom access link delivered to your email. This helps to ensure that you are the only person accessing your account.


Carlgo11 (approved):
By your quote, it sounds like Scribd could do with some reading up on the authentication basics :eyes:

hkamran80:
That they definitely could!

PR #7423, opened 2023-02-20
Update Stanford University

hkamran80:
They now support U2F through Duo ([source](https://uit.stanford.edu/service/authentication/twostep/securitykey))

### DIFF
--- a/entries/s/stanford.edu.json
+++ b/entries/s/stanford.edu.json
@@ -11,6 +11,9 @@
     "custom-software": [
       "Duo"
     ],
+    "custom-hardware": [
+      "Fortitoken"
+    ],
     "documentation": "https://uit.stanford.edu/service/webauth/twostep",
     "categories": [
       "universities"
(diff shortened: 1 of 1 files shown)


entries/s/stanford.edu.json
```diff
@@ -1,12 +1,12 @@
 {
   "Stanford University": {
     "domain": "stanford.edu",
-    "url": "https://www.stanford.edu/",
     "tfa": [
       "sms",
       "call",
       "custom-software",
-      "custom-hardware"
+      "custom-hardware",
```
Carlgo11, line 8: Looks like the Fortitoken option is deprecated so it seems wise to remove it from the list.
```suggestion
```

hkamran80: Where does it say deprecated? [All I see](https://uit.stanford.edu/service/authentication/twostep/token) is that they recommend security keys over it.

Carlgo11: That was what I concluded from the message but you're right, it's not spelled out, so either remove the custom-hardware method or add `Fortitoken` as a custom-hardware element.

hkamran80: Sure thing

Carlgo11 (approved):

PR #7427, opened 2023-02-20
Update Virginia Tech

hkamran80:
The Duo D100 token is mentioned on the main docs page

### DIFF
--- a/entries/v/vt.edu.json
+++ b/entries/v/vt.edu.json
@@ -6,7 +6,14 @@
       "sms",
       "call",
       "custom-software",
+      "custom-hardware",
       "u2f"
+    ],
+    "custom-software": [
+      "Duo"
+    ],
+    "custom-hardware": [
+      "Duo D100 Token"
     ],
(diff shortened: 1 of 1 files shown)


entries/v/vt.edu.json
```diff
@@ -1,13 +1,19 @@
...
       "call",
       "custom-software",
+      "custom-hardware",
       "u2f"
     ],
+    "custom-software": [
+      "Duo"
+    ],
+    "custom-hardware": [
+      "Duo D100 Token"
```
Carlgo11: Looks like Yubico OTP is also being used.
>* When prompted for a second factor, click Enter a Passcode.
>* Click to place the cursor in the text box.
> * Tap your YubiKey.

On the other hand, the only mention of U2F on the page is the following:
> #### To authenticate with a YubiKey:
>  The YubiKey must be previously registered with Duo as AES. **U2F** tokens will not work.

hkamran80: I'll fix that

Carlgo11 (approved):

PR #7447, opened 2023-03-05
Add TOTP to Adobe ID

gbdlin:
Adobe ID does support TOTP, adding this information to the website.

### DIFF
--- a/entries/a/adobe.com.json
+++ b/entries/a/adobe.com.json
@@ -4,11 +4,12 @@
     "tfa": [
       "sms",
       "email",
-      "custom-software"
+      "custom-software",
+      "totp"
     ],
-	  "custom-software": [
-	    "Adobe Account Access"
-	  ],
+    "custom-software": [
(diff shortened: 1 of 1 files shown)


hkamran80:
Can you prove it? Because their documentation says they don't and new accounts weren't able to enable it, which is why #7305 was merged. Also, there was [this post in their forums](https://community.adobe.com/t5/account-payment-plan-discussions/does-anyone-know-how-to-setup-2fa-with-totp-without-secure-signin-app/td-p/12806773):
> Sorry for the late response but I talked to Adobe Support and it seems that 3rd party Authenticator apps cannot be set up to be used anymore. So if anyone wants to use an app they have to use the official Adobe one.

gbdlin:
Looks like it was available for old accounts and it's still available for me, they removed this option.

Sorry for a wrong report.

hkamran80:
No worries! Thanks for contributing!

PR #7453, opened 2023-03-08
Add Oracle NetSuite

Carlgo11:
PR to add the HR/Finance platform NetSuite.

Documentation page mentions that SMS is depreciated so only TOTP is marked as available for the entry.


img/n/netsuite.com.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" style="background:#000;fill:#fff"><path d="M4.45 16.57V3.43h2.61L12.75 12V3.43h2.8v13.14h-2.62L7.25 8v8.57Z"/></svg>
```
hkamran80, line 1:
```suggestion
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="#fff" style="background:#000"><path d="M4.45 16.57V3.43h2.61L12.75 12V3.43h2.8v13.14h-2.62L7.25 8v8.57Z"/></svg>
```

hkamran80 (changes requested):

img/n/netsuite.com.svg
Carlgo11: Whilst non-style elements are nearly always preferred, in this case the background requires the use of `style` and so including fill in the already existing style element actually results in a smaller file than declaring a new element for the color filling.

Good work catching the warning though 👍

hkamran80: Ah, thanks for explaining!

hkamran80 (approved):

PR #7461, opened 2023-03-09
Add SurveyMonkey

hkamran80:


entries/s/surveymonkey.com.json
```diff
@@ -0,0 +1,13 @@
+{
+  "SurveyMonkey": {
+    "domain": "surveymonkey.com",
+    "contact": {
+      "twitter": "SurveyMonkey",
+      "form": "https://help.surveymonkey.com/en/contact/"
+    },
+    "categories": [
+      "marketing",
+      "other"
```
Carlgo11: As the `other` category is only for entries that don't fit in any other category, it's probably best to just use the `marketing` category.
```suggestion
      "marketing"
```

Carlgo11 (changes requested):

Carlgo11 (approved):

PR #7463, opened 2023-03-09
Increase workflow VM cores

Carlgo11:
This PR increases the amount of cores available to the workflow VM from 2 to 8. This should increase the speed of multithreaded tests (i.e. Ruby scripts using the `parallel` gem)


Carlgo11:
Never mind. VMs with more than 2 cores aren't included in our price plan 😒

PR #7486, opened 2023-03-21
Add Integrate.io

dakolech:
Xplenty was rebranded to integrate.io


hkamran80:
Hello @dakolech,

Unfortunately, as this site is below our minimum [Similarweb](https://www.similarweb.com/website/integrate.io/#overview) global rank of 200,000, this site cannot be added. This rule is in place because maintaining the lists takes time, and a list of all websites on the internet just isn't feasible. In the future, please review our [site guidelines](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#guidelines) and [exclusion list](https://github.com/2factorauth/twofactorauth/blob/master/EXCLUSION.md) before adding a site.

msmallcombe (commented):
We recently switched domains, we will resubmit when we meet this threshold - thanks

PR #7496, opened 2023-03-27
Fix Similarweb queries

Carlgo11:
Recently it was discovered that `simlarweb.rb` only checks the root domain and not the full FQDN.
The proposed changes in this PR fix that behavior.

### DIFF
--- a/tests/similarweb.rb
+++ b/tests/similarweb.rb
@@ -49,7 +49,7 @@ diff = `git diff origin/master...HEAD entries/ | sed -n 's/^+.*"domain"[^"]*"\\(
 # Strip and loop through diff
 diff.split("\n").each.with_index do |site, i|
   sleep 2 if i.positive?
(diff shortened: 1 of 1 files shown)


phallobst:
"Discovered"? It was designed that way, because Similarweb does not support subdomains.
```
> curl -s "https://api.similarweb.com/v1/similar-rank/mail.google.com/rank?api_key=KEY"
{"meta":{"status":"Error","error_code":401,"error_message":"Data not found"}}

> curl -s "https://api.similarweb.com/v1/similar-rank/google.com/rank?api_key=KEY"
{"meta":{"request":{"domain":"google.com","limit":10,"format":"json"},"status":"Success","last_updated":"2023-02-28"},"similar_rank":{"rank":1}}
```

Carlgo11:
Oh, I should have tested with a real subdomain before I made the PR :facepalm:

PR #7497, opened 2023-03-27
Fix CircleCI Bundler cache

Carlgo11:
See in-line comments for a changelog.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
       - run:
           name: Bundler
           command: |
-            bundle install && bundle cache
+            bundle cache
 
       - run:
           name: Check Twitter handles
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,11 @@ jobs:
 
     docker:
       - image: ruby:latest
+    environment:
+      BUNDLE_PATH: '/tmp/bundler/vendor'
+      BUNDLE_CACHE_PATH: '/tmp/bundler/cache'
+      BUNDLE_WITH: 'tests'
+
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -18,9 +23,7 @@ jobs:
       - run:
           name: Bundler
           command: |
-            bundle config set --local path '/tmp/bundler'
-            bundle config set --local with 'tests'
-            bundle install
```
Carlgo11, line 23: Cleanup of commands to run

.circleci/config.yml
```diff
@@ -18,9 +23,7 @@ jobs:
...
+            bundle install && bundle cache
```
Carlgo11: Force caching of gems

.circleci/config.yml
```diff
@@ -37,12 +40,12 @@ jobs:
       - save_cache:
           key: similarweb-cache-{{ epoch }}
           paths:
-            - "/tmp/similarweb/*"
+            - "/tmp/similarweb"
 
       - save_cache:
-          key: bundler-cache-{{ epoch }}
+          key: bundler-cache-{{ checksum "Gemfile.lock" }}
```
Carlgo11, line 46: Only cache if Gemfile.lock changes (= updated Gems)

.circleci/config.yml
hkamran80: Why are both commands run? The [`bundle cache` docs](https://bundler.io/v2.4/man/bundle-cache.1.html#REMOTE-FETCHING) say:
> By default, bundle cache fetches and also installs the gems to the default location.

hkamran80 (approved):

PR #7512, opened 2023-04-05
Add raindrop.io

chdleoj:
Raindrop.io now supports 2FA.

PR #7513, opened 2023-04-05
raindrop.io icon

chdleoj:
Adding raindro.io svg icon.


hkamran80:
Please add the SVG directly to #7512

PR #7543, opened 2023-04-23
Add myRMV

lazytownfan:
Add myRMV for the RMV of the state of Massachusetts

## Note on image

[This](https://atlas-myrmv.massdot.state.ma.us/myrmv/Image/ENG/MR.RMV_logo_small_white.png) was the best PNG I could find from the myRMV site (after making the image square, resizing it down to 128x128 in GIMP, and running it through the TinyPNG site).

However, I prefer this [one](https://i2.wp.com/www.mass.gov/files/2017-08/rmv%20logo.png), as the logo colors have white text on darker backgrounds and has much better background contrast, but "Registry of Motor Vehicles" is spelled out on the bottom.

(I also haven't figured out how to convert PNG images into SVG yet.)


hkamran80:
If you want to use the second image, you could crop it to be just "RMV" then resize it.

lazytownfan:
I reuploaded the image and [this](https://imgur.com/a/QTu5c9l) is the largest PNG I have (before shrinking the image) of the RMV logo without the words.

Carlgo11 (changes requested):
I think the letters need to be white.
myRMV logo

Carlgo11 (approved):

PR #7548, opened 2023-04-25
Added descriptions for VPN Providers and Crowdfunding

soodaayush:


Carlgo11:
Do these categories really need an explanation?

soodaayush:
I filled in a description because it seemed inconsistent that other categories had a description but not those two.

PR #7549, opened 2023-04-26
Move workflow secrets

Carlgo11:
Moved the workflow secrets to a separate [environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) for better security.

### DIFF
--- a/.github/workflows/repository.yml
+++ b/.github/workflows/repository.yml
@@ -56,6 +56,7 @@ jobs:
     permissions:
       pages: write
       id-token: write
+    environment: production
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


hkamran80 (approved):

.github/workflows/repository.yml
```diff
@@ -75,17 +76,17 @@ jobs:
         run: bundle exec ruby ./scripts/Algolia.rb
         env:
           ALGOLIA_APP_ID: ${{ secrets.ALGOLIA_APP_ID }}
-          ALGOLIA_INDEX_NAME: ${{ secrets.ALGOLIA_INDEX_NAME }}
+          ALGOLIA_INDEX_NAME : ${{ vars.ALGOLIA_INDEX_NAME }}
```
hkamran80, line 79: Is there a reason this was a changed from `secrets` to `vars` when everything else remained the same?

Carlgo11: I didn't really see a reason for encrypting this variable as it's publicly available already.

PR #7551, opened 2023-04-28
Add Placeit

soodaayush:


hkamran80 (changes requested):
Placeit seems to use Envato's authentication backend based on the "Log in with Envato" label. In that case, I would recommend adding this as an additional domain to [the Envato entry](https://github.com/2factorauth/twofactorauth/blob/master/entries/e/envato.com.json).

Also, per [Apple's shared credentials list](https://github.com/apple/password-manager-resources/blob/main/quirks/shared-credentials.json#L3-L14), it's a shared backend.

soodaayush:
Changes made.

hkamran80 (approved):

PR #7555, opened 2023-04-28
Add Betterment to the banking category (Betterment Cash management)

boconnell:
Betterment has a cash management product


Carlgo11 (approved):
Thank you! 🙂

PR #7574, opened 2023-05-07
Update USPS

hkamran80:
The USPS is an independent executive agency of the federal government


Carlgo11 (commented):
Whilst I understand that USPS is technically a government agency, I don't think it's a service that would commonly and primarily be regarded as a government agency.
Just like we don't list universities, public transit, or infrastructure companies in the _Government_ category, I don't think USPS should be listed there either.

I am however not a U.S. resident and might lack some cultural understanding that could be crucial in determining if USPS should be listed in the _Government_ category. Therefore, if any other U.S. member of 2factorauth would like to review this PR, it would be much appreciated.

PR #7576, opened 2023-05-07
Add TickPick

hkamran80:


entries/t/tickpick.com.json
```diff
@@ -0,0 +1,13 @@
...
+  "TickPick": {
+    "domain": "tickpick.com",
+    "contact": {
+      "twitter": "TickPick",
+      "facebook": "TickPick",
+      "form": "https://support.tickpick.com/hc/en-us/requests/new"
+    },
+    "categories": [
+      "tickets"
+    ]
```
Carlgo11, line 16: Seems like they only sell tickets for events in the US and UK.

hkamran80: [For sellers](https://support.tickpick.com/hc/en-us/articles/360007566254):
> Please note, we only support payouts to residents within the United States (excluding US territories) with United States bank accounts and Canadian residents with Canadian bank accounts.

As far as I know, anyone outside the U.S. and Canada can purchase tickets for these events too, although [all prices are in U.S. dollars](https://support.tickpick.com/hc/en-us/articles/360007554694).

Carlgo11: That's true but the target audience would be residents of the US, UK (and maybe Canada?)
Many sites with region tags are technically accessible from outside the listed regions but don't benefit users outside the listed regions.

For example, [clippercard.com](https://www.clippercard.com/) is available worldwide but is of little interest outside the U.S.

hkamran80: I'll update the regions to include the U.S., Canada, and the UK

Carlgo11 (approved):
Thanks! :smiley:

hkamran80:
Thank you!

PR #7582, opened 2023-05-10
Add Kajabi

Carlgo11:
Kajabi appears to be an e-commerce platform specifically designed for selling access to online courses and podcasts. I don't think it fits `retail`, `education` or `development` so I put it in the `other` category.

### DIFF
--- a/entries/k/kajabi.com.json
+++ b/entries/k/kajabi.com.json
@@ -1,0 +1,12 @@
+{
+  "Kajabi": {
+    "domain": "kajabi.com",
+    "tfa": [
+      "totp"
+    ],
+    "documentation": "https://help.kajabi.com/hc/en-us/articles/8293506425115",
+    "categories": [
+      "other"
+    ]
+  }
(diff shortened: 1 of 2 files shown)


entries/k/kajabi.com.json
```diff
@@ -0,0 +1,12 @@
+{
+  "Kajabi": {
+    "domain": "kajabi.com",
+    "tfa": [
+      "totp"
+    ],
+    "documentation": "https://help.kajabi.com/hc/en-us/articles/8293506425115",
+    "categories": [
+      "other"
```
hkamran80, line 9: Could this be classified as `retail` and `marketing`?

Carlgo11: A service for retailers isn't really a retail service in and of itself.
I don't see that they offer marketing or analytics services either.

It's more like Shopify which is already in the `other` category.

hkamran80 (approved):

PR #7593, opened 2023-05-15
Update Home Depot

lazytownfan:
* Add call & e-mail as 2FA methods for Home Depot
* Discovered this after experimenting with new account

### DIFF
--- a/entries/h/homedepot.com.json
+++ b/entries/h/homedepot.com.json
@@ -1,7 +1,6 @@
 {
   "Home Depot": {
     "domain": "homedepot.com",
-    "url": "https://www.homedepot.com",
     "tfa": [
       "sms",
       "call",
--- a/entries/h/homedepot.com.json
+++ b/entries/h/homedepot.com.json
@@ -1,11 +1,10 @@
 {
   "Home Depot": {
     "domain": "homedepot.com",
-    "tfa": [
-      "sms",
-      "call",
-      "email"
-    ],
+    "contact": {
+      "facebook": "homedepot",
+      "twitter": "HDCares"
+    },
     "regions": [
       "us",
       "ca",
--- a/entries/h/homedepot.com.json
+++ b/entries/h/homedepot.com.json
@@ -3,7 +3,9 @@
     "domain": "homedepot.com",
     "url": "https://www.homedepot.com",
     "tfa": [
-      "sms"
+      "sms",
+      "call",
+      "email"
     ],
     "regions": [
       "us",


entries/h/homedepot.com.json
```diff
@@ -3,7 +3,9 @@
     "domain": "homedepot.com",
     "url": "https://www.homedepot.com",
```
hkamran80:
```suggestion
```

entries/h/homedepot.com.json
```diff
@@ -3,7 +3,9 @@
...
     "tfa": [
-      "sms"
+      "sms",
+      "call",
+      "email"
```
hkamran80: Can you provide a screenshot of the activation page? Please blur any personal information.

hkamran80 (changes requested):

entries/h/homedepot.com.json
lazytownfan: * Done (in another commit - I should've just done it here)
* I didn't create the Home Depot entry, just using what was already here

hkamran80: Thanks!

lazytownfan:
@hkamran80 Could you use the "update site" label, instead of "add site"?

I'm only editing an existing entry (and not submitting a new entry).

hkamran80:
> @hkamran80 Could you use the "update site" label, instead of "add site"?
> 
> I'm only editing an existing entry (and not submitting a new entry).

Thanks for catching that!

entries/h/homedepot.com.json
lazytownfan: Here's a screenshot:

upload

(Fun fact: Signal Desktop has some more marker thickness controls for editing photos than Signal Android; not sure how Signal iOS is like.)

Also, I used manual redaction, as the [warning](https://www.privacyguides.org/en/data-redaction/#privacyblur) for PrivacyBlur on Privacy Guides indicates that it's possible to recover text that has been blurred.

So, drawing opaque rectangles (as [implied](https://nitter.1d4.us/Snowden/status/1423018727361060865#m) by Snowden on Twitter) or opaque redaction (such as the old-school way of redacting paper copies of classified documents) is the safer way of redacting text.

hkamran80: Is there another page that shows that phone calls and emails act as two-factor authentication?

lazytownfan: So I had to turn off 2FA in order to see this again, which Home Depot calls "logging in without a password".  It's sort of laconic and I don't think Home Depot has any support page explaining what's going on.

This is what the page looks like before turning on Home Depot's 2FA, which is called "Sign In with a Password":
upload1

When you select to activate 2FA, Home Depot calls this "Sign In without a Password":
upload2

Although Home Depot requires a cell phone number to create an account, which will be verified with a code in an SMS message (or a SMS capable VOIP number); Home Depot will default to sending the 2FA code for initial setup and all subsequent logins via e-mail:
upload3

This is what the page will look like upon successful activation of 2FA:
upload4

hkamran80: What their sign in without password describes is not 2FA.
> Instead of entering a password you may sign into your account using a one-time passcode that we will send to you.

What happens if you toggle the 2FA switch?

Carlgo11:
> **Warning**
> Unless there's activity within the next 7 days I'll close the PR.

Carlgo11:
Closing due to inactivity.
> **Note**

entries/h/homedepot.com.json
lazytownfan: This is the login screen after turning on Home Depot's 2FA switch:
upload cleaned

You only need to enter the username (i.e., your e-mail) and then you receive the code from Home Depot.  Usually the site defaults to e-mail, but you can manually choose a SMS text or automated voice call to your phone number.

The site does not request for your password when using this method.

lazytownfan:
Could this be opened again?

I was inactive because I got busy at work.

entries/h/homedepot.com.json
Carlgo11: @lazytownfan if a password isn't required then it's still just 1 factor.

lazytownfan: So, would this not fall under 2FA Directory's definition of 2FA?

Carlgo11: Exactly

hkamran80 (approved):

PR #7599, opened 2023-05-20
Add AU Telcos

fletcherdodd:
-Adding Amaysim, Belong, More & Woolworths Mobile telcos

-Updating Optus with MFA doco


Carlgo11:
I think the SimilarWeb ranking requirement can be overlooked in this instance as 217K is very close to the limit.

entries/m/mobile.woolworths.com.au.json
```diff
@@ -0,0 +1,17 @@
+{
+  "Woolworths Mobile": {
```
hkamran80, line 2: Are the Woolworths and Woolworths Mobile authentication backends different?

fletcherdodd: Yeah, it doesn't seem like I can log into Woolworths Mobile with my Woolworths (supermarket) account. The only way the two are linked is through the [rewards program](https://www.woolworthsrewards.com.au/).

Carlgo11 (approved):

PR #7608, opened 2023-05-25
Update MAX

fletcherdodd:
Adjusting brand name and adding regions as per [their central site](https://www.maxburgers.com/). The brand name isn't just MAX as per [Wikipedia](https://en.wikipedia.org/wiki/Max_Hamburgers) and also most locations include hamburger/burger in their brand name. Just went with 'Burgers' as their central site was named that.

Resolves issue in #7607

PR #7609, opened 2023-05-25
Remove name duplication test

Carlgo11:
PR addressing #7607 and #7608.

As the duplicate name test doesn't take regional sites into account, I think it's worth removing for the time being.

### DIFF
--- a/tests/validate-json.rb
+++ b/tests/validate-json.rb
@@ -4,7 +4,6 @@
 require 'json_schemer'
 require 'parallel'
 @status = 0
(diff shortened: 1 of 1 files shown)


phallobst:
API v1 and v2 use the name as the key in the dictionary. These may break if there are duplicate names/keys.

I think in general, the name should be unique. For other entries, we add `[UK]` or something to distinguish them. Dropping this requirement for a rather low number of potential conflicts seems unnecessary to me.

PR #7624, opened 2023-06-02
Update Experian

aarontorres0:


phallobst:
Was this supposed to go into the `Equifax` entry?

aarontorres0:
I tend to get both of these companies mixed up and did it here too. Should be the domain for Experian.

Carlgo11 (approved):

Carlgo11:
@aarontorres0 can you confirm that 2FA is supported on Experian.com?

aarontorres0:
Yes, I can confirm. I just enabled it yesterday and my options were to receive a text message or phone call.

PR #7646, opened 2023-06-08
Remove URLs

fletcherdodd:
Removing URLs where the domain is the same.

I was going to do heaps but not sure on/if there is a limit to how any site edits I could make

### DIFF
--- a/entries/a/aetna.com.json
+++ b/entries/a/aetna.com.json
@@ -1,7 +1,6 @@
 {
   "Aetna": {
     "domain": "aetna.com",
-    "url": "https://www.aetna.com/",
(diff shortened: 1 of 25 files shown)


phallobst (approved):

entries/a/aetna.com.json
```diff
@@ -1,7 +1,6 @@
 {
   "Aetna": {
     "domain": "aetna.com",
-    "url": "https://www.aetna.com/",
```
Carlgo11, line 4: `https://aetna.com` redirects to `http://www.aetna.com` so the URL tag should stay.

PR #7654, opened 2023-06-11
Add EU Login

hkamran80:


Carlgo11:
> EU Login is the European Commission's user authentication service. It allows authorised users to access a wide range of Commission web services

Maybe the name "European Commission" would better describe the service?

Carlgo11 (approved):

PR #7670, opened 2023-06-28
Fixed multiple doc urls that returned 404 status

40504:
Also, added recovery documentation for BlockFi

### DIFF
--- a/entries/b/blockfi.com.json
+++ b/entries/b/blockfi.com.json
@@ -5,6 +5,7 @@
       "totp"
     ],
     "documentation": "https://help.blockfi.com/hc/en-us/articles/360049343611",
+    "recovery": "https://help.blockfi.com/hc/en-us/articles/360048862652",
     "categories": [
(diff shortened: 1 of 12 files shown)


entries/b/blockfi.com.json
```diff
@@ -5,6 +5,7 @@
       "totp"
     ],
     "documentation": "https://help.blockfi.com/hc/en-us/articles/360049343611",
+    "recovery": "https://help.blockfi.com/hc/en-us/articles/360048862652",
```
hkamran80: I'm getting a page that says that their help center no longer exists.

hkamran80 (changes requested):

entries/b/blockfi.com.json
40504: The same for me, but the page was available when I opened the PR. For now, I will remove both the doc and rec URLs.

hkamran80 (approved):

PR #7681, opened 2023-07-18
Update IONOS

bdonvr:
I had initially looked for this under "hosting" and didn't find it. 

They have some odd region restrictions. They don't list these anywhere so the regions I excluded aren't exhaustive.


Carlgo11:
From looking at their websites, only these countries seem to be their target demographic.
Maybe list those regions? Also, your PR contains invalid JSON that needs to be fixed before it can be merged.
screenshot

bdonvr:
> From looking at their websites, only these countries seem to be their target demographic. Maybe list those regions? Also, your PR contains invalid JSON that needs to be fixed before it can be merged. screenshot

Okay so their region support is messy to say the least. I found they have other domains such as ionos.hu - and if you go to the "regions" link there, it shows some other regions not listed at ionos.com! Additionally there's https://www.ionos.com/en-au/servers/server-hosting-australia but no ionos.com.au. Similar for South Africa. 

In any case I've made the best list I can. But it probably isn't exhaustive.

hkamran80:
Can you log into the `ionos.hu` portal with credentials from `ionos.com`? If so, it might be worth it to add `ionos.hu` as an additional domain.

bdonvr:
> Can you log into the `ionos.hu` portal with credentials from `ionos.com`? If so, it might be worth it to add `ionos.hu` as an additional domain.

I can't, nor .co.uk or .cz. 

The .co.uk site seems to have basically the same setup for 2FA, I can't read the other sites so I can't confirm.

Carlgo11 (approved):

PR #7682, opened 2023-07-19
Add Bluesky

eternalyperplxed:


hkamran80:
Given that the default Bluesky provider is invite-only at the moment, should it be listed now or should we wait for general availability?

eternalyperplxed:
Fair point, I've seen it being talked about frequently in the public discourse, but it may not have the user base to warrant listing.

hkamran80:
As have I. I would recommend closing this PR for now, then reopen it when it goes GA.

eternalyperplxed:
Sounds good, appreciate the feedback.

PR #7694, opened 2023-07-31
Added various additional domains that Schibsted owns

DandelionSprout:
They own quite a whole lot of sites, especially in Norway and Sweden, so I attempted to update the file to reflect that.


hkamran80:
Do they all have the same authentication backend, that is, can you use the same account on all of those sites?

DandelionSprout:
Correct.

Carlgo11 (approved):

PR #7722, opened 2023-08-22
Update Pixiv

sckemu:
I put a link to the [help page(available in English and Japanese)](https://www.pixiv.help/hc/en-us/articles/20876082054553-What-is-two-factor-authentication-) in the "documentation", but the [new feature announcement page(only Japanese)](https://www.pixiv.net/info.php?id=10043) has more details.
Which is desirable?

### DIFF
--- a/entries/p/pixiv.net.json
+++ b/entries/p/pixiv.net.json
@@ -1,10 +1,6 @@
 {
   "Pixiv": {
     "domain": "pixiv.net",
-    "contact": {
-      "twitter": "pixiv_en",
-      "email": "support@pixiv.net"
-    },
     "tfa": [
       "totp"
     ],
--- a/entries/p/pixiv.net.json
+++ b/entries/p/pixiv.net.json
@@ -4,7 +4,7 @@
     "tfa": [
       "totp"
     ],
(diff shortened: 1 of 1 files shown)


entries/p/pixiv.net.json
```diff
@@ -5,6 +5,10 @@
       "twitter": "pixiv_en",
       "email": "support@pixiv.net"
     },
+    "tfa": [
+      "totp"
+    ],
+    "documentation": "https://www.pixiv.help/hc/en-us/articles/20876082054553-What-is-two-factor-authentication-",
```
Carlgo11: Just like @fletcherdodd commented (although I can't seem to find that comment now?) the part to the right of the numbers isn't needed and can be removed.
```suggestion
    "documentation": "https://www.pixiv.help/hc/en-us/articles/20876082054553",
```

Carlgo11 (changes requested):
Thanks for the pull request @sckemu,
please remove the `contact` part as that's not needed when the service supports 2FA/MFA.

sckemu:
Thanks! @fletcherdodd @Carlgo11

Carlgo11 (approved):
Thank you! 🙂

PR #7735, opened 2023-09-03
Add oikotie.fi and tori.fi to Schibsted. Remove Tori as a separate entry.

momentti:
Both sites use Schibsted account to log in.


phallobst:
Since `Schibsted` has no documentation link, but `Tori` has (or had), I wonder if we should include that url here. It describes the common login backend, after all.

momentti:
I added FAQ page as documentation. It used to be behind login, but not anymore.

entries/s/schibsted.com.json
```diff
@@ -24,12 +24,15 @@
...
-      "tek.no"
+      "tek.no",
+      "oikotie.fi",
+      "tori.fi"
     ],
     "tfa": [
       "sms",
       "totp"
     ],
+    "documentation": "https://login.schibsted.com/profile-pages/faq",
```
Carlgo11: The entire page is rather long and it is primarily this part that's of interest.
```suggestion
    "documentation": "https://login.schibsted.com/profile-pages/faq#how-to-activate-and-deactivate-2fa",
```

Carlgo11 (changes requested):

Carlgo11 (approved):
Thank you! 🙂

PR #7750, opened 2023-09-06
Update Airtable

40504:


entries/a/airtable.com.json
```diff
@@ -6,7 +6,8 @@
       "call",
       "totp"
     ],
-    "documentation": "https://support.airtable.com/hc/en-us/articles/219409917",
+    "documentation": "https://support.airtable.com/docs/enabling-two-factor-authentication",
+    "recovery": "https://support.airtable.com/v1/docs/contacting-airtable-support",
```
hkamran80: Maybe link to the FAQ section of the 2FA documentation instead, since it provides some context?

hkamran80 (changes requested):

entries/a/airtable.com.json
40504: Could you suggest the URL? :)

hkamran80:
```suggestion
    "recovery": "https://support.airtable.com/docs/enabling-two-factor-authentication#faqs",
```

hkamran80 (changes requested):

Carlgo11 (approved):

phallobst:
Do we generally want to have the `documentation` and `recovery` urls point to the same page, even if anchors are used?

hkamran80:
I think we should have the URLs point to the same page if both pieces of information are present on that page. [WeTransfer's entry](https://github.com/2factorauth/twofactorauth/blob/master/entries/w/wetransfer.com.json) has the same URL with anchors because they have the information on one page.

hkamran80 (approved):

PR #7759, opened 2023-09-09
Add Bluesky

Carlgo11:


hkamran80:
Given that the main server is still invite-only, should we add it?

Carlgo11:
@hkamran80 I think so. It's not super hard to get an invite. Even I have an account there :sweat_smile:

hkamran80 (approved):

PR #7792, opened 2023-10-03
Update Shopify

santibarclay:
Shopify is an e-commerce platform


phallobst:
We had _Shopify_ listed as `retail` before, but changed it to `other` - probably because it's not a retailer itself. (_BigCommerce_ and _ThriveCart_ are in the same category as well.)

PR #7814, opened 2023-10-17
Update letterboxd

sn0opy:
[letterboxd](https://letterboxd.com/) now supports 2FA via OTP.

The changelog of the iOS app even mentiones "Apple Keychain". I assume this is an iOS-only feature so I didn't add it here.

### DIFF
--- a/entries/l/letterboxd.com.json
+++ b/entries/l/letterboxd.com.json
@@ -6,6 +6,9 @@
 			"facebook": "letterboxd",
 			"twitter": "letterboxd"
 		},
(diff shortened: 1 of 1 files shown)


fletcherdodd:
FYI - Keychain is the password, MFA and passkey manager for Apple devices. So you're correct, no need to add 🤠

hkamran80 (approved):

hkamran80:
Thank you! Your contribution will be visible in around 10 minutes, although it may take longer due to server and browser caches.

PR #7822, opened 2023-11-01
Update Chase

nsbuman:
Added custom hardware

additional documentation for security token

https://www.chase.com/content/dam/chaseonline/en/legacy/online/Help/jpmol_hwtf_logon_token.htm

### DIFF
--- a/entries/c/chase.com.json
+++ b/entries/c/chase.com.json
@@ -4,10 +4,14 @@
     "tfa": [
       "sms",
       "call",
-      "custom-software"
+      "custom-software",
+      "custom-hardware"
     ],
     "custom-software": [
       "Chase app"
+    ],
+    "custom-hardware": [
+      "Security Token"
     ],
     "documentation": "https://2fa.directory/notes/chase/",
     "categories": [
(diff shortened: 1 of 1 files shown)


Carlgo11:
Thanks for the pull request, @nsbuman!
It sounds like Chase is using Symantec VIP Hardware Authenticators (see image below). Would you happen to know if that's the case? If so, please change the name of the custom hardware to `Symantec VIP`.
image

A note would also be good in this case. Perhaps something like "Hardware tokens can be obtained through Chase support for high-risk accounts."

nsbuman:
Hi Carlgo, from what I was able to find online it looks like Chase uses a branded RSA SecurID device
Img_2023_11_01_14_27_45

I'll update the custom hardware to 'RSA SecurID' and use your suggested note

Thanks for the feedback

Carlgo11 (approved):
Thank you! 🙂
Your contribution will be visible in around 10 minutes. Sometimes, it takes longer due to [server and browser caching](https://github.com/2factorauth/twofactorauth/wiki/FAQ-Caching).

PR #7828, opened 2023-11-08
Update Canada Post

thomassth:


Carlgo11 (commented):
What's described sounds more like authorization for specific user actions than authentication at the login phase.
We distinguish between the two in the section about authorization in [Contributing.md](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#authorization).
Please confirm if this is the case or if 2FA is also prompted on login.

thomassth:
Can confirm the 2FA only appears “when needed”

PR #7851, opened 2023-11-27
Add Fakturoid

marketauml:

PR #7850, opened 2023-11-27
Create fakturoid.cz.svg icon

marketauml:

PR #7852, opened 2023-11-29
Add Fakturoid

marketauml:


hkamran80:
As noted in #7850 and #7851, please add the SVG directly to #7851.

PR #7862, opened 2023-12-12
Clean up SVG code

Carlgo11:
Minor cleanup of SVG files because I had nothing better to do.


kmpoppe (approved):
Superb work, as always l.

Issue #7869, opened 2023-12-18
Update University of Colorado Boulder


person1234-ux:
### Site name

University of Colorado Boulder

### Site URL

https://canvas.colorado.edu/

### Update reason

Other (please describe below).

### Additional information

univeristy is going to implement 2fa via duo on January 9th so I believe its appropriate to remove the links to their twitter and email in the mean while
[
<img width="991" alt="Screenshot 2023-12-17 at 7 09 23 PM" src="https://github.com/2factorauth/twofactorauth/assets/76968588/b1d00400-c7b5-4951-ad08-74fc335cfbad">
](url)

https://oit.colorado.edu/services/identity-access-management/duo-multi-factor-authentication#useduo
Documentation of what methods will be supported on January 9th. 
### Issue Eligibility

- [X] The issue I'm creating is not a duplicate of an existing [issue](https://github.com/2factorauth/twofactorauth/issues).
- [X] The issue I'm creating is not a duplicate of an existing [pull request](https://github.com/2factorauth/twofactorauth/pulls)

noahggarrett-1:
https://oit.colorado.edu/services/identity-access-management/duo-multi-factor-authentication#useduo link to documentation of what methods will be supported starting January 9th.

PR #7870, opened 2023-12-18
Update CU Boulder

person1234-ux:
see issue #7869 for why


hkamran80 (changes requested):
Please add links to documentation and the supported MFA methods if you're removing the `contact` object.

PR #7887, opened 2024-01-18
Updated Kungliga Tekniska Högskolan image

soodaayush:


hkamran80:
The unminified SVG is 52 KB, the minified SVG is 38 KB, and the current PNG is 1.6 KB. Because of the differences in file size between the SVGs and the PNG, I think it is best to keep the PNG at this time. If you can optimize the SVG further, please either reopen this pull request or create a new one.

Thank you for contributing to the 2FA Directory.

PR #7915, opened 2024-02-08
Update Deutsche Telekom

arxo:
Setting up an authenticator app unfortunately requires providing a mobile phone number as fallback.

### DIFF
--- a/entries/t/telekom.de.json
+++ b/entries/t/telekom.de.json
@@ -9,7 +9,7 @@
     "categories": [
       "utilities"
     ],
-    "notes": "Mobile phone number required as fallback when enabling authenticator app.",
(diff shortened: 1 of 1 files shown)


entries/t/telekom.de.json
```diff
@@ -9,6 +9,7 @@
     "categories": [
       "utilities"
     ],
+    "notes": "Mobile phone number required as fallback when enabling authenticator app.",
```
Carlgo11: I suggest this rephrasing as it will better align with the notes for sites with similar requirements.
```suggestion
    "notes": "Software 2FA requires setting up SMS 2FA as a fallback.",
```

Carlgo11 (changes requested):

Carlgo11 (approved):

PR #7919, opened 2024-02-10
Update uber.com.json

RedRubble:
Uber now appears to have TOTP available. See attached images.
Uber 2FA1
Uber 2FA2


RedRubble:
Just saw the previous issue.

PR #7935, opened 2024-03-01
Update Spaceship

fletcherdodd:
Not sure if it's too early since MFA is currently being rolled out.

### DIFF
--- a/entries/s/spaceship.com.au.json
+++ b/entries/s/spaceship.com.au.json
@@ -5,7 +5,11 @@
 			"spaceshipinvest.com.au"
 		],
 		"tfa": [
-			"sms"
+			"sms",
+			"custom-software",
+		],
+		"custom-software": [
+			"WhatsApp"
 		],
 		"documentation": "https://help.spaceship.com.au/en/articles/8695426",
 		"categories": [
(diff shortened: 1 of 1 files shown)


entries/s/spaceship.com.au.json
```diff
@@ -4,11 +4,10 @@
 		"additional-domains": [
 			"spaceshipinvest.com.au"
 		],
-		"contact": {
-			"facebook": "SpaceshipAU",
-			"twitter": "SpaceshipAU",
-			"email": "hello@spaceship.com.au"
-		},
+		"tfa": [
+			"sms"
```
hkamran80:
```suggestion
			"sms",
			"custom-software",
		],
		"custom-software": [
			"WhatsApp"
```

hkamran80 (changes requested):

entries/s/spaceship.com.au.json
fletcherdodd: Wasn't sure whether or not it counted as a different method to SMS so I left it out.

hkamran80 (approved):

entries/s/spaceship.com.au.json
hkamran80: I think it's definitely a different method.

hkamran80 (approved):

PR #7956, opened 2024-03-16
Add Resend

hkamran80:
There is no documentation that I can find, but they support TOTP.

Resend - MFA


entries/r/resend.com.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Resend": {
+    "domain": "resend.com",
+    "tfa": [
+      "totp"
+    ],
+    "categories": [
+      "communication"
```
Carlgo11: Is it not more of a developer product? Or perhaps both categories should be used?

hkamran80: It is a developer product, but I put in communication because tools like SendGrid are in there. I'll add the developer category.

Carlgo11 (approved):

PR #8012, opened 2024-05-02
Add Spaceship

fletcherdodd:
No public documentation exists for the 2FA.

Proof of the two methods:
spaceship2fa


Carlgo11:
@fletcherdodd I think you need to update your branch before the test can pass.

hkamran80 (approved):

PR #8014, opened 2024-05-02
Remove Yammer

fletcherdodd:


Carlgo11 (commented):
I'm not sure if listing Viva Engage is the best thing. As it's a part of the Microsoft 365 suite it might be better to remove Yammer completely as Microsoft 365 is already listed here.

fletcherdodd:
> I'm not sure if listing Viva Engage is the best thing. As it's a part of the Microsoft 365 suite it might be better to remove Yammer completely as Microsoft 365 is already listed here.

I definitely agree, I thought about doing this but opted to updated it instead. Will remove it shortly.

hkamran80:
Should `yammer.com` be added to the Microsoft 365 entry as an additional domain?

fletcherdodd:
> Should `yammer.com` be added to the Microsoft 365 entry as an additional domain?

Good thinking! Done.

entries/o/office.com.json
```diff
@@ -5,7 +5,9 @@
       "login.microsoftonline.com",
       "office.microsoft.com",
       "microsoft365.com",
-      "office365.com"
+      "office365.com",
+      "cloud.micrsoft",
```
hkamran80:
```suggestion
      "cloud.microsoft",
```

hkamran80 (changes requested):

hkamran80 (approved):

PR #8022, opened 2024-05-06
Update Swyftx

fletcherdodd:

PR #8023, opened 2024-05-06
Update Swyft logo

Carlgo11:
Continuation of #8022


hkamran80:
Should we use the logo from their website or the favicon?

Carlgo11:
@hkamran80 I call it logos, but I guess the correct terminology for the images we add would be [icons](https://www.namecheap.com/guru-guides/logo-vs-icon-explained/#:~:text=what%20it%20needs.-,What's%20the%20difference%20between%20a%20logo%20and%20an%20icon%3F,full%20logo%20won't%20fit.). The image attached in #8022 was part of their logo (with the name cropped out), and the image here is the icon that they use on Facebook, Twitter, and Reddit, as well as on their website as their favicon. As the logo and icon differ, I think this version should be the preferred one.

hkamran80:
Makes sense to me!

hkamran80 (approved):

PR #8024, opened 2024-05-09
Update Element

fletcherdodd:
Wasn’t sure if I should have included the note about it not applying to accounts on EMS servers as stated in the doco, so I left it out.

### DIFF
--- a/entries/e/element.io.json
+++ b/entries/e/element.io.json
@@ -1,9 +1,10 @@
 {
   "Element": {
     "domain": "element.io",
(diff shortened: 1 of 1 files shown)


Carlgo11:
Can you confirm that TOTP is used? The documentation doesn't mention any specific 2FA methods.

fletcherdodd:
I made a bold assumption it's totp based on the step below:

image

Carlgo11 (approved):

PR #8030, opened 2024-05-17
Remove Carbon Black Community Exchange

fletcherdodd:


Carlgo11 (changes requested):
Reviewing the site, It looks to just be some sort of forum and thus don't think community.broadcom.com meets our [site criteria](https://github.com/2factorauth/twofactorauth/blob/master/EXCLUSION.md#forums)

I'm not really sure how Carbon Black Community Exchange managed to be added in the first place, but I think it's better to just delete it completely now.

fletcherdodd:
> Reviewing the site, It looks to just be some sort of forum and thus don't think community.broadcom.com meets our [site criteria](https://github.com/2factorauth/twofactorauth/blob/master/EXCLUSION.md#forums)
> 
> I'm not really sure how Carbon Black Community Exchange managed to be added in the first place, but I think it's better to just delete it completely now.

Fair enough, I have updated the PR :)

Carlgo11 (approved):

PR #8034, opened 2024-05-18
Update Quora

fletcherdodd:
No doco on the setup, see below

quora2fa


hkamran80:
Is that on every login or only on unrecognized logins?

fletcherdodd:
> Is that on every login or only on unrecognized logins?

Seems to be every login, I turned the setting on and then logged out and back in and got that message

hkamran80 (approved):

PR #8055, opened 2024-05-26
Create planetminecraft.com.json

Machou:


fletcherdodd:
@Carlgo11 I would almost argue that this site falls within the forum category and thus falls outside of the criteria.

Carlgo11:
Indeed. Closing as it doesn't meet the site criteria.

PR #8060, opened 2024-05-31
Add Energy Australia

fletcherdodd:


entries/e/energyaustralia.com.au.json
```diff
@@ -0,0 +1,12 @@
+{
+  "Energy Australia": {
+    "domain": "energyaustralia.com.au",
+    "tfa": [
+      "email"
+    ],
+    "documentation": "https://www.energyaustralia.com.au/home/help-and-support/faqs/multi-factor-authentication",
+    "categories": [
+      "utilities"
+    ]
```
hkamran80, line 14:
```suggestion
    ],
    "regions": [
      "au"
    ]
```

hkamran80 (changes requested):
I can't find an actual SVG of their logo, so maybe extract the PNG from the SVG and just add that?

fletcherdodd:
> I can't find an actual SVG of their logo, so maybe extract the PNG from the SVG and just add that?

I removed the reference image in Illustrator and created a new path same as the image it was referencing. Everytime I export the SVG add adds the image reference back. So I'll export it as a PNG

Carlgo11 (changes requested):
6KiB is a bit much for an icon :grimacing:
Try this 64x64px icon instead:
energyaustralia com au

hkamran80 (approved):

Carlgo11 (approved):

PR #8079, opened 2024-06-06
Add Adelaide Bank

fletcherdodd:


Carlgo11 (approved):
This might be one of the smallest PRs in the history of 2FA Directory 😂

PR #8084, opened 2024-06-06
Add Suncorp Bank & Insurance

fletcherdodd:


Carlgo11:
@hkamran80 What do you think about the category here? :thinking:

hkamran80:
> @hkamran80 What do you think about the category here? 🤔

I'm assuming you're talking about Suncorp Insurance's category. I would recommend the finance category because the existing insurance providers we have are categorized with that and the description says "Financial, insurance and pension services".

Carlgo11 (approved):

PR #8096, opened 2024-06-14
Split GitHub workflows

Carlgo11:
This PR splits the `repository.yml` workflow into separate workflows: one for pull requests containing different tests and one for deployment.

I also started a Node.js job inside the pull requests workflow. My idea is to gradually rewrite all the tests in JavaScript for better performance and easier collaboration as more people know JavaScript than Ruby

I have tested `tests/languages.js` independently but I'm unsure if I've missed something in the workflows.

__Edit:__ Obviously, the workflow wasn't functioning when I made the PR. It should work now. Sorry for the commit spam 😅


hkamran80:
If you want to keep the `Test PR #XXXX` run name, I think it'll work if you wrap it in quotes.
```yaml
run-name: "Test PR #${{ github.event.number }}"
```

Carlgo11:
@hkamran80 You think that will work?
Also, sorry about the commit spam! I tried to do this with the act CLI but it didn't work 🙈

hkamran80:
I think so. The problem with the original method is that it was being treated as a comment. Wrapping it in quotes should prevent YAML from recognizing it as a comment.

No worries about the commits!

Carlgo11:
It did work! :partying_face:

.github/workflows/pull_request.yml
```diff
@@ -1,15 +1,87 @@
...
+
+  external-tests:
+    name: External Tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
       - name: Call PR Validator
         run: |
           curl -s --fail-with-body "https://pr-validator.2factorauth.workers.dev/${{ github.event.repository.name }}/${{ github.event.number }}/" \
```
hkamran80: Should we use the `workers.dev` URL or the subdomain on `2fa.directory`?

Carlgo11: Whoops! Yeah the `2fa.directory` domain is the preferred one because the WAF can be used then. Fixed in d9ec537

hkamran80:
Do you want to merge this PR now and convert the rest of the tests later or convert them all now and merge this after?

I can help with the conversion if you'd like.

Carlgo11:
Let's merge it and then convert more of the tests.

>I can help with the conversion if you'd like.

I'd love some help! :sweat_smile: We can discuss it on Slack.

hkamran80:
Sounds good!

hkamran80 (approved):

Issue #8139, opened 2024-07-29
v3 API URL 'https://api.2fa.directory/v3/tfa.json' returns 404


joshuacant:
- [X] I have searched [open issues and pull requests](https://github.com/2factorauth/twofactorauth/issues?q=is%3Aopen). The issue I'm creating is not a duplicate of an existing open issue or pull request.

### Information about the bug: ###
**Bug description:**
The All 2FA-supporting sites 'https://api.2fa.directory/v3/tfa.json' v3 API url returns a github 404 error document instead of valid JSON

**Bug reproduction:**
Open https://api.2fa.directory/v3/tfa.json

**Impacted devices:**
All

**Expected behavior:**
https://api.2fa.directory/v3/tfa.json returns valid JSON

Carlgo11:
Thanks for the bug report @joshuacant!
_tfa.json_ and _tfa.json.sig_ are now back.

Just out of curiosity, what's your use-case where you need to use _tfa.json_ instead of a narrower output, like _totp.json_ or _u2f.json_?

joshuacant:
Thanks for fixing this so fast!

The question about use-case would have to be answered by @Rookiestyle who develops the KeePassOTP plugin (https://github.com/Rookiestyle/KeePassOTP) for an application called KeePass. I am just an end user of that plugin. It showed me an error message today when using it. I did a little digging and found this 404 was the root cause, so I raised the issue here.

Carlgo11:
Oh, I didn't realize that you were an end-user 😅

@Rookiestyle I went through [your code](https://github.com/Rookiestyle/KeePassOTP/blob/e7cdd931966da84daa3f92672e09dff9b407c975/src/DAO/TFASites.cs) and I don't see any reason why it would need to fetch all 2FA-supporting sites instead of just those with TOTP support. Am I missing something?

Rookiestyle:
@Carlgo11 [This code](https://github.com/Rookiestyle/KeePassOTP/blob/e7cdd931966da84daa3f92672e09dff9b407c975/src/KPOTP_Details.cs) is executed when a KeePass user opens the details of an entry. My plugin checks whether 2FA is possible and shows all the available options to the user.

Additionally, in the entry overview my plugin shows a column with the calculated OTP value or "Setup TFA" if this is possible - based on tfa.json.
[...]

PR #8141, opened 2024-07-30
Add API directory test

Carlgo11:
I made a small test to validate that all the expected API files are present before publishing a build.
That would prevent future #8139-events from happening :sweat_smile:


kmpoppe (commented):
Would it make sense to also check if the files are reachable via https?

Carlgo11:
Yes it would, but I have no idea how that could be achieved as the files are only online after the publication workflow is done 😅

kmpoppe:
Ah i see the problem... Well then, file existence only it is then 😬

kmpoppe (approved):

PR #8150, opened 2024-08-05
Ignore non-SVG images for SVG check

Carlgo11:


tests/svg.js
```diff
@@ -22,6 +22,9 @@ async function main() {
 
   await Promise.allSettled(
     files.map(async (file) => {
+      // Skip SVG checks if the extension isn't svg
+      if (!file.endsWith(".svg")) return null;
```
hkamran80: Do you need to return anything? If not, try just `return;`

Carlgo11: Didn't even know that was possible 😅

kmpoppe (approved):

PR #8173, opened 2024-08-28
Remove Laybuy

fletcherdodd:
Another BNPL service gone into receivership, as per the notice on their website.


hkamran80:
Is Laybuy NZ fully under Klarna or is it still a separate service? We might consider removing the other regions if its authentication system is still separate.

fletcherdodd:
> Is Laybuy NZ fully under Klarna or is it still a separate service? We might consider removing the other regions if its authentication system is still separate.

Good point, didn't see that announcement. I would assume they merge into one authenication system, but I can make changes and remove other regions for now. Or remove it and add it back later if it's not merged into Klarna completely

hkamran80:
Let's remove the other regions for now. If Klarna merges it, which they very likely will, we can remove the entry then.

fletcherdodd:
Welp, it's just out of the minimum site ranking

hkamran80:
Guess it's getting removed then

hkamran80 (approved):

PR #8185, opened 2024-09-03
Update ee

dalDevelo:
Updates ee.co.uk to include 2FA support via SMS, documentation + notes that this is a 1 way operation.

Screenshot 2024-09-03

### DIFF
--- a/entries/e/ee.co.uk.json
+++ b/entries/e/ee.co.uk.json
@@ -4,10 +4,11 @@
     "tfa": [
       "sms"
     ],
+    "documentation": "https://ee.co.uk/help/profile/registration-setup/account-authentication",
+    "notes": "Once enabled you cannot disable it again. You can only change the phone number you receive the code on.",
     "categories": [
       "utilities"
     ],
-    "notes": "Once enabled you cannot disable it again. You can only change the phone number you receive the code on.",
     "regions": [
       "gb"
     ]
--- a/entries/e/ee.co.uk.json
+++ b/entries/e/ee.co.uk.json
@@ -1,13 +1,13 @@
 {
   "EE": {
     "domain": "ee.co.uk",
-    "contact": {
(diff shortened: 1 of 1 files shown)


entries/e/ee.co.uk.json
```diff
@@ -1,10 +1,12 @@
 {
   "EE": {
     "domain": "ee.co.uk",
-    "contact": {
-      "facebook": "ee",
-      "twitter": "EE"
-    },
+    "url": "https://www.ee.co.uk/",
```
Carlgo11: Thanks for the pull request @dalDevelo!
Could you please revert this change?
I'm not sure why the build failed earlier, but it seems to have been due to some random unrelated event, as the URL element shouldn't affect the domain rank.
```suggestion
```

Carlgo11 (changes requested):

dalDevelo:
Done :)

Not sure why this happened either, but it's not the first time. I have had previous commits that failed for same / similar issues and I added URL, then removed it and it passed the 2nd time without it, as it has this time too. Might be something that needs looking at? 🤷🏻‍♂️ I was also confused as when I clicked on the test link the results displayed without issue on the output page.

Anyway, good to know the URL shouldn't be included for future ref, even if I add it to pass the test, as per above.

There are quite a few entries with the URL included, presumably from a time when this was required? If these can be removed I will happily submit updates and PR's for some of those too.

Carlgo11 (approved):

hkamran80:
> There are quite a few entries with the URL included, presumably from a time when this was required? If these can be removed I will happily submit updates and PR's for some of those too.

We currently require the URL if going to `https://[domain]` fails. Feel free to submit PRs to update those that don't need it!

Carlgo11:
What hkamran80 ^

I clearly type too slow

PR #8189, opened 2024-09-05
Add Cisco Network Academy

dalDevelo:
Added Cisco Network Academy Site 'Netacad'. Cisco's education platform.


dalDevelo:
I added this as a separate site instead of adding as an additional domain to the Cisco entry, mainly due to the purpose of the site being different from Cisco's core website [Education and training]. If you'd prefer I edit the Cisco entry instead, I'll go ahead and do that.

*Edit*
 Or add cisco.com as an additional domain to this one, or both? 😅

entries/n/netacad.com.json
```diff
@@ -0,0 +1,16 @@
+{
+  "Cisco Network Academy": {
+    "domain": "netacad.com",
```
Carlgo11, line 3: Instead of having two identical images, it's better to just link to the other image :wink: 
```suggestion
    "domain": "netacad.com",
    "img": "cisco.com.svg",
```

dalDevelo:
I was unsure about that so thanks for the info :)

fletcherdodd (approved):

PR #8191, opened 2024-09-07
Add RACQ

fletcherdodd:


entries/r/racq.com.au.json
```diff
@@ -0,0 +1,15 @@
+{
+  "RACQ": {
+    "domain": "racq.com.au",
+    "contact": {
+      "twitter": "RACQOfficial",
+      "facebook": "racqofficial"
+    },
+    "categories": [
+      "insurance"
```
hkamran80, line 9:
```suggestion
	  "banking",
      "insurance"
```

Looks like they offer [banking services](https://www.racq.com.au/banking) as well

hkamran80 (changes requested):

entries/r/racq.com.au.json
fletcherdodd: That is true, and I was going to make a separate listing for it as it seems to be a different authentication system. There seems to be 2FA available for the banking login, the domain racqbank.com.au doesn't meant the rank criteria though.

hkamran80: Got it.

hkamran80 (approved):

PR #8198, opened 2024-09-11
Update McAfee

fletcherdodd:


hkamran80 (approved):
Why do they only support SMS MFA?????

fletcherdodd:
> Why do they only support SMS MFA?????

Haha I was disappointed and not surprised that it was SMS

PR #8215, opened 2024-10-08
Add SubscribeStar.com

WesVanVugt:


entries/s/subscribestar.com.json
```diff
@@ -0,0 +1,14 @@
+{
+  "Patreon": {
+    "domain": "subscribestar.com",
```
Carlgo11, line 3: If you go to `https://<domain>` and it redirects to the value of the `url`-element then the `url`-element isn't needed.
```suggestion
```

entries/s/subscribestar.com.json
```diff
@@ -0,0 +1,14 @@
+{
+  "Patreon": {
```
Carlgo11: Please change the name to SubscribeStar.com

Carlgo11 (changes requested):

entries/s/subscribestar.com.json
```diff
@@ -0,0 +1,14 @@
...
+    "domain": "subscribestar.com",
+    "url": "https://www.subscribestar.com/",
```
Carlgo11: If you go to `https://<domain>` and it redirects to the value of the `url`-element then the `url`-element isn't needed.
```suggestion
```

Carlgo11 (changes requested):

entries/s/subscribestar.com.json
WesVanVugt: I updated the name to SubscribeStar, since that seemed more in line with the other sites. If I should be including the root domain, let me know and I'll add it.

Carlgo11: Good thinking!
That would normally be the preferred option, however as the name on the website is "SubscribeStar.com", it's better to stick with their preferred brand name.

When it comes to branding and trademarks it's always better to just go with what the company does. Makes for less legal trouble 😅

WesVanVugt: I have corrected the name now, and I'll try to follow this rule for future PRs as well.

Carlgo11 (approved):
Thank you! 🙂
Your contribution will be visible in around 10 minutes. Sometimes, it takes longer to show up due to [server and browser caching](https://github.com/2factorauth/twofactorauth/wiki/FAQ-Caching).

PR #8226, opened 2024-10-21
Remove region from Let's Do This

haroldadmin:
Let's Do This is expanding into more regions globally, so the regions field does not make sense anymore.


Carlgo11:
Hello Kshitij!
You can just append the new regions or use `-<region>` if you serve users globally except in some countries.
As your service seems to be very geographically restricted, I would advise against removing the regions array completely.

PR #8297, opened 2024-12-03
Add Dollar General

Rabadash8820:
Add dollargeneral.com, for the Dollar General store.

This dialog shows when creating a Dollar General online account. A phone number is required, and is then used to send SMS verification codes on every login. This step cannot be skipped.

image


entries/d/dollargeneral.com.json
```diff
@@ -0,0 +1,12 @@
+{
+  "Dollar General": {
+    "domain": "dollargeneral.com",
+    "img": "dollargeneral.com.png",
+    "tfa": [
+      "sms"
+    ],
+    "categories": [
+      "retail"
+    ]
```
hkamran80, line 15:
```suggestion
    ],
    "regions": [
      "mx",
      "us"
    ]
```

entries/d/dollargeneral.com.json
```diff
@@ -0,0 +1,12 @@
+{
+  "Dollar General": {
+    "domain": "dollargeneral.com",
+    "img": "dollargeneral.com.png",
+    "tfa": [
+      "sms"
```
hkamran80:
```suggestion
      "sms",
      "call"
```

Your screenshot says that calls might also be used.

hkamran80 (changes requested):

entries/d/dollargeneral.com.json
Rabadash8820: Good catch

entries/d/dollargeneral.com.json
Rabadash8820: Also good catch!

hkamran80 (approved):

PR #8326, opened 2024-12-10
Add Roborock

fletcherdodd:


entries/r/roborock.com.json
```diff
@@ -0,0 +1,13 @@
+{
+  "Roborock": {
+    "domain": "roborock.com",
```
Carlgo11, line 3: https://roborock.com/ doesn't work.
```suggestion
    "domain": "roborock.com",
    "url": "https://www.roborock.com/",
```

Carlgo11 (changes requested):

entries/r/roborock.com.json
fletcherdodd: Maybe i'll change it to their global URL (https://global.roborock.com/) instead, the www one goes to their forums 🫤

Carlgo11: Really? To me https://www.roborock.com/ redirects to https://us.roborock.com/ :laughing:

fletcherdodd: Weird, the 3 times I tried that link it was going to https://forum.roborock.com/ now it is going to the US site 🤷‍♂️

Carlgo11 (approved):

PR #8332, opened 2024-12-13
Update 500px

Yang-Nankai:
Update the 500px additional domains.


Yang-Nankai:
I tested and found that `web.500px.com` redirects to `500px.com`, so I added `500px.com` as the `domain`, `web.500px.com` as the additional domain, and changed the location where it is stored (starting with 5).

entries/5/500px.com.json
```diff
@@ -1,6 +1,9 @@
 {
   "500px": {
-    "domain": "web.500px.com",
+    "domain": "500px.com",
+    "additional-domains": [
+      "web.500px.com"
+    ],
```
Carlgo11: Since the parent domain is listed, the subdomain (www) isn't necessary.
```suggestion
```

Carlgo11 (changes requested):

entries/5/500px.com.json
Yang-Nankai: 👌

Carlgo11 (approved):

PR #8342, opened 2024-12-17
Update Rockstar Games

Yang-Nankai:
`rockstargames.com` and `socialclub.rockstargames.com` both use the same login, so make the former the primary domain


hkamran80 (changes requested):
Please change the filename and directory to `entries/r/rockstargames.com`.

entries/r/rockstargames.com.json
```diff
@@ -1,6 +1,9 @@
 {
-  "Rockstar Games Social Club": {
-    "domain": "socialclub.rockstargames.com",
+  "Rockstar Games": {
+    "domain": "rockstargames.com",
+    "additional-domains": [
+      "socialclub.rockstargames.com"
+    ],
```
Carlgo11: The socialclub subdomain isn't needed as rockstargames.com is the parent domain.
```suggestion
```

hkamran80 (approved):

PR #8343, opened 2024-12-18
Update UniFi

Yang-Nankai:
`ubnt.com` redirects to `ui.com`.


Carlgo11 (approved):
Since the domain just points to ui.com, there's not much gained from adding it as an additional domain. But as you've already done all the work, I'll merge it.

PR #8363, opened 2025-01-11
Create Travel category

### LINKED ISSUE #4799: Travel Category
Not sure if this has been discussed before, but, would there be an appetite to add a 'Travel' category? For example, Expedia provides both transport and accommodation, it would fit into either of these categories but I think 'Travel' would fit better. The same can be said for TripIt, which is currently listed under Transport.
I am happy to implement this change if people think this is a good idea.

Carlgo11:
Resolves #4799

* Moved all airlines and hotels to the travel category.
* Removed flygbra.se as they no longer operate.
* Added regions to DB (Deutsche Bahn)
* Moved car rentals to travel category.


fletcherdodd:
Might be wise to update the categories list in [CONTRIBUTING.md](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md) with this PR 🙂

hkamran80:
Should we include rental car services in travel or leave them in transport?

hkamran80:
For some reason, I thought we had way more rental car entries. The only one I could find was SIXT.

kmpoppe:
I believe the last time I thought about adding Europcar there wasn't a category they fit ... But that's been quite some time ago 😄

hkamran80:
Side note: the PR validator is failing due to [too many subrequests](https://developers.cloudflare.com/workers/platform/limits).

Carlgo11:
Yeah 54 changed files is a lot :sweat_smile: 
I can override it if/when the PR is approved.

fletcherdodd (approved):

hkamran80 (approved):

PR #8364, opened 2025-01-11
Update banking category

Carlgo11:
Looked at the banking category and found some entries that needed to be updated.

* __Credit Suisse__ - Removing the entry as they have been acquired by UBS Group
* __Bankera__ - Their sales pitch is that they're _not a bank_ and only a payment middleman. I thus switched the category to payments.
* __pleo.io__ - Added all the regions they operate it. (Sourced from their website)


Carlgo11:
Apparently Bankera is ranked >400K on SimilarWeb. Previously we haven't removed existing sites unless they've dropped substantially, but I leave that choice up to the reviewing maintainer. I'm fine with either leaving it in or removing the entry :shrug:

entries/p/pleo.io.json
```diff
@@ -10,6 +10,25 @@
...
+	  	"gb",
+	  	"ie",
+	  	"it",
+		  "lu",
+	  	"nl",
+	  	"no",
+		  "pt",
+	  	"se",
+	  	"us"
+	  ]
```
hkamran80:
```suggestion
    "regions": [
      "at",
      "be",
      "de",
      "dk",
      "ee",
      "es",
      "fi",
      "fr",
      "gb",
      "ie",
      "it",
      "lu",
      "nl",
      "no",
      "pt",
      "se",
      "us"
    ]
```

entries/b/bankera.com.json
```diff
@@ -6,10 +6,14 @@
...
+      "payments"
     ],
     "regions": [
-      "-us"
+      "-us",
+      "-kp",
+      "-ru",
+      "-ir",
+      "-by"
     ]
```
hkamran80: Given that it gives users a European IBAN with SEPA payment support, should we restrict Bankera to EU regions?

hkamran80 (changes requested):
> Apparently Bankera is ranked >400K on SimilarWeb. Previously we haven't removed existing sites unless they've dropped substantially, but I leave that choice up to the reviewing maintainer. I'm fine with either leaving it in or removing the entry 🤷

I'm fine with keeping it in for now.

hkamran80 (approved):

PR #8367, opened 2025-01-14
Update YNAB

fletcherdodd:


Carlgo11:
Rank of 1 million :thinking: 
Maybe it would be better to just remove the entry?

hkamran80:
`youneedabudget.com` redirects to `ynab.com` which has a rank of 6,822. Maybe we should update the entry's primary domain to `ynab.com`?

fletcherdodd:
> `youneedabudget.com` redirects to `ynab.com` which has a rank of 6,822. Maybe we should update the entry's primary domain to `ynab.com`?

Say no more :)

Carlgo11 (approved):

PR #8384, opened 2025-02-09
Adjust note on Vanguard 2FA, SMS backup is not required

cvoege:
On Vanguard Brokerage 2FA, you do not need an SMS backup if you have multiple hardware keys (i.e. two yubikeys). However doing so will break the mobile app and the mobile website login, they will say something like "Your login method is not supported on this device). If you view the mobile website in "Desktop Preview" however, login does in fact work, and you can then go back to non-preview and be okay. I've updated the note on Vanguard to include both of these factoids. I removed my SMS myself and it's working. 

Here's public a source (just an online forum): https://www.bogleheads.org/forum/viewtopic.php?t=349826

I think someone in that forum mentions that they could bypass 2FA, on mobile with security questions instead of Security Keys, but this does not appear to be the case anymore.

Let me know if there's anything else I can do! Could provide some screenshots if needed.

### DIFF
--- a/entries/v/vanguard.com.json
+++ b/entries/v/vanguard.com.json
@@ -13,7 +13,7 @@
       "Voice Verification"
     ],
     "documentation": "https://investor.vanguard.com/security/security-keys",
-    "notes": "Hardware based 2FA requires SMS / Phone call 2FA as a backup.",
+    "notes": "Hardware based 2FA requires two separate security keys or SMS / Phone call 2FA as a backup. Security keys can not be used to log in to mobile app or mobile website unless \"Preview as Desktop\" is enabled.",
     "categories": [
       "investing"
     ],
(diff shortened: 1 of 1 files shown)


entries/v/vanguard.com.json
```diff
@@ -13,7 +13,7 @@
       "Voice Verification"
     ],
     "documentation": "https://investor.vanguard.com/security/security-keys",
-    "notes": "Hardware based 2FA requires SMS / Phone call 2FA as a backup.",
+    "notes": "Hardware based 2FA requires two separate security keys or SMS / Phone call 2FA as a backup. Security keys can not be used to log in to mobile app or mobile website unless \"Preview as Desktop\" is enabled.",
```
Carlgo11: As the requirement of 2 keys is a common standard, it doesn't warrant a note.
The switching of user-agent by using "Preview as Desktop" is a bit of a hack and would be better left out as it's not an officially supported mode.
```suggestion
    "notes": "Security keys can not be used to log in to mobile app or mobile website.",
```

Carlgo11 (changes requested):

cvoege:
@Carlgo11 works for me, updated!

Carlgo11 (approved):

PR #8408, opened 2025-04-16
Add Brightline

hkamran80:


Carlgo11 (changes requested):
I think it would be better to use this icon rather than their logo.

Carlgo11 (approved):

PR #8422, opened 2025-05-08
Add Texas A&M

hkamran80:
For Texas A&M's documentation URL, I went with [their MFA page](https://it.tamu.edu/duo/), but they have a better documentation page in [their knowledge base](https://service.tamu.edu/TDClient/36/Portal/KB/ArticleDet?ID=364). The knowledge base article is linked from the MFA page. Which should we use?


Carlgo11 (approved):
I like the doc page you used :+1:

PR #8425, opened 2025-05-08
Add George Washington University

hkamran80:


entries/g/gwu.edu.json
```diff
@@ -0,0 +1,21 @@
+{
+  "George Washington University": {
+    "domain": "gwu.edu",
```
Carlgo11, line 3: I can't find a DNS record for gwu.edu :thinking:

hkamran80: When will people start adding root domain DNS records?
Thank you for catching that.

Carlgo11: I guess they don't want to assume people use HTTP and not [Gopher](https://youtu.be/b4aApVkvrNU)?

Carlgo11 (approved):

PR #8444, opened 2025-06-22
Updated info about Truist's broken authenticator app

pleasespammelater:
Screenshot_20250621_200311_Firefox

### DIFF
--- a/entries/t/truist.com.json
+++ b/entries/t/truist.com.json
@@ -2,15 +2,11 @@
   "Truist": {
     "domain": "truist.com",
     "tfa": [
-      "custom-software",
       "sms",
       "call",
       "email"
     ],
-    "custom-software": [
-      "Custom app doesn't work - see reviews at https://play.google.com/store/apps/details?id=com.truist.authenticator and https://apps.apple.com/us/app/truist-authenticator/id1568427655"
-    ],
-    "documentation": "https://www.truist.com/truist-authenticator-app",
+    "notes": "They have a custom app, but users on both iOS and Android report that it doesn't work.",
     "categories": [
       "banking"
     ],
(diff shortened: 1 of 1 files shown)


entries/t/truist.com.json
```diff
@@ -8,7 +8,7 @@
       "email"
     ],
     "custom-software": [
-      "Truist Authenticator"
+      "Custom app doesn't work - see reviews at https://play.google.com/store/apps/details?id=com.truist.authenticator and https://apps.apple.com/us/app/truist-authenticator/id1568427655"
```
hkamran80: This should not be in the `custom-software` array. This should be in the [`notes` string](https://github.com/2factorauth/twofactorauth/blob/master/CONTRIBUTING.md#exceptions--restrictions).

However, I don't think this should be included at all. If the app is nonfunctional, it should be removed.

pleasespammelater: > If the app is nonfunctional, it should be removed

Yeah, I would agree. The only problem then is the documentation link to their website claiming that they have an app!

How about I:
1. Remove the custom-software array
1. Remove the custom-software entry from the tfa field
1. Remove the documentation entry
1. Add a note about them having an app, but it doesn't work

What do you think of that plan?

hkamran80: The first three sound good to me. I'm not sure about the note though.

@2factorauth/maintainers, any advice?

Carlgo11: Yeah sounds good :+1:

pleasespammelater: > Yeah sounds good 👍

okay - changes are in!

hkamran80 (approved):

hkamran80:
Thank you! Your contribution will be visible in around 10 minutes, although it may take longer due to [server and browser caching](https://github.com/2factorauth/twofactorauth/wiki/FAQ-Caching).

PR #8520, opened 2026-01-01
Add `u2f` to openai.com

hydrabeer:
OpenAI now supports hardware keys.

https://help.openai.com/en/articles/20001039-passkeys-to-secure-your-openai-account


hkamran80:
It looks like they only support passkeys, not FIDO2/U2F for MFA. This should be created in the [2factorauth/passkeys](https://github.com/2factorauth/passkeys) instead.

PR #8525, opened 2026-01-07
Update ameriprise.com.json

nelson-old-man:
Ameriprise supports TOTP (Google Authenticator) which isn't noted yet on 2fa.directory. 

Additionally, I found a better documentation link which goes directly to the 2FA configuration settings, rather than needing to click a link in the 1st documentation page to be send to the 2nd documentation page.

Thank you for maintaining this, I can't even imagine how many people this has saved by making them aware of more secure options!

### DIFF
--- a/entries/a/ameriprise.com.json
+++ b/entries/a/ameriprise.com.json
@@ -2,9 +2,10 @@
   "Ameriprise Financial": {
     "domain": "ameriprise.com",
     "tfa": [
-      "sms"
+      "sms",
+      "totp"
     ],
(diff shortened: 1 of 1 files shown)


Carlgo11 (approved):
Thanks for the pull request @wnelson03! 🙂
I added the Ameriprice app too since push notification based authentication appeared to also be available.

Your contribution will be visible in around 10 minutes. Sometimes, it takes longer to show up due to [server and browser caching](https://github.com/2factorauth/twofactorauth/wiki/FAQ-Caching).

PR #8528, opened 2026-01-09
Update monzo.com.json to include Ireland and note about magic links

Jacob-Roberts:
Include note that explains that Monzo only uses magic links, so 2FA work transitively


hkamran80:
Using a magic link does not constitute two-factor authentication, so there's no need to add a note about it.

PR #8538, opened 2026-02-07
Update Virgin Plus

### LINKED ISSUE #8537: Update Virgin Plus
### Site name

Virgin Plus

### Site URL

https://virginplus.ca/

### Update reason

It now supports 2FA.

### Additional information

Virgin Plus now supports MFA by voice call, SMS or email per this documentation:

https://www.virginplus.ca/en/support/myaccount-mfa.html

That being said, there is no option for customers to opt in or out of MFA. Instead, the customer is prompted for a second factor only on "unusual activity", whatever that means.

### Issue Eligibility

- [x] The issue I'm creating is not a duplicate of an existing [issue](https://github.com/2factorauth/twofactorauth/issues).
- [x] The issue I'm creating is not a duplicate of an existing [pull request](https://github.com/2factorauth/twofactorauth/pulls)

fletcherdodd:
Closes #8537

SVG is a bit on the large side. Tried to optimise as much as possible, maybe sticking with the PNG is the best approach 🤔


hkamran80 (commented):
I think we should keep the PNG.

hkamran80 (approved):

PR #8558, opened 2026-03-31
Add Fintual

melipass:
Adds Fintual to the directory as an investing platform available in Chile and Mexico. Login uses a one-time code sent by email; users can optionally register a passkey (listed as u2f) with user verification (mandatory second factor).


melipass:
Seems like the External Tests job failed due to a Cloudflare error. pr-validator.2fa.directory returned Cloudflare 1102 (“Worker exceeded resource limits”). Looks transient / service-side.

image

fletcherdodd:
fintual com
I found an SVG icon that would be preferred over the PNG :)

melipass:
Thanks @fletcherdodd, just made a commit with that change.

I also have a PR in the passkeys repo without an icon (https://github.com/2factorauth/passkeys/pull/74) so that the svg here is retrieved.

entries/f/fintual.com.json
```diff
@@ -0,0 +1,23 @@
...
+    "categories": [
+      "investing"
+    ],
+    "regions": [
+      "cl",
+      "mx"
+    ],
+    "additional-domains": [
+      "fintual.cl",
+      "fintual.mx"
```
fletcherdodd, line 20:
```suggestion
```

entries/f/fintual.com.json
```diff
@@ -0,0 +1,23 @@
+{
+  "Fintual": {
+    "domain": "fintual.com",
+    "img": "fintual.com.svg",
```
fletcherdodd, line 4:
```suggestion
    "additional-domains": [
      "fintual.cl",
      "fintual.mx"
    ]
```
Moving this up just so it's arranged out like other entries. The `img` field can also be removed as it's only required when you are using a PNG file.

entries/f/fintual.com.json
```diff
@@ -0,0 +1,23 @@
...
+    "tfa": [
+      "email",
+      "u2f"
+    ],
+    "documentation": "https://fintual.cl/seguridad-confianza/",
+    "notes": "A one-time code is emailed on each login. You can add a passkey instead and use device biometric or PIN as the second factor.",
```
fletcherdodd, line 10: This can either be removed all together or reworded. What you have noted is usually already implied, if the 2FA email method is on by default you could something like 'Email 2FA is always enabled. U2F security keys replace Email 2FA if setup' (if that is the case, otherwise you would just highlight that Email 2FA is always enabled).

entries/f/fintual.com.json
```diff
@@ -0,0 +1,23 @@
+{
+  "Fintual": {
+    "domain": "fintual.com",
+    "img": "fintual.com.svg",
+    "tfa": [
+      "email",
+      "u2f"
+    ],
+    "documentation": "https://fintual.cl/seguridad-confianza/",
```
fletcherdodd, line 9:
```suggestion
    "documentation": "https://ayuda.fintual.cl/articles/10421693",
```
This is a better link as it provides steps on setting it up rather than just providing methods available

fletcherdodd (commented):
Nice work on the PR. A couple of other things need to be addressed before we can approve and merge.

I am struggling to find any documentation or evidence that email 2FA is an option. Are you able to provide a screenshot of this process? Either from the login screen or your account settings if it requires setup. Remember to blur out any sensitive information!
Also could you provide a screenshot of the FIDO2 (U2F) MFA setup, just to confirm it is available and not just passkey support?

PR #8572, opened 2026-04-25
Update BMO

4fosaefp:
Add support for app push notification & remove email

### DIFF
--- a/entries/b/bmo.com.json
+++ b/entries/b/bmo.com.json
@@ -4,10 +4,13 @@
     "img": "bmo.svg",
     "tfa": [
       "sms",
-      "email",
-      "call"
+      "call",
+      "custom-software"
     ],
-    "documentation": "https://www.bmoinvestorline.com/General_Info/MFA_preEnroll_SD_EN.html",
+    "custom-software": [
+      "BMO App Push Notification"
+    ],
+    "documentation": "https://www.bmo.com/en-ca/main/personal/ways-to-bank/alerts/",
     "regions": [
       "ca"
     ],
--- a/entries/b/bmo.com.json
+++ b/entries/b/bmo.com.json
@@ -10,7 +10,7 @@
     "custom-software": [
       "BMO App Push Notification"
     ],
-    "documentation": "https://www.bmo.com/en-ca/main/personal/ways-to-bank/alerts/",
+    "documentation": "https://www.bmo.com/en-ca/main/personal/security-centre/how-to-protect-yourself/",
     "regions": [
       "ca"
     ],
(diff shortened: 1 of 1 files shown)


entries/b/bmo.com.json
```diff
@@ -4,10 +4,13 @@
...
-      "email",
-      "call"
+      "call",
+      "custom-software"
     ],
-    "documentation": "https://www.bmoinvestorline.com/General_Info/MFA_preEnroll_SD_EN.html",
+    "custom-software": [
+      "BMO App Push Notification"
+    ],
+    "documentation": "https://www.bmo.com/en-ca/main/personal/ways-to-bank/alerts/",
```
dsoegijono: I might be missing something but that page seems to be for alerts on banking activities instead of 2fa?

fletcherdodd: I read that too, but it does seem to be the only place they detail steps on setting up the app. This page talks about both alerts and verification codes (but provides no setup steps) https://www.bmo.com/en-ca/main/personal/security-centre/how-to-protect-yourself/?icid=tl-FEAT2953BRND4-AJBMOH296

image

entries/b/bmo.com.json
```diff
@@ -4,10 +4,13 @@
     "img": "bmo.svg",
```
fletcherdodd, line 4:
```suggestion
```
I know you didn't put this in here but it can be removed, as this is only needed if the file is a PNG.

entries/b/bmo.com.json
dsoegijono: Even that verification code is for "activity that isn't typical", so still transaction based. It doesn't really imply second factor during login?

4fosaefp: BMO doesn't really document it well. But it is listed along side SMS and calls for 2FA.

The settings page in the app also mentions it:
Screenshot_20260502-114447

Here's what it looks like in practice during login:
Screenshot_20260502-114330
Screenshot_20260502-114344
Screenshot_20260502-114116

Updated documentation link to https://www.bmo.com/en-ca/main/personal/security-centre/how-to-protect-yourself/

entries/b/bmo.com.json
4fosaefp: Got `Error: Image img/b/bmo.com.svg not found` if I remove this line.
I guess it's because the names are different: `bmo.com.json` vs. `bmo.svg`
And `bmoinvestorline.com.json` uses the same svg

hkamran80: We need to keep the `img` here. BMO's logo is used by three entries.

fletcherdodd (approved):

PR #8579, opened 2026-05-07
Update wolf.bet.json

markwolf92-92:
change domain to new address Wolfbet.com


fletcherdodd:
See #8581 as to why this was closed without merging.

PR #8580, opened 2026-05-07
Update and rename wolf.bet.json to wolfbet.com.json

markwolf92-92:
Change all for new brand and domain


fletcherdodd:
See #8581 as to why this was closed without merging.

PR #8624, opened 2026-08-03
Merge Samsung and Smartthings

HandyHat:
smartthings.com is no longer served on its own domain. Also updates docs link


Carlgo11 (changes requested):
Thanks — merging these is the right call, and removing `img/s/smartthings.com.svg` alongside the entry is correct (an orphaned image is a hard error in `tests/images.js`).

One change needed before this can go in: **please add `smartthings.com` as an `additional-domain`.**

`smartthings.com` does still resolve — it redirects to `https://www.samsung.com/us/smartthings/`. As the PR stands, deleting the entry means anyone searching for `smartthings.com` in the directory gets no result at all. Keeping it as an additional domain preserves that lookup while still merging the two entries:

```json
{
  "Samsung": {
    "domain": "samsung.com",
    "additional-domains": [
      "smartthings.com"
    ],
    "tfa": [
      "sms",
      "totp"
    ],
    "documentation": "https://v3.account.samsung.com/guide/two-step-verification/introduction",
    "categories": [
      "iot",
      "retail"
    ]
  }
}
```

Adding `iot` to the categories is a good catch, since the merged entry now covers SmartThings.

HandyHat:
Done - thanks @Carlgo11 !

Carlgo11 (approved):

PR #8639, opened 2026-08-15
Update TfL docs link

HandyHat:
Old link redirects to new

### DIFF
--- a/entries/t/tfl.gov.uk.json
+++ b/entries/t/tfl.gov.uk.json
@@ -4,7 +4,7 @@
     "tfa": [
       "sms"
     ],
-    "documentation": "https://tfl.gov.uk/corporate/privacy-and-cookies/protecting-your-account",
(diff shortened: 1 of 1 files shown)


entries/t/tfl.gov.uk.json
```diff
@@ -4,12 +4,12 @@
     "tfa": [
       "sms"
```
fletcherdodd, line 5:
```suggestion
      "sms",
      "totp"
```
Their documentation makes mention of the ability to setup an authenicator app/totp. Do you have an account with them that you can very this?

Image

HandyHat: I do have an account with TfL and can confirm that at the moment, TOTP is not an option. When I opened this PR, that paragraph was not on the website - it looks like this is an early indication that TOTP is coming soon, which is very exciting!

fletcherdodd: Ah gotcha! Fair enough :)

fletcherdodd (approved):

PR #8653, opened 2026-09-10
Fix formatting issue in codeql.yml

mydickassbusiness:


Carlgo11:
Not sure what you're trying to do here. CodeQL isn't something that would benefit a repository of JSON files.

PR #8654, opened 2026-09-12
Add Personify Health

fletcherdodd:
As per [this article](https://personifyhealth.com/press/virgin-pulse-and-healthcomp-introduce-combined-company-as-personify-health/), Virgin Pulse is now Personify Health. The domain currently does not point to personifyhealth.com though.


entries/p/personifyhealth.com.json
```diff
@@ -0,0 +1,17 @@
+{
+  "Personify Health": {
+    "domain": "personifyhealth.com",
+    "additional-domains": [
+      "healthcomp.com"
+    ],
```
hkamran80, line 6: Should we add `virginpulse.com`?

fletcherdodd: It doesn't redirect there currently it seems. Could do anyways though

hkamran80: Do you know if they use the same accounts?

fletcherdodd: Virgin Pulse seems to be completely shutdown.

hkamran80 (approved):

PR #8661, opened 2026-09-19
Update Fitbit

fletcherdodd:
- Wasn't sure whether or not to including '(Fitbit)' at the end of the name like the app stores have it listed as. The website only says 'Google Health' currently.
- Google Fit is a seperate app still but will eventually be merged into the Health app later in the year ([as per](https://9to5google.com/2026/05/07/google-fit-shut-down-health-replacement-migration-tool-coming/))


hkamran80 (approved):
I think it's fine without the Fitbit suffix

